PR #2, opened 2017-07-15
Added script to move all libraries

SchrodingersGat:
(Filters need work)

The added script provides a framework for transferring the entire set of libraries.

Filter pairs are added in a dict, with the format:

```
{
"lib:part" : output_lib
}
```

* Supports wildcard filters (`?` / `*`)
* Move entire library by leaving partname blank


SchrodingersGat:
@poeschlr @jkriege2 this is the script I was talking about [here](https://github.com/KiCad/kicad-library/issues/1415).

It works well for copying entire libraries and also selecting parts to move to other libs. I have not added many filters but it provides some good starting examples.

If we get the script ready in parallel with maintaining 4.x libraries, we should be ready to go when v5.0 comes around.

SchrodingersGat:
Example usage:

```
python move_symbols.py ~/kicad/share/kicad-library/library/ ../output --schlib ~/kicad/utils/schlib --real > log.txt
```

hackscribble:
Is it worth merging the script as is, even if there are still fixes or changes needed to the code?  That way, people could fork and help with the job of adding the filters.

SchrodingersGat:
I think it would be worth merging as it - so others can contribute to the filter development as you say.

SchrodingersGat:
This rearrangement process is one that is of some priority before v5. If anyone wants to help with the filtering and rearranging all the libraries, that'd be great!

@jkriege2 @poeschlr @hackscribble @Shackmeister

SchrodingersGat:
Merged so others can easier help with the filters. We'll delete the scripts once we release the libs

PR #9, opened 2017-11-27
DSP

SchrodingersGat:
Transferred DSP_Microchip library


evanshultz:
This DSPIC library seems fine, but the check scripts seems to be busted. Also, should dsp.lib/dsp.dcm be brought over here or were you going to do that separately?

SchrodingersGat:
I am working on the scripts currently. They have never been enabled for this repo :)

Thanks for the hint on dsp.lib - I have added that now too

SchrodingersGat:
Yay scripts are working again (and better than ever!)

evanshultz:
@SchrodingersGat Great! There are some ugly issues with missing or wrong default footprints and FPfilters. Do you want to merge as-is or correct those now?

SchrodingersGat:
Please merge as-is and we will raise an issue to fix. Thanks

SchrodingersGat:
@evanshultz @poeschlr bump - ok to merge as is?

evanshultz:
@SchrodingersGat Can you fix the conflict?

Do we merge with major errors like this? I will merge, but CvPcb won't work (well). Admittedly, it was already broken so this isn't a regression...

SchrodingersGat:
Yes it will be a lot of work to fix all those. We will create an ongoing issue. As long as we don't make it worse :)

antoniovazquezblanco:
I use Microchip DSPs and the looks are not consistent between different symbols. Also microchip, depending on the package changes pin numbering so I think it would be a good idea to provide a symbol for each packaging option otherwise users may find problems in the future but this requires a good amount of work I cannot contribute right now.

SchrodingersGat:
@antoniovazquezblanco valid points, there could be a lot of improvements made here

PR #10, opened 2017-11-27
MCU

SchrodingersGat:
This PR adds all the `MCU_xx` libraries. Incorrect footprint associations have been (somewhat) fixed, as far as can be done automatically.


evanshultz:
@SchrodingersGat Can you check the script output?

Many of these have a "/" in the name, since Microchip likes to do that. I also see opening and closing parentheses. All 3 chars are in violations of KLC. Should those be fixed before merging?

poeschlr:
From mailing list i gather that "/" is a problem with the new symbol-lib-table. So we should try to avoid this char if at all possible.

SchrodingersGat:
I have fixed the `/` issues across these libs. I have also changed all instances of (e.g.) **PIC16(L)F** to simply **PIC16F**

- I do not have the time to add aliases for **L** variant
- They are all pin compatible anyway

evanshultz:
Sounds like you've done all that you want to do right now.

evanshultz:
However, there are still some issues that may be serious. Missing FPfilters, incorrect footprint fields, missing DCM entries (at least some of these are because the DCM still has the `/`), etc. I assume at least some of that should be fixed before merging.

SchrodingersGat:
I had an extra commit sitting on my computer that I had not pushed, fixing those DCM files. Should be ok now (at least for the slashes)

SchrodingersGat:
If there are any issues that are "worse" than current v4 libs please raise them, otherwise merge :)

SchrodingersGat:
@poeschlr I think this is OK to merge. I shall create an ongoing issue for this after merge.

poeschlr:
Yup looks good as well.

PR #11, opened 2017-11-27
Logic

SchrodingersGat:
Added Logic libraries


evanshultz:
Here are some things that could be fixed.
- There are no packages or package types (just "SSOP" without pin count) listed in part descriptions
- In the descriptions "Schmidt" should be "Schmitt"
- No supply voltages given in descriptions
- 74LVC1G11 uses "3in" but I think "3 input" is a nicer way to write it; also see 7400 and 7402
- Some hyphens in FPfilter, such as 74LVC1G02
- Some descriptions use all lowercase letters
- 74LS09 is describes as "Open Collect" instead of "Open Collector"
- The datasheet links in Logic_74xx don't work for me
- At least in the IEEE libs there are some lowercase letters for pins names; do we allow that?

There was a recent discussion about having a separate power symbol for even single-gate logic. I still support this so I want to resolve that discussion. I believe @jkriege2 was with me on this point.

Also, the check scripts seem not to work which makes me too nervous to merge since we could use this as an example to fix it and check that they work.

Let me know if you don't care to fix these, or at least not now. Bob's script may address some of them.

SchrodingersGat:
Scripts are running now and I have touched a couple of the smaller issues

evanshultz:
This lib really needs a lot of work to get it all cleaned up. Ready to merge now as-is?

poeschlr:
I fear we don't have time to get the libs perfect before the v5 release. I would suggest we transfer the libs with minimal changes but document the problems we find as issues. If we then have time before the v5 release we can fix the problems then.

(We could even ask for help over at the forum if we have well documented things to do.)

SchrodingersGat:
As long as there are no critical issues that will break compatibility (such as the slashes in the other lib) then let's merge, raise an issue, and fix it at a later date. The priority is to transfer all the libs :)

bobc:
I reworked 74xx.lib and cmos4000.lib, I didn't touch the others but noticed that they could do with rework too.

I am not surprised URLs don't work, URLs are notorious for becoming stale as web sites are reworked, I would be surprised if any expects URLs to keep working :) In fact, why do we bother with the datasheet URL anyway? All I do is google the part number and put it into the component, which is exactly what the user could do, and they will get an up to date link. The datasheet URL is just a very poor web cache.

SchrodingersGat:
@poeschlr ok to merge? We should create an issue that these libs need to be fixed :)

evanshultz:
See https://github.com/KiCad/kicad-symbols/issues/27.

PR #13, opened 2017-11-27
Singles

SchrodingersGat:
Move across a bunch of symbol libraries that can be mapped 1:1

(i.e. simple renaming)


poeschlr:
Should the espressiv lib be named a bit differently? (Example Rf_Wifi)

You have a lib called Interface_RF but according to https://github.com/KiCad/kicad-library/issues/1402 i think this lib should be split between Rf_Wifi and either Rf_am_fm or Rf_misc.

The lib reference voltage has some reference current sources in it.
Do we want the reference diodes in there?

evanshultz:
Some of these issues look severe. Missing footprints, wrong footprint formatting, etc.

I'm not trying to slow this process down and fix everything now, but if CvPcb breaks because of these library changes that seems bad. We want v5 to be just as good as v4 out-of-the-box, right? Breaking the symbol->footprint connection seems like a step backwards that shouldn't be allowed.

SchrodingersGat:
Yes they should be "just as good". Can you point to any specific issues here where the new changes are explicitly worse? If so, I shall fix those :)

poeschlr:
@SchrodingersGat do you want to address any of the questions i raised?

SchrodingersGat:
Yep, I'll deal with those on Monday :)

SchrodingersGat:
- [x] Split Interface_RF into RF_WiFi and RF_AM_FM
- [x] Split Reference_Voltage into Reference_Voltage and Reference_Current

@poeschlr any other issues, or OK?

poeschlr:
I don't see any issue that would prevent a merge.

PR #15, opened 2017-11-30
Transferred memory.lib

SchrodingersGat:
This PR splits the `memory.lib` library into:

- EEPROM
- Flash
- NVRAM (FRAM / MRAM)
- RAM

There libraries need a lot of work. If there is nothing critical that anyone can see, please merge and we can raise an issue to fix.


poeschlr:
Do we have a script that can at least change the Housings_xxx lib name to Package_xxx and Pitch -> P?
This should repair most footprint associations.

SchrodingersGat:
Yep, I'll run that later on. Probably not until Monday, I'm away all weekend

SchrodingersGat:
@poeschlr the incorrect footprint library names have been fixed. Ok to merge?

poeschlr:
I think this is ok for now. Any further fixes can be made later on.

PR #26, opened 2017-12-05
Transfer and split existing connector lib

poeschlr:
I would suggest to split the connector lib between the completely generic symbols and the more specialized symbols.
I am not sure if the male/female symbols should move towards the specialized lib.

I am also not really sure about the name.

In the near future i will also add symbols with shielding into a third lib called `Connector_Generic_Shielded`. This will simply hold the same symbols that are currently in the Connector_Generic lib with the addition of on pin for connecting the shield. (Pin name still to be decided.)


poeschlr:
I decided to add the shielded versions for generic connectors.
I used SH as the pin number for the shield connection pin.

The dual row version does have the shield pin not on 100mil grid but on 50 mil grid.

screenshot from 2017-12-05 20-01-23
screenshot from 2017-12-05 20-01-40

SchrodingersGat:
Are the `160`mil pins intentional?

poeschlr:
yes.
For shielded versions i reduced the main body size (yellow rectangle) by 10 mil in all directions.
I can either place a short 10mil line to elongate the pins or make the pins longer by 10 mils. (I decided on the later)

I did this to ensure that even the 3 char pin numbers have enough space.

SchrodingersGat:
That's fine, probably ok to leave as is. Any further changes to be made or OK to merge?

poeschlr:
I don't know of any changes. If the symbols and lib names are ok for your then you are free to merge.

SchrodingersGat:
Thanks :)

PR #31, opened 2017-12-06
add 4P4C connector (aka RJ9, RJ10 or RJ22)

lorem-ipsum:
Add the (not really standardised) 4P4C connector. It is often called RJ9, RJ10 or RJ22, but neither is correct as it not technically a registered jack.
The drawing is slightly asymmetric and I find it ugly, but there is no way to fix it as the drawing has to be on the 50 mils grid (or I am too stupid to find the override).
There is no specification or data sheet for this connector I could find.
A PR for a specific footprint will follow in the next days if I have decided which one to use for my project.

4p4c


SchrodingersGat:
@lorem-ipsum looks good to me, thanks for the contribution :)

poeschlr:
@SchrodingersGat since when do we went I/O pins on the top/bottom of a symbol?

poeschlr:
@reportingsjr you can change the grid by right clicking on the canvas. (At least this works in kicad stable.) This might not work while a tool is active.

lorem-ipsum:
@poeschlr I've just followed the existing RJ connectors. But If there is an agreement on how they should look I can also go through them add change/add all the RJ connectors (they are not that many).
Personally I think it would make sense to add all the nPmC names as well as the matching RJ as aliases. Wikipedia has a lengthy list with all the associatons and the existing confusions.

SchrodingersGat:
I thought it best to match the current format, plus I think that it's a fairly good layout for these. We could rotate the symbols perhaps...

PR #32, opened 2017-12-06
Moved more symbol libs

SchrodingersGat:
This PR moves across some more old libraries, merged into a variety of existing and/or new ones:

* brooktre
* Decawave
* RF_Bluetooth
* nordicsemi
* Bosch
* silabs
* supertex
* Switch


evanshultz:
A few thoughts:
- Is `MCU_SiLabs` too brief? I feel when I've abbreviated that much it was discouraged. `MCU_Silicon_Labs`?
- Is just `Sensor.lib` too generic? Perhaps `Sensor_Misc` as Rene proposed at https://github.com/KiCad/kicad-symbols/pull/34#issuecomment-349643369.
- Is `Espressif` supposed to be removed? I don't see that lib already merged.

Other issues seem minor and can be handled later. The pin offset rule needs a KLC update.

SchrodingersGat:
* I can rename to `MCU_SiliconLabs` if that is preferable
* See my response at #34 
* Espressif has been integrated into RF_WiFi

evanshultz:
I prefer renaming the library for clarity. Otherwise, thanks for the info and I'll merge when ready.

evanshultz:
@SchrodingersGat One more that you may want to address now since this is a breaking problem: `ESP8266EX` specifies footprint library `Housings_DFN_QFN`.

PR #33, opened 2017-12-06
add Schrack RT2 DPDT relays

lorem-ipsum:
This adds the Schrack RT2 relays. Schematic was copied from an existing relay, only pin numbers were modified. [Datasheet](http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Data+Sheet%7FRT2%7F1014%7Fpdf%7FEnglish%7FENG_DS_RT2_1014.pdf%7F6-1393243-3). The footprint was already created by someone else.

As I was done I found out that the naming I've chosen (I haven taken everything, but dropped the last 3 components designating the coil) is not unique, as there exist [bistable versions](http://www.te.com/commerce/DocumentDelivery/DDEController?Action=showdoc&DocId=Data+Sheet%7FRT2_bistable%7F1116%7Fpdf%7FEnglish%7FENG_DS_RT2_bistable_1116.pdf%7F4-1393243-5)  which only differ in the third last component. Now I am not quite sure what to do. I could either add all variants (which are a lot) or add a suffix (possibly only to the bistable ones). Any suggestions on how to deal with this?
rt424


poeschlr:
Regarding symbol naming. We can use clever wildcarding to communicate what symbol represents what.
As far as i can tell the bistable versions start with either F or A.

I think this would mean we have the following symbols:

- RTx2xxxxx
- RTx4xxxxx
- RTx4xxAxx
- RTx4xxFxx

@SchrodingersGat, @jkriege2 and @evanshultz what do you guys think?

lorem-ipsum:
Ok, then I will rename them and add the manufacturer prefix.
I didn't knew that overlapping aliases are fine (RTE4xxxx and RTE4xAxx).
I will also add a PR for the modified housings for the bistable versions soon.

poeschlr:
Just make sure that the description field is filled out accordingly. It should be clear to the user by only reading the description what the symbol represents.

lorem-ipsum:
As there is no reference to copy from (and this is my first time creating relay footprints) is this fine for the bistable, polarized relays?
rt42xfxx

poeschlr:
Thanks for your contribution.

PR #34, opened 2017-12-06
Created AS3935 Lightening sensor

openelectron:
Created AS3935 Lightening sensor
Created Sensor .lib to hold new part.

This was originally a PR to the old library, moved at the request of @SchrodingersGat 

Datasheet is free but behind a registration link, however digikey has a mirror;
https://media.digikey.com/pdf/Data%20Sheets/Austriamicrosystems%20PDFs/AS3935.pdf

screenshot_2017-12-04_09-49-59


openelectron:
I think I need to update the symbol table for the newly added library but I'm not sure.

poeschlr:
This should go into the existing `Sensor_Optical` lib

openelectron:
I'm happy to move it but it isn't an Optical sensor so that doesn't seem like the right place to me. It detects Lightening up to 40km away, I think by measuring electrical field strength.

poeschlr:
Hm ok i read it as light. sorry for the confusion.
Well then either we make a Sensor_Misc lib for sensors types that do not merit a separate lib or we make a Sensor_Lightning lib.
@SchrodingersGat, @evanshultz, @jkriege2 what would you guys prefer?

evanshultz:
Ugh. I dislike `misc` since it's so generic and can be a garbage heap, but sensors are so varied that it seems best to me. I vote for creating `Sensors_Misc` for this part and using judicious supervision of that lib.

SchrodingersGat:
The thinking on my part was as follows:

* Sensor_<function> libraries contain sensors with a very specific function (e.g. temperature sensing)
* Any others (e.g. multi-function or hard to categorize) can go in `Sensor.lib`
* The same goes for other classes of symbols e.g. `MCU.lib` contains assorted or uncategorized microcontrollers.

Adding `misc` or `unassigned` is messy and I think the approach above provides a clean solution

openelectron:
Sensor.lib seems the best place to put this to me. 
I have updated the PR to pull in upstream changes.

I don't understand why the CI tests are failing, if someone can help me I will fix whatever is wrong.

SchrodingersGat:
@openelectron your symbols have no KLC errors, the CI tests are failing because we have not yet got the sym-lib-table file fully worked out.

Now on to your changes:

1. Please remote the `.swp` and `.save` files that have made it into this PR

openelectron:
@SchrodingersGat  I have removed those files. I know the swp files come from git when doing a merge.  I have missed them before.

Would there be any help in adding that to the helper scripts?. I'm happy to do a PR to add that.

SchrodingersGat:
It would be helpful as sometimes it does not get spotted.

Thanks for the contribution :)

PR #35, opened 2017-12-06
4P4C center drawing

lorem-ipsum:
Center the drawing in the just merged 4P4C connector. No further changes.

4p4c


poeschlr:
Thanks.

Issue #36, opened 2017-12-06
RJxx symbols are oriented with their pins downwards instead of towards the right like other connector symbols


poeschlr:
See discussion in https://github.com/KiCad/kicad-symbols/pull/31

lorem-ipsum:
I've started a cleanup of all the registered jack / modular connectors in a [branch](https://github.com/lorem-ipsum/kicad-symbols/tree/rj-cleanup) of my fork. I will fix the orientation at the same time.

lorem-ipsum:
I have updated all connectors, but I am currently not sure what to do with the RJ45-TRANSFO. According to the old description it is based of a specific component from Wurth which I couldn't find with a quick search. In general the RJ45 with transformer is a generic component, but i don't think the pin numbering is the same for all connectors. Proper handling will most likely require manufacture specific components, which can hopefully be handled mostly by aliases.

![rj45_transfo](https://user-images.githubusercontent.com/1010629/33672005-65115f6c-daa9-11e7-9335-8a6a58d6098b.png)

PR #37, opened 2017-12-06
RJxx / modular connector cleanup

lorem-ipsum:
Update of all the modular connectors. Previously only selected connectors (RJ12 and RJ45) existed, also the orientation of many was wrong (see issue #36). This update replaces all these connectors with their correct modular connector (xPyC) and aliases to all the RJ connectors (based off [Wikipedia](https://en.wikipedia.org/wiki/Registered_jack)).
I have removed the shielding from all connectors as it is not part of the connector itself (and many without shield exist). I will most likely create RJxx_Shielded variants where I find them.

4p4c
6p6c
8p8c_led


poeschlr:
Could you resolve the merge conflict?

jkriege2:
About the shielding: At least the RJ45 are often (in most cases? ... see https://de.rs-online.com/web/c/steckverbinder/netzwerk-und-telekommunikations-steckverbinder/rj-45-steckverbinder/?applied-dimensions=4294796229,4294807825,4294782227) available with shield, so I think we should have variants with extra shield pins (see e.g. the D-SUB-connectors, which are available with and without shield-pin each).

poeschlr:
We now also have them for generic connector symbols.
I think we should also use SH as pin number here.

lorem-ipsum:
I've played a bit with shielded variants and came up with two different approaches. A think the first version is a bit more clear, the second version has the advantage that it is the same size as the unshielded variants. Which versions do you prefer?
4p4c_shielded_1
4p4c_shielded_2

lorem-ipsum:
I rebased the branch onto the current master. But I am having a few more questions/issues:
- The pin numbering is currently not consistent between footprints and symbols. If a connector has less connected than positions the numbering is different. The symbol removes the outer pins, dropping pin number 1 and starting at higher numbers. The footprints however all start at 1 (that's for the footprints I am making right now, based on the Molex notation). I think the symbol notation makes more sense, but there should be an agreement before finishing those.
- Should the name be _SHIELDED or _Shielded?
- The description is not yet consistent. Is `8 positions, 8 connected, (shielded) modular connector` fine?

poeschlr:
About the pin numbering:
I am a bit confused what you mean.

The footprints should use the pin numbers as they are defined in their respective datasheets. The symbol should follow the same naming scheme.
You do not need to take into account users who connect the symbol to a footprint that has a different number of pins than the symbol. (This usecase is not supported. The netlist import will complain about the missmatch)

lorem-ipsum:
I should have added a picture in the first place. Below is a 4P2C connector (please ignore all the other issues as rotation). Following the connector labeling the remaining pins have numbers 2 and 3 as the two outermost positions (1 and 4) are not connected. However the plugs I have found so far use numberings always starting at 1 (I couldn't find any 4P2C to be honest, but 6P4C exist with the same issue). 
screenshot_20180115_133538

evanshultz:
@poeschlr 
Are you able to take a look at this PR again? Note my PR above which changes the name of the library to Connector.lib and also updates many symbols. One PR being merged will break the other one.

evanshultz:
@poeschlr 
Can you take a look at these symbols and also my question above?

evanshultz:
@poeschlr @jkriege2 
Do you have more comments on these symbols? This branch is already broken and I can add these symbols to my PR above if you think they're ready to review/merge.

jkriege2:
they look fine to me!

Schlumpf:
Any news on this? I would like to push some footprints.

evanshultz:
@lorem-ipsum 
Do you have any more changes to make or outstanding questions? If we don't hear a response, then I'll take these symbols and start a new PR.

@Schlumpf 
You should be free to make footprints. I can imagine only pin numbers might change but otherwise thes symbols won't affect your footprints. Thanks for offering to submit footprints!

lorem-ipsum:
Besides time difficulties I am still wondering about the pin labeling for the 4P2C connecters (and similar) posted above. Should they be labeled 2-3 (including the non-connected positions) or 1-2?

Otherwise if you have time, feel free to take these symbols and include them into your footprint.

evanshultz:
I've imported these symbols, with the changes above, to my PR at https://github.com/KiCad/kicad-symbols/pull/310. I'll close this in a day or so if there's no objection.

Ratfink:
Merged via #310.

PR #42, opened 2017-12-07
Transistor sorting

SchrodingersGat:
The transistor library was too large, and had some parts that were not actually transistors.

This PR sorts transistors into:

* BJT
* FET
* IGBT

And also moves non-transistor parts to where they should go.

`Transistor.lib` has been deleted.


poeschlr:
Is it on purpose to move transistors into power management?
Yes they can be used for power management but this is not really what i would search in that lib. (I would expect the power management lib to contain the chip that tells the fet to turn on or off.)

evanshultz:
@poeschlr Only transistors with supporting circuitry moved into that lib. Discrete transistors are in the 3 libs mentioned above and that all looks fine to me.

There are loads more intelligent switches in the old ir.lib. Do you want to take care of those right now?

poeschlr:
So you would say that this transistor: https://www.infineon.com/dgdl/ips6011pbf.pdf?fileId=5546d462533600a4015355c3db3d1625
should be in power management?

From the feature list i would say it is the same as all other infineon high side smart switches. (minus spi control which some of them have) Would you move all high side smart switches into power_management?

evanshultz:
My comment above only said that discrete transistors were in the libs mentioned above and that transistors with extra stuff were in power_management. I did not state any belief or preference.

It does make sense to me that discrete transistors and these intelligent/smart transistors are separated. The user is free to use a discrete transistor in any way they like, and that doesn't apply to these smart switches. I think that power_management would be for switch controllers, PMICs, current monitors, etc., and my brain would not look for not smart switches there.

Other than perhaps separating high-side and low-side, I can't think of anything better than your suggestions above for new lib names. My gut says there may be an even better lib name with further consideration, though.

SchrodingersGat:
Digikey categories all these under `PMIC - Power Distribution`.

My feeling was to group these and other power devices (e.g. ideal diode controllers) into a single category to avoid too much splintering of the libs...

SchrodingersGat:
Any further issues preventing merge?

poeschlr:
I will merge it for now, but i still think power_management is not the best place to put smart fets. But this can be discussed in a separate issue.

PR #45, opened 2017-12-10
Moved Lattice

SchrodingersGat:
To Logic_Programmable.lib


poeschlr:
These have invisible power pins. I will merge it as is, but will create a pull request that does fix that particular problem.

PR #46, opened 2017-12-10
Memory_NVRAM cleanup

lorem-ipsum:
This series updates the entries in the Memory_NVRAM library to fix the existing KLC issues.

MB81461ZIP was not updated as it has no datasheet and I couldn't find any (from a manufacturer). I would just delete it if it is ok.

nvram


SchrodingersGat:
Thanks! These changes are sorely needed :)

PR #17, opened 2017-11-30
Converter_DCDC_ACDC

antoniovazquezblanco:
Work in progress.

Issues:
- [ ] FSDH321L filters & footprint
- [ ] FSDL321L filters & footprint
- [ ] FSL136MRT filters & footprint
- [ ] FSQ0565RSLDTU filters & footprint
- [ ] FSQ0565RSWDTU filters & footprint
- [ ] FSQ0565RQLDTU filters & footprint
- [ ] FSQ0565RQWDTU filters & footprint

PR #47, opened 2017-12-11
Transfer dc-dc

SchrodingersGat:
Moved dc-dc library across


poeschlr:
Isn't @antoniovazquezblanco working on this in PR https://github.com/KiCad/kicad-symbols/pull/17?

antoniovazquezblanco:
I will get back to it in a couple of weeks. I will pick on wherever you leave it.

SchrodingersGat:
I believe he was working on the Converter_DCDC lib which is standalone converter modules. The existing dc-dc lib is for switchmode regulators. However the dc-dc lib did also contain some which should be moved into Converter_DC-DC

antoniovazquezblanco:
I see two different libs in your comment: `Converter_DCDC` and `Converter_DC-DC`. Shouldn't we avoid this?

SchrodingersGat:
Yes, we should. Thanks for pointing that out. I have update my comment.

antoniovazquezblanco:
Given your preferred name is `Converter_DC-DC` should I open issues to rename footprint libraries to that from the current `Converter_DCDC` and `Converter_ACDC` names?

SchrodingersGat:
@antoniovazquezblanco given that they are already in place, I will rename the file in my PR here. My preference is not that strong. We will need to rename the 3D models though - https://github.com/KiCad/kicad-packages3D/tree/master/Converters_DCDC_ACDC.3dshapes

antoniovazquezblanco:
My preference would be to regenerate the files with the script. I am also supposed to do that but again, not until I meet my work deadline.

poeschlr:
@SchrodingersGat could you merge the sym lib table? This merge is a bit much for the web interface.

SchrodingersGat:
Finally had a chance to fix conflicts

poeschlr:
Thanks

PR #51, opened 2017-12-12
Correct tilde in dcm files

MitjaNemec:
Graphics.dcm and power.dcm files have tilde prepended to the component name filed. KiCad issues errors when opening these libraries (e.g. 21:06:55: Alias '~Earth_Protective' not found in library: 'D:\Mitja\Plate\Kicad_libs\official_libs\kicad-symbols\power.dcm' at line 465 offset 23)

Looking at the original files from https://github.com/KiCad/kicad-library I concluded that tilde was inserted when porting libraries.

### DIFF
--- a/Graphic.dcm
+++ b/Graphic.dcm
@@ -1,174 +1,174 @@
 EESchema-DOCLIB  Version 2.0
 #
-$CMP ~Logo_Open_Hardware_Large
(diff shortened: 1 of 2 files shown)


jkriege2:
Thanks for spotting!

PR #55, opened 2017-12-13
Added STM32F413ZHTx symbol

Resaj:
STM32F413ZHTx symbol added to MCU_ST_STM32 library.

I've designed the F413 symbol basing on another STM included in the library.

This is the web of the component, where you can see the microcontroller characteristics: http://www.st.com/en/microcontrollers/stm32f413zh.html
And this is the datasheet (pins' distribution on page 47): http://www.st.com/content/ccc/resource/technical/document/datasheet/group3/f6/1d/4a/d9/f9/96/43/2a/DM00282249/files/DM00282249.pdf/jcr:content/translations/en.DM00282249.pdf

The component is very big, so I can't attach a screenshoot with good resolution.
stm32f413zhtx


Resaj:
Hi,

after actualyzing Kicad and commiting to this new repository, I receive a failed pull request. The inform marks that I've modify a lot of symbols and added one more, but I only have added one symbol, not modify any other.

Is it probably that my operating system has modify those symbols? I use Windows 8, but not for much more time...

Regards,
Rubén

poeschlr:
Could you remove the alternative functions from the pin names?
See mcu_nxp lib for an example of how this might look like. (Long pin names in reality help no one. They make it hard to generate a readable schematic.)

It might even be a good idea to split this component into multiple units. (one unit per port plus one for supporting stuff like xtals, plus one for power pins.)

Yes i know other symbols in this lib include the full list of alternative functions. But this is because nobody took the time to make the symbols more readable. (some symbols are larger than a A4 page.)

Resaj:
Hi Rene,

thanks for your comments. I'll try to follow your advice for modifying my symbol, but I can't do this for all of the STM32 library. Maybe it will casuses a failed pull request again.

Regards,
Rubén

poeschlr:
What kicad version are you using?

Resaj:
Version 4.0.7
kicad
I updated Kicad tonight to avoid some problems between version 4.0.6 and the repository.

poeschlr:
4.0.7 is ok to use. You can ignore all symbols you did not commit. I think this lib has been modified using some nightly build version. This is the reason why the line ordering (for all symbols of this lib) is different in your branch than in the master branch.

If you don't want to change other symbols you can create your symbol in a local lib of yours and use a text editor to copy the symbol it self into the pull request branch. (Not really necessary. But it would still make the work easier for us maintainers because we can more easily check if you changed something important in symbols you did not intent to change.)

bobc:
@Resaj  It looks like your line endings got messed up, probably because the .gitattributes file was wrong. Git now thinks you changed every line. Maybe if you run unix2dos on the file it will fix the line endings, but you might need to get a fresh copy of the repo.

Resaj:
Thanks for your comments! I'll try it again. I hope this problems will be solved once I migrate everything to Ubuntu, but anow I'm going to test the things you tell me.

Resaj:
Hi!
I've modified the symbol with your advices and divided it in some modules.

Also, I've been investigating about the UTF-8 codification. Windows change it to ANSI. So I opened the .lib file with Notepad2 on Windows, selected View/Show line endings and yes... line endings were "CR+LF".
I was able to change the encoding by selecting File/Encoding/UTF-8, and last the line endings are "LF" now.

If this PR causes more problems because of the encoding, I will close the branch and open another one on Ubuntu PC for finishing this symbol contribution.

ua
ub
uc
ud
ue
uf
ug
uh

PR #57, opened 2017-12-14
Microchip

SchrodingersGat:
This PR moves the existing `microchip.lib` across.

The easy to categorize symbols have been allocated.

The remaining ones are in `legacy/microchip.lib`

My strategy is to move the easy ones first which I can do very quickly. Then for the few remaining ones we will have to go through and decide on categorization.


poeschlr:
Should the I/O expanders go int interface_i2c and interface_spi? (currently you have them in the legacy lib.)

SchrodingersGat:
Interface_xxx libs are for parts that use an xxx interface to communicate with something off-board and xxx is their primary functionality. In this case the primary function is I/O expansion and not SPI or I2C. We just need to develop a good category.for these.

poeschlr:
Ok i misunderstood the note in the lib reorg discussion thread. In that case thanks

Issue #56, opened 2017-12-13
Should smart fets (intedned for switching batteries) be placed under power_management or transistor_smart (inteligent)?


poeschlr:
Some intelligent FETs have been moved into power_management via pull request https://github.com/KiCad/kicad-symbols/pull/42. I merged it to get all the other contributions in there. But i feel the question about smart fets in general still stands.

Infineon (which hosts the datasheets for internal rectifier datasheets, might mean they own them.) places all their smart (intelligent) fets into the smart fet category. (No matter their originally intended application.)
https://www.infineon.com/cms/de/product/power/smart-low-side-high-side-switches/

[ITS428L2-DS](https://www.infineon.com/dgdl/Infineon-ITS428L2-DS-v01_01-en.pdf?fileId=db3a30431c69a49d011cdcc9a1fd41df) has the exact same features as [IPS6011](https://www.infineon.com/dgdl/ips6011pbf.pdf?fileId=5546d462533600a4015355c3db3d1625) (including reverse battery protection.) The later is one of these transistors placed into power_management.

poeschlr:
@SchrodingersGat, @jkriege2, @evanshultz what do you guys think?

jkriege2:
Hi!  from the Infineon page I see that they have them in `Power | Smart Low-Side & High-Side Switches` ... why not use that? To me that makes sense, so I would put them either into `power_management` (they are used to manage power-supplies), or into a new lib `power_switch` (maybe split by low-side/high-side).

I would not expect them in transistors-libs, as they are much more involved than a "simple" discrete transistor. Also they are not simply plugin-replacements for a transistor with all the intelligence hidden, so I would rate them their own lib.

Best,
JAN

poeschlr:
`power_switch` is definitely better than power_management. Maybe `power_switch_smart` to indicate that there is some intelligence build in.

SchrodingersGat:
Should power-switch also include things like ideal diode controllers? I would imagine so.

poeschlr:
I don't think ideal diode controllers are similar enough to such smart fets to be thrown into the same lib.
One is intended to replace transistors (plus the transistor driver), whereas the other is meant to replace diodes.

I am not even sure any more that switch is the best name for smart fets. True switches (even transistor switches) allow current flow in both directions. Whereas smart fets (like fets) are designed for the current flowing in one direction.

SchrodingersGat:
[...]

PR #59, opened 2017-12-14
ir

SchrodingersGat:
Moved `ir` library (International Rectifier)


poeschlr:
According to https://github.com/KiCad/kicad-library/issues/1402:
- The audio amplifiers should go into amplifier_audio, the rest of the audio stuff into audio_misc
- driver_fet should be called driver_gate (Both names are equally valid. But we might want to fix the table in the issue or use the name we used there.)
- smart fets in power_distribution lib see issue: https://github.com/KiCad/kicad-symbols/issues/56
- maybe we should use Power_FactorController (or Correction) instead of Power_PFC

evanshultz:
> maybe we should use Power_FactorController (or Correction) instead of Power_PFC

I think this breaks the paradigm. It's a controller for a PFC supply, so I suggest `Regulator_Controller_PFC`. That sticks closely to the existing lib naming and is clear. A PFC controller and non-PFC controller are not to be confused and would not be confused by someone designing a SMPS using one.

SchrodingersGat:
* Amplifier_Audio - Fixed
* Driver_FET - I think Driver_FET is a better name, but happy to change if the consensus is otherwise.
* I put these into `Power_Distritbution` so that we can easily rename this lib if we come up with a better name. At least they are in the one place.
* Would `Regulator_PFC` be an acceptable compromise?

poeschlr:
Merge required.

About PFC lib: I suggest to use less appreciations. (I think i will not be the only person who has no clue what PFC stands for. Took me some trips to google to find this out.)

SchrodingersGat:
I think that people who are using a PFC device would know what it stands for, though.. 

I have inquired about adding library descriptions here - https://lists.launchpad.net/kicad-developers/msg32497.html

We also have other abbreviations in the library names, FET, BJT, CMOS, etc. There is at some level always going to be required knowledge.

(I had to Google PFC at some point, too).

SchrodingersGat:
I will have to restart this PR as the merge conflicts are very complicated. I will have to only transfer one lib at a time now that they are all meshing together.

Once we arrive at a consensus for the PFC lib I'll close this PR and start again.

SchrodingersGat:
Given that there are not enough PFC devices to really warrant their own lib, I'll put these under `Regulator_Controller`. If that library ever balloons to hundreds of items we can reclassify at that stage.

SchrodingersGat:
Once #71 and #73 are merged I shall resubmit this one.

SchrodingersGat:
Closing, replaced by https://github.com/KiCad/kicad-symbols/pull/78

PR #60, opened 2017-12-14
add Mini-Circuits RF Transformer (CDxxx footprints)

lorem-ipsum:
Add Mini-Circuits, unbalanced to balanced center tap RF transformers which use the Mini-Circuits [CDxxx](https://ww2.minicircuits.com/case_style/CD542.pdf) footprint. All symbols share the same housing (except height), PR for housing will follow soon.
Example datasheet: [ADT2-1T](https://ww2.minicircuits.com/pdfs/ADT2-1T+.pdf)

adt1-1wt


poeschlr:
Would it be a good idea to add the dot from the datasheet?

And if you only want to show pin numbers it might be better to disable show pin names in the symbol properties dialog and enable pin names inside.
(You then can give the pins more descriptive names without compromising the look of the symbol.)

lorem-ipsum:
I am not really sure about the dots. The only purpose I see is to distinguish the pin assignment.
I do agree that the changed display options look better.

lorem-ipsum:
I am not really sure about the dots. The only purpose I see is to distinguish the pin assignment.
I do agree that the changed display options look better. As the footprints are split into the 4 different heights, should I also split the symbol into four based on the different housings?

poeschlr:
yes split the symbols into 4 symbols that link the correct footprint. (each of these can of course have aliases for parts that differ in other things than the footprint.)

lorem-ipsum:
I did the split. Only three symbols were created as the thinest is not used by any of the transformers.

poeschlr:
Thanks

PR #66, opened 2017-12-16
added symbol library with more complete set of Jumpers & SolderJumpers

jkriege2:
- removed old solder-jumpers (GS2/GS3) 
- removed one non-present lib RFSolutions from sym-lib-table
- touched libs were resorted to new lib-format by KiCAD
- simple symbols, so they are on a 50mil grid!
- This commit touches some older symbols, so they light-up red in the check, but I will only change Jumper* and SolderJumper*!
- new footpritns will follow in the next days/weeks!

2017-12-16 15_02_19-eeschema devproject _ d__kicad_devproject

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing *NOT APPLICABLE, GENERIC SYMBOLS*


poeschlr:
Is the connector lib the best place to put these? They are not meant to connect something to the outside world like other connector symbols.

I think they might fit better in either switch or device lib. (switch because they can be seen as a one time switch, device because well they are generic symbols that don't really fit anywhere else.)

poeschlr:
Thanks jan.
These look nice.

Issue #69, opened 2017-12-16
Memory_EEPROM contains some NXP MCU


pointhi:
https://kicad.github.io/symbols/Memory_EEPROM

* for example: MC68HC11A0CC

PR #70, opened 2017-12-16
Moved HC11 microcontrollers from Memory_EEPROM

SchrodingersGat:
Fixes https://github.com/KiCad/kicad-symbols/issues/69


poeschlr:
Ok i made a funny discovery.
This lib has lib version 2.4
My stable build can open it but for some reason it does not show the content of the dcm file. (If i change the first line to show version 2.3 i can open it normally.)

Should we for now use lib version 2.3 or should i switch to nightly for reviewing the v5 libs. (Would then also mean all contributors would be required to use nightly.)

SchrodingersGat:
Let's stick with v2.3 for now at least until v5 is released. This was due to me manually creating the new library...

poeschlr:
Thanks

PR #71, opened 2017-12-16
ADC / DAC

SchrodingersGat:
This PR brings the `adc-dac` library across, splitting into multiple different libraries.


poeschlr:
Why did you put UVC3100 into analog.lib instead of analog_adc? (Should we even keep this component? It does not seem one can buy it any more.)

should AD5593R be in a different lib? (according to the description it is a DAC/ADC and GPIO.)

There are also video dac components. Should there be some sort of video standard to standard converter lib that holds such things?

SchrodingersGat:
* UVC3100 - I have removed this entirely as it is obsolete
* AD5593xx - Moved to `Analog.lib` which is the home for miscellaneous analog devices
* Moved video DAC chips to `Video.lib`

jkriege2:
1. Could you move the AD558-variants from Analog.lib to Analog_DAC.lib? These are simple DACs!
2. The devices you put into Driver_Display, I would put into `Interface_TouchScreen` or `Interface_HID`/`Interface_HumanInputDevice`as proposed in https://github.com/KiCad/kicad-symbols/pull/73
3. PT2399 should go to Audio.lib
4. I think we can savely remove TDC1408 ... I could not find any clue of a datasheet for it ... it looks like a multi-input ADC ... but there is no usablemetadata, not package, no datasheet ... + the symbol would really need a major work-over!

Best,
JAN

SchrodingersGat:
1. Ok, moved
1. The more generic `Driver_Display` is more multi purpose and matches the current scheme. Not all items have any inputs (e.g. touch screen). I think the current categorisation is fine.
1. Done
1. Done

Thanks for the in-depth feedback on these - hopefully we can speed these through and get the rest transferred across :)

SchrodingersGat:
Remember that I am making a concerted effort to get the majority of the symbols moved across. If there are only some very small issues we can raise an issue that can be fixed at a later date :+1:

poeschlr:
I will merge for now. @jkriege2 please document your suggestions for enhancements in an issue.

SchrodingersGat:
Thanks guys :)

evanshultz:
Just a couple thoughts as I'm able to hopefully contribute something conversation-worthy today:
1. I would err on the side of keeping stuff like UVC3100, even if it's not readily available and it doesn't break our library. Rochester Electronics and NTE, for example, will sell older ICs which might be of use to the hobbyist. And the hobbyist is a target user of KiCad. Furthermore, there might be a drop-in replacement available which a user could identify and continue on with their design in KiCad. Again, if this symbol because a bad apple then it may be time to prune.
1. TDC1408, I suspect, is the same as MC1408. This is sort of a hybrid part since it's like a DAC but without a fully implemented controller. I've seen it used in some purely analog applications, but it's reasonable to keep it here.
1. What defines Analog_DAC vs Digital_DAC vs Analog_ADC. This is for clarification for us as well as contributors. An ADC or DAC, by description, must have an analog and a digital section so how have the KiCad libraries been organized? Perhaps once the parts are all ported it's a good time to make a new chart to help maintain organization of the symbols in each lib?

jkriege2:
3. I think ADC/DACs are classical mixed-signal and in most (if not all cases) periphery/interface devices ... so I would rename the libs maybe to `Interface_ADC/DAC` or `MixedSignal_ADC/DAC` (the latter may be useful in its won right and basic form `MixedSignal.lib` to store stuff like ADE7758

PR #73, opened 2017-12-17
Interface

SchrodingersGat:
This is quite a large and complicated transfer. 

Moved interface.lib across.

Any weird or one-off symbols which would be placed in libraries by themselves have been placed in `Interface.lib` but otherwise they have been categorized based on device functionality.


jkriege2:
Hi!

Great work on that one :-)))))

a few notes from my side:
1. What about having a lib `FunctionGenerator.lib` or `Waveformgenerator.lib`, which could contain classical analog waveform generator (like MAX038) + DDS-ICs (like AD9834, AD9951 from this PR) ?
2. move SI9986 from interface.lib to `Driver_HBridge` or `Driver_Power` (or something like that ... we need that anyways for motor drivers, ...)
3. I would put the JoyWarrior/SpinWarrior/... into `Interface_HumanInputDevices.lib` or Ìnterface_HID.lib` ... they do not ONLY read encoder ... there are also variants for keyboards, mouses  ...
4. Maybe we could put the port-expanders (STMPE1600, TCA9534) into `Interface_DigitalIO` or `Interface_DigitalInputOutput` (ADC/DAC are in separate libs anyways) and the bus switches and hubs (TCA9544A, TCA9548AMRGER, TCA9548APWR, TCA9548ARGER) into `BusManagement_I2C`, `BusManagement_SPI`, ... We could then also put USB-Hubs into `BusManagement_USB` ...
5. Should we maybe split `Interface_LineDriver.lib` into `Interface_DifferentialLineDriver.lib` and `Interface_SingleLineDriver.lib`? (If tehre are enough of the second kind ;-) ...  maybe it could also make sense to split out the differentials for specific formats, e.g. `Interface_LVDS.lib`, `Interface_RS485.lib`, `Interface_RS422.lib`, `Interface_RS232.lib` ... (UART is very generic and could be anything with several types of protocols (Some "UARTs" in µControllers may even be reconfigured for synchronous SPI/I2C in addition to the asynchronous RS232/...)
6. Should we splot `Isolator.lib` by analog and digital? I think we have several analog isolator too in the legacy lib ...


Best,
JAN

SchrodingersGat:
I do not want to make lots of little libraries with only a few parts in them. Once we get too many of a certain type of symbol, we can split. I also find that using DigiKey categories as a rough guide works pretty well. With this in mind, what about:

1. These function generator parts could just live in `Timer.lib`. https://www.digikey.com/products/en?keywords=MAX038
1. The old library already has `motor_drivers` - https://github.com/KiCad/kicad-library/blob/master/library/motor_drivers.dcm - I say these parts are merged into `Driver_Motor`.lib
1. I will move these to `Interface_HID` as suggested
1. Is `Interface_InputOutput` sufficient? There is no need to narrow the focus only to digital I think. The bus switches / hubs should just go into `Interface` which is essentially "other" interface devices. Once again, until we reach a point where we need to split.
1. I think line-driver is good as a single library for now
1. Same with the UART library - I think this should stay together for now
1. Same for the Isolator lib.

I'll make the `Interface_HID` change right now, let's see if we can come to a fast resolution on the other issues.

poeschlr:
This needs a merge now.

jkriege2:
1. I wouldn't put them into Timer ... that does not really fit ... then better leave them in Analog until we find a better solution
2. OK
4. OK ... I added Digital as I think most of the others will end up in ADC or DAC ... but keep it more general ;-)
5. OK
6 OK
7. OK

Finally: Could you fix ther conglicts?

SchrodingersGat:
Given your comments above I do not think there is any more do do at this stage?

jkriege2:
Looks good to me and I will merge ... could you raise an issue maybe for a rename/split `Inetrface_Expansion` --> `Interface_InputOutput`?

PR #78, opened 2017-12-20
Moving ir.lib

SchrodingersGat:
Replacement PR for https://github.com/KiCad/kicad-symbols/pull/59

Notes:

The load switches are now in the "generic" Power_Management lib. We have not arrived at a consensus, let's move on and rearrange if we come up with a better solution.

PFC controllers are now in Regulator_Controller


evanshultz:
@SchrodingersGat There are a few broken footprints and one bad FPfilter. Want to fix?

I think it's helpful to have a wide range of folks with varying areas of expertise on the team. As someone who has primarily designed only high-power electronics, a PFC controller is certainly different from a generic SMPS controller to warrant it's own lib. There's a major mental separation between a controller meant for PFC (usually with a multiplier or some transition-mode detection) and a "normal" controller without those features. I'll merge, but anyone skilled in the art of discrete SMPS design will see an obvious reason to keep them separate. We must have at least a half dozen.

SchrodingersGat:
If we get more than a handful PFC controllers then I'm happy to split them. But remember we are now above 100 symbol libraries...

SchrodingersGat:
Fixed footprints - OK to merge now?

SchrodingersGat:
Bump - any further changes or can this be merged?

poeschlr:
If you fix the merge conflict then yes.

SchrodingersGat:
Ok, conflicts have been addressed

SchrodingersGat:
Bump - OK to merge this now?

poeschlr:
Thanks

PR #80, opened 2017-12-20
Added the VISHAY IRF740 N Channel FET

johnfwhitmore:
Added the VISHAY IRF740 N Channel FET

I've run .../kicad-library-utils/schlib/checklib.py Transistor_FET.lib

Checking symbol 'IRF740':
  Violating S3.1
  Violating S3.2
  Violating S3.6
  Violating S6.3

But doing a pull request will give me a better description of what the problems are with the component. To be fair a lot of other components in that lib give a lot of the same errors.

Datasheet is - http://www.vishay.com/docs/91054/91054.pdf

screenshot from 2017-12-20 14-52-19


johnfwhitmore:
One of the reasons why I'm failing Travis CI is "Pin offset (0) should not be below 20mils" I can't change that value because I've unchecked the "Place pin names inside" option If you uncheck that then the offset get reduced to 1 and I can't change it. I can check that check box to put the pins inside but the pin names, G,S &D are then all stacked on top of each other. Do that, change offset to the required 20 and then uncheck the checkbox to move the names outside again and it reverts to "1" again failing the test.

johnfwhitmore:
There two of the remaining errors which I'm confused by. I've tried to search on the problem with Alias but can't see what I'm doing different, as compared to the 2N7002 device.

My footprint is from the older footprint library but at present there doesn't appear to be a TO-220 in the newer footprint library?

  Violating S5.1
    filename Specified footprint library does not exist
     - Footprint library 'TO_SOT_Packages_THT' was not found

  Violating S6.3
    Part meta-data is filled in as appropriate
    ALIAS IRF740PBF has metadata errors:
     - Missing all fields on 'Properties > Description' tab
    ALIAS IRF740B has metadata errors:
     - Missing all fields on 'Properties > Description' tab
    ALIAS IRF740BPBF has metadata errors:
     - Missing all fields on 'Properties > Description' tab

jkriege2:
Hi!

about the different errors:
Checking symbol 'IRF740':

```
  Violating S3.1
    Origin is be centered on the middle of the symbol
    Symbol not centered on origin
     - Center calculated @ (-50, 0)
```
This can be ignored for small components


```
  Violating S3.6
    Pin name position offset
    Pin offset ouside allowed range
     - Pin offset (0) should not be below 20mils
```
This can be ignord (a warning anyways)


```
  Violating S5.1
    filename
    Specified footprint library does not exist
     - Footprint library 'TO_SOT_Packages_THT' was not found
```
The library has been renamed to `Package_TO_SOT_THT` Footprint is `TO-220-3_Vertical`

Please fix the latter two things and I can merge!

Thanks,
JAN

```
  Violating S6.3
    Part meta-data is filled in as appropriate
    ALIAS IRF740PBF has metadata errors:
     - Missing all fields on 'Properties > Description' tab
    ALIAS IRF740B has metadata errors:
     - Missing all fields on 'Properties > Description' tab
    ALIAS IRF740BPBF has metadata errors:
     - Missing all fields on 'Properties > Description' tab
```
You need to add metadata to the DCM-file for each ALIAS ... they are not inherited!

johnfwhitmore:
Package_TO_SOT_THT?  Oops hadn't correctly rebased my fork of footprints!

I've actually removed those aliases. I'm not sure about them. To be honest I only got the IRF740B as I bought one of these from farnell and that's what they had called it. They also used IRF740BPBF but the PbF just signifies that it's lead free.

johnfwhitmore:
Oops I removed the aliases but kicad didn't. Have to force it!

jkriege2:
Nice work!

thanks,
JAN

PR #81, opened 2017-12-20
STM32F413ZHTx symbol added

Resaj:
STM32F413ZHTx symbol added to MCU_ST_STM32 library.

This is the web of the component, where you can see the microcontroller characteristics: http://www.st.com/en/microcontrollers/stm32f413zh.html
And this is the datasheet (pins' distribution on page 47): http://www.st.com/content/ccc/resource/technical/document/datasheet/group3/f6/1d/4a/d9/f9/96/43/2a/DM00282249/files/DM00282249.pdf/jcr:content/translations/en.DM00282249.pdf

ua

ub

uc

ud

ue

uf

ug

uh


poeschlr:
I assume this is the replacement for:
https://github.com/KiCad/kicad-symbols/pull/55

Resaj:
Yes, I preferred close the other branch and actualyzed the repository for avoiding problems with the encoding modifications with Windows 8.1

In fact, I had to do copy-paste with a text editor because if I save the library modifications on Kicad, the program substitutes all endlines by ANSI format.

poeschlr:
Sorry for the late reply. (We maintainers where a bit busy with the whole lib transfer stuff)

Could you stack the vdd and vss pins?
Rule http://kicad-pcb.org/libraries/klc/S4.3/ describes what you should consider.

Resaj:
I've just stacked the pins. That's the new image of the last module.

sin titulo

poeschlr:
You forgot to set the line thickness for the outline rectangle in unit A

Resaj:
I'm trying to delete the last commit. While I changed the thickness, the lines' ending changed again.

Resaj:
I think there are some conflicts between my repositories on Ubuntu and Windows. I tried to solve them but reversing commits is difficult once I've pushed them.
Do you think it can be solve easily? If not, I think the best way is deleting my repositories, fork the principal and remake the branch again.

Resaj:
I was able to solve the conflicts. I'm going to change the rectangle outline thickness now.

poeschlr:
thanks

PR #84, opened 2017-12-20
Regul

SchrodingersGat:
Another large one, this patch moves the `regul.lib` across - mostly to `Regulator_Linear.lib` but some symbols were in the wrong place.


jkriege2:
Hi!

looks mostly good, I just spotted three problems:
1. There is no DCM-data for MAX874 and MAX872 in ` Reference_Voltage.dcm`
2. LTC3406B-1.2 is missing package in description
3. MIC4690 was already present in `Regulator_Switching.lib`.Now it exists twice and you crumbled the description (package is missing, ..)

... for all three of them: Where did they come from? They were not present in regul.lib ...

Best,
JAN

SchrodingersGat:
Not sure what happened with the duplicate symbol. Should be fixed now

jkriege2:
Hi!

please have a look at the files changed tab ... the MAX874 is still cripled (pins/FPFilters missing) compared to the old version in https://github.com/KiCad/kicad-library/blob/master/library/references.lib and https://github.com/KiCad/kicad-library/blob/master/library/references.dcm. Could you revert it to the old state?

The same goes for the description of the MIC:
2017-12-22 07_31_11-greenshot

aimylios:
This PR seems to be incomplete. Is it possible that some symbols from [regul.lib](https://github.com/KiCad/kicad-library/blob/master/library/regul.lib) got lost? I was looking for the [MIC5504](https://github.com/KiCad/kicad-library/blob/master/library/regul.lib#L2495) and its variants, but can't find them. It should be part of the new Regulator_Linear library.

SchrodingersGat:
Ah, I was working of an out of date version of kicad-library on this computer. Fixed now, I've opened a new PR here - https://github.com/KiCad/kicad-symbols/pull/88

PR #86, opened 2017-12-22
Analog devices

SchrodingersGat:
PR for analog-devices library


jkriege2:
I think the ADE7758 could either go into Analog_ADC.lib or in a sensor-lib ... What do you think?

SchrodingersGat:
I moved it to `Sensor.lib` - OK to merge?

PR #89, opened 2017-12-23
Philips

SchrodingersGat:
This PR moves old `philips.lib` across


jkriege2:
- PCF8563 should go to `Timer_RTC`
- in order to stick to the semi-hirarchic approach, shouldn't RFID.lib be called `RF_RFID.lib`?

Otherwise looks fine to me (we should definitly clean-up the Interface_Expansion (pull out ADC/DAC-combi, I2C-hubs ...), but that can be done in a second step later.

SchrodingersGat:
Fixes made as suggested. Thanks

PR #91, opened 2017-12-23
Transferred xicor

SchrodingersGat:
All into `Potentiometer_Digital.lib`


jkriege2:
Hi!

looks good the only thing I stumble upon is X9258 which is present twice: Once as a single-symbol and once with units (one for control&power and four for the pots). Do we want to keep both? I would vote to remove X9258-PART ... as it would be the only digitla pot that is split i this way ...

If you want to keep it, let me know and I will merge.

SchrodingersGat:
Fixed, thanks

jkriege2:
Hi @SchrodingersGat  could you fix the conflict, so I can merge?

Thanks,
JAN

SchrodingersGat:
Fixed conflicts

PR #92, opened 2017-12-23
Motor drivers

SchrodingersGat:
This PR moves the motor drivers across:

**Old Library:** motor_drivers
**New Library:** Driver_Motor


SchrodingersGat:
Only remaining KLC errors are for footprint filters, these need to be addressed in greater detail than I am willing to spend here.

Issue #25, opened 2017-12-05
Issues regarding micro controller libs


poeschlr:
- [ ] Some have very long pin names. Remove alternative functions from them.
- [ ] Some footprint fields don't include the lib name

evanshultz:
From https://github.com/KiCad/kicad-symbols/pull/10#event-1372553510: Missing FPfilters, incorrect footprint fields, missing DCM entries (at least some of these are because the DCM still has the /), etc.

poeschlr:
Added to the list.

antoniovazquezblanco:
Every library has been checked with the library utils for rules S5.1 and S5.2 and the corresponding issues have been created in this repo. That should mark the last three bullet points out of this issue.

As for task number 1, should it be prebooked for v6?

evanshultz:
If the symbols aren't touched, but only named shortened, then schematics don't break. I vote for consistency of pin names if any changes are made by renaming pins to be uniform across all libraries, and then shrinking/moving/changing symbols only at a later date. But IMO doing anything now isn't bringing big value to the library for users as opposed to other librarian tasks. Just my feeling.

On this topic... It may not be worth the effort to do anything manually, though. I believe we should be able to extract all the symbol information from the existing libraries (we will assume for now it is all correct) and then generate YAML files for the library which would then be inputs to a script-based symbol generator. If we only do this work when we are ready/able to change symbols we shouldn't be wasting any effort.

PR #83, opened 2017-12-20
Cleaned up micro controller lib symbols

DerekGn:
Cleaned up symbols in following libraries

- MCU_Microchip_PIC10.lib
- MCU_Microchip_PIC12.lib
- MCU_NXP_S08.lib

from issue https://github.com/KiCad/kicad-symbols/issues/25

PIC libary example:

image

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [X] An example screenshot image is very helpful
- [X] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #96, opened 2017-12-23
Cleaned up issues for MCU_NXP_LPC.lib

DerekGn:
Cleaned up build issues for MCU_NXP_LPC.lib from https://github.com/KiCad/kicad-symbols/issues/25

image


jkriege2:
Hi!

THanks for the contribution. I had a quick look at it ... is there a reason why you removed the footpritn fields? Are these devices meant for different footprints each? If there is only one footprint per device, set the footprint field (see http://kicad-pcb.org/libraries/klc/S5.1/ and http://kicad-pcb.org/libraries/klc/S5.2/) to point to that footprint AND the footprint filter (FPFilter) to match that footprint (+possible variants, like _HandSoldering ...). An example:
LPC1102UK pointed to `Package_CSP:WLCSP-16_4x4_2.17x2.32mm_P0.5mm`, which was renamed to `Package_CSP:WLCSP-16_4x4_B2.17x2.32mm_P0.5mm`, so the footprint field should be set for the second (newer) version. 
The FPFilter then does not have to contain the package lib. In the case above it should be simply `WLCSP*4x4*B2.17x2.32mm*P0.5mm*`.

Could you modify your PR accordingly?

Thanks,
JAN

DerekGn:
Ok I missed the fact that the FP should be set with the FP filter. I will fix and update PR.

jkriege2:
Hi!

nice work, thanks!
One last thing: Could you remove the pin-count 16 from the WLSCP-package FPFilter?

Thanks,
JAN

PS: Could you also take a second look at your older work there: https://github.com/KiCad/kicad-symbols/pull/83

DerekGn:
I am working on #83 at the moment
Part 1 here https://github.com/KiCad/kicad-symbols/pull/97

you mean change this WLCSP*16*4x4*B2.17x2.32mm*P0.5mm* to this WLCSP*4x4*B2.17x2.32mm*P0.5mm*?

jkriege2:
yes, that's what I meant. 

Pin count is not a "problem" ... we distinguish two cases here:
1. Symbols that have all pins (also NC pins) have the same number of pins in the symbol as are found in the package. Therefore KiCADs own pin-count filter is used to match out the correct footprint (e.g. for `DIP*W7.62mm*` will match all naroow DIPs, but the pin-count filter will narrow it down to say a DIP-8). Tehrefore pin-count is not necessary
2. In contrast there may be symbols where you don't place all pins (e.g. a reference diode in DIP-8, as you really cannot place all those 6 NC pins ...). Then we need the pin-count in the filter as the pin-count filter in KiCAD does not work for such cases!

jkriege2:
PS: The KLC is not that specific about this, it just states to include all "dimensional information" ...
PPS: Bear in mind: This is one of the less important rules in KLC ;-)

DerekGn:
Right that all makes sense I had forgotten about KiCADs own pin-count filter. So there is no need set the pin count explicitly in the Fpfilter. I will fix.

jkriege2:
Thanks,
JAN

PR #97, opened 2017-12-23
Updated MCU_Microchip_PIC10.lib and MCU_Microchip_PIC10.lib

DerekGn:
Removed device name from description
Added footprint
Updated footprint filter

Added footprints missed in #83


jkriege2:
great work!

Thanks,
JAN

PR #101, opened 2017-12-24
Fix MAX453 according to KLC

jcyrax:
Fixed symbol of MAX453 to KLC. This device is obsolete so maybe it even shouldn't be added to library.
But if you do, hope this will be better.

Test fails on S4.4, but this is because op amp output type is marked as output and not power output. Pin name is VOUT (as in datasheet). I decided to leave it output.

Datasheet: [https://datasheets.maximintegrated.com/en/ds/MAX452-MAX455.pdf](url)
symbol
descr
filter
prop

Alias is empty.
No associated footprint - comes in DIP and SOIC

------------


jkriege2:
Looks good, thanks for that contribution. Could you please change a few minor issues, before I can merge:
1. please add the packages (`, DIP-8/SOIC-8`) to the description, separated with a comma
2. The FPFilters should also include the dimensional information, i.e. `DIP*W7.62mm*` and `SOIC*3.9x4.9mm*Pitch1.27mm*`

Thanks ... 
JAN

jcyrax:
Regarding 2. there is a problem that, datasheet has no exact description of package sizes. Only DIP and SOIC. Thats why I left it without sizes.

jkriege2:
The sizes I gave above are for the standard footprints, which usually fit the SOIC-8 and DIP-8 packages as they are pretty standardized. Also we use these for most (if not all) other Maxim ICs in DIP8/SOIC8

best,
JAN

jkriege2:
Thanks,
JAN

PR #109, opened 2017-12-27
Texas

SchrodingersGat:
This is quite a substantial PR, moving across the contents of `texas.lib`.

managed to find a home for all of them!


poeschlr:
I will merge this and create a pull request to fix the errors in the footprint assignment. (A lot of them just point to the wrong lib.)

PR #110, opened 2017-12-28
Add ReflectiveOpto sensors (2 pin orderings)

MatthewStephenson:
Couldn't find a symbol for a reflective opto sensor, so I've made some. It strikes me as the sort of component which ought to have a generic symbol, but there are many different pin orderings. So I'd like to propose the following symbol, with variants according to the order of the Anode, Cathode, Collector, and Base.

image

I've included two variants in this PR - AKEC (Anode, Cathode, Emitter, Collector) - for example for the commonly used [QRB1114](https://www.sparkfun.com/datasheets/Sensors/QRB1114.pdf) device, and an AKCE version, for example for the [Kingbright KRC011](http://www.kingbright.com/attachments/file/psearch/000/00/00/KRC011(Ver.15).pdf). I can add more if the generic symbol approach is adopted.

As always, comments and suggestions welcome :)


poeschlr:
The specialized libs should not hold generic symbols. They are meant for fully specified symbols (symbol name = manufacturer part number plus footprint field set to the correct footprint.)

Generic symbols should only be in the Device lib.

MatthewStephenson:
Thanks, I didn't know this. I've moved them to the Device lib.

The Travis check is failing only because there's no documentation file - I think that's right for generic symbols.

poeschlr:
Fill it with a single "~" character. This symbolizes that it is intentionally left blank.

jkriege2:
I'm not sure whether a generic symbol for this class of devices makes sense ... Most comparable devices have specific symbols in the special opto lib. I would rather have the reflective coupler you use as a full-gledged device with correct footprint and symbol.

The main reason is that I'm not aware of any standardized pinout and pin-numbering. To cover all variants we would need 4*3*2=24 variants + I know of reflective couplers that have K and E tied together, i.e. only 3 pins, so that would add another 6 variants ...

MatthewStephenson:
Ok then. So just to be clear, is Sensor_Optical the right library to put a specific symbol in?

jkriege2:
yes

MatthewStephenson:
Closing this PR now, will submit a new one for the specific part.

PR #113, opened 2017-12-28
Fix footprint errors for texas symbols

poeschlr:
Fix for travis errors reported in https://github.com/KiCad/kicad-symbols/pull/109


poeschlr:
Ok my kicad version did change the line ordering compared to the current state of the lib. This means more errors are reported. I will fix them as well.

poeschlr:
We don't have a correct footprint for the SSOP-20 FT231XS and FT221XS symbols. (I removed the footprint field and the filter for now, will open an issue about this.) I think the correct footprint name would be Package_SSOP:SSOP-20_3.9x8.7mm_P0.635mm (same body size as ssop-24)

jkriege2:
Then I would say: set the footprint field accordingly with the new name and we should create the footprints for it ... If I find some time next week, I can do that.

poeschlr:
Done

jkriege2:
added missing footprint: https://github.com/KiCad/kicad-footprints/pull/206

PR #115, opened 2017-12-29
Transfer linear lib

poeschlr:
LF398* : sample and hold unity gain amplifier, Is Amplifier_Buffer buffer ok for this?

rf mixer LT5560 -> RF_Mixer ? (or should it be in rf_misc or Amplifier_RF?)

NE567 is Amplifier_Audio ok for this? (It is a tone decoder/PPL)

LTC4414 should this be in Power_Management? (for now in legacy/linear)


poeschlr:
Branched from the wrong branch.

PR #116, opened 2017-12-29
Tansfer old linear lib

poeschlr:
LF398* : sample and hold unity gain amplifier, Is Amplifier_Buffer buffer ok for this?

rf mixer LT5560 -> RF_Mixer ? (or should it be in rf_misc or Amplifier_RF?)

NE567 is Amplifier_Audio ok for this? (It is a tone decoder/PPL)

LTC4414 should this be in Power_Management? (for now in legacy/linear)


poeschlr:
OPA330xxYFF: There is no footprint in our lib that fits this part. (5 pin dsbga)

jkriege2:
Hi!

I would put the Sample&Hold directly into Analog.lib (for lack of a better lib ... it's also a buffer amplifier, but that's not the main function, I think) and the NE567 maybe into Timer.lib (which seems t currently be the place for anything that is remotely a timer/oscillator/...)? I will maybe try and find a better split for both libs lateron.


OPA330xxYFF: Should we simply keep it and add the footprint later? Maybe I will have time during january to create it ...

LT5560 RF_Mixer seems fine

LTC4414 OK, I would say.

Best,
JAN

poeschlr:
Moved symbol for the sample and hold parts over to the analog lib as requested.

OPA330xxYFF is still in the lib and also has the footprint field pre filled. So if somebody makes a footprint it should work without changing the symbol. (Well if i have the footprint name correct.)

jkriege2:
nice work!

Thanks,
JAN

PR #120, opened 2017-12-31
Transfer old audio lib

poeschlr:
Well as the title says. This PR does transfer the audio lib.

3 symbols are in legacy. The matched transistors are marked as obsolete by analog devices, the transformer does not have any ordering information attached to it.

The LM386 was duplicated in the audio lib and linear lib. I chose to keep the version from the linear lib. (looks nicer i think.)

The symbols i moved into the audio lib don't really merit a separate lib at this time. (If we get more similar components into our lib than we can think about splitting them off from the audio lib.)
The symbols are:
  - Equivalizer
  - multi channel volume control (could also fit into amplifier audio)


jkriege2:
Hi!

I think the trafo is nearly generic, but we have to work out a way to represent trafos generically in a better way anyways, so I would remove it also from the legacy libs.

Why not move the transistors to Transistor_BJT ... even if they are obsolete they might be of value to someone + the symbols are in good shape (except the FPFilter should be `SOIC*3.9x4.9mm*P1.27mm*` without Pitch)?

Best,
JAN

poeschlr:
Moved the transistors as requested.
Now there are much more KLC errors reported. It seems most BJT transistors are missing the datasheet. But this is something for a later date.

jkriege2:
What about the transformer?

poeschlr:
Done

jkriege2:
Thanks,
JAN

PR #121, opened 2017-12-31
Add Kingbright KRC011 Reflective Opto Sensor

MatthewStephenson:
This is the Kingbright KRC011 Subminiature Reflective Opto sensor.

Datasheet: http://www.kingbright.com/attachments/file/psearch/000/00/00/KRC011(Ver.15).pdf

image

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/204

As always, comments and suggestions welcome.


jkriege2:
Thanks for your contribution. 
Could you please change a few minor things:
1. rename to simply `KRC011` (footprint name should stay as it is! We have different naming conventions for symbols and footprints)
2. The footprint field should also contain the lib, i.e. `OptoDevice:Kingbright_KRC011`

MatthewStephenson:
Thanks, I clearly don't do this often enough, as there's always something I forget, however carefully I read the KLC. Have fixed both things - travis check is now failing because the footprint is not there yet.

jkriege2:
Great work!

Thanks,
JAN

PR #122, opened 2018-01-01
Transfer opto lib

poeschlr:
This does transfer the opto lib from the old repo.

I created the following new libs:
- Interface_Optical
  - Holds IR and other optical interface parts.
- Sensor_Proximity
  - Currently only holds optical proximity (optical interruption) sensors
- Diode_Laser
- Relay_SoldiState
  - Holds mosfet and triac based solid state relays


jkriege2:
Sensor_Proximity  does not really sound natural for photo interrupters (maybe for reflective couplers ...) ... what about putting them into `Sensors_LightBarrier` or `Sensor_PhotoInterrupter` (less general)?

poeschlr:
Yea my idea was to find a name that could be used for photo interrupters and proximity sensors (photo, capacity, induction, ...) Because they are kind of similar. (They detect the presence of some external part)

jkriege2:
Hmmm ... also makes sense ... let's stick to the names you chose for now?

One more thing: Could you move the `TSSP58*`-parts from ` Interface_Optical` to `Sensor_Proximity'? I think that's where they belong ;-)

Also there are several problems with footprint-names, as the symbols still use the old library nomenclature.

And Please add the new libs to the `sym-lib-table`:
- Library 'Diode_Laser' 
- Library 'Interface_Optical' 
- Library 'Isolator_Analog'
- Library 'Relay_SolidState'
- Library 'Sensor_Proximity'

Finally, could you please move `KRC011`, which I merged today into `Sensor_Optical` to the new `Sensor_Proximity`?

Best,
JAN

poeschlr:
Remaining travis errors are:
  - missing footprint (not yet merged)
  - pins not on 100mil grid (small devices, should be ok)

PR #125, opened 2018-01-02
Transfer of the nxp lib

poeschlr:
Transferring the 6 symbols from the nxp lib. (4 interface expanders, 2 led drivers.)

PR #126, opened 2018-01-02
Fix travis errors found when transferring the nxp symbols

poeschlr:
This does fix the travis errors found in https://github.com/KiCad/kicad-symbols/pull/125
(Wrongly assigned footprints and visible datasheet fields)


poeschlr:
Remaining travis errors are about not yet merged footprints

PR #136, opened 2018-01-06
rfcom

SchrodingersGat:
This PR pulls across all symbols from the old `rfcom` library


pointhi:
there are some bluetooth modules in RF as well as Lora/Sigfox and GSM which could be moved

SchrodingersGat:
@pointhi can you point out which ones, exactly?

pointhi:
For bluetooth for example ```SPBTLE-RF```

SchrodingersGat:
Ok, these have now been moved. Thanks for spotting

SchrodingersGat:
I have opened an issue here - https://github.com/KiCad/kicad-symbols/issues/149

PR #137, opened 2018-01-06
Video

SchrodingersGat:
This PR moves all symbols from the old `video` lib into the new `Video` lib


SchrodingersGat:
I have opened an issue here to deal with the various KLC issues - https://github.com/KiCad/kicad-symbols/issues/138

PR #140, opened 2018-01-06
Add Altera Max10 FPGA.

krtkr:
Hi all!

Here are Altera Max10 FPGAs. Check them out.

The symbols are automatically generated. Check them carefully.
1. I have grouped IO pins to IO banks and order by pin number. From my experience it is most useful way to create FPGA symbols.
2. I have removed 'NC' pins from symbols.
3. Tried to keep symbols small separating them for huge FPGA (256+ pins).

Well, I do **not** like current symbols. Same for Max2 series I add recently. Currently KiCAD does not allow symbols of different size due to [a bug](https://bugs.launchpad.net/kicad/+bug/1334502). Once it has been fixed I would like to regenerate all components in more traditional way:
1. Separate unit per IO pins of IO bank
2. Separate unit for power and system (JTAG etc) pins

Anyway, until now I made this. Here are some screen shots (if you need all of them, ping me in comments):

10m02dcu324-a
10m02dcu324-b
10m02dcv36
10m04dau324-a
10m04dau324-b
10m16daf484-a
10m16daf484-b
10m16daf484-c
10m16daf484-d
10m40dcf672-a
10m40dcf672-b
10m40dcf672-c
10m40dcf672-d

------------


jkriege2:
What about stacking power pins? THat should make the symbols a bit smaller...

bobc:
In order to workaround the field placement issue, I put the ref/name in the center of the symbol. This works well for large symbols that have plenty of space in the middle.

poeschlr:
Is the source for the pin number to pin name assignment from an official document or did you hand assign them?

krtkr:
Hi all! Sorry for long response, was busy with other task. I do kicad contributions in my free time...

krtkr:
@jkriege2 
> What about stacking power pins?
Though KLC suggest pin stacking for footprints, I would like to avoid this.

My concerns are following:
1. when I do a scheme with such huge FPGA I want to know actual number of power pins. This is important while usually I place same number of decoupling capacitors on power lines. Pins count information on schematic saves time. Otherwise I would have to look into documentation searching for power pins count.
2. I think that schematic should be verbose. If one have Kicad's scheme opened in Eeschema he can check that symbol has stacked power pins connected to a single power source. But what should he guess if he has only PDF version of the scheme? That symbol miss some power pins and they are not connected? Of he should know that exact Kicad symbol has stacked power pins?

Anyway, if community insist on stacking power pins I will do it, though I think it is better to have larger but verbose symbols.

krtkr:
@bobc 
> I put the ref/name in the center of the symbol
Hey! Good suggestion, though I remember I tried this approach [here](https://github.com/KiCad/kicad-library/pull/1418). But I got following comment:
> Move the reference and value field to the outside. (reference on top, value on bottom)

Anyway, for time of my absence at Kicad there is a [fix](https://bugs.launchpad.net/kicad/+bug/1334502) for the bug got committed. I will monitor it's status. I will rework symbols if the fix got accepted in new release.

krtkr:
@poeschlr 
> Is the source for the pin number to pin name assignment from an official document or did you hand assign them?

I used pins list from official documentation. Pin names are generated from fields in the official pins table. Generator and a copy of pins tables are available at [altera_kicad_gen](https://github.com/krtkr/altera_kicad_gen).

Pins are groupped by power bank and sorted by pin number. I have tried to use original pins order and it drove me crazy.

evanshultz:
@krtkr 
- The decision to stack pins, regardless of what you read KLC to be saying, happened a long time ago. Pins have been stacked for quite some time and no decision has been made to not stack. I do understand what you're saying about decoupling. So... please stack.
- After stacking, shrink the symbols is possible and it doesn't clutter things.
- Put the ref des and MPN within the symbol if the symbol isn't cluttered. These all look fine.
- So you didn't hand enter the pin names? If you extracted from the datasheet, we can save a ton of time by not checking each pin individually.

Would you be willing to contribute your symbol generator to the official repo? Surely we can find a nice place to put it where it can hopefully get more use in the future.

poeschlr:
The generator would fit nicely into the kicad-library-utils repo:
https://github.com/KiCad/kicad-library-utils/tree/master/schlib/autogen

evanshultz:
@krtkr 
We've recently started allowing non-stacked power pins if the datasheet gives a reason, such as decoupling on each power pin. This is what you mentioned above. I don't see specific info in the Max10 datasheet, but the PDN calculator from Intel will help to pick the capacitance and exact capacitors to achieve a desired impedance vs freq characteristic. That's good enough for me.

Here are a few other things: for now
- The description does not include all the nice info at https://www.altera.com/content/dam/altera-www/global/en_US/pdfs/literature/hb/max-10/m10_overview.pdf for each PN. Put the pin number at the end of the description and remove the work "package". I suggest something like `FPGA, MAX 10, 2K logic elements, Dual Supply - Compact Features, UBGA-324` for 10M02DCU324.
- For other NSMD footprints, we did not put _NSMD at the end of the footprint since NSMD was preferred and so the official KiCad footprint was the preferred one. Where is footprint info for these parts located? I didn't see it at https://www.altera.com/products/fpga/max-series/max-10/support.html.

evanshultz:
@krtkr 
Are you interested in continuing this PR?

krtkr:
@evanshultz
* Good. I will stack pins then.
* I took pin names/numbers from text documentation. I did some random pins check. I hope this way we could find out errors in my parser if any.
* My generator is available at my github repo. Well, I am willing to contribute it, but it does not look really nice for now. It might require some refactoring to be really used.

krtkr:
@evanshultz 
* Packaging information for Altera devices can be found here: https://www.altera.com/support/literature/lit-index/lit-pkg/package.html?family=MAX_10
* It is a bit tricky to find footprint recommendations for Altera devices. Check out this appnote: https://www.altera.com/content/dam/altera-www/global/en_US/pdfs/literature/an/an114.pdf

krtkr:
@evanshultz 
What about NSMD? Do you want me to remove this postfix? I add a few packages https://github.com/KiCad/kicad-footprints/pull/152 with this postfix. Do you want to rename them?

evanshultz:
Thanks for coming back to this!

If you can stack the pins, put the ref des and value inside the symbol, etc. I'll work on a real review.

Correct, you do no need to put `_NSMD` in the footprint name. It is clearly stated on page that this is preferred for the footprints on page 6.

krtkr:
@evanshultz 
Hi again!

I was looking at KLC for NSMD. Here it is stated to be required (search for 'Ball Grid Array Packages'):
http://kicad-pcb.org/libraries/klc/

I am a bit confused here. Maybe official KLC have moved somewhere else? I used KLC from link above.

evanshultz:
@krtkr 
I'm not quite sure what's you're asking. It's my fault not to understand.

The datasheet always wins over KLC. And the datasheet says use NSMD pads.

Besides, we prefer NSMD if given the choice because it gives a tighter tolerance and more consistent PCBs. So due to that preference, there is no need to specifically call out NSMD in the footprint name.

Does that answer everything? What am I missing?

evanshultz:
@krtkr 
There are now branch conflicts. Will you be coming back to this PR soon?

antoniovazquezblanco:
Ping @krtkr!

antoniovazquezblanco:
I fixed the conflict.

At http://kicad-pcb.org/libraries/klc/#_ball_grid_array_packages the keyword NSMD seems to be mandatory in the footprint name.

Does this solve your last comment @evanshultz ?

Thanks!

evanshultz:
@antoniovazquezblanco 
This may be a place where KLC was written at one time and things changed later or information was added to KLC without expert input. I see that, yes, KLC does show that but it needs an update here.

SMD is poor footprint design and increases tolerance without any advantage. NSMD is recommended everywhere and it does not need to be added. For example, see the YAML entries in our BGA script and you will find they are all NSMD but this isn't included in the footprint name.

Also, two things from a quick look through the changed files:
1. The line for the rectangle for the symbols has weird spacing in it. Opening in KiCad and saving should fix that if it doesn't cause an error.
1. The FP filters need some work. For example, `BGA-324_15.0x15.0mm_Layout18x18_P0.8mm_Ball0.5mm_Pad0.4mm_NSMD` (let's say it's correct) should result in a FP filter of `BGA*15.0x15.0mm*Layout18x18_P0.8mm*`.
1. Many footprints are missing. Before merging these symbols, we want to include the footprints. They can be scripted now so please let us know if you're willing to contribute them.

Once the above is settled, I'll open this PR in KiCad and do another check. I didn't review all comments above to get a good feel for the current status of this PR.

krtkr:
Hi all! Sorry for a delay. A lot of stupid problems consumed all my free time for quite a bit. I guess I finally can continue contributing to the opensource.

Let me check latest KLC first.

chmorgan:
Hi @krtkr, checking in to see if you are still active with this pull request.

PR #141, opened 2018-01-06
A version of the ATmega328P-PU symbol with Arduino pin names.

ppelleti:
As before, in KiCad/kicad-library#1770:

I'm not entirely sure this even belongs in the KiCad library, so please let me know if it doesn't.

This is a symbol for the [ATmega328P-PU](http://www.atmel.com/images/Atmel-8271-8-bit-AVR-Microcontroller-ATmega48A-48PA-88A-88PA-168A-168PA-328-328P_datasheet_Complete.pdf) microcontroller, but instead of using the manufacturer's pin names, it uses the Arduino pin names.  This makes it easier when designing a circuit whose software is meant to be programmed with Arduino.

The Arduino pinout is [here](https://www.arduino.cc/en/Hacking/PinMapping168), although most of the exact pin names came from [this sticker](https://cdn-shop.adafruit.com/970x728/554-04.jpg).

atmega328p-pu_arduinopinout


ppelleti:
So, I don't know how to make Travis happy.  First it failed with:

```
  Violating S5.1
    filename
    Specified footprint library does not exist
     - Footprint library 'Housings_DIP' was not found
```

So I changed the footprint field and footprint filters to match the ones on `ATMEGA168A-PU`.  But now it fails with:

```
  Violating S5.1
    filename
    Footprint field 'DIL28' must be of the format '<Library>:<Footprint>'
  Violating S5.2
    Footprint filters should match all appropriate footprints
    Footprint filter '28DIP-ELL600' not correctly formatted
     - Does not contain wildcard ('*') character
    Minuses and underscores in footprint filter '28DIP-ELL600' should be escaped with '?' or '*'.
    Footprint filter '28dip600' not correctly formatted
     - Does not contain wildcard ('*') character
```

I'm at a loss.  What do I need to do to make Travis happy?

jkriege2:
The lib was renamed from `Housings_DIP` (plural) to `Housing_DIP` (singular)!

About the symbol itself: I'm not so sure whether we should have it, as it is essentially a doubled-device, "just" with other names that are also more volatile, as the pin-ordering is defined by software only and might change ... but then for designing with the Arduino-IDE it might make sense ... But do we then also need second variants of all the different µCs and form-factors usable together with the Arduino-IDE? I think if you use that IDE together with a normal µC and not an Arduino you will at some point anyways refer to the official datasheet and then (at the latest) you will be confronted with the real pin-names ... 

BTW: Do they sell this µC preprogrammed with the Arduino-Bootloader? Then we could name it after that?

@SchrodingersGat @poeschlr @evanshultz What are your thoughts on this?

poeschlr:
@jkriege2 said:
> The lib was renamed from Housings_DIP (plural) to Housing_DIP (singular)!

nope it was renamed to `Package_DIP`

diggit:
I think, there shouldn't be symbols modified just to match some software library.

ppelleti:
> About the symbol itself: I'm not so sure whether we should have it, as it is essentially a doubled-device

Yeah, I wasn't sure myself, so I wanted to start that discussion.  (Though I would have preferred to have the discussion on the original PR.  😀)

> but then for designing with the Arduino-IDE it might make sense

Yes, that's why I made it.  I designed a board that I was going to program with the Arduino IDE, and it was helpful to be able to look at the schematic and see what pins the wires connect to, using the names I would actually use in the software.  Otherwise I would have to constantly look up the pin names, which seemed like a pain.

> BTW: Do they sell this µC preprogrammed with the Arduino-Bootloader? Then we could name it after that?

Adafruit does sell [a preprogrammed version of this chip](https://www.adafruit.com/product/123), although I'm not sure that helps any with the naming.

Misca1234:
I am not sure this is the right way to go, ATmega328-PU already exist in the library

ppelleti:
No, I don't object.

PR #145, opened 2018-01-07
Added Microchip MCS3142

swbr:
Add Microchip MCS3142.

This IC is used by many car keyfobs. When pressing one or more  of the four buttons, it transmits the button information, a serial number and a hopping code as manchester or pwm encoded AM or FM signal in one of the ISM bands. Beneath buttons, battery and antenna, only two crystals and two caps are needed. Optionally, an LED can be attached.

The IC has IN and OUT pins for CTRL and DATA, but the datasheet does not mention them, except showing in a schematic that they should be tied together.  So, they were grouped to the right here for a clear schematic.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/40001747A.pdf

mcs


jkriege2:
Hi!

thanks for your contribution. I have a few requests:
1. please add the datasheet link to the DCM-file, i.e. with the menu entry Symbol | Properties (not as a field!)
2. you could stack the power-pins to clean up the symbol a bit more.
3. The footprint link is wrong. Note that some libs have been renamed in the new footprint-repo: https://github.com/KiCad/kicad-footprints The contents of Housings_SSOP is now found in Package_SSOP (https://github.com/KiCad/kicad-footprints/tree/master/Package_SSOP.pretty)
4. You need to add a footprint filter

Best,
JAN

jkriege2:
Please also fix the merge conflict by rebasing your branch onto master (or pulling master and hand-merging)

swbr:
OK, I've done what you wrote.

Just wondering: After the last commit, I got a "failed check" from Travis CI, which is now missing completely. Is this normal?

jkriege2:
I think you first need to resolve the conflicts ...

swbr:
OK, for this last commit, I just replaced my RF_AM_FM files with those from KiCad. Next one will contain my device again.

swbr:
So I end up with this:

- Copy the last version of RF_AM_FM.* from the KiCad repo to last version of my fork.
- Commit/push to my fork ("I have no idea what I'm doing here"), which has no conflicts
- Immediately open that lib in KiCad Library Editor and insert my symbol there
- Commit/push to my fork, which results in a conflict.

I think I'll give up for now.

jkriege2:
Something went wrong ... the DCM only contains your component's data now and the lib is still resorted. Also the conflicts persist. Maybe easiest ist to open a fresh new branch and only copy in your component (in a text editor). This way there also are not so many reorganizations in the file, which makes review much easier ...

poeschlr:
It might be a better idea to make a fresh start. The only problem is that you used your own master branch as a basis for this pull request. In the future always make a dedicated branch for each pull request.

---

Now a step by step explanation about what you need to do for a fresh start:

First add the official repo as a secondary upstream repo:
`git remote add upstream https://github.com/KiCad/kicad-symbols.git`

remember to copy your symbols to some place outside your local repo. (this will make your live easier)

then get the data from the our repo:
`git fetch upstream`

and checkout its master branch for a fresh start:
`git checkout upstream/master`

start a fresh branch;
`git checkout -b [branch name]`
`git push --set-upstream origin [branch name]`


add your changes, add the git history, commit it and push it to the server
`git checkout [branch name]` # only needed if you changed to a different branch in the meantime
do your changes
`git add -u` #(or `git add [changed or new file path]`)
`git commit -m "Descriptive commit message"`
`git push`

All of this should be doable in a git gui. But the command line stuff is universal to all systems.

If you need more information about git stuff, either look directly in the git documentation for commands (simply google for example "git add", or use `git add --help`)
You can also look at the tutorials on https://www.atlassian.com/git

swbr:
Thanks, Rene for your really comprehensive posting. I will work through this and try again the next days. And I'll definitely bookmark this page for later!

swbr:
OK, I followed your guide, which of course results in a [new pull request](https://github.com/KiCad/kicad-symbols/pull/163).

Just wondering:
I added my device with kicad again, and saw that it messes up the files. It inserts the new device at the beginning of `.lib`, which is fine, but it reorders several lines of the other symbols. As far as I checked, it does not corrupt anything, it just reorders some pin lines within their symbol.
Much worse, the `.dcm` file is overwritten completely, and contains my symbol only.

For now, I've added my symbol to both files by hand, but this behavior is strange. Is this known? I'm running Kicad 4.0.6 on openSuse 42.3, which is the latest version in the repository. But the changelog does not mention something related to this.

poeschlr:
Sadly the reordering of lines seems to be a reoccurring problem. This is just how different versions of kicad handle libraries. As long as you do not use one of the really old versions of kicad that also insert a lot of whitespace characters it is ok.

If you want to avoid it, you can insert your symbol using a text editor instead of using kicad. But this is a very dangerous workflow that might result in a corrupt library if done wrong.

swbr:
Hmh, ok, but the complete removal of other devices from the `dcm`-file is a problem.

poeschlr:
yup that is a problem. It means that someone modified it with a recent nightly.
If you did not already revert the changes to the dcm file and change the first line in the lib file to say version 2.3 instead of 2.4

@SchrodingersGat could we have a temporary (until v5 release) travis check that shouts if somebody changes the lib version to 2.4 which results in it being incompatible with contributors that use the stable version or even a slightly older nightly version.

jkriege2:
closed, as this is replaced by https://github.com/KiCad/kicad-symbols/pull/163

Issue #149, opened 2018-01-08
KLC issues in RF libraries


SchrodingersGat:
The various RF libraries have outstanding KLC issues which need to be addressed

jkriege2:
fixed it here: https://github.com/KiCad/kicad-symbols/pull/150

SchrodingersGat:
Closed as (mostly) fixed

PR #150, opened 2018-01-08
fixed several KLC-issues 

jkriege2:
- wrong FPFilters/footprint-links
- `PitchXXmm` was used in several footprints and FPFilters -> replaced wit `PXXmm`
- some datasheet-links (by hand)
- some pin-placement-issues (by hand, e.g. moved NC-pins onto border ...)
- fixed some non-existent footprints
- emptied datasheet-fields
- made datasheet/footprint fields invisible
- should fix most issues of https://github.com/KiCad/kicad-symbols/issues/149


SchrodingersGat:
Nice work Jan, this fixes a LOT of issues! Thanks :)

PR #151, opened 2018-01-09
Converter_ACDC

antoniovazquezblanco:
Replaces old PR: https://github.com/KiCad/kicad-symbols/pull/17

Rework of the AC DC converter symbols.

Progress:
- [x] FSDH321
- [x] FSDH321L
- [x] FSDL321
- [x] FSDL321L
- [x] FSL136MRT
- [x] FSQ0565RQLDTU
- [x] FSQ0565RQWDTU
- [x] FSQ0565RSLDTU
- [x] FSQ0565RSWDTU
- [x] HS-40003
- [x] HS-40005
- [x] HS-40009
- [x] HS-40012
- [x] HS-40015
- [x] HS-40018
- [x] HS-40024
- [x] IRM-02-12
- [x] IRM-02-15
- [x] IRM-02-24
- [x] IRM-02-3.3
- [x] IRM-02-5
- [x] IRM-02-9
- [x] IRM-03-12S
- [x] IRM-03-15S
- [x] IRM-03-24S
- [x] IRM-03-3.3S
- [x] IRM-03-5S
- [x] KA5H02659RN
- [x] KA5H0265RCTU
- [x] KA5H0265RCYDTU
- [x] KA5H0280RTU
- [x] KA5H0280RYDTU
- [x] KA5L0265RTU
- [x] KA5L0265RYDTU
- [x] KA5M02659RN
- [x] KA5M0265RTU
- [x] KA5M0265RYDTU
- [x] KA5M0280RTU
- [x] KA5M0280RYDTU
- [x] LT1248
- [x] LT1249
- [x] LT1509
- [x] NCP1200D
- [x] NCP1200P
- [x] SG3525
- [x] SG3527
- [x] TL494
- [x] TMLM04103
- [x] TMLM04105
- [x] TMLM04109
- [x] TMLM04112
- [x] TMLM04115
- [x] TMLM04124
- [x] UC3525
- [x] UC3527
- [x] UC3842_DIP8
- [x] UC3842_SOIC8
- [x] UC3842_SOIC14
- [x] UC3842_SOIC16
- [x] UC3843_DIP8
- [x] UC3843_SOIC8
- [x] UC3843_SOIC14
- [x] UC3844_DIP8
- [x] UC3844_SOIC8
- [x] UC3844_SOIC14
- [x] UC3845_DIP8
- [x] UC3845_SOIC8
- [x] UC3845_SOIC14
- [x] UC3846X-D08-T
- [x] UCC3800
- [x] UCC3801
- [x] UCC3802
- [x] UCC3803
- [x] UCC3804
- [x] UCC3805
- [x] UCC3808D
- [x] UCC3808N
- [x] UCC3813-0
- [x] UCC3813-1
- [x] UCC3813-2
- [x] UCC3813-3
- [x] UCC3813-4
- [x] UCC3813-5
- [x] VIPer22ADIP-E
- [x] VIPer22AS-E
- [x] VIPer22ASTR-E
- [x] VIPer25HN
- [x] VIPer25LN
- [x] VTX-214-015-103
- [x] VTX-214-015-105
- [x] VTX-214-015-106
- [x] VTX-214-015-107
- [x] VTX-214-015-109
- [x] VTX-214-015-112
- [x] VTX-214-015-115
- [x] VTX-214-015-118
- [x] VTX-214-015-124
- [x] VTX-214-015-148

Related issues:
https://github.com/KiCad/kicad-symbols/issues/153
https://github.com/KiCad/kicad-symbols/issues/154
https://github.com/KiCad/kicad-symbols/issues/155
https://github.com/KiCad/kicad-symbols/issues/157
https://github.com/KiCad/kicad-footprints/issues/235
https://github.com/KiCad/kicad-footprints/issues/240
https://github.com/KiCad/kicad-footprints/issues/241
https://github.com/KiCad/kicad-footprints/issues/242


antoniovazquezblanco:
I have realised that I cannot specify different footprints for a component and its alias 🤦‍♂️ 

Am I right or just missing something?

poeschlr:
Yup that is why we have the one symbol per footprint policy in place.

antoniovazquezblanco:
Most of the KLC issues have been fixed and quality is a little bit better than the original.

Things that should be reviewed please:
- KA5M0265RTU footprint may be wrong
- I have deleted UC3846X-D08-T. I could not find it anywhere.
- MeanWell footprint filters
- I don't know if I did well in the VIPer25LN filters

evanshultz:
First, the old library names were, frankly, stupid. With the new library names, all the ICs should go to https://github.com/KiCad/kicad-symbols/blob/master/Regulator_Controller.lib or https://github.com/KiCad/kicad-symbols/blob/master/Regulator_Switching.lib (ask if you would like help).

Second, can I ask you to be more detailed about your commit logs? When you write "rework..." but the changes are updating datasheet URL or changing the default footprint that doesn't qualify as "rework" to me and makes it harder to figure out what's being done in each commit.

Third, you can leave out the pin count in the FPfilter if all pins are located on the device. This allows the pin filter in CvPcb to work. If you want to leave out NC pins due to just having a huge number of them, it's fine to include the pin count in the FPfilter. In at least some cases all pins are shown but the pin count is listed as well. That's either discouraged or disallowed by KLC.

Fourth, you've done an awesome job! I reviewed the non-module (or the IC) changes down to UC3842 and a cursory look seems good. Let's nail down the question below, move the ICs to the right library, and then get a final review. This will really help!

@SchrodingersGat @poeschlr 
We had previously written into KLC to use a "_DIP8" suffix, for example, when multiple manufacturers had the same part but with different suffixes. For example, UC3842 comes in different, and conflicting, suffixes for the DIP8 package. I don't see mention of this in http://kicad-pcb.org/libraries/klc/ and some of the commits above pick a single suffix. While this might work for one given manufacturer, it won't necessarily be correct for all compatible manufacturers. How are we handling this situation as of KLC 3?

antoniovazquezblanco:
Thanks @evanshultz!

I've tryed to do everything you suggested. Sill waiting on the UC3842 issue though.

evanshultz:
@SchrodingersGat @poeschlr 
Can you comment on my post above? Are we still using the package name as a suffix when multiple vendors sell the same generic part with different suffixes?

SchrodingersGat:
Yes I believe that this is the way that we decided to go for non manufacturer specific part numbering.

Please note that @poeschlr is now the lead on library matters, I am stepping away from KiCad for a while :)

antoniovazquezblanco:
I will be renaming the UC3842 family back to the footprint suffix. I am sorry but I will not be able to do this until the next week.

antoniovazquezblanco:
Ready for review @poeschlr @evanshultz 

Thank you for your work!

evanshultz:
@antoniovazquezblanco 
UC3846 is a valid PN, but it was shown as in DIP8 and I don't see that package. TI datasheet is at http://www.ti.com/lit/ds/symlink/uc3846.pdf.

Would you mind to reinstate and fix that symbol? It's not required, but it would be nice to avoid losing this symbol. This should also be generic because at least two vendors make this IC. Also not required, but you could add UC3847 too if you like.

Everything else looks OK, but I'll take a closer look once you answer my above request. Thank you!!

antoniovazquezblanco:
> UC3846 is a valid PN, but it was shown as in DIP8 and I don't see that package. TI datasheet is at http://www.ti.com/lit/ds/symlink/uc3846.pdf.
>Would you mind to reinstate and fix that symbol? It's not required, but it would be nice to avoid losing this symbol. This should also be generic because at least two vendors make this IC. Also not required, but you could add UC3847 too if you like.

When I was refactoring the lib I found no information about `UC3846X-D08-T` or about `UC3846` in any 8 pin package. That is the reason for me to drop it. `UC3846` cannot be an alias of any of the existing components. That would mean creating 5 symbols and 2 alias to support the family described in the datasheet which for me would be quite an amount of work right now. I would rather spend the time I have contributing to other symbols I will be using in the near future.

Sorry @evanshultz

evanshultz:
@antoniovazquezblanco 
That's fine. As I said, it wasn't required.

But there are two things I'd like you to do before I merge:
1. I see that `UC3842_SOIC14` and `UC3842_SOIC16` don't have all pins shown, but they have the pin count in the FPfilter. The original `UC3842_SOIC14` had all pins listed, and I'd prefer it goes back exactly as it was since that matches the majority of the library. Can you do that?
2. Did you add any UC28xx parts? I don't see any at https://github.com/KiCad/kicad-library/blob/master/library/ac-dc.lib. KLC S2.2 says to replace with an `x` but please go back to just the 38xx PNs as before. This will be more clear than `x8xx` in this case.

Thanks again for all your work on this!

antoniovazquezblanco:
@evanshultz 

1. I will put the NC pins back and fix the FPFilter.

2. I did not realize that I mistyped one PN. Will fix that asap.

antoniovazquezblanco:
@evanshultz  Done

evanshultz:
@antoniovazquezblanco 
Sorry for all the changes and surprise that this lib was holding parts that belong in another lib (using our new organizations scheme). This lib started out a big mess, and you've done a heroic job of improving the libraries!

Can you please review the CI output above? In Converter_ACDC.lib there are a few symbols with messed up footprints for FPfilters. Could you fix that? The warnings in Regulator_Controller.lib look OK to me.

antoniovazquezblanco:
Regarding Converter_ACDC.lib CI output:

1. When a part is atomic and the PN contains `-` or `_` symbols, should they also be scaped?
2. See https://github.com/KiCad/kicad-footprints/issues/328

poeschlr:
I would escape them in the footprint filter. (via *) This is such that the user can choose a slightly different footprints that can have "fields" in the footprint name at any position. (It can for example define differing pad dimensions)
This is the only reason why we would even need a footprint filter. It makes the lib a bit more future proof.

antoniovazquezblanco:
@evanshultz Done

evanshultz:
@antoniovazquezblanco 
There are two symbols with a FPfilter but no default footprint. Can you add the default footprint and then I will merge?

antoniovazquezblanco:
Those symbols are available in two versions. THT and SMD so I left it without a default one so that the user can assing whatever suits the application better. Are you sure I should still specify a default one?

evanshultz:
OK. I see how the footprints for those converters are named. That's fine.

Three `KA5* symbols are without the footprints mentioned at the 3 links in your first post above and also https://github.com/KiCad/kicad-footprints/issues/244. Just FYI, were you going to create those footprints?

Lastly, `VIPer25LN` rises two questions for me:
1. It seems to have a mismatch between the default footprint and FPfilter. Can you take a look at that? I don't see any discussion about this above. It seems `Package_DIP:PowerIntegrations_PDIP-8B` will fit this package but the FPfilter doesn't match this default footprint.
2. Do you know the difference between `VIPer25LN` and `VIPer25HN`? I don't see it in the datasheet (http://www.st.com/content/ccc/resource/technical/document/datasheet/52/fe/bc/a1/69/06/49/08/CD00231127.pdf/files/CD00231127.pdf/jcr:content/translations/en.CD00231127.pdf).

antoniovazquezblanco:
> Just FYI, were you going to create those footprints?

Not currently using this components so I didn't plan to create them.

> It seems to have a mismatch between the default footprint and FPfilter. Can you take a look at that? I don't see any discussion about this above. It seems Package_DIP:PowerIntegrations_PDIP-8B will fit this package but the FPfilter doesn't match this default footprint.

https://github.com/KiCad/kicad-footprints/issues/243

Forgot to remove the DIP 7 filter. Will do now.

> Do you know the difference between VIPer25LN and VIPer25HN? I don't see it in the datasheet (http://www.st.com/content/ccc/resource/technical/document/datasheet/52/fe/bc/a1/69/06/49/08/CD00231127.pdf/files/CD00231127.pdf/jcr:content/translations/en.CD00231127.pdf).

HN - 225 kHz 
LN - 136 kHz

evanshultz:
@antoniovazquezblanco 
Thanks so much for your work and patience!

With such a small difference I'd like to include both LN and HN variants, but I don't want to drag this on any longer.

antoniovazquezblanco:
Thank you very much to everyone. I hope to have helped a little bit with my contribution.

radek-senfeld:
Thank you very very much, gentlemen! Indeed it helped quite a lot!

PR #152, opened 2018-01-09
Add OPA1641 operational amplifier + fix erros in Amplifier_Operational.lib

jcyrax:
New part OPA1641 symbol.
No aliases!
http://www.ti.com/lit/ds/symlink/opa1641.pdf
symbol
descr
fp_filter
options

------------


jcyrax:
Travis fails because of other parts. My kicad version touches pin ordering. I could fix those errors if its ok to include them together with this PR.

jcyrax:
I left S3.6 because otherwise pin names are very close and otherwise symbol will become very large. This is practically for all opamps in lib.

poeschlr:
A lot of the problems you fixed might have already been fixed by the recently merged pull request https://github.com/KiCad/kicad-symbols/pull/150

For this reason your PR will need manual merging with the current master.

jcyrax:
Should I do a merge locally in my branch and then push update to my PR?

jkriege2:
yes, like that ... you can also rebase your branch onto a current master, so there is no extra commit for the merge.

Best,
JAN

poeschlr:
Just a heads up: rebase requires a forced push.

jcyrax:
OK, this I need help with. OPA330xxYFF fails because it has footprint set, but it doesn't exist. If I add only generic filter like \*DSBGA\* it throws me error that it expects exact footprint. Also part was left like this in #150 so I'm thinking there is some reason to leave it like this even if there is no footprint.

evanshultz:
@jcyrax 
Please use **Texas_DSBGA-5_0.822x1.116mm_Layout2x1x2_P0.4mm** and a matching FPfilter. I've created https://github.com/KiCad/kicad-footprints/issues/253 to capture the need for this missing footprint.

jcyrax:
Changed footprint and filter

jkriege2:
Looks good up to one slight thing: Please addd the package options to the description, i.e. add `, SOIC-8/VSSOP-8`

Then I can merge!

Thanks,
JAN

jcyrax:
Done, I somehow forgot it again.
Could you please explain a bit this rule? It says

> For symbols with a default footprint, the human-readable footprint name should be appended to the description e.g. SOIC-8

Would this mean practically every symbol needs this added? I don't quite get what is default footprint. My intuition was that if you fill footprint field then this would apply.

poeschlr:
Yes, if you fill the footprint field this does apply. And you are also right that most symbols require filling of the footprint field. Only generic symbols are allowed to leave it empty. (New generic symbols are not allowed in most libs. They should only be in libs like device, connector_* and switch.)

jkriege2:
Thanks, Jan

PR #158, opened 2018-01-09
Merge old modules.lib

jkriege2:
- Moves over the old modules.lib
- MCU modules (Arduino etz.) ended up in new MCU_Module.lib
- Two motor-drivcer-breakouts ended up in  Driver_Motor.lib

Hope this sorting is OK ... One could also think about having the MCUs in other libs (e.g. if they are specific to one MCU-type, put it in the MCU_...-lib for that type, and all Arduinos into MCU_Arduino ...

But I think that will crowd the libs and this way we get away with a minimum of extra libs. OK?

JAN


SchrodingersGat:
I agree with the organization, and conforming to the existing grouping. Can you please fix the footprint filter errors for the ST nucleo boards? Otherwise I"m happy to merge.

jkriege2:
Done!

JAN

SchrodingersGat:
Thanks :) Almost done for the symbols!

PR #161, opened 2018-01-11
Fix Generic opamps

jcyrax:
I added missing outline boxes for generic op amps in Device.lib

------------


jkriege2:
DO I see this correctly, you added a rectangle around the OPAMP? This is NOT REQUIRED for Opamps ... compare to the other OPAMP symbols throughout the lib (is an exception to KLC!)

jcyrax:
No, only for power unit.
It was like this
old
And I added:
new
Is this also not needed? Are you supposed to somehow overlay power unit?

jkriege2:
The old style was OK, because this way you can combine an OPAMP-unit with the poer-unit to a single symbol by just moving the over each other!

PR #163, opened 2018-01-12
Mcs3142

swbr:
__Since things went wrong [last time](https://github.com/KiCad/kicad-symbols/pull/145), here is a clean retry.__

> Add Microchip MCS3142.
> 
> This IC is used by many car keyfobs. When pressing one or more of the four buttons, it transmits the button information, a serial number and a hopping code as manchester or pwm encoded AM or FM signal in one of the ISM bands. Beneath buttons, battery and antenna, only two crystals and two caps are needed. Optionally, an LED can be attached.
> 
> The IC has IN and OUT pins for CTRL and DATA, but the datasheet does not mention them, except showing in a schematic that they should be tied together. So, they were grouped to the right here for a clear schematic.
> 
> Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/40001747A.pdf

mcs3142


jkriege2:
Hi!

please correct the FPFilter to simply `TSSOP*4.4x6.5mm?P0.65mm*` (we usually don't want the pin-count in there!).

Best,
JAN

jkriege2:
Thanks,
JAN

PR #167, opened 2018-01-13
Add Microchip CAN FD components MCP2517FD and MCP2558FD

menschel:
CAN FD is an update to the Contoller Area Network standard.

This patch adds schematics for Microchip devices:
1.
CAN FD controller MCP2517FD
https://ww1.microchip.com/downloads/en/DeviceDoc/20005688A.pdf
with 2 package options, where 1 package was not found in the official footprint library.

MCP2517FDx-H/JHA is VDFN-14 according to spec sheet but the official footprint library's naming seems to omit the V letter thus package name given as
Package_DFN_QFN:DFN-14-1EP_4.5x3mm_P0.65mm
This package is not yet included in the official footprint library.

2.
CAN FD transceiver MCP2558FD 
https://ww1.microchip.com/downloads/en/DeviceDoc/20005533A.pdf
with 3 package options where 1 package was not found in the official footprint library.

MCP2558FDx-H/MNY is TDFN-8 according to spec sheet but only DFN footprint was found in official footprint library, thus changed to 
Package_DFN_QFN:DFN-8-1EP_2x3mm_P0.5mm

Pin groups are adapted from predecessor devices MCP2515 and MCP2551 respectively.


SchrodingersGat:
Why did you close this?

menschel:
1. I don't approve CLAassistant to access my account data while the commit is already signed-off, so no CLA needs to be signed any more.
2. I missed adding the DFN14-1EP footprint thus the PR is not mature enough for integration

PR #169, opened 2018-01-14
Add ADF4002 frequency synthesizer

lorem-ipsum:
This adds the ADF4002 frequency synthesizer from Analog Devices. Two variants as they use different pin assignments. Pin stacking is used to maintain the the same symbol shape. [Datasheet](http://www.analog.com/media/en/technical-documentation/data-sheets/ADF4002.pdf)

One question regarding the ADF4002BCPZ variant: Analog labels the package as LFCSP. For me it looks like a standard QFN. Would you agree with this? I need to create a footprint variant anyway as the pad size does not exist.

adf4002bcpz


calebns:
QFN uses different landing pad size than the one recommended by Analog for LFCSP. Analog recommends having pads bigger than the physical leads of the package to reduce parasitics.
Please take a look on BXL file http://www.analog.com/media/en/symbols-footprints/symbols/ADF4002BCPZ.bxl
The are supposed to be ~0.3x0.8mm

Timer.dcm
```diff
@@ -36,6 +36,18 @@ K single opamp
 F http://www.analog.com/media/en/technical-documentation/data-sheets/AD797.pdf
 $ENDCMP
 #
+$CMP ADF4002BCPZ
+D 400MHz Bandwidth Frequency Synthesizer, TSSOP-16
```
calebns: Should be LFCSP here

calebns:
By the way, do you think it would be good to create a new library file?
I was planning to add some ADF435x fractional PLLs and I'm not quite convinced that Timer.lib is appropriate. What about Clock_PLL.lib ?

lorem-ipsum:
Ok, thanks for looking into this. I will create the LFCSP variant and fix the mistake in the description.

poeschlr:
Are these really clocks?
Might it not be better to use timer_pll and put every pll device in there?
This would fit the original plan layed out in https://github.com/KiCad/kicad-library/issues/1402

lorem-ipsum:
Didn't check with the plan, but Timer_PLL is fine as well. It is probably more correct as they don't have a clock source. I created the lib with the two parts and added it to the symbol-lib-table. Please double check that I did not make any stupid mistakes.
I also fixed the description and changed the name of the CSP footprint.

jkriege2:
Hi!

the symbols look good. Will you also post a PR for the package? 
@poeschlr @SchrodingersGat  Is the proposed name `Package_CSP:LFCSP-20-1EP_4x4mm_P0.5mm_EP2.1x2.1mm` for the package OK?

If yes, I will merge this and we can see who does tha footprints.

Best,
JAN

poeschlr:
The name you propose looks good.

@jkriege2 Another thing (in case you missed it): I took over Olivers position. (Oliver has other obligations that mean he will not have time to continue working for kicad related stuff the same way as he did in the past.)

calebns:
Here is a footprint
https://github.com/KiCad/kicad-footprints/pull/271

Once it's merged, I'll push a PR for two more ADx PLLs to Timer_PLL lib.

jkriege2:
f

jkriege2:
Please fix the merge conflicts!

lorem-ipsum:
I did the rebase and it looks like I didn't break anything. At least checklib is happy.

jkriege2:
Thanks,
JAN

PR #170, opened 2018-01-14
Added symbol for the Ai-Thinker Ra-01 LoRa Module

johnfwhitmore:
Added a symbol for the Ai-Thinker Ra-01 LoRa Module

I've previously had a footprint accepted by kicad-footprints so there is a footprint for this module I'm not sure I've managed to tie this symbol to that footprint as yet.

I'm afraid that a datasheet has been hard to find for this module. Well I've struggled this might be of some help: https://mikroelectron.com/Product/10KM-433M-LORA-LONG-RANGE-WIRELESS-MODULE-RA-01

screenshot from 2018-01-14 15-49-22

This is a first attempt I'll be checking for Travis errors and fixing


johnfwhitmore:
Finished cleaning up errors from Travis so hope this is more or less correct.

screenshot from 2018-01-14 16-29-13

jkriege2:
Some hints on the symbol:
1. you need to fill the box with background (yellow)
2. You could think about stacking the Gnd-pins to clean up further
3. I think inverted pins without the circle, but with a bar (use `~Reset` as label to achieve that) look better

Best,
JAN

johnfwhitmore:
Cheers JAN, sorry about those mistakes.

johnfwhitmore:
screenshot from 2018-01-14 17-17-07

jkriege2:
thanks, looks good now ... a final request: maybe you can move the NAME-label to some better location, maybe on the rhs of the GND-pin, below the symbol?

About work: You see ... it took me already longer to respond to you PR ;-)

johnfwhitmore:
It ain't a race so you can take as long as you like. And the duration of this is largely down to my not being very "smooth". There's a lot of this I should know better but I don't do it often enough. You might argue that I do it too often as it is ;-)

screenshot from 2018-01-15 09-08-33

jkriege2:
OK, the scripts are still buggy as it does no show that there are errors left here ... In a power-pin stack, you need to have one visible POWER I/O-pin and all invisivble pins passive to satisfy the ERC. See http://kicad-pcb.org/libraries/klc/S4.3/ for details. Could you please still change that?

jkriege2:
Thanks,
JAN

johnfwhitmore:
Thank you Jan, sorry it took me a few goes

poeschlr:
It is not expected that everyone gets it right at the first try. Even we maintainers sometimes forget about some rule.

PR #171, opened 2018-01-14
LT RS485 driver and receiver

matyro:
Added four different variations of linear tech RS485 high speed driver and receiver

LTC1688/LTC1689:
http://cds.linear.com/docs/en/datasheet/16889fa.pdf

LTC1518/LTC1519:
http://cds.linear.com/docs/en/datasheet/15189fa.pdf

ltc1518
ltc1519
ltc1688
ltc1689


jkriege2:
Hi!

thanks for your contribution. Could you add some (very) simplified internal schematic, as is already provided for comparable symbols in the lib: 
2018-01-15 07_19_29-symbol library editor - interface_uart d__kicad_kicad-symbols_interface_uart li
2018-01-15 07_19_17-symbol library editor - interface_uart d__kicad_kicad-symbols_interface_uart li

Thanks,
JAN

matyro:
Should I include the enable line as well? Currently its looking like this:
grafik

With enable lines it will be similiar to this (I would move the enable ports to untangle it a bit)
grafik

jkriege2:
I think the gate would be helpful ... so I would add it

matyro:
This style ok?
grafik


Another question, should I increase the distance between A,B inputs to 200 or 300 to allow filter elements and termination in between? With the symbols now I think I do not need the current spacing pattern to seperate the individual blocks.

jkriege2:
Wouldn't the symbol get a bit large?

matyro:
It would get a bit bigger, alternative I can split it in 4 units. But I think that would not work very good with the enable pins.

jkriege2:
I would do it that way: 
2018-01-18 18_41_35-symbol library editor - interface d__kicad_review pretty_kicad-symbols_interfac

Splitting is not a good idea, as then we have the question: What do we do with the EN/~EN-pins? Duplicate in each unit, which poses the danger of connecting multiple signals to them unintentionally?

matyro:
grafik

EN should stay in the center, because of EN12/EN34 for the alternativ part. 

I am not sure if 200mill distance should be used:
Benefit: You can put ESD Protection and termination resistor in between. No need for labes to fanout the part.
Disadvantage: Symbol looks less polished

jkriege2:
I'll leave that style decision up to you ... I see the advantages of both version ...

Let me know which one you chose.

JAN

jkriege2:
@matyro Are you still on this? Could you please also rebase you branch/pull master to fix the conflicts?

Thanks,
JAN

matyro:
Hi here is the update:

ltc1518
ltc1519
ltc1688
ltc1689

jkriege2:
Hi!

A final request: The EN-pins allow to switch the drivers into a High-C-mode ... thus the outputs should actually have the pin-type Tri-State instead of output. Could you still change that?

The rest looks awesome and I will merge after that bugfix ,-)

JAN

jkriege2:
One other thing: I noted that the LTC1689/1519 contain two independent drivers (saw that just now). For these devices you could also think about splitting them into three units, two for the drivers and one for the power-pins ... as is done for logic-ICs or dual-OPAMPs.

But if you think these will mostly be used near to each other in a schematic, we can also leave them as 1-unit symbols.

Best,
JAN

matyro:
Hi,
like this:
ltc1519_c
ltc1519_a
ltc1519_b

I am not sure how KiCad decides which line should be shared between units or when "synchronized pin edit" should auto toggle on nightly, but it does not work as I expect it should.

jkriege2:
yes looks good ... I would remove the box from the power-symbol though ...

poeschlr:
@matyro wrote:
> I am not sure how KiCad decides which line should be shared between units or when "synchronized pin edit" should auto toggle on nightly, but it does not work as I expect it should.

the lines that are shared is determined with a flag in the line properties.
About the syncronized edit: It should only be automatically enabled if you did not select: "all units are not interchangeable". In this case you should set this flag. See: https://lists.launchpad.net/kicad-developers/msg33444.html

matyro:
> I would remove the box
So freefloating power pins?

> the lines that are shared is determined with a flag in the line properties.
That worked, but part of the lines had the flag set others not, so I needed to manual check each line.

jkriege2:
I think freefloating power-symbols are kind of widely used

jkriege2:
Please lay out the pins as follows: 
2018-01-28 17_23_50-symbol library editor - interface d__kicad_review pretty_kicad-symbols_interfac

Also I would propose to set the pin-label offset to 20 (see screenshot). And it is customary to make the power-unit the last unit!

jkriege2:
Also there is a connection to the EN-lines mising: 
2018-01-28 17_26_30-symbol library editor - interface d__kicad_review pretty_kicad-symbols_interfac

matyro:
Is there a fast way to change units or only copy/paste?

jkriege2:
Easiest way is to edit the library with a text editor like notepad++ ... 
2018-01-28 17_37_18-_d__kicad_review pretty_kicad-symbols_interface lib - notepad

jkriege2:
Looks mostly good (units and all). A final request: Please modify the symbols to conform with the standard-style for inverted pins, which is an unfilled circle, e.g.
2018-02-02 21_51_38-symbol library editor - interface d__kicad_review pretty_kicad-symbols_interfac

Thanks,
JAN

jkriege2:
nice work!

Thanks,
JAN

PR #175, opened 2018-01-18
Revert lib version updates for some libs. Again.

poeschlr:
Some libs have again been stored using a recent nightly. This creates problems for everyone using older versions. (dcm file content gets invisible for some reason.)

I reverted the lib version header for all libs that had problems.

### DIFF
--- a/Amplifier_Operational.lib
+++ b/Amplifier_Operational.lib
@@ -1,4 +1,4 @@
-EESchema-LIBRARY Version 2.4
+EESchema-LIBRARY Version 2.3
 #encoding utf-8
 #
 # ADA4807-1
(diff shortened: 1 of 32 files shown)


jkriege2:
Hmmm ... I'm not sure how much sense this makes. In the end we're building libs for v5 here (or the current nightlies). So at some point we will have to allow the new format. In footprints we already create incompatible footprints by e.g. using new features like polygon pads!

I would find it cleaner to declare this lib for v5 (and current nightlies), while the old kicad-library is our legacy lib for the old versions?

This will also help us in the PRs, because the new KiCAD versions resort the drawing commands in the .lib-files and if we allow symbols from a stable version here, we will always get a lot of noise just by that resorting in the diffs.

PR #196, opened 2018-01-20
Max30102

aurabindo:
Add max30102 heart rate sensor. I added it to Sensor lib
Datasheet: https://datasheets.maximintegrated.com/en/ds/MAX30102.pdf

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/277

max30102_sym
max30102_sym_1
max30102_sym_0


jkriege2:
- FPFilter should be `Maxim*OLGA*3.3x5.6mm*P0.8mm*`
- REFDEC should be IC
- `R_DRV` and `IR_DRV` should not be NC-pins ... but OUTPUT + thex should not be invisible and moved inside!

aurabindo:
>`R_DRV` and `IR_DRV` should not be NC-pins ... but OUTPUT + thex should not be invisible and moved inside!

@jkriege2 It is shown as NC in the reference design in the datasheet, which is why I kept it NC. Could you please confirm if this pin should be NC or not ?

nc_max

aurabindo:
The footprint associated is to be in OptoDevice. Shouldn't the symbol also be in a lib (which doesnt exist now) OptoDevice ? Although because of default footprint association, this would not confuse the user.

jkriege2:
OK. I would make them standard pins (SHOW, OUTPUT). In the end the user will refer to the datasheet anyways and maybe there is a use-case for them (e.g. some measuement).

About the lib: footprint-libs and symbol-libs are not necessarily connected. Both domains use different sortings. I think sensor.lib is OK, and `Sensor_Optical.lib` would be OK too ... If there are many sensors specifically used to measure medical data, one could think about having `Sensor_Medical`, but I think there are not enough of them (especially ones that do not have a second non-medical use-case). So stick to `Sensor`or move to `Sensor_Mediacl`... whatever seems more befitting for you.

JAN

jkriege2:
Did you change as proposed? Please update the screenshot when you're done!

Thanks,
JAN

aurabindo:
@jkriege2 sure, I'm just trying to fix them one by one :P

aurabindo:
max30102_sym
max30_0
max30_1

@jkriege2  done.

evanshultz:
@jkriege2 
Just a couple comments as I was without internet this weekend:
- This symbol seems excessively large. Would we want it smaller?
- Could the VLED+ pins (9 and 10) be stacked? I quickly looked at the datasheet and it seems to be OK.

PR #200, opened 2018-01-20
moved symbols from old Motorola-lib

jkriege2:
please merge this way ... I won't have time to go over all KLC-Issues (I fixed a few thow) ... we can open an Issue for the other KLC-violations as done for several other libs.


jkriege2:
OK; I think I'm as far as I will go with this (most KLCissues are resolved ... I didn't research datasheets and packages (only minor fixes on those)

Can be merged?

poeschlr:
Did you notice the stuff regarding duplicated pins? (Symbols with duplicated pins result in unpredictable netlists generation when used in a schematic.)
We might separate the affected symbols into a lib marked as broken. (And create an issue asking for help) Or we go further and remove them completely. (Maybe make an issue that details which symbols have been deleted for this reason. Somebody interested in fixing them could copy them over from the old repo before fixing them.)

Symbols that show this particular problem:
- MCF5212-LQFP64
- P4080-BGA1295

jkriege2:
OK, duplicated pins are fixed

poeschlr:
Looks good, thanks

PR #201, opened 2018-01-21
Add Microchip CAN FD components MCP2517FD and MCP2558FD V2

menschel:
CAN FD is an update to the Contoller Area Network standard.

This patch adds schematics for Microchip devices:
1.
CAN FD controller MCP2517FD
https://ww1.microchip.com/downloads/en/DeviceDoc/20005688A.pdf
with 2 package options, where 1 package was not found in the official footprint library.

mcp2517fd_example

MCP2517FDx-H/JHA is VDFN-14 according to spec sheet but the official footprint library's naming seems to omit the V letter thus package name given as
Package_DFN_QFN:DFN-14-1EP_4.5x3mm_P0.65mm
See PR 
https://github.com/KiCad/kicad-footprints/pull/280

2.
CAN FD transceiver MCP2558FD
https://ww1.microchip.com/downloads/en/DeviceDoc/20005533A.pdf
with 3 package options where 1 package was not found in the official footprint library.

mcp2558fd_example

MCP2558FDx-H/MNY is TDFN-8 according to spec sheet but only DFN footprint was found in official footprint library, thus changed to
Package_DFN_QFN:DFN-8-1EP_2x3mm_P0.5mm

Pin groups are adapted from predecessor devices MCP2515 and MCP2551 respectively.

Checking symbol 'MCP2517FDx-H-JHA' - No errors
Checking symbol 'MCP2517FDx-H-SL' - No errors
Checking symbol 'MCP2558FDx-H-MF' - No errors
Checking symbol 'MCP2558FDx-H-MNY' - No errors
Checking symbol 'MCP2558FDx-H-SN' - No errors

------------


Misca1234:
when you get your foot print accepted, you do not need to two variants of the MCP2517FD, 
just add both foot prints to the same schematic (MCP2517FD)

menschel:
The 2nd variant has been removed and added as alias to the 1st variant
instead.

evanshultz:
General:
- Wildcard temp, and also the T&R package is optional so it can be omitted. For example, MCP2517 will be just `MCP2517FD-xJHA` and `MCP2517FD-xSL`. (Sorry that you got bad advice earlier.)
- The package type is always the last element of the description.
- Active-low pins with one function need only a leading tilde (~) and not a trailing one (not a big deal).

MCP2157:
- Stack the EP and VSS pins of DFN symbol (datasheet section 1.2).
- Fix DFN footprint name as mentioned at https://github.com/KiCad/kicad-footprints/pull/280.
- OSC2 pin is an output.
- Pins 8 and 9 are bidirectional.
- Can you shrink this symbol a bit vertically? It's overly large.

MCP2558:
- Would you be willing to add MCP2557? Not required, just asking.
- Is there a reason why you noted the level shifter on the inputs? Can this be left out?
- TDFN: The footprint you picked does not match. Use `TDFN-8-1EP_3x2mm_Pitch0.5mm_EP1.80x1.65mm.kicad_mod`. Can you make this footprint?

menschel:
> General:
> - [x] Wildcard temp, and also the T&R package is optional so it can be omitted. For example, MCP2517 will be just MCP2517FD-xJHA and MCP2517FD-xSL. (Sorry that you got bad advice earlier.)
> - [x] The package type is always the last element of the description.
> - [x] Set pin text offset to 20mil.
> - [ ] Active-low pins with one function need only a leading tilde (~) and not a trailing one (not a big deal).

The trailing tilde is necessary as the other pin functions are not negative logic.

> MCP2517:
> - [x]  Stack the EP and VSS pins of DFN symbol (datasheet section 1.2).
> - [x]  Fix DFN footprint name as mentioned at KiCad/kicad-footprints#280.
> - [x]  OSC2 pin is an output.
> - [x]  Pins 8 and 9 are bidirectional.
> - [x]  Can you shrink this symbol a bit vertically? It's overly large.

> MCP2558:
> - [x]  Would you be willing to add MCP2557? Not required, just asking.
> - [-]  Is there a reason why you noted the level shifter on the inputs? Can this be left out?

The level shifter input voltage VIO technically is a power input.

> - [ ] TDFN: The footprint you picked does not match. Use TDFN-8-1EP_3x2mm_Pitch0.5mm_EP1.80x1.65mm.kicad_mod. Can you make this footprint?

I'll see to it once I have time.

evanshultz:
Thanks for all the fixes!

General:
- Remove the word `package` from the description. End with, for example, `TDFN-8`.

MCP2157:
- MCP2517FD-xSL cannot be an ALIAS of MCP2517FD-xJHA because it does not have pin 15 (the exposed pad) and is a different package. You will need a separate symbol.

MCP2557/8:
- `MCP2558FD-xMNY`, not `MCP2558FDx-MNY`.
- MCP2557FDx-MF and MCP2558FDx-MF symbols did not get the pin text offset set to 20mil.
- Regarding the level shifter, I meant why this is in the description. It doesn't seem important enough to be listed there to me. What was the reason?
- My mistake on the footprint name. "Pitch" should just be "P". Can you correct that? It will be OK to call out the footprint on the symbol without the footprint being merged as long as we note it's missing as an issue in the footprint repository at https://github.com/KiCad/kicad-footprints/issues.

menschel:
Done.

The level shifter is mentioned in the description because it distinguishes the MCP2558 from the MCP2557.

evanshultz:
Thank you!

"Internal level shifter on digital I/O pins, " is an awfully long amount of text. "Internal level shifter, " would be fine or even leaving it all out is fine too, but I'll merge as it is.

I just noticed one thing I missed earlier: MCP2517 has " Extended temperature range," in the description, but that's not correct since we've put a wildcard "x" in the temp range location of the PN. Please remove that text string and then I'll merge. Sorry about that!

evanshultz:
I didn't check the Travis output earlier because I had before, but it's now found something new due to your changes. The symbols with a SOIC package need to have the library name "Package_SO" instead of "Package_SOIC".

menschel:
Done,
but are you sure
TDFN-8-1EP_3x2mm_Pitch0.5mm_EP1.80x1.65mm.kicad_mod
is the right Footprint?

I came up with 
TDFN-8-1EP_3x2mm_P0.50mm_EP1.4x1.3mm.kicad_mod
while I'm not sure what actually are X and Y as ST does it the other way 2x3.
http://ww1.microchip.com/downloads/en/DeviceDoc/8L_TDFN_2x3_MN_C04-0129E-MN.pdf

evanshultz:
KiCad footprints like this have pin 1 in the upper right, so it is rotated. That does change the dimensions from the Microchip docs so we do need to think about it. But we agree on 3x2mm.

I corrected "Pitch" to "P" above. So now we agree up to `TDFN-8-1EP_3x2mm_P0.50mm_` and the only thing left is the copper size of the big center pad.

Actually, http://ww1.microchip.com/downloads/en/DeviceDoc/8L_TDFN_2x3_MN_C04-0129E-MN.pdf and https://ww1.microchip.com/downloads/en/DeviceDoc/20005533A.pdf give conflicing information. The first link would tell us 1.5x1.6mm. But the datasheet shows 1.80x1.65mm, and so that's what I chose since I looked at the datasheet. And since it's the larger size, but still gives adequate clearance to perimeter pads, I'm inclined to go with that size.

So yes, I do believe `TDFN-8-1EP_3x2mm_Pitch0.5mm_EP1.80x1.65mm` is the right footprint. However, we don't current have that footprint so it would need to be created.

How did you come up with a 1.4x1.3mm exposed pad?

menschel:
1.4x1.3 are the dims of the exposed pad, mentioned in the description and dims E2, D2. 
1.80x1.65 are max values for the exposed pad, mentioned in the recommended land pattern.

evanshultz:
Ah! Now I see. Our footprints are named by the copper pad size, and so we'll have to use the max from the datasheet (no nominal is given and we don't want to go smaller and affect the expected thetaJA.

Certainly an argument could be made for using the exposed pad on the package in the footprint name, especially as that also affects the 3D model (assuming the footprint and 3D model have the same name), but that's not what we have done.

It looks to me like your latest commit is correct. To give you a chance to respond, please let me know if you're OK with everything and I will merge.

menschel:
Yes, i think the branch is mature enough to be merged.
Nice working with you. Thanks.

evanshultz:
@menschel 
Alright. Thanks for all your work! This is your first contribution, at least to the symbol library, and I hope to see more from you in the future!

Unfortunately, with all the talk about the TDFN above we forgot to add the EP size to the DFN footprint. While the datasheet clearly says it's optional, if the symbol doesn't support it the user won't be able to attach to the EP. And you already have pin 9 added correctly. The default footprint should be `DFN-8-1EP_3x3mm_P0.65mm_EP1.55x2.40mm`.

We will need to update the footprint filter for this DFN-8 footprint above. And for the TDFN package, please add the EP size to the footprint filter. Add `EP1.80x1.65mm*`. Similarly, the VDFN should add `EP1.65x4.25mm*`.

I'm closing and re-opening to that Travis will see the DFN-14 footprint has been added.

Ratfink:
@evanshultz For future reference, if you're logged in to Travis CI, you can re-run the tests from the page with the output log by clicking "Restart build" near the upper right.

evanshultz:
Argh! I should have thoroughly checked the SOICs along with the DFNs. One more thing: MCP255x needs the dimensions of the SOIC in the footprint filter. Change it to `SOIC*3.9x4.9*P1.27mm*`.

I assume by "tutorial for Travis CI" you mean how to interpret the results of the auto-checking? (Travis CI has nothing to do with KiCad but is a generic framework for continuous integration regardless of the application.) KLC (http://kicad-pcb.org/libraries/klc/) is the best place to check. All rules in the check scripts should be based on meeting the guidelines from KLC. That doesn't mean everything is perfect, or that all cases have a guideline, but it's the intention for KLC to define how the official libraries are built.

evanshultz:
@menschel 
Thanks for this final fix. I appreciate your patience to make these symbols meet the official library requirements. I hope you will contribute more to the KiCad libraries in the future!

PR #203, opened 2018-01-21
Add A2982 8-Channel Source Driver by Allegro

Schlumpf:
https://www.allegromicro.com/~/media/Files/Datasheets/A2981-2-Datasheet.ashx

image

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #204, opened 2018-01-21
KLC-fixes for Transistor_Array.lib:

jkriege2:
- better split symbols by available package types
- improved description
- added missing variants of existing (ULQ... as copy of existing symbol due to different footprints)
- removed MC1417 for which no information is available
- reworked A2982 symbol (made pins shorter, now 100mil, were 200mil)


jkriege2:
see my comment there: https://github.com/KiCad/kicad-symbols/pull/203#issuecomment-359492302

jkriege2:
@evanshultz This also fixes the issue you raised in #203

jkriege2:
I reworked the style, does this adress everything?
2018-01-23 06_41_09-eeschema devproject _ d__kicad_devproject

Note that The A2982 is actually open-emitter instead of open-collector as all others. Thus the "inverter"/"driver"-symbol difference

evanshultz:
@jkriege2 
Sorry that I forgot to review this earlier! Thanks for making all the updates!

PR #206, opened 2018-01-21
reworked rotary encoder symbols

jkriege2:
- added internal drawing of switches 
- added shielded versions
- all pins are PASSIVE now (were INPUT)
- all other components were only touched by the lib-file version update to 2.4
- ALPS footprints available at https://github.com/KiCad/kicad-footprints/pull/288

2018-01-21 18_03_27-eeschema devproject _ d__kicad_devproject


bobc:
Looks good. Just a minor thing, should the switches indicate momentary "push to make" mode rather than a toggle?

poeschlr:
For generic switches we use "SH" as the pin number for the shield pin. This might be an option here as well.

jkriege2:
Yes possible (since we don't ahve any footprints except the ones I posted earlier yet, we can decide rather freely ;-) ... what do we use for connectors?

BTW: For housings and heat-sinks we use 0 I think, maybe we should unify that over the whole lib? There are also crystals that have a shield pin ...

evanshultz:
@jkriege2 
Can you take a look at https://github.com/KiCad/Buttons_Switches_THT.pretty/pull/7 and see if those footprints will work with these symbols? Here is what I see:
- PEC11R-4xxxF-Nxxxx / EC11B15202AA (Rotary_Encoder): OK
- PEC11R-4xxxF-Sxxxx / EC11B15242AE / EC11B15242AF (Rotary_Encoder_Switch): OK
- EC12E: You got it at https://github.com/KiCad/kicad-footprints/pull/288

This would add the horizontal Alps EC11 footprint with switch and without, and also Bourns PEC11 with and without switch.

If you don't find any other changes that need to be made, I'll fix up the footprints according to my comments in the old PR and create a new PR for them.

jkriege2:
To me it looks like they fit ... Only the mounting holes seem to be without number, so cannot be connected to. Also note that I "merely" modified the unshielded variants (they were already there before) and just added the shielded variants. ... i.e. go ahead with the old footprints ... I assume @SchrodingersGat won't have any objections ;-)

So any decision on pin-naming?

jkriege2:
Thanks for your contribution!

poeschlr:
Jan did you merge your own PR and thank your self for it? Or is there a bug in github?

jkriege2:
Uupppps ... sorry, yes it seems like it :-(((
I meant to merge another PR and had several tabs open, now I understand why I was sure to have merged, but had to merge again!

I will revert, so we can go on with this discussion!

jkriege2:
I opened a new PR to replace this: #220. Please go on discussing over there. I reverted thes erroneous merge of this PR!

PR #218, opened 2018-01-24
Added dual-row connector symbols with odd pin-counts

jhol:
20180123-backflip-symbol-screenshot-small

This patch adds symbols for dual-row style connectors with odd numbers of pins as found in dual-row FFC connectors.

The list of pin-counts was compiled from a simple survey of FFC products listed on Mouser.

A similar patch was previously submitted to the kicad-library repository: https://github.com/KiCad/kicad-library/pull/1755


poeschlr:
Could you change the naming to: `Conn_2Rows-nnPins` (where nn is the number of pins)
The second footprint filter could be changed to *FCC*2Rows* (Remove the "_"). This filter is only needed for kicad 4 compatibility as in kicad 5 the first filter will already result in them being shown.

jhol:
@poeschlr Done!

poeschlr:
Thanks

PR #220, opened 2018-01-24
reworked rotary encoder symbols (added internal drawing of switches +…

jkriege2:
- replaces burnt PR https://github.com/KiCad/kicad-symbols/pull/206
- added internal drawing of switches 
- added shielded versions
- all pins are PASSIVE now (were INPUT)
- all other components were only touched by the lib-file version update to 2.4
- ALPS footprints available at https://github.com/KiCad/kicad-footprints/pull/288

- **WHAT NEEDS TO BE DISCUSSED:** (from discussion in #206)
  **- name/number of shielding pin (should be 0 or SH?)**
  **- style of push button and switches**

2018-02-19 20_49_20-symbol library editor - device d__kicad_kicad-symbols_device lib
2018-02-19 20_49_10-symbol library editor - device d__kicad_kicad-symbols_device lib


jkriege2:
One proposal about style: I could change the style of the internal switches to this: 
2018-01-24 07_07_11-symbol library editor - device d__kicad_kicad-symbols_device lib

- one of the two ABC-switches is open, one is closed 
- additional contact 4-5 is drawn as momentary switch

What do you all think?

evanshultz:
@jkriege2 
Pin Numbers
After some thought, I think "SH" is better than "0" for the shield pin. Using "0" gives only one option, whereas "SH" would allow several other pins in case they were needed. For example, this is similar to the thermal pad having a pin called "PAD" if the connection is not called out by the part vendors. Furthermore, having some intelligence in the name is nice.

I do realize it's more burden to confirm footprints have matching pin numbers.

What do you think?

Style
I personally like the original A/B/C drawing. The time when one is open and the other is closed is small. Most of the time (always when the shaft isn't moving and most of the time during a turn) they are both open.

Regarding the switch, I think momentary is right if the encoder has a switch that closes when the shaft is pushed. Are there other types of switches on a rotary encoder? I suppose you could have one where the first detent is a switch, as I've seen some encoders with endpoints, but usually encoders are freely-spinning and you see this on pots. So I think momentary is best.

jkriege2:
1. About the drawing: I like having the ABC-switch inside, as it shows which one is the common pin and thus makes the schematic clearer Alps uses a schamtic symbol that does just show two switches (see http://www.alps.com/prod/info/E/HTML/Encoder/Incremental/EC09E/EC09E1520407.html towards the end) ... But I agree, we could also have them both open ... I'm not aware of other rotary encoders with permanent contacts or opening contacts ... there are varieties with integrated LEDs (also RGB), see e.g. https://www.sparkfun.com/products/10982 ... but those are specialized (I think), so they are better suited by specific symbols?

2. Since we discussed this elsewhere ... which name should be picked? 
  - 0 (I like that idea most, as it does not require to use the pin for a specific usage), 
  - SH (shield, thinking about this a bit more ... is a shield for such encoders really necessary?), 
  - MT (They are mostly for mounting?) ... really not sure... 
even pondering whether that symbol with extra pin should be called SHIELD at all ... Please have a look at these discussions, before we come to a conclusion: https://github.com/KiCad/kicad-symbols/issues/268 https://github.com/KiCad/kicad-footprints/issues/117

poeschlr:
About pin naming.

In my opinion, shield and pure mounting pins should look different in the symbol to communicate the fact that the later is purely for increased mechanical strength and has no electrical purpose. That is why i am opposed to the generic use of pin number 0 for both of them.

In the [discussion about connectors](https://github.com/KiCad/kicad-footprints/issues/117) we decided to use the pin number SH for shields and **MP** for mounting pads.

We also decided to add -[number uniquely numbered pads]SH (or MP) to the footprint name after the pin count. This is to make it easier to define a footprint filter for generic components. (I also mentioned this in https://github.com/KiCad/kicad-symbols/issues/268)

---

The pins in https://github.com/KiCad/kicad-footprints/pull/288 might be shield pin. But i am not sure about that.
In this image from the product site it looks as if they are:

jkriege2:
So I removed the shielded versions, because I figure that for THT-components the mounting pins do not really need to be connected anywhere (so no pin required) and I have no clue how we should make the mounting-pins look like in the symbol. I searched a bit and found no encoder that has a shield that makes sense (with those switches there shouldn't really be any EMC-problems).

I will make the footprint-pins have name MT though, so people can use them that way if they want or simply ignore those pins (most photos of PCBs I found on google seem to be using that approach) even if there is a GND-plane!

... and reworked the symbol a tiny bit
2018-02-07 21_32_40-symbol library editor - device d__kicad_kicad-symbols_device lib

jkriege2:
@poeschlr How do we proceed here?

jkriege2:
One more suggestion (since I saw @strombom 's PR #299): We could shift from the pin-numbers to letters, i.e. A,B,C and maybe SW1 and SW2 .... This way we could also have generic symbols with at least one LED (e.g. LEDA, LEDK). The function assignment would then move into the footprint, but this allows us to describe most rotary encoders easily with one symbol and the function-pin-assignment is a bit more obvious than with numbers (Also several manufacturers don't give pin-numbers for these devices, but just put a label with the function).

@evanshultz @poeschlr @strombom What do you all think?

evanshultz:
I like the idea of using intelligent pin names, but I'm scared that it opens Pandora's box. It will be very difficult to control this and decide when and how to approve PRs. And then things may turn into a mess later. So my feeling is against pin names in general, with only a limited number of exceptions (SH, MP, etc.)

For the sake of pin length, if we do this I think we use the shortest names available. So instead of "LEDA" just use "A" as I assume the symbol will have the LED inside.

poeschlr:
My opinion on "intelligent" pin names is as follows:
If there is a datasheet that defines pin names (or numbers) then these should be used.
I would not touch things old symbols and footprint at this point in time as i fear we would not finish till the v5 release (hold of till v6.)


And i am a bit opposed to the idea of having general symbols for everything. I think it is better to have a few symbols for the most used purposes in the device lib but then go with specialized symbols per part. 
I assume that very few of the parts are both symbol and footprint compatible. 

---

I really would like libs other then the device lib (plus connector libs) to house only fully specified parts (=parts with MPN or Base MPN + footprint suffix as name and footprint field set to point to the correct footprint).
Generic symbols are ok for hobbyist users but most users of kicad who do professional design prefer the option to generate a bom and to have the footprint field assigned already in the symbol such that one point of user error is removed. (Which means use fully specified symbols)

jkriege2:
The two types of rotary encoders are very common (made by many different companies, all with 3pins for ABC (2-switches, 1 common pin) and optionally 2 extra pins for the push-switch. I think having a generic symbol for these would be good (note we also have generic symbols for switches that are compatible with most footprints!). We can still use these as a basis to derive fully specified symbols for a series. Having such a generic symbol has the additional advantage of beeing a template, so the style might (I hope) be more uniform for future additions to the lib (note: in #299 a very different style was proposed!).

About pin-naming: **Personally** the 1-2-3 (=A-C-B) and 4-5 (=SWITCH) is fine for me for these generic symbols. I saw that @strombom uses A-B-C pin"number" so I wanted to discuss this question here with a broader audience.

In sum, I like the symbols in this PR better, as they show the switches AND show that it's a rotating knob, showing only the switches does not hint at the function of a rotary encoder. But we should decide for a sytle and then we can for all other PRs accordingly. I think I would have proposed specialized symbols for the footpritns that I PR elsewhere in a second PR anyways ;-)

jkriege2:
As an explanation: About the named pins I see the same pros/cons as @evanshultz does ... and because of them I tend towards the numbers.

jkriege2:
@evanshultz @poeschlr Could we finish this off? What's you opinion on my comments above?

evanshultz:
I'm OK using A, B, and C as the pin number. Perhaps the switch is 1 and 2. I like the symbols above and would merge them.

@jkriege2 
Can you summarize? What are the outstanding items you want resolved?

evanshultz:
And you could add version with a LEDs (switch + LED and also no switch and just LED) to have all the rotary encoder symbols I can think of. But I'm OK to merge without any LED versions.

jkriege2:
I want a CHECK on the following items:
1. updated symbols
2. no schielded version or versions mit mounting-pins (symols!!!)
3. pin-numbering 1-2-3 or A-C-B für the encoder and 3-4 or 1-2 for the push-button
4. I didn't add LEDs, as they are relatively diverse (1 LED, 2LEDs, RGB-LEDs with/without switch) and I'm no clear whether the internal connections (often 1 pole of the push-button is connected to all LED cathodes, but for 1-LED they are separate ... ). In sum I think these need to be made as fully specified symbols. Also finding a pin-naming scheme with numbers only is not easy.

I think the most important thing is item 3.

Best,
JAN

poeschlr:
As long as we allow fully specified symbols and atomic parts in addition to these generic symbols then i am ok with having a limited number of generic symbols to give users a starting point.

For fully specified the pin number discussion is moot as the symbol there is made for one particular footprint. I don't think making each fully specified symbols (or even atomic parts) compatible to a generic symbol makes sense.

In this particular case, the pin number discussion here is independent of the contribution by @strombom. (As his contribution can be seen as an atomic part. That footprint is not meant to be used with any other symbol. -> It is atomic.)

jkriege2:
OK, so should we keep the 123-numbering or change to ACB?

I now tend to ACB, because that seems to be used by most manufacturers:
- http://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=1-1773449-1&DocType=DS&DocLang=English
- https://www.ctscorp.com/wp-content/uploads/290.pdf
- http://www.bourns.com/docs/Product-Datasheets/pec09.pdf
- http://docs-europe.electrocomponents.com/webdocs/0f4c/0900766b80f4c221.pdf

For the switches there are different standards. I've seen 1-2, S1-S2, D-E, ...


For @strombom 's addition: If there are symbols for each footprint, the devices are covered as fully specified or even atomic components, but if we adopt the 123-style here, the footprints will NOT MATCH the generic symbols, even if they could easily.

So all in all, the ACB-"numbering" makes most sense (now with S1,S2 for the switch, as I feel that if the kind of name by function we should do it for all pins) and I changed the PR accordingly. Can be merged then?

2018-02-19 20_49_20-symbol library editor - device d__kicad_kicad-symbols_device lib
2018-02-19 20_49_10-symbol library editor - device d__kicad_kicad-symbols_device lib

poeschlr:
Seems reasonable.

My thinking about fully specified and atomic parts is that they do not necessary need to match up with generic symbols. (This allows these to follow datasheets more closely.) 
We should only allow footprints with different pin numbers then the generic symbol if the datasheet gives different pin numbers. If no pin numbers are given in the datasheet then the footprint and symbol should use the same numbering scheme as the generic symbol. 
(But we would still allow adding a fully specified symbol as it will make creating BOMs easier and does reduce the steps that can introduce a design error.)

I will check the PR tomorrow if it has not been merged until then. (Don't have access to kicad right now.)

jkriege2:
Yes, I just though that if we adopt a widely used naming-scheme for the generic device, there is a good chance that even the footprints for specific parts will match (if they don't there's still their fully-specified symbol ;-)

PS: I change my footprints accordingly

evanshultz:
@jkriege2 
Are you ready? Anything left you want to do? Seems everybody is OK to merge.

jkriege2:
I'm ready!

poeschlr:
Looks good in the current nightly. Thanks

PR #223, opened 2018-01-25
TMP36xS symbol added

Resaj:
TMP36 symbol added for SOIC-8 footprint.

This is the datasheet: http://www.analog.com/media/en/technical-documentation/data-sheets/TMP35_36_37.pdf

imagen


jkriege2:
Hi!

thanks for you contribution. Could you please fix a few more things:
1. The footprint filter should include the geometry `SOIC*3.9x4.9mm*P1.27mm*`
2. The device has 4 NC-pins in that package. There are two approaches to handle these: 1. add all NC-pins as invisible NC-pins (their end should be on the box of  your symbol to prevent unwanted connections), or you leave them out, then you'll have to add the pin-count to the FPFilter, i.e. `SOIC?8*3.9x4.9mm*P1.27mm*`
3. Finally please add the package-name to the description (separated by comma at the back), not the tags field.

Thanks,
JAN

Resaj:
I think it's right now.
I've chosen the second method to fix the NC-pins, so I've changed the footprint filter to SOIC?8*3.9x4.9mm*P1.27mm*

evanshultz:
@Resaj 
The active-low pin should start with a "~" instead of a "#".

@jkriege2 
Sorry for popping in. It's all yours!

jkriege2:
@evanshultz no problem ;-)

@Resaj Please do as @evanshultz commented ... Then this should be OK for merging!

Resaj:
Thanks for your comments! I've just changed the pin.

jkriege2:
Thanks,
JAN

PR #226, opened 2018-01-26
AD9142

calebns:
1.6GSPS 16bit dual-channel DAC, LFCSP-72
 16bit DAC 2CH
 http://www.analog.com/media/en/technical-documentation/data-sheets/AD9142.pdf

image

------------


jkriege2:
Hi!

thanks for your contribution. Please fix a few minor things:
1. remove the pin-count from the FPFilter ... KiCAD has a pin-count filter for such checks
2. If you want to clean-up the symbol a bit, you could think about using power-pin stacks for the power-pins, see http://kicad-pcb.org/libraries/klc/S4.3/ (just a proposal, no requirement!)

Best,
JAN

calebns:
Good point, thanks.
I've stacked power pins as per KLC. Also, removed pin-count from the filter.

image

evanshultz:
Sorry to pop in, but a few comments:
1. Can this symbol be a bit smaller? There is definitely dead space at the top and bottom, and it might be able to be less wide as well.
1. Adding a simplified internal block diagram would be nice and will take up some room inside the symbol body. Many other ADI DACs have this, but it's not required unless @jkriege2 requires it.
1. Is "GND_EP" the best name for the thermal pad pin? The datasheet calls it AVSS, which matches with the positive supply pins. It is a ground, but I think there's a better name in this case.

calebns:
Made the symbol smaller.
As for simplified diagram, I'd rather avoid it. The part is very complex, it has NCO, FIFO, I/Q equalization path... it's definitely not a typical DAC as seen in the library.

image

jkriege2:
I like the idea of an internal schematic drawing (just roughly, as done for other DACs). Would you be able/willing to provide that (not a requirement!)?

calebns:
image

jkriege2:
THanks,
JAN

PR #227, opened 2018-01-26
Add TI LM25085MM and LM25085MY Buck controllers

strombom:
2V Constant On-Time PFET Buck Switching Controller
http://www.ti.com/lit/ds/symlink/lm25085.pdf

image

image

------------


strombom:
I think the tests failed because the EESchema-LIBRARY Version 2.3 was increased to 2.4 and many changes was made to the lib file so that all symbols were reevaluated. Should I try to revert to version 2.3 and paste my changes there or is there any other way to pass the tests?

poeschlr:
Did you close this because you did not get a reaction fast enough or was there a different problem?

Regarding the many reported errors when saving the lib in a different kicad version: This is a known problem. Sadly different versions of kicad use different line ordering in their lib output. (Even if the lib file version is the same.)
It is allowed to contribute to the lib in any file format version as long as the contribution does not contain invalid files (and follows KLC)
The danger is by going form 2.4 back to 2.3. (If done by kicad this results in an empty dcm file. This is another known problem.)

If you really only want your component checked by travis you can use a text editor to insert it into the files. (But this is kind of dangerous.)

strombom:
Yes I'm just learning git. I think I'll create a new branch for each component. And use 2.3 file format this time so that there is only a minimal difference in the commit.

strombom:
Many regulators in the library have the power in and out pins on left/right instead of on the top as stated in KLC.

Example:
image

I tried to follow KLC and put them on top:
image

What is preferable?

On all datasheets the power pins are on left/right...

poeschlr:
The KLC is not written to encompass all parts. (We also do not list all known exceptions in there as it will make the KLC unreadable)

I think sometime in the past it was decided that regulators do not need to follow the "power pins only on top/bottom" rule.
Maybe @jkriege2 does remember more.

jkriege2:
For regulators we often accept power-pins on the left/right as it is often the "natural"/commonly used pin-layout (especially for e.g. 3-pin linear regulators) ... in the end they can be thought of as power-converters as compared to usual components that are more like power-sinks.

Keep in mind that exceptions to KLC can be allowed if there is a good reason for them! + our check scripts are naturally not always right

PR #230, opened 2018-01-27
fixed opto libraries to match renamed in PR https://github.com/KiCad/kicad-footprints/pull/308 for OptoDevice.pretty

jkriege2:
This PR changes some footpritns and FPFilters only so the symbols follow the renames in PR https://github.com/KiCad/kicad-footprints/pull/308

For details, see PR https://github.com/KiCad/kicad-footprints/pull/308

remaining KLC-violations:
- most should go away when footprints are renamed/PR is merged
- some will remain, as the symbols are for small components that may use 50mil pin-grid


poeschlr:
C12880MA Has a spelling mistake in the footprint filter. (OPtoDevice instead of OptoDevice)

jkriege2:
problems should be fixed!

poeschlr:
Looks good. Thanks

PR #233, opened 2018-01-28
fixed I2C-pintypes (SCL in slaves INPUT) 

jkriege2:
- affected devices:
  - ADS128D818
  - BQ78350 
  - AS3935 
  - HDC1080
  - TMP411 
  - DS1621 
  - LTR-303ALS-01 
  - APDS-9301

- some other KLC-violationson the way (wrong FPFilters, made some pin-stacks, fixed broken datasheet, moved NC-pins to where they cannot be erroneously connected)


Shackmeister:
Perhaps we should make SCL bidirectional if the device supports clock stretching https://www.i2c-bus.org/clock-stretching/
Which I believe most do

evanshultz:
@jkriege2 
Sorry to let this linger! Can you fix branch conflicts and I'll take a look?

jkriege2:
done!

evanshultz:
@jkriege2 
Great! There is a good number of errors still. Do you want to fix anything else or merge this now?

@Shackmeister 
Holding the SCL bus "low" is not outputting a signal. SCL is only an input for all slave devices.

jkriege2:
fixed several of the errors, some footprints are simply missing (don't have time to add them right now!).

i.e. please merge in this state, if there aren't any serious objections right now.

Thanks
JAN

PR #235, opened 2018-01-28
Lm25085

strombom:
LM25085MM

Datasheet: http://www.ti.com/lit/ds/symlink/lm25085.pdf

image

image

------------


jkriege2:
Nice work!

I noticed the datasheet also contains other package-options. Would you be willing to add these to this PR too?

strombom:
Thank you.

Yes, however the other packages have exposed pads, so I think I need to make a separate symbol for those?

jkriege2:
yes, ideally one for each type of package (could be a copy, an ALIAS does not work, as you cannot assign different packages to different ALIASes

jkriege2:
Please also fix the merge-conflicts

Issue #182, opened 2018-01-20
Triac symbols needs correction, or new symbol


poeschlr:
See old issue: Triac symbols needs correction, or new symbo

jkriege2:
posted PR as a fix: https://github.com/KiCad/kicad-symbols/pull/239
![2018-01-28 16_41_11-symbol library editor - device d__kicad_kicad-symbols_device lib](https://user-images.githubusercontent.com/11967523/35484048-55f84c46-044a-11e8-933a-bc1c5560f3d1.png)

PR #239, opened 2018-01-28
made full set of TRIAC-sybols with visible A1/A2-names and all possible pin-assignments, explicitly naming A1&A2

jkriege2:
fixes https://github.com/KiCad/kicad-symbols/issues/182

2018-01-28 16_41_11-symbol library editor - device d__kicad_kicad-symbols_device lib


poeschlr:
For reference: original discussion over at https://github.com/KiCad/kicad-library/issues/1814

@evanshultz you seem to have a better understanding about this than i do. Could you take a look at this if @Samonitari does not chime in.

jkriege2:
remaining KLC-violations on symbols from this PR (TRIACs only, not the LED) are due to using small text for pins to make labels NOT interfere with the symbol. Should be covered by http://kicad-pcb.org/libraries/klc/S3.2/

evanshultz:
@jkriege2 
Thanks Jan! These look nice and should resolve the issue. The user will have to think through which symbol to pick and then match up a footprint on their own, but TRIACs aren't common and this burden exists for many other generic symbols as well.

Comments:
1. I don't like using a Wikipedia link for the datasheet. I'd rather have it blank for these generic symbols.
1. While you're in there, can you change `Q_Thyristor*` symbols to `Q_SCR*`?
1. I assume making this lib version 2.4 is intentional?

evanshultz:
@Samonitari Can you please see if this resolves the issue you brought up? If we don't hear from you in a couple days, I propose this PR gets merged.

jkriege2:
Hi!

1. I removed the Wikipedia-links and added ~ to all devices in device.dcm that didn't have it before.
2. I could ... but as I'm not an expert I consulted Wikipedia, which says (https://en.wikipedia.org/wiki/Silicon_controlled_rectifier):
```
The name "silicon controlled rectifier" is General Electric's trade name for a type of thyristor. The SCR was developed by a team of power engineers led by Gordon Hall[4] and commercialized by Frank W. "Bill" Gutzwiller in 1957.
[...]
According to Bill Gutzwiller, the terms "SCR" and "controlled rectifier" were earlier, and "thyristor" was applied later, as usage of the device spread internationally.
```
So which name should we choose ... if SCR is really "only" a marketing name, I would tend to keeping thyristor ... if not I'm open to rename them! Both names are in the description/keywords anyways!
3. Change to 2.4 was done because I use KiCAD nightly to edit ... and we decided it will be fine as there will be a cleaned up v4-version for download on a branch and this is meant for v5

evanshultz:
1. Thanks!
2. "Thyristor" encompasess many types of parts. SCRs are one type, as are TRAICs. "Thyristor" is too generic to describe a diode with a gate. Some supporting info: ST's classification system at http://www.st.com/content/ccc/resource/sales_and_marketing/promotional_material/selection_guide/a5/52/c3/da/5f/68/4e/97/sgthyracs.pdf/files/sgthyracs.pdf/jcr:content/translations/en.sgthyracs.pdf (note that the term "SCR" is always provided so you know which type of thyristor), ST datasheet http://www.st.com/content/ccc/resource/technical/document/datasheet/group3/d0/28/b3/ff/a7/b4/4c/e5/DM00314016/files/DM00314016.pdf/jcr:content/translations/en.DM00314016.pdf, Digi-Key's discrete semiconductor taxonomy https://www.digikey.com/products/en/discrete-semiconductor-products/19, and Wikipedia's page for TRIAC https://en.wikipedia.org/wiki/TRIAC (if you believe that). Using only "thyristor" in the symbol does not indicate which type of thyristor, while using "SCR" does, so I believe that is the better way to go for users so they can more easily find the part they want.
3. Yep. Just making sure it was intentional so no reverting is necessary later. :)

jkriege2:
renamed as requested.

jkriege2:
@evanshultz Can be merged?

Samonitari:
Sorry again, I broke my hand last week, so I had my queue of task building up...

Triacs are fine this way, thak you!
- Labels are a nice addition to the symbols, it makes easier to re-check the schematic, or makes the reviewer's life easier - I think.
- A2 moved up - it lines up now with the empirical convention manufacturers use
- A1/2 or M1/2 on the other hand is interchangeable, even T1/2 appears in datasheets - glad you didn't rename them, as it is less of a change (although already a breaking one). Accidentally, common layout of TO-220, TO-252 p\ackaged triac is matching with the first in alphabetic order.

SCR is also better this way.
Have a nice day!

evanshultz:
@Samonitari 
It sounds like you're happy now, and thanks for pointing out this issue.

@jkriege2 
Yep, I was just waiting to see if we got a response from anyone else. And we did! Thank you!!

PR #242, opened 2018-01-28
Add BUK9M53-60EX

strombom:
BUK9M53-60EX

https://assets.nexperia.com/documents/data-sheet/BUK9M53-60E.pdf

image

image

Footprint pull request: https://github.com/KiCad/kicad-footprints/pull/313

------------


jkriege2:
since you removed all whitespace before the A and V-unit-sybols, could you please do the same for mOhm?

strombom:
Ok removed whitespace

evanshultz:
The footprint name change at https://github.com/KiCad/kicad-footprints/pull/313 will cause this PR to need a corresponding change.

evanshultz:
Thanks again! I really appreciate your hard work to want to improve the library and follow through to get it done!

@poeschlr 
Can you check out the CI output here? It's complaining about the pin text offset being 1mil on this symbol. But many other symbols are the same. Is our script not correctly catchinging all these warnings?

jkriege2:
no, the scripts work fine. I noticed that several "small" symbols (also in device.lib) have the wrong pin-offset set. It's not a visible problem, as they don't have pin-labels ... bt still the scripts check that and complain. We should do a cleanup for that at some point.

jkriege2:
I opened an issue for taht: https://github.com/KiCad/kicad-symbols/issues/260

strombom:
Is it this value?

image

I'm unable to change the value, it's not saved...

evanshultz:
@strombom 
That's fine. Jan's PR will clean up the whole lib anyway. Thank you!

PR #243, opened 2018-01-28
AdUM12XN

matyro:
Added 150 Mbs Isolator

http://www.analog.com/media/en/technical-documentation/data-sheets/ADuM120N_121N.pdf

adum121n
adum120n


jkriege2:
nice work!

Thanks,
JAN

PR #247, opened 2018-01-29
Add nRF52810-QFxx, nRF52810-QCxx and fix small issues with nRF52832-QFxx

jcyrax:
nRF52810-QFxx and nRF52810-QCxx (package difference only 48/32 pin) based very much on existing symbol nRF52832-QFxx

Datasheet nRF52810-QFxx and nRF52810-QCxx:
[http://infocenter.nordicsemi.com/pdf/nRF52810_PS_v1.1.pdf](http://infocenter.nordicsemi.com/pdf/nRF52810_PS_v1.1.pdf)

**nRF52810-QFxx**
symbol_52810-qf
descript_52810-qf
filter_52810-qf

**nRF52810-QCxx**
symbol_52810-qc
descript_52810-qc
filter_52810-qc

Question for this part footprint. According to datasheet it has exposed pad 3.5x3.5mm. We have footprint in kicad-footprints: QFN-32-1EP_5x5mm_P0.5mm_EP3.45x3.45mm It is within tolerance range of part, but nominal is below. Should I create new footprint with exact EP size? As of this I created entry for exact footprint that is not found in library.

Regarding fixes to **nRF52832-QFxx**:
nRF52832-QFxx was missing footprint type in description.
Symbol was not centered (a bit asymmetric) for no obvious reason.
Some pin names used where not exactly as in datasheet:
1) Used SWCLK, but in datasheet it has only reference as signal, but in all pinouts pin name is SWDCLK.
2) Used RST, but in datasheet pin name is RESET or nRESET so changed to ~RESET

symbol_52832-qf
descript_52832-qf

Non of them have aliases.

------------


jcyrax:
Updated symbols
image
image
image

evanshultz:
Regarding the QFN package question, the copper pad size is typically larger than the slug size. The slug is 3.5mm square, so generally the copper pad size would be 3.8-4.0mm square. The copper pad cannot be smaller or you might put traces or vias or something under the slug on the part. So a new QFN32 footprint will need to be created.

Similarly, the QFN48 footprint is simply too tight with only 0.06mm oversize on the Cu pad. We will need a new footprint here with a bigger Cu pad.

Review notes:
- Would it make more sense for the ANT pin to the Passive type? It only goes to passive components.
- It seems weird for the GPIO pins to wrap around the part clockwise. What if they went top-to-bottom on both sides, left first and then right?
- Part of me would like the GPIO all on the right side, with the left side for the crystal (pins spaced apart a bit), clocks, etc. What do you think?
- Pin 44 of the QFN48 package is named "NC", not "N.C".
- Can you put the ref des above the PN instead of below it?
- For me, like most people, I feel left-to-right is natural so it's odd that the DEC pins are RTL. Can you flip them? It will also make the pin numbers increment LTR.
- In the MPN, the first "x" seems to always be an "A", while the second may be another "A" or a "B". Do you know what this means? Temperature rating? At any rate, would it make sense to change the first "x" to an "A"?
- On nRF52832, pins 11 and 12 should have the pin names swapped around the "/".

jcyrax:
1) To be honest I basically copied existing nRF52832 without much consideration. I actually agree on your points. I'll rework those symbols. 

2) Regarding footprints there may be a problem. I can't seem to find no official dimensions from Nordic. I could extract from demo board gerbers.
There was this PR for Nordic specific QFN https://github.com/KiCad/kicad-footprints/pull/184 but I cant seem to find source from which it was made. Maybe he used gerbers also?

evanshultz:
@jcyrax 
1. Thanks! You can just pile on the updates to this PR if you don't mind. Please continue adding new screenshots of the symbols as that helps review.
2. Yes, Nordic doesn't provide it. Pretty lame. Can you take a look at https://github.com/KiCad/kicad-symbols/pull/340 and work out the minimum EP size we will need for the new footprints?

jcyrax:
Made most of changes.

> In the MPN, the first "x" seems to always be an "A", while the second may be another "A" or a "B". Do you know what this means? Temperature rating? At any rate, would it make sense to change the first "x" to an "A"?

This is function code. I think used only for FLASH and RAM size. I think we should leave it as is or maybe remove trailing xx altogether. Added only because other symbols in lib from Nordic used this style.

New symbols:
image
image
image


About package names.
I recalculated using IPC formula and now I have:
Package_DFN_QFN:QFN-32-1EP_5x5mm_P0.5mm_EP3.55x3.55mm
Package_DFN_QFN:QFN-48-1EP_6x6mm_P0.4mm_EP4.65x4.65mm

There is QFN-48-1EP_6x6mm_P0.4mm_EP4.66x4.66mm in library allready used previously for this part.
Should that be used?

I could make these footprints, but what should be the source. I can try to use that older IPC document or maybe gerbers from Nordic.

evanshultz:
This looks so much better. Thank you!
- Can you space apart the crystal pins so a crystal symbol fits nicely there? There is plenty of room.
- I'm unsure of The combo GPIO/reset pin. Is it best on the left or the right?

It's fine to keep the PNs like they are with "xx". I just didn't see the meaning in the datasheet. Thanks for checking!

There was a small error in the calculation from the linked PR. Can you check again and verify? The 4.66mm EP will be OK for the QFN48 if that matches the IPC guidelines. We can always use a bigger EP, as long as we have adequate clearance from the perimeter pins, if a suitable footprint is available in our library.

jcyrax:
Pin spacing changed for xtal:
image
image
image
image

1. About reset, I personally use left, because I try to not use reset pin as long as I can somehow fit my needs. Then debug interface stays somewhat together. For somebody who will need all pins it may be not so comfortable. Either way I really don't care that much.

2. I recalculated EP size, 4.65mm for QFN-48 and 3.55mm for 32pin. So I switched to original footprint with 4.66mm.
I found thread on Nordic dev forum about footprints in general: https://devzone.nordicsemi.com/f/nordic-q-a/19860/qfn-land-pattern-in-reference-design/77244#77244
Relevant:

> The QFN footprint we're using should follow the IPC standard, and is generated using the IPC footprint generator in Altium. But if you need to extend the pads, feel free to do so.

evanshultz:
@jcyrax 
Thanks! Everything is good except that we're missing one footprint. I'd just like the ask one question about that before I merge. Are you able to contribute the footprint? It's not required but would be nice to have a footprint for this symbol that we know is valid.

jcyrax:
I actually already made one I will use, but I'll open PR tomorrow to see if it can be merged in official libs.

evanshultz:
Thank you!!

PR #249, opened 2018-01-29
Add A4403GEU symbol

jcyrax:
**A4403GEU** buck regulator:
https://www.allegromicro.com/~/media/Files/Datasheets/A4403-Datasheet.ashx
There is no link for actual file in manufacturer site!
symbol
descript
filter

Regarding footprint:
There is footprint that is matching in size, but it is UQFN type. As a quick lookup showed UQFN is 0.5mm, but this chip is 0.75mm. Thats why I created entry for footprint that matches even height, but is not in library. I will probably later create it. I guess question is if I should have maybe just used UQFN?

------------


jcyrax:
I've stacked GND pins 
symbol-fixed

poeschlr:
Adding a new qfn footprint is the correct idea. Would you be willing to make a pull request for it over on the footprint library?

jcyrax:
Yes, I will try to create footprint.

jcyrax:
I've fixed some things for this symbol. Datasheet calls for NC pins to be tied to GND. So I renamed them to GND and stacked with other GND pins.

image

jcyrax:
Latest symbol screenshot:
image

Misca1234:
Any objections to merge this ?

evanshultz:
I don't see that this ever got a review.
- What does the "G" in the part number mean? If it's for temp we would wildcard it; I looked a bit and didn't find out what it indicates in Allegro MPNs.
- To help users choose this part, could you add `46V, 3A, 2MHz, ` before the package type in the description?
- The footprint filter needs `1EP` in it.
- I'm inclined to make pin 12 Passive type. It's only connected to passive parts and it's not the actual output of a buck converter; the other side of the inductor is. If the LX pin is connected to parts that source current they won't get it since a buck converter needs the inductor. So making pin 12 Passive type will catch this mis-connection and the user will need add PWR_FLAG after the inductor anyway. I don't see a downside but I do see a possible upside.

Misca1234:
> What does the "G" in the part number mean? If it's for temp we would wildcard it; I looked a bit 
> and didn't find out what it indicates in Allegro MPNs.

It is the part number, top on page 2

> The footprint filter needs 1EP in it.
This component have a dedicated foot print

jcyrax:
> What does the "G" in the part number mean? If it's for temp we would wildcard it; I looked a bit and didn't find out what it indicates in Allegro MPNs.

I also couldn't find any decoding of them when created this.

> To help users choose this part, could you add 46V, 3A, 2MHz, before the package type in the description?

Added, but for frequency specified "up to", because you can set it with resistor.

> The footprint filter needs 1EP in it.

Fixed. 

> I'm inclined to make pin 12 Passive type. It's only connected to passive parts and it's not the actual output of a buck converter; the other side of the inductor is. If the LX pin is connected to parts that source current they won't get it since a buck converter needs the inductor. So making pin 12 Passive type will catch this mis-connection and the user will need add PWR_FLAG after the inductor anyway. I don't see a downside but I do see a possible upside.

I agree.

PR #251, opened 2018-01-29
Add symbol for ADM1491EBRZ

jcyrax:
ADM1491EBRZ
[http://www.analog.com/media/en/technical-documentation/data-sheets/ADM1490E_1491E.pdf](http://www.analog.com/media/en/technical-documentation/data-sheets/ADM1490E_1491E.pdf)
symbol
description
filter

No Alias parts

*Edit text formatting

------------


evanshultz:
@jcyrax 
Thanks for submitting this PR. A few requests:
1. By "No Alias parts" do you mean that you checked the existing symbols and none of them are package- and pinout-compatible with this component?
1. Can you stack the VCC and GND pins? Search recent PRs for examples and explanation.
1. Please remove the trailing "Z" from the symbol name, leaving `ADM1491EBR`.
1. The pin line for ~RE isn't lined up with the circle. Please fix that.

When you make the changes, please post another image of the symbol for easy checking. Thanks!

jcyrax:
1. I meant that this part has no aliases, because I didn't include screenshot of Alias section. But now I went and checked that in fact there aren't parts with matching symbol pinout.
2. Mostly didn't do because I imagined that if symbol size isn't influenced there is no need to stack power pins. Should I just do it always no matter what? I have some other PR that would then need adjustment.
3. done
4. changed

symbol_fix

evanshultz:
@jcyrax 
Sorry! I missed your reply.

2. Yes, please always stack pins. It's at least colloquially done everywhere. Thank you.

One point that I didn't notice earlier: my mind typically sees positive as being "above" negative. Pin 10 is above pin 9, however, and that seems backwards to me. But since the datasheet shows this layout, I'm fine to leave it like this. Just a note as I looked at your image above.

And lastly, would you be willing to make any of other parts in the same datasheet? There is an MSOP package for 1491 plus two packages for 1490. If not, it's fine and I can merge this.

jcyrax:
@evanshultz yes, I'll add the rest of parts. Also question for style. I used same thickness for internal drawings as box outline. Do you think it looks ok or maybe I should use smaller thickness for internal drawings?

evanshultz:
Thank you! I think the symbol is OK like with lines all the same thickness, but let's get more feedback.

@poeschlr @jkriege2 
I see several in our lib that are the same thickness. But I also see some where the outline is thicker. KLC S3.3 and S3.4 specify the outline thickness but are silent on the line thickness of internal lines. What do you other librarians think?

poeschlr:
I think for inner details we can allow a range of line thicknesses as long as any thickness is specified.The reasoning behind this is that lines with no thickness are rendered differently in depending on kicad version. (And yes we should put that into the KLC.)

Example:
In my connector symbols i used 6mil for inner details and 10 mil for the outside rectangle. This is because i have small details that would not look good otherwise.

evanshultz:
@jcyrax 
You've made the outer symbol box 10mil and that's the only real requirement we have now. Thinner inside lines may look better, but that's personal preference and many symbols have the same line thickness.

Sorry this took so long to answer all the questions. Thank you!

PR #257, opened 2018-01-30
Add FAN7371

strombom:
FAN7371
https://www.fairchildsemi.com/datasheets/FA/FAN7371.pdf

image

image

------------


evanshultz:
Thanks for your continuing contributions!
1. Set pin text offset to 20mil or less.
1. Add NC pins 3 and 5.
1. Pin 8 should be of type Passive (`P`).
1. Please add output current and voltage to the description.
1. Add package type to the end of the description.

strombom:
1. OK
2. KLC S4.5 states that unconnected pins may be omitted from the symbol?
3. OK
4. OK
5. OK

image

evanshultz:
The wording of S4.5 is poor, and worse than the previous version. Thanks for pointing that out. Unconnected pins must be shown unless their count is so great that we instead put pin count in the FPfilter. The reason to include NC pins is to the pin count filter in CvPcb will identify only matching footprints. This is dealt with on a case-by-case basis.

Thanks once again for making the libraries better!

PR #259, opened 2018-01-30
Added 2 UBIQ mosfets

matyro:
Added two high power ultra low GateCharge MOSFETS's.
The center pin is cut off, so I am not sure if this is the correct footprint. But it should work

http://www.jaolen.com/images/pdf/QM6006D.pdf
http://www.jaolen.com/images/pdf/QM6015D.pdf

qmd6015d
qmd6006d


evanshultz:
@matyro 
Thanks for contributing!
1. Merge your branch with master and I believe these can be an ALIAS of `IPD50R380CE`. That symbol was added to master today. (At the very least one your symbols should be an ALIAS of the other.)
1. Typo of `QMD6015D`; it has an extra 'D' and should be `QM6015D`.
1. Because the D connection doesn't have a gullwing pin and only the tab, use the package `TO-252-2`.
1. Description fixes: Add parameters to `QM6006D` (the other one has them), edit `QM6006D`'s description and keywords to look like the other FETs, change Rds units to mOhm as other FETs use, capitalize the `O` in `Ohm`, add package name at the end.

matyro:
Q6015 is a PMos 
Q6006 is a NMos 
I do not think that i can alias them to each other.

I can alias them but why is it necessary here?

evanshultz:
Oh yes. My mistake. You can ALIAS only the N-channel to the symbol I mentioned above. This ensures consistency between symbols, makes the file size smaller, and is easier on librarians. That is why it is necessary.

matyro:
Hi, I checked IPD50R380CE and it is not pin compatible to Q6006D. Drain and source are exchanged.

evanshultz:
@matyro 
You're mistaken. They are pin compatible, using the standard FET pinout for DPAK.

https://www.infineon.com/dgdl/Infineon-IPD50R380CE-DS-v02_03-EN.pdf?fileId=db3a30433ecb86d4013ed0a2ef580f38
image

http://www.jaolen.com/images/pdf/QM6006D.pdf
image

The pins are labeled differently in each datasheet, but they're electrically the same. These FETs can and should be as ALIAS of each other in the KiCad library.

matyro:
ipd50r380ce
footprint

Maybe I understand something wrong, I am only a hobby person but source is here conneted to the tab.

evanshultz:
I see now. These FETs can be an ALIAS of each other since they are electrically compatible, but the IPD50R380CE has pins 2 and 3 swapped in the symbol. I've looked at too many PRs today and need to take a break...

If you could make your N-channel an ALIAS and swap those pins in the IPD50R380CE symbol, it should work and all will be well. Right?

matyro:
Fine for me,
should I make a seperate PR for the fix?

evanshultz:
OK. Good. If you merge master into your branch, you will have the new symbol and then you can modify it here. No need to make another PR.

evanshultz:
4. Can you please address this item from my first post?
5. QM6015D's pin 1 is not on grid. Move it 50mil to the left, make it 100mil long, and extend the line.

matyro:
>5.

I copied from FDS6890A so i did not notice the error

>4. 

$CMP QM6006D
D 60V Id, 35A Vds, N-Channel Power MOSFET, 18mOhm Ron, Qg (typ) 19.3nC, TO252
K N-Channel MOSFET
F http://www.jaolen.com/images/pdf/QM6006D.pdf
$ENDCMP
#
$CMP QM6015D
D -35A Id, -60V Vds, P-Channel Power MOSFET, 25mOhm Ron, Qg (typ) 25.0nC, TO252
K P-Channel MOSFET
F http://www.jaolen.com/images/pdf/QM6015D.pdf
$ENDCMP


Some mosfets specify Qg others do not. Some have Ronon instead of Ron.

evanshultz:
Let's re-arrange gate charge to `19.3nC Qg (typ)` on both FETs and then it's all good. If you can commit that I can merge. Unfortunately there are quite a number of minor cleaning up that needs to be done.

evanshultz:
KLC (http://kicad-pcb.org/libraries/klc/) covers the library requirements and suggestions for KiCad's official libraries. S4.1 (http://kicad-pcb.org/libraries/klc/S4.1/) says that pins need a 100mil length except in rare cases.

I'll merge this now to get in your symbols and the pin fix. In the future all pin lengths should be fixed.

Thanks again and sorry for the confusion!

PR #265, opened 2018-01-31
Move opto gate drivers to Driver_FET

El3ctr0nic:
Move:
ACPL-P343
ACPL-W343
TLP250
HCPL3120

Add:
HCPL-314J

Fix FP filters, update doc. links
------------


evanshultz:
Thanks! It does make sense to place them here since they are intended for driving gates and no coupling signals across a barrier. Review notes:
1. Please add package type to end of the description, as is done for `HCPL-3120`. (Some of them lost the package name through this PR.)
1. `HCPL-314J` needs a 12-pin footprint while you've assigned a 16-pin version. This will not afford the creepage/clearance between the sections. Will you be able to contribute an appropriate footprint? If not, that's OK, but we'll still need to change the default footprint in anticipation of the correct footprint being added.
1. Please split `HCPL-314J` into two units. It will be easier to place this way and each symbol will match the existing HCPL part.

evanshultz:
@El3ctr0nic 
Please improve your commit messages. "Fix" and "Quick Fix2" are very unhelpful.
2. Are you willing to generate the necessary footprint?
4. HCPL-314J has two pin 14s.
5. VEE pins should be of type Power Output.
6. A footprint with missing pins that counts the full number of pins (so pins are missing but their number is not missing) flip the pin counts in your footprint. It should be `SOIC-16W-12_7.5x10.3mm_P1.27mm`.

El3ctr0nic:
I will generate missing footprint.
Sorry for mess but i was a little bit confused - lib name change from SOIC to SO "just in time" :).

About footpirnt name  KLC :
F2.4 - Footprint naming for non-standard pin numbering - says remaining pins first next remaining+numner of removed - if i undestand correctly. Sorry my English isnt good ;)

evanshultz:
Unfortunately, the text on the website is incomplete or wrong.

See page 2 of http://ohm.bu.edu/~pbohn/__Engineering_Reference/pcb_layout/pcbmatrix/IPC-7x51%20&%20PCBM%20Land%20Pattern%20Naming%20Convention.pdf, the section Hidden and Deleted pins. This symbol falls into the Deleted Pins category since only 12 pins are used.

@poeschlr 
Can you please check this out to be 100% sure I'm correct? We should reference the IPC doc above (at least by the publication number) and capture both Deleted and Hidden pins in KLC. For another example, see https://github.com/KiCad/Housings_DFN_QFN.pretty/pull/66.

evanshultz:
Thank you!

I believe all the SOIC footprints are script generated, and that is the best way to make footprints.

evanshultz:
I won't merge now so that comments can still be submitted to get @poeschlr's response, but this PR is ready to merge.

FYI, it appears the footprint generator script for this type of footprint is at https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_SOIC/package_soic_sop.py. Or you can do it manually.

PR #267, opened 2018-02-01
Add PS8802-x

strombom:
PS8802-1
PS8802-2
http://www.cel.com/pdf/datasheets/ps8802.pdf

image
image

image
image

Footprint: https://github.com/KiCad/kicad-footprints/pull/333

------------


strombom:
All other optoisolators have all pins of passive type.

Should I remove the names of the pins so that travis allows them to be of passive type instead of power pins (which travis wants on top/bottom but which does not fit those components)?

jkriege2:
Hmmm .. I would leave the pin-names, but make types to PASSIVE, we can then accept this exception to KLC. I think passive makes sense here, as people might want to use this OC in another way than connecting the emitter diectly to a power-source/GND and keeping the names makes sense for reference to the datasheet.

strombom:
Ok now they are passive again.

image

image

jkriege2:
Please fix the footprint library names and add a strike-through-line through the photodiodes in PS8802-2

Thanks,
JAN

strombom:
Updated library and strike-through lines.

image

image

jkriege2:
Thanks,
JAN

PR #271, opened 2018-02-02
Add DSPIC33FJ256MC710A

antoniovazquezblanco:
Add DSPIC33FJ256MC710A

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/70594C.pdf

imagen

imagen

imagen

------------


poeschlr:
Is this symbol script generated from official data or did you hand assign the pin numbers and pin names?

antoniovazquezblanco:
Hand made. No script.

Misca1234:
I think you should close this push and recreate it , you have affected way more components than just adding the DSPIC33FJ256MC710A

You description line should be change from
40 MIPS, 256k Flash, 30k RAM 
into 
40 MIPS, 256k Flash, 30k RAM, TQFP-100

With little effort you could add 
dsPIC33FJ128MC710A
dsPIC33FJ256MC510A

As aliases

poeschlr:
His PR does touch more symbols then adverdiced because he used a newer version of kicad then the last contributor.

There is a difference in line ordering between different kicad version.

As he seems to use one of the newer nightlies his version of the line order is expected to more closely resemble kicad version 5. (Unless the file save lib algorithm is changed until then.)

antoniovazquezblanco:
> I think you should close this push and recreate it , you have affected way more components than just adding the DSPIC33FJ256MC710A

> His PR does touch more symbols then adverdiced because he used a newer version of kicad then the last contributor.

Sorry but I am not willing to redo this contribution under an older version of KiCad. That would mean to uninstall my current setup to do something and then go back to the original state to be able to still work with the new projects.

Description line changed.

Multiple aliases added.

poeschlr:
@antoniovazquezblanco as i already stated using your version of kicad is totally fine (even prefered), the discussion is mood anyways.

antoniovazquezblanco:
@Misca1234 Anything else you think I should modify? :)

evanshultz:
@antoniovazquezblanco 
Sorry that this lingered for so long. You've done great work! A couple more things:
- Please stack power and ground pins.
- The description is only a few parameters, add `High-Performance, 16-bit Digital Signal Controller` and then the current description.
- Pin 21 is AN4 (not AN5).
- Pin 28 is VREF-.
_ Please move VCAP to the top of the symbol. This will give more room for a capacitor symbol and also put it with the other power supply pins (this is a power supply output but I don't see that any current can be drawn from it so Passive type is correct).

antoniovazquezblanco:
VCAP is not exactly a power supply output. AFAIK it should not be used as a power supply output and it should not be connected to anything imposing a voltage in that node. It is there to connect a capacitor that can ensure the correct operation of the voltage regulator for the uC core.

Updated everything as requested:
imagen
imagen

evanshultz:
Thank you once again for your great work! I really appreciate you adding the extra TQFP100 parts too!

antoniovazquezblanco:
Thank you!

antoniovazquezblanco:
Sorry to reopen this. Specially now that I have seen other threads where the following topic is of extreme toxicity.

I believe we made a mistake when stacking the VCC pins due to the fact that each pin should have a capacitor as suggested in Microchip datasheet. Now the user does not know how many VCC pins there are thus having to go back to the documentation again.

I think I will open an Issue to remind me to fix this if people does not provide arguments for me to not do this.

PR #274, opened 2018-02-02
Added LM5166 alias to LM5165

matyro:
Hi,
I added the alias for LM5166 to the symbols and filled in the necessary informations.

http://www.ti.com/lit/ds/symlink/lm5166.pdf


matyro:
grafik

This footprint is available for me

evanshultz:
@matyro 
Revert the change to the footprint as it should have the EP size specified. There may be other sizes of pads in the future and we want the one originally there. The rest looks good. Thanks!

evanshultz:
Actually, we have this exact manufacturer-specific footprint at https://github.com/KiCad/kicad-footprints/tree/master/Package_SON.pretty. Use `Texas_S-PVSON-N10_ThermalVias` with an FPfilter of `Texas*PVSON*N10*`.

matyro:
Whats interesting for me is that KiCad did change the footprint itself and I could not find the old one anymore.

evanshultz:
Have you updated your libraries? There's been a lot of changes recently.

Thanks for this contribution!

PR #278, opened 2018-02-03
Added the optocoupled isolator TLP2770

Misca1234:
Hello guys

Here comes TLP2770 

https://toshiba.semicon-storage.com/info/docget.jsp?did=53548&prodName=TLP2770

bild

bild


Misca1234:
Have something changed?
TLP2745 points to the foot print library Package_SO for the SO-6L foot print 
TLP2770 wants the same foot print

But the SO-6L is not in Package_SO but instead in Package_SOIC

Shall I make a push and update all symbols that points to Package_SO to point to Package_SOIC instead ?

jkriege2:
I think you should not overlay text and internal lines ... best break the line to leave some space for the pin-labels, like here: 
2018-01-28 17_26_30-symbol library editor - interface d__kicad_review pretty_kicad-symbols_interfac

Misca1234:
Changed the symbol into this

bild

jkriege2:
Thanks,
JAN

PR #279, opened 2018-02-03
Add UCC21520DW to Driver_FET

El3ctr0nic:
Add UCC21520DW to Driver_FET

------------


El3ctr0nic:
obraz

strombom:
Only a comment on your schematics, pin 4 on Q207 is not connected ;)

strombom:
The sombol is very well drawn.

In an example in KLC they leave the W in the footprint filter:
SOIC*W*7.5x10.3mm*P1.27mm*

Small spelling error: the description says "Output Cirrent".

Digikey normally specifies worst case parameters, I'm not sure how it's normally done in the KiCad library, but worst case propagation delay is 30ns. Usually one have to use the worst case parameters during design.

I have been asked to write electrical specifications on the form `<param><unit> <value>`, that would make the description look something like this:
"Isolated Gate Drive, 4.0A I_o+, 6.0A I_o-, 30ns t_pd, SOIC-16W"

But I think your description is easier to read:
"Isolated Gate Drive, Output Current 4.0/6.0A, Propagation Delay 19ns, SOIC-16W"

Would be nice to hear some comments on above.

It can also be a good idea to attach a screenshot from the symbol editor so that NC pins can be seen:
image

evanshultz:
@El3ctr0nic 
When submitting a PR, always link the datasheet and give an image. You posted an image above (@strombom's is better) but no datasheet. It is http://www.ti.com/lit/ds/symlink/ucc21520.pdf.

1. Would you mind to add the 'ADW' version as an ALIAS? Not required, but quick and easy. It appears the difference is the UVLO threshold.
1. Stack the VCCI pins since they're internally connected.
1. Propagation delay isn't important to have in the description, but isolation is. Please change it to `Isolated Dual-Channel Gate Driver, Output Current 4.0/6.0A, 5.7kV Isolation, 8V UVLO, SOIC-16W`. Each library may have different formatting for the description to highlight important parameters for that library.
1. The keywords can be simply `dual isolated gate driver`.
1. All power pins should be Power Input type (`W`). Power Output requires a power flag to avoid an ERC error.

One last note: instead of making changes to master (head of the repo), I suggest you use a branch. You can use master, but it's generally not recommended. Your choice.

El3ctr0nic:
obraz

Should be ok now.

evanshultz:
Thank you!

PR #280, opened 2018-02-03
Ds pic33 ev32 gm002

Misca1234:
This push consist of all CPU's in the data sheet
http://ww1.microchip.com/downloads/en/DeviceDoc/70005144e.pdf

The QFN symbols have the suffix _QFN

A typical view of the metadata is 
bild

bild

bild

bild


strombom:
Hi,

The power pins are not located on a 100mil grid so they need to be moved. The symbols should also be centered.

Furthermore the pin names should be shorter, I have been told that only pin names should be displayed and at most one major function. 

When the pin names are shorter it should be possible to have pins on the left and right sides in order to make the symbols smaller and symmetric.

Misca1234:
Jee, I was a little uncertain about that, 
The problem is that the "one major function",depends on what application the CPU is going to be used for

poeschlr:
It is not about keeping only one major function, but only keeping enough information such that each pin is uniquely identifiable via the datasheet.

In general the pin name alone should be enough. But some processors have some major funcitons (like example a reset) on a gpio pin. In that case it is allowed to add that major function as well.

Misca1234:
Centred the objects, made the pin names shorter 

bild

bild

bild

bild

evanshultz:
@Misca1234 
I'll let someone else do the official review, but here are a couple things:
1. Separate the packages in the description with forward slashes, like `DIP-20/SOIC-20`.
1. Stack power and ground pins that are tied together.

Misca1234:
I have changed the ',' into a '/' in the description to separate the different packages

But I wont change the pin stacking thing because I intensively despise that sort of thing, 
It is no good at all or in any way and it just confusing for the user.

I did it once for a component last year and I still regret it, I know that I have to live with it for the rest of my life.

poeschlr:
The thing is that the kicad library is build by multiple people. Everyone has (sometimes strong) opinions how symbols or footprints should look like. If we accept every contribution that is build to someones personal standard we will never achieve a consistent library. This is why the official kicad lib has rules in place to ensure uniformity.

One of these standards is that pins that are always connected should be stacked. (I can't remember when this was decided. It might have been decided before i started to contribute.)
This is one of the things where there is no right answer as both sides have equally valid arguments. I am not prepared to reverse this decision as it would require a lot of work to change all symbols back to using non stacked pins. (And a few month later another contributor will be adamant that all pins have to be stacked. Should we then reverse again?)

jkriege2:
I think the stacked pins are not a must (KLC says stacks are "desirable", not a requirement)? ... Anyways I thin they make sense as they clean up the symbol and the pins ought to be connected together in most cases anyways. For special cases, where the pins need to be separated (e.g. because they should be on different nets for some reason or other), I think the user can alter the symbol???

Misca1234:
At every review meeting or any time a person see the schematic the designer will hear
"wtf is this, you have forgot to connect several pins of the chip"

If the schematic is redrawn in a new CAD tool based on the schematic drawing with stacked pins the new designer will say
"What was this for teenage amateur designer, he forgot to add several pins of the chip"

This stacked pin thingy invalidates traceability or the possibility to review the schematic with the comment
"I promise you, they are there, they are just invisible"

Like a programmer who use a language with invisible code would get away in a RISE/TÜF certification of the software
"Yes, the security algorithm that prevents people for being squeezed is there, it is just invisible"

jkriege2:
As a side note (not really wanting to start a discussion ... and not beeing a professional electronics designer, but "merely" a physicist and software developer): From a software POV the stacked pins make sense: The schematic is a logical representation of the circuit. In this view, the chip actually has ONE GND-pin, as all pins in the footprint are internally connected. so I wouldn't expect to hoave to conect each and every of potentially several GND-pins separately. Just say: This chip should be connected to that GND-signal and let the software figure out, which pins of the footprint that are. the info is encoded in the lib and if I'm interested and want to check, I can go to the lib and have a look, but to understand the circuit there is no need to have all the GND-pins. This way (with the future KiCAD lib-format) we will be able to have one logical symbol and you simply select the desired package. then the software can figure out from the pin-pad-mapping, which pads to connect to which pin. Then e.g. changing from a simple 7805 in TO-92 to SO-8 is very simple, you don't need to remove and readd another symbol, but merely select a new footprint and reroute. That's it.

This left aside: If you read carefully, I said that IMHO the pin-stacking is not a requirement per se, but more like a strong suggestion/wish for symbols in the lib. And the points that @poeschlr made about homgeneity in the lib and that not everybody will get what he/she likes most, we can just find a way to do things that satisfies most, are valid IMHO.

Therefore: Please don't let this get into a lengthy discussion fo what's the best way to do things in the world, as we will never agree on that thing as everybody has good arguments on his/her side!

@poeschlr @evanshultz any final remarks? How to proceed?

evanshultz:
Nope. I would discourage more discussion on this topic right now as I feel it's easily flammable. It is what it is now and there are much bigger tasks to tackle in the near term.

bobc:
My 2c : The current KLC (3.0.11) doesn't appear to state power pin stacking is mandatory, by omission the reader would assume it is optional.

I don't really like hidden pins for any reason, nor changing the type of power pins to passive for the purpose of stacking, but I understand the limitations of KiCad.

If there are a few power pins, I prefer those to be shown explicitly. It does not detract from the symbol nor cause undue effort to the user, but retains maximum traceability. 

OTOH, humongous symbols like 1184 ball FPGAs have so many power pins it becomes impractical to show them all. In that case, stacking power pins seems like a better compromise.

Either way, I have added a new optional feature to symgen to automatically stack power pins.

>  It is what it is now and there are much bigger tasks to tackle in the near term.

+1 to that, if we could agree what the current policy is and continue that for now. Maybe we should take the opportunity to get changes in KiCad for v6 when the file format changes, to better deal with the issues we encounter. I've started generating files in the draft "sweet" format to get an idea of the implementation.

Misca1234:
Agree, let us focus on the content of the push instead.

poeschlr:
The screenshots look good to me. (Ignoring the pin stack stuff)

Sadly i am a bit short on time regarding checking the symbol for pin correctness.
Did you generate them via a script from official data or did you make these by hand?

Misca1234:
They are hand made

evanshultz:
@Misca1234 
Because pin stacking has been decided to do, even if KLC isn't an explicit as it could be, and regardless if you like it, not stacking breaks with the current design of symbols in the lib. Please stack the power pins or close this SR.

Misca1234:
I wont participate in that stacking stupidity

evanshultz:
As a volunteer community, we must agree to agree. You may not like the result of some discussion if you're in the minority, but if you've presented your viewpoint well and what you want done isn't accepted then that's simply too bad. That's "agree to agree". We all come out of a decision together, even if we don't like the result.

Pouting and cursing doesn't help. A well-structured argument and mature behavior is your most powerful ally.

On this particular topic, you have shared why you don't like pin stacking but that decision was made some time ago. My reading of recent conversations is that Rene, most importantly, is continuing pin stacking. I believe much depends on the new symbol format which we hope brings a lot of improvements!

Misca1234:
Do you have the link to that conversation where this was decided so I can read what the arguments are for this decision.

bobc:
> Because pin stacking has been decided to do, even if KLC isn't an explicit as it could be, and regardless if you like it, not stacking breaks with the current design of symbols in the lib. Please stack the power pins or close this SR.

"Agree to agree" = "do it my way or take a hike".

You've turned into quite a petty little dictator haven't you? I can see why people are put off contributing with such arrogant behaviour.  Who put you in charge anyway?

poeschlr:
Ok as this seems to explode in our face i will make a statement.
I acknowledge that there are competing "worldviews". I can think of two:

---

Worldview 1: The symbol - footprint pair is approved separately to the circuits. It makes sense in such a case that the symbol ensures that it is hard to miss-use the component.  The circuit designer assumes it is correct. 
In this case it makes sense to have stacked pins as it removes a possible point of error. (The point of error being that the circuit designer connects one of pins that would be stacked in this workflow to the wrong net.)

---

Worldview 2: The circuit is the master of all information. It is the thing that is approved. All information must be visible in it. This workflow does not assume the correctness of symbol - footprint pairs. It makes it easy to check printed versions of the circuit.

---

I also agree that there are legitimate reasons to show all positive power pins. (Outside the scope of my defined worldviews)
If the datasheet suggests that a decoupling cap should be placed near each of these pins then they should be shown. Otherwise it is nearly impossible to communicate to the layout guy that a specific capacitor should be near some specific pin.

---

As you might have already guessed i don't think there will be one rule that encompasses every component. I prefer stacked pins. But if there is a good reason (outside the scope of the worldviews defined above) then i would say they should not be stacked.

---

TlDr: there are competing views. I prefer stacked unless in cases where i prefer all being shown.

evanshultz:
I did not intend to inflame anything. I was not vitriolic when I wrote anything above. You can read into it whatever you like, though. And it is easy to read text without being face-to-face and miss something that a real, live conversation can convey.

I have also felt strongly about some particular decision, but was on the losing side, so I went along with the final outcome. This is agree to agree. When the votes are tallied, I go with the outcome of the group and help contributors to make their submission align with the convention.

That doesn't mean, however, we can't revisit any decision. This isn't the right time to change, as Rene wrote above, but if we decide to not stack in the future I'm on board with that. Or if we table this discussion and come back to it when the library format is changed I'm fine with that too. Especially as the librarians change and time moves on we (I) should be ready to open discussion on any points. Hopefully keeping the perspective of KiCad's users in mind.

Re-reading the above, I'm not quite sure why this became so toxic so quickly. The same group of people covered stacking pins in another PR recently and Stefan made it clear he wouldn't stack. In this thread on Feb 5 Rene clearly said that pins should be stacked. There is no controversy. I only asked Stefan to make a choice about how he would like to proceed with this PR.

@bobc 
Nobody put me in charge. Oliver put Rene in charge when he stepped away.

Stefan's dislike for stacked pins is well-stated a few times and it clearly wasn't going to change with this PR. I also wouldn't be so blunt with a new contributor but Stefan has been around a while and proven to do great work. And was just promoted to a librarian. So I was more direct.

Because you feel empowered to point out when you think I've overstepped any bounds, please continue to do so. I intend to be welcoming if somebody is just getting the hang of things so they can grow into a strong contributor. If I haven't done that then I need to be made aware.

evanshultz:
I have been working hard over the past few days trying to get as many PRs resolved so contributors will see their work in the v5 release. This case lingers open and pin stacking has no resolution (either short- or long-term). Perhaps we can get an official decision?

@poeschlr 
You mentioned above:

> I prefer stacked pins. But if there is a good reason (outside the scope of the worldviews defined above) then i would say they should not be stacked.

Unfortunately, this is ambiguous. What is a good reason? The above has valid perspectives but no natural resolution or a mandate. Contributors are now desiring to create issues for pins to be unstacked (https://github.com/KiCad/kicad-symbols/pull/271#issuecomment-371102560). There's no issue if pin stacking is always required, for example. So upon what criteria is unstacking or not stacking allowed? How to know when to merge or not with regards to pin stacking? Do you want symbols to stack pins unless the datasheet calls out individual pins that need decoupling? What about if some pins need individual decoupling but others don't; are there several individual power pins and then one stack of the remaining pins? Or stack all pins for now until v5 is released? Or anything is accepted for now?

Feel free to break this off rather than keep in this PR body if warranted. This topic has been contentious many times and it doesn't need to be. Let's get an official decision and move on.

jcyrax:
I know its very late for this proposal, but I'll say it anyway.
Maybe stacked pins should all be set visible. You would see mess with overlapped pin numbers, but the reader of schematic would guess there is something more than one pin. On paper you would be left not knowing exactly, but on computer you can then check if needed.

Maybe we could ask for V5 change in eeschema to not display pin numbers if there are stacked pins, or maybe put ellipsis instead.

poeschlr:
@jcyrax before we do that i would rather not stack pins. A schematic should always be readable.

---

@evanshultz 
If we are lucky the new file format will give us more options. But until then we need to at least have some idea what we want to do.

I don't think a simple rule that is valid for everything can be found. 
Well one simple rule would be to disallow stacking but that would make most of our symbols non complaint and i want to avoid this. (And as i said i prefer stacked where it makes sense.)

---

I would define the rule set as follows.
Stacking pins is required for pins if:
- they have the same name in the datasheet
- and they need to be connected for the part to function correctly

An exception to this rule are pins where the datasheet states that each of these pins needs to be connected to a separate component (Example one decoupling capacitor per pin)
- This allows to at least hint at what pin should be connected to what other component. (Allows the schematic designer to communicate the intent to the person designing the layout.)
- Examples are the positive supply pins of micro controllers and other logic devices.

antoniovazquezblanco:
I personally agree with @poeschlr. The stated guidelines look fine to me and can be a temporal solution until better support for more options is developed.

evanshultz:
@poeschlr 
A few clarification questions as I try to anticipate future questions:
1. What if pins don't have the same name? Like there is a GND pin and the exposed pad is EP? In the past we would name them all the same, changing the EP to GND, so they could stack. Still do this?
1. If the datasheet says that some pins need caps but not others, are the leftover pins stacked?

I know you're busy at least this week, but it seems important to put the result of this decision into KLC for easy reference. Can you do that when you have time soon?

poeschlr:
For 1: Stacking pins with different names might create problems with kicad. (Even more then having differing electrical types.) So if pins are stacked they should have the same name.

For 2: If there are some left in such a case then yes these may be stacked.

As soon as we finalize the wording i will draft a KLC update. (But i still have to look closely at the comments of my last update. I hope i will have time sometime next week.)

Misca1234:
The discussion in this thread seems to have ended,
This thread will be deleted some time after 2018-03-14 21:00 CET

I do not know if the conversation will be deleted or will still be readable.
If someone have an interest to keep the conversion or move it elsewhere please do so before the above stated time.

Ratfink:
There were some important decisions made here that would be good to keep around for future reference.  We don't delete other closed PRs, why would we want to delete this one?

poeschlr:
Such discussions stay forever connected to the pull request even if it is closed and its branch deleted. (no matter if it is closed because it is rejected, closed because the contributor did not want to continue or closed because it is merged.)

It might have been better to have it as part of an issue with a meaningful name (limited to the scope of the discussion) That would have made it easier to search for it. But sadly there is no way to convert part of a pull request conversation to an issue. (github is very limited in this regard.)

The decision will be recorded in the KLC. I will link this discussion in the KLC pull request. So it should be possible to find it later on. I might even link it from the KLC history page. (Not sure about that as the discussion was kind of toxic at times and i am a bit uncertain if we should make it that public.)

---

The discussion ended because there was no further input. I don't know if my suggested compromise is acceptable to anyone. (I assume nobody is truly happy about it.)

PR #281, opened 2018-02-03
Add Si864x series

strombom:
Si864x series
https://www.silabs.com/documents/public/data-sheets/si864x-datasheet.pdf

Symbol copied from ISO7340C.

- 4 packages: 
  - QSOP-16
  - SOIC-16
  - SOIC-16W
  - SOIC-16W with 8 mm creepage. Has same footprint as SOIC-16W according to the datasheet: "allowing for extra creepage distance while maintaining standard WB SOIC-16 package dimensions and land pattern"
- 4 I/O configurations: 
  - 4 outputs, with EN pin
  - 3 outputs, 1 input
  - 2 outputs, 2 inputs
  - 4 outputs, no EN pin
- 2 Fail safe versions
  - Default output state = Low
  - Default output state = High
- 4 Isolation versions:
  - 5.0kV
  - 3.75kV
  - 2.5kV
  - 1.0kV

image

image

image

image

------------


evanshultz:
1. In the FPfilter, please use `*` everywhere instead of `?`. This allows new fields to be added while the regex still works. The only place a question mark should be used is when a single character is used to separate within a single field and the next part of the footprint is known, like with `SOT-23`.
1. Please remove the automotive grade parts and parts with 10kV surge isolation. This falls under [KLC S2.2](http://kicad-pcb.org/libraries/klc/S2.2/). (Should they be ALIASes anyway? Isn't `Si8640BD-AS` functionally the same as `Si8640BD-B-IS`?)
1. The `W` isn't required in the FPfilter because the package XY dimensions will capture that this is a wide body part and is a better filter for package size.
1. Stack ground pins that are the same potential.
1. From pages 2-3 and page 28 we can see there are 4 different logical variations (*0, *1, *2, *5) which you identified above. There are 3 package versions, also identified above, so that's 12 total symbols (IS1 and IS2 are the same package and can be an ALIAS). Fail safe state and isolation do not require unique symbol. But I count more than 12 symbols which aren't automotive or 10kV.
1. So that any future filtering works, let's leave the last field of the description as just the package(s). Can you move the text about extra creepage for IS2 parts within with description instead of inside a paren at the end?

Just a note that our wide SOIC-16 footprint has 7.8mm creepage between the rows, so it's good for their recommended footprint. The IS2 package says it has 8mm creepage, but since their footprint is only 7.5mm creepage we are still good to go.

This is another example where not having dashed lines is so annoying...

strombom:
I guess if temperature ratings are non-functional then isolation ratings are also non-functional and should be removed completely?

strombom:
- Changed ? to * and removed W in FPfilter
- Removed most variants, only kept I/O-configuration and 3 package types
- Added IS2 (8mm creepage) as an ALIAS
- Stacked pins
- Move creepage text in description

image

image

image

image

(not sure whats good about stacking pins, it is nice to see all pins on the schematics when debugging circuits and when placing decoupling capacitors while routing the board)

evanshultz:
One fix: Only one of the 8mm creepage parts has the package `SOIC-16W` in the description. The rest are missing the `W`. Can you please add that?

Yeah, the pin stacking was a big discussion with adamant people on both sides. But that's how it is.

Temperature ratings and automotive qualification are typically manufacturing process and testing differences which aren't changes to the die. Of course some are different designs or packaging differences. Isolation, in my experience, is usually a product decision which drives the die design. Again, this is not guaranteed. So, I'm considering isolation to be worthwhile to include while automotive is not. This is possibly a grey area since KLC isn't super detailed, I don't see any past history of similar discussions, and we're not trying to add a symbol for every single orderable PN while still providing a nice and robust library.

Default output state is definitely functional. It will determine how the state of the output circuit when the modulation signal is not preset. The die will be different (perhaps fused during assembly, but still functionally different in the finished part) so I would like to keep this. And isolation voltage as well. I believe the number of symbols is good and all other changes above are fine, but I would like to see more ALIASes. I think isolation voltage and default state is something important that a user may search for if they were looking for a part. Plus they are a functional differences. They can be an ALIAS and not grow the number of symbols, but I think it will be helpful to users if the symbols aren't as abstract as you have here.

So, what do you think? I do not intend to drive away a willing and prolific contributor like you. You've helped greatly to improve the library and I look forward to lots more input from you. If you were looking for a part, and considering both the exact wording and spirit of KLC, how do you suggest this part should be in the library. I feel strongly that only the 12 current symbols are required, but I'm open to your thoughts on how many ALIASes should be created.

jkriege2:
I know that the ISO-symbols, that these symbols are based on, have all the power-pins in the upper part, but wouldn't it be better to have VCC-pins towards the top (but pointing left/right) and GND-pins ate the bottom (pointing left/right)? Many comparable isolator symbols do it that way:
2018-02-05 07_36_01-symbol library editor - isolator d__kicad_review pretty_kicad-symbols_isolator
2018-02-05 07_35_44-symbol library editor - isolator d__kicad_review pretty_kicad-symbols_isolator

poeschlr:
The reason we include more variations then some might find necessary is such that we can offer better results for people who want to use official symbols for bom creation. There are a lot of people out there who feel we don't go far enough with this approach. Some want to include ordering information to each part. Some want true atomic parts where even the footprint is unique for each component.

So in short, the official lib rule set is a compromise between different workflows that might in the end fit nobody perfectly. The hope is that it will fit everyone well enough such that they can use it with only minimal modifications.

strombom:
Committed is new ALIAS for all default state and isolation configurations except for automotive and reinforced parts. Total number of unique symbols is 12.

image

image

image

image

evanshultz:
@strombom 
Thanks for all the work on this PR and other ones!

strombom:
No problem :)

PR #283, opened 2018-02-05
Max5741

Misca1234:
Here is the D/A MAX5741
https://datasheets.maximintegrated.com/en/ds/MAX5741.pdf

bild

bild


jkriege2:
Nice work,

Thanks,
JAN

PR #284, opened 2018-02-05
Add symbol for TS3A24159DGSR analog switch

byronhopps:
http://www.ti.com/lit/ds/symlink/ts3a24159.pdf

screen shot 2018-02-05 at 7 24 56 pm

------------


jkriege2:
Please split the symbol into sub-units (one for each switch + one power-symbol) and make the style like the other symbols in the same lib!

byronhopps:
Done! Here are screenshots of each unit in the new symbol:

screen shot 2018-02-05 at 8 01 07 pm

screen shot 2018-02-05 at 8 01 12 pm

screen shot 2018-02-05 at 8 01 18 pm

jkriege2:
Thanks, the symbol is OK for me now and I will merge.

Only one request: Would you also be willing to add the VSON-variant (footprint: VSON-10-1EP_3x3mm_P0.5mm_EP1.65x2.4mm_ThermalVias ... I think the EP should be an additional PAD-pin in the power-unit, as I didn't find anything in the datasheet on what signal the EP is connected to) and MAYBE the DBSGA-variant (there is no matching footprint as far as i could see, but you could easily modify Texas_DSBGA-12_1.36x1.86mm_Layout3x4_P0.5mm by removing two balls ...)

Let me know if you can't add the variants, then I'll merge in the current form!

Best,
JAN

byronhopps:
When I made this symbol a while ago, the DFN footprint didn't match what TI suggested in their datasheet. A few dimensions for the pads and pad spacing were off by a few hundred µm, which was too much for me to feel comfortable using that footprint. Of course, with the new footprint libraries, there might be something that better matches TIs suggestions. I'll look into that!

For the BGA footprint, that does seem like a very straightforward modification.

jkriege2:
Let me know ... or would you open a second PR, then I'll merge this one for now ...

JAN

byronhopps:
As it turns out, I was able to find the correct footprint for the DFN/VSON one. It was very straightforward to modify the Texas_DSBGA-12 footprint into a DSBGA-10 one, so I did that as well. I'm now working on creating the pull request for kicad-footprints, and will add a link here once it's up.

jkriege2:
Thanks,
JAN

PR #289, opened 2018-02-07
CY7C68013A/14A USB MCU SSOP-56/QFN-56

calebns:
CY7C68013A-56LTX

FX2LP 8-bit USB MCU QFN-56
CY7C68013A, MCU, USB

http://www.cypress.com/file/138911/download

image

------------


Misca1234:
If your screenshoot of the symbol also include the pin type
like in this push (the text describing the pintype)
https://github.com/KiCad/kicad-symbols/pull/283

It is easier to review

calebns:
image

calebns:
Moved the ref field to the top.
image

evanshultz:
- Spread apart the XTAL pins so the crystal symbol will fit there better; in fact, all pins on the left side could be spread out a bit.
- Where did you get the dimensions for the exposed pad? You picked a footprint with a 4.5x5.2mm exposed copper pad, but that's the exact dimensions of the package's thermal pad and usually the copper is a bit bigger.
- The QFN footprint has 57 pins, with the 57th being the thermal pad; you haven't added a 57th pin to the design and you also need to find out how to connect the thermal pad (or if it should float).
- Pins 1 and 2 are input; 29-31 are output; 15 is an input.
- Pins 13 and 54 are not GPIO pins on the '13A MCU so change the pin names; at least pin 54 should go on the right side.
- Please give pin 14 the full name and move it towards the bottom since it should be grounded.
- The description is weak compared with other parts in this lib; I suggest something like `FX2LP USB Microcontroller, 48MHz 8051, 16KB RAM, USB 2.0, I2C, QFN-56`.

Because of all the fixes above, please post another screenshot of the symbol when you're done.

Lastly, and this is not required, would you be willing to make the other parts in this family?

calebns:
In fact, Cypress does not provide guidelines for their packages. I've prepared a TI-compatible footprint, which seems to fit to dimenstions of CY768 MCU.
https://github.com/KiCad/kicad-footprints/pull/400
Also, added the SSOP version. I'm not planning to add TQFP/BGA chips in this PR, as the footprints need to be prepared first.

image

evanshultz:
Even just adding the SSOP is great. That was no requirement but adding the 13/14/15/16 or any other packages is nice. Thank you!
- We really need a correct QFN footprint. If the copper pad is too small or too big there will be problems. This is just slightly bigger than the pad, which is usually correct, so unless you can find the recommended footprint we'll go with it.
- Why did you connect pin the thermal pad of the QFN to GND? Is that in the datasheet?
- CLKOUT should go on the right side. IFCLK can go on either side.
- RDYx should be on the left side. Sorry that I missed that earlier.
- You got the description and keywords mixed up. Please fix that.
- Would you mind making the '14A version of both packages? it is pin-compatible so it can be done as an ALIAS and no new symbols need to be created. If not, that's fine.

Please post a screenshot of both symbols for ease of review. Thank you.

calebns:
The footprint from TI should be fine, both vendors' QFN56 has exactly the same ep dimensions. However, ADI uses an exact EP dimensions for EP pad in their LFCSP, but that's rather an exception than a rule.

EP is recommended to be soldered to ground, but not necessarily. I've moved it away to leave a choice.

I've also moved CLKOUT to the right, but I disagree to move anything else. All pins on right are creating a consistent FIFO interface to FPGA, and having such pins on both sides will make a mess on a schematic.

Updated with ALIAS to 14A as well and fixed keywords.

image
image

evanshultz:
Thank you for all the changes!
- Please further adjust the description. It's still not like I requested above. `FX2LP USB` should be there.
- Remove the package from the keywords and only have it in the description.
- Where does Cypress recommend EP electrical connection? I still don't find it in the datasheet. Page 52 talks about soldering it to a pad on the board but I don't find anything about tying to GND.
- Page 56 shows a LFX package that is the punched version of the QFN package, but the table only lists LTX so let's leave this alone. Also, Digi-Key shows the LFX as obsolete.
- OK on pin placement.

calebns:
Thanks for all comments, I hope everything is OK now.

As for the EP, I've read it somewhere, but not in a datasheet for sure. It might have been PCB layout guide or reference design docs... Generally speaking, for best thermal management, pad "can" be soldered to a solid ground plane with a bunch of vias all over it. But I admit, the source of this revelations is suspicious for me either, so it's better to leave the pin to let the user to decide what to do with it.

evanshultz:
Description and keywords are OK now. And it's best to leave the pad separate, unless we find a reference for the connection, as you've done and for the reason you state.

Thank you!

whitequark:
There are a few issues with this symbol.

1. It references a footprint `QFN-56-1EP_8x8mm_P0.5mm_EP4.8x5.5mm` is not currently merged.
2. The EP electrical connection should be to ground. Also, the thermal vias aren't described as optional in the datasheet. See the excerpt from the datasheet below.
3. There is a footprint `QFN-56-1EP_8x8mm_P0.5mm_EP4.5x5.2mm`, which has the same EP dimensions as the datasheet. See drawing below.
screenshot_20180410_210644
screenshot_20180410_210253
screenshot_20180410_205809

@evanshultz Questions:
1. Should the symbol be switched to use `QFN-56-1EP_8x8mm_P0.5mm_EP4.5x5.2mm`?
2. Should the symbol be switched to use ``QFN-56-1EP_8x8mm_P0.5mm_EP4.5x5.2mm` with added thermal vias?

whitequark:
@evanshultz Thanks for such a quick and thorough reply! For the project I'm currently working on, I took a decision to get every footprint upstream in KiCAD libraries, and I am very happy with it.

> Ground is by far the most common in my experience, but I didn't find that in the datasheet. Did you ask Cypress?

It's on Figure 41: "Via hole for thermally connecting the QFN to the circuit board ground plane", plus "The QFN metal die paddle must be soldered to the PCB's thermal pad" in the paragraph of text above, which are unambiguous in my opinion.

Regarding the footprint with the thermal vias, should it be a generic variant of the footprint? Or do you think it should be done as a Cypress-prefixed variant?

evanshultz:
Yeah, now that you put that all together it does seem to indicate it must be connected to ground. Still rather ambiguous compared with most datasheets. If you don't mind, let's give Cypress a day or two to respond and then decide.

whitequark:
Sounds good. Cypress is extremely thorough in documenting the software side of their microcontrollers but it seems that mechanical side is lacking for some reason.

whitequark:
@evanshultz It's been 9 days--any answer from Cypress? We made a footprint, anyway, do you want me to submit that?

evanshultz:
Thanks for pinging me on this topic! Unfortunately I have not gotten anything useful back. I have raised this issue through another avenue. Let's give it a couple days and hope this approach works better.

evanshultz:
@whitequark 
Alright. Here are a couple of helpful items:
- http://www.cypress.com/file/140006/download
- http://www.cypress.com/file/45261/download

I'm told the CY8CKIT-50 design is the best reference design from Cypress, but it has only a few randomly-located vias in the thermal pad and it also has a single square paste pad half the edge side of the copper pad. They said a 5x5 via pattern would be better but with only a ~6mm thermal pad a single paste pad is best. But since KiCad typically uses smaller paste pads and it's also in AN72845 we'll do it.

So all together, here is the best info I've gathered about the footprint (note that I've rounded things since we work in mm but the Cypress designs are all in mils):
- Pad shape: Oval
- Pad size: 1x0.28mm
- Row/Column spacing: 8mm
- Center pad copper shape: Square
- Center pad copper size: 6.22mm
- Center pad paste shape: 4 squares on 1.555mm grid
- Center pad paste size: 2.6mm (70% paste coverage)
- Center pad via drill/ring/mask: 0.3/0.6/0.4mm (Cypress specifically shows mask over vias)
- Center pad via layout: 5x5 on 1.27mm grid

What are the parameters of your footprint?

Sorry this took so long. Thanks for bringing it up.

awygle:
Hi! I think most of these parameters are totally reasonable (although the thermal pad does seem a bit large to me, but never mind). I did run into a couple of things however.

First, to get the via cap, I had to get creative with the artwork as you can see here:

image

This is because the plane is negative and we want mask to be removed everywhere _but_ over the vias. There's nothing wrong with this exactly but it is odd (and Kicad should probably support this butter but never mind).

Second, I'm not a big fan of the proposed stencil layer (seen here):

image

Specifically, having paste over top of vias is something I typically try to avoid. Our footprint went a different way:

image

This gave us 55% paste coverage without being over the top of the vias (actually this image is slightly wrong because of other changes but you can see the idea), and we could easily get to 70% with this technique. That said, I'm open to doing it either way,  just let me know what you'd like to go with. I'll submit a new footprint PR, then a symbol PR updating this symbol to point to the new footprint.

Sound good?

evanshultz:
I'm fine with a modified paste layout. There are many ways to get there. :)

I understand what you mean by keeping the mask over vias to avoid paste drainage. Since Cypress calls out mask over vias in the datasheet I'm OK with it here. We default to no mask over vias, so please make this a Cypress-specific footprint if it's customized.

PR #291, opened 2018-02-08
Centered and added foot print filter

Misca1234:
Fixed some symbols, they was either

Centred,
Increased Pin name position offset to 20
Fixed foot print filter

bild

bild

bild

bild

bild

bild

bild

bild


Misca1234:
There are more errors in the library, i will fix them when this push is merged

evanshultz:
The symbol images are much less useful when you cover them up with dialogs.

Ugh. The v2.3 -> v2.4 makes a lot of noise so it's harder to check exactly what changed and that nothing else was changed. But we'll have to do it sooner or later...

General:
- `SOIC*7.5x12.8mm?P1.27mm*` works; you don't need the `W?` in `SOIC*W?7.5x12.8mm?P1.27mm*`. The package dimensions ensure only wide packages are selected and not narrow ones.
- Further, I would like you to use asterisks instead of question marks within the FPfilter strings. The regex doesn't matter if only a single character is present, but if fields are inserted it gives more flexibility. A recent example was when we wanted to add the EP size; it went at the end because there was always a trailing asterisk.

LTC3890:
- Stack the SGND pins.

NID30S24-05 / TMR_1-0522SM / NXE2S0505MC:
- This symbols should move to Converter_DCDC library.

TPS61277:
- Stack pin 33 with AGND as per page 27 of the datasheet (http://www.ti.com/lit/ds/symlink/tps62177.pdf).
- Also on the same page, make the NC pin visible and move it next to the ANGD pin.

MIC2178:
- Capitalize all letters of pin 19.

ADuM6000:
- You gave a default footprint to this symbol. This part is available in two different packages with different suffixes. Either add a suffix to this symbol or do not select a default footprint and add both available footprints in the FPfilter.

evanshultz:
Also check the Travis output: apparently the Murata DCDC converter has a busted FPfilter.

Misca1234:
> I would like you to use asterisks instead of question marks within the FPfilter strings.
I changed it for 
TMR_1-0522SM
NID30S24-05
MIC2178

> Also check the Travis output: apparently the Murata DCDC converter has a busted FPfilter.
Yes, I know, I was pushing this partly so it wont be to much to review at the same time.

> NID30S24-05 / TMR_1-0522SM / NXE2S0505MC:
> This symbols should move to Converter_DCDC library.
I can do it in push part 2, when this is merged



> LTC3890:
> Stack the SGND pins.
It will not happend


> TPS61277:
> Stack pin 33 with AGND as per page 27 of the datasheet (http://www.ti.com/lit/ds/symlink/tps62177.pdf).
It will not happend

> Also on the same page, make the NC pin visible and move it next to the ANGD pin.
Made it visible


> MIC2178:
> Capitalize all letters of pin 19.
Fixed

> ADuM6000:
> You gave a default footprint to this symbol. 
I removed the specific one and added both as FP filter


bild

bild

bild

evanshultz:
If you can fix the FPfilter and move the DC-DC symbols in a separate PR, all looks good to me. Since the symbols I requested stacking above weren't stacked already this is fine. Thanks!

PR #292, opened 2018-02-09
RFM9xW symbol added

Resaj:
[Datasheet](http://www.hoperf.com/upload/rf/RFM95_96_97_98W.pdf)
Pin diagram on page 10
Pin description on page 11
Dimensions on page 120
RFM9xW variants specified on page 121

I don't specify any footprint because there are 2 different footprints on [kicad-footprints](https://github.com/KiCad/kicad-footprints/tree/master/RF_Module.pretty):
- Hoperf_RFM9XW_SMD.kicad_mod
- Hoperf_RFM9XW_THT.kicad_mod

imagen


Resaj:
Seeing the available footprints, should I have to fill the footprint field? Travis failed because of that.

poeschlr:
Yes if a symbol is designed for one footprint, then the footprint field should point to that footprint.
The footprint filter should then allow slight variations to be shown to the user. (Example a handsolder version or one with thermal vias, ...)

Misca1234:
This is just a personal suggestion, I would have shortened the pin length to half, it is just an look and feel suggestion

By the way, this component is already in the library
bild





Also, switch on the "show pin type" in the symbol editor, then it will be more directly visible in the screen shoots and easier/quicker to review

bild

Resaj:
Thanks for the advices. It's a little confusing, because the component is a RF module. So I didn't see it and I made it on RF_module.lib.

I'm going to exclude my component and modify the existing modules on RF_AM_FM, but there are some mistakes. I've been investigating and there are 2 RFM9x types:
- [RFM96HW](http://www.hoperf.com/upload/rf/RFM69HW-V1.3.pdf)
- [RFM9xW](http://www.hoperf.com/upload/rf/RFM95_96_97_98W.pdf). Includes 95 to 98 modules
Only the pins distribution changes between types. And independently of the frequency, the modules are the same inside of each type. So I think it's not necessary to maintain 9 different symbols for RFM9xW.
Also, distribution of RFM96W is wrong, because it has the pin distribution of HW extension.

I you are agree with this changes, I'll proceed to simplify and fix the errors of this components.

Misca1234:
there is more differences between 
RFM95/96/97/98(W) 
and 
RFM69HW 
It is the size of the PCB

I would suggest you change the existing RFM69HW  so it support both type of pin layout as the 
RFM95/96/97/98(W) do

I also suggest you remove the alias RFM69W from RFM69HW because it is already covered by 
RFM95/96/97/98(W) and RFM69W have a different pcb size than RFM69HW 
and is not a part of the datasheet 
http://www.hoperf.com/upload/rf/RFM69HW-V1.3.pdf

Resaj:
Yes, the modules have different dimensions, but I think there is a little confusing about "96" and "69".
The RFM96W is covered by [RFM95/96/97/98(W)](http://www.hoperf.com/upload/rf/RFM95_96_97_98W.pdf).
The [RFM69W](http://www.hoperf.com/upload/rf/RFM69W-V1.3.pdf) is not covered by RFM95/96/97/98(W). This module has the same dimensions, footprint and pin distribution than [RFM69HW](http://www.hoperf.com/upload/rf/RFM69HW-V1.3.pdf).

Misca1234:
Indeed, you are right, amazing how a company can mess it up

Resaj:
As the RFM69W is included with the alias in the component RFM69HW and the THT footprint is not included in kicad-footprints, I think it's better to close this PR and don't modify the library. Is it OK?

Ratfink:
Yes, it looks like symbols for these parts are already in the library in RF_AM_FM.lib, so to avoid duplication, I'm closing this PR.  I do see a couple problems with the ones in RF_AM_FM.lib (at least a typo in the footprint filters), so if you'd like to improve those that would be welcome!

PR #293, opened 2018-02-09
Restructuring generic connector symbols plus better fp filters.

poeschlr:
Made sure the filters work with the addition of mounting and
shield pins.
Moved Male and Female versions into connector specialized.

Footprint renaming is done in https://github.com/KiCad/kicad-footprints/pull/359

script: https://github.com/KiCad/kicad-library-utils/pull/210


Schlumpf:
If you are touching the connectors, whats about a new 1x50 connector? Currently I need this part...

poeschlr:
We need a cutoff at some point. Currently the cutoff is at 40 connections per row. (So a dual row connector goes up to 80 Connections)
The generator is build such that connectors with at most 3 chars look good. So in general it should work for up to 99 pins per row. (limiting factor are the dual row connectors with letters for the row and number for the column.)

50 might indeed be a more natural cutoff. It will increase the lib size by a lot. (Not necessarily the component count but the line count.)

evanshultz:
@poeschlr 
Whoa... a lot was changed here. Do you want to do anything else, like add 1x50? From the above I'm not sure if you're ready to merge this. If so, I assume minimal checking is fine since this is scripted.

poeschlr:
I don't think adding new symbols should be done in the same pull request.

It would be easy for @Schlumpf  to extend the scripts such that they create the symbols he wants to have.

evanshultz:
I'll just merge as is. My PR at https://github.com/KiCad/kicad-symbols/pull/310 will fix a lot of the remaining errors.

PR #295, opened 2018-02-10
Add SK6812MINI

fauxpark:
I learnt the hard way that the Neopixel Mini pinout is not the same as the 5050 pinout. I've copied and pasted the WS2812B symbol, warts and all, and fixed the pin numbers -- correcting the KLC issues for all the Neopixel symbols should be a different PR I think.

sym

Datasheet: https://cdn-shop.adafruit.com/product-files/2686/SK6812MINI_REV.01-1-2.pdf

------------


Misca1234:
I suggest you do either following things

A) Check in the library  LED_SMD.pretty or OptoDevice.pretty 
To see if the foot print already exist.

B)
Create a foot print, get it pushed, merged and then update this push

Another tip would be that you should make the pin type more visible

bild

fauxpark:
Wasn't sure what you were talking about until I realised it should be LED_SMD, not LEDs - but I probably would have caught that in the next PR cleaning up all the Neopixel symbols anyway.

jkriege2:
Thanks for you contribution!

Before we can merge:
1. Please provide the link to a datasheet inside this PR (thanks)
1. Could you make the symbol look a bit more like the existing Neopixel-symbol we already have (i.e. move power-pins top/bottom and add the small drawing of an LED):
2018-02-12 07_22_28-symbol library editor - led d__kicad_review pretty_kicad-symbols_led lib

fauxpark:
Datasheet added to OP.

As for moving the pins around etc, the idea was to do that in a separate PR which dealt with the other SMD Neopixel symbols all at once.

jkriege2:
Please fix the pin-location+internal drawing in this PR (there really is no point in merging a symbol that needs improvement, when you're willing to improve anyways) and fix the others in a separate PR.

Best,
JAN

jkriege2:
Also the REFDES should be D as with all other LEDs

fauxpark:
My thinking was that a PR should do only one thing, this is how I learned to use Git - I guess a better approach probably would have been for me to fix the existing symbols before adding this one.

Travis says it can't find the footprint but once kicad/kicad-footprints#343 is merged this shouldn't be an issue.

RE the internal drawing: there is only one LED symbol drawn but NeoPixels have three separate LEDs inside - should there be three drawings, or the text "RGB" in there? Or is it not that useful?

jkriege2:
If you find a way to integrate 3 LEDs that would be nice, but maybe adding the (small) text RGB might also be sufficient? ... because the symbol should not be too large (could you shrink you'rs a bit, like the Neopixel_THT-symbol?). 

An option could also be to place the pin-labels with 20-30mil text-size next to the pins ... we use that sometimes if the symbol would get out of hand in size otherwise, although the style above is the default.

fauxpark:
I started off with the THT symbol - this is the smallest I could make it without either the pin names overlapping the LED drawing or the script complaining about it not being on the 100mil grid. It's only able to be so small because the clearance between the pin name position offset is incorrect.

jkriege2:
I know that the lib format has its limits, but please try to somehow fit this symbol into the size of the old THT-symbol ... An idea could also be to move the LED symbol into one corner. Also if you need it, we can make an exception for the pin-label-offset (Note that we DO make such exceptions to KLC, if the need arrises and there is a good reason for it). A symbol that has comparable problems is e.g. 
2018-02-13 21_15_08-greenshot capture form

jkriege2:
@poeschlr @evanshultz Do you have an opinion how we should treat LEDs with included controller like this?

fauxpark:
Why does it need to be the same size as the THT symbol? I can't find anything in the KLC about rectangle sizes.

I'm more than happy to reduce the pin number size - if I reduce the pin name size to 40mil I can make it 600x600mil. The THT symbol is slightly less tall at 600x400mil, but in the process of cleaning it up after this PR reaches a resolution it would also have to be increased to 600x600mil, as according to [this datasheet](https://cdn.sparkfun.com/datasheets/Components/LED/COM-12877.pdf) "5V" should in fact be "VDD".

poeschlr:
If two symbols for similar parts have the same size and pin placement,. then it is easier for the user to exchange them.

---

There is some rule in the KLC that states that similar things should look similar. (At least this rule has been in the KLC in the past. If it did not survive the reorganization of the KLC then it should definitely be reintroduced.) 
The reason for this rule is because we do not have the manpower to make the KLC such that it has a rule for everything. (There is also the danger that such a detailed ruleset confuses contributors more then it helps them.)

evanshultz:
Yeah, this one is a bit tough to make both clear but also succinct.

Here's my thoughts:
- The main purpose of this part is to emit light, so that should be the concept the symbol communicates most clearly.
- Perhaps add a box around the LED with short lines going to DIN and DOUT to indicate the non-LED circuitry.
- A single LED is fine with me, especially if "RGB" text is present.

I haven't seen this type of LED before, but it's quite cool!

fauxpark:
> If two symbols for similar parts have the same size and pin placement,. then it is easier for the user to exchange them.

Yes, this is my goal for all of the Neopixel symbols. The THT symbol has some KLC issues (but fewer than the SMD symbols - see the image in the OP) that I have fixed in this one, which can then be used as a template for the rest of them.

jkriege2:
Also I think we should try to make the symbols as small as possible (that's why I repeated the example of the THT-Neopixel). A good guideline is to put the LED next to say a transistor, resistor, or the µC you use to control it and see how it looks (that's the design angle to this). Also these are (as I understand) often chained, so it would be good to get mayn of them onto a single schematic sheet.

I was thinking a bit: We have some symbols, where we have power-supply on left/right, e.g. some isolators:
2018-02-05 07_36_01-symbol library editor - isolator d__kicad_review pretty_kicad-symbols_isolator
2017-06-06 08_22_57-part library editor d__kicad_test_addlin_ _kicad-library_library_regul lib
2018-02-14 07_28_24-symbol library editor - interface_optical d__kicad_kicad-symbols_interface_opti


Maybe it would be a good idea to use such an approach and use a symbol as shown in your first comment (maybe power on the left, DIN right-top, DOUT right-bottom) and then the LED-symbol with "RGB" inside?

What do you all think? Would that be an option for this symbol?

fauxpark:
> Also I think we should try to make the symbols as small as possible

Which is what I have done, the end result being that it is only 200mil taller than the THT symbol - that seems pretty good to me, considering that if the THT symbol were to be given the same treatment, it would end up the exact same size. This also happens to make the symbol a square shape, which is typically what these components look like.

I know it doesn't have to resemble the package though - I could offset VCC/GND to the left/right so that the pin names are not intersecting the body of the LED, but again, I'm not sure the THT symbol is actually a good example for size.

> Maybe it would be a good idea to use such an approach and use a symbol as shown in your first comment

...So you want me to go back to the old symbol, that you wanted me to make more like the THT symbol?

jkriege2:
>> Maybe it would be a good idea to use such an approach and use a symbol as shown in your first comment
>...So you want me to go back to the old symbol, that you wanted me to make more like the THT symbol?

Yes, that MIGHT be an option ... let's see what the other Maintainers think.

poeschlr:
I don't really have a style preference here.

But i would like similar symbols to stay exchangeable. So whatever choice you make, ensure that all similar symbols stay exchangeable.

I am assuming that the old symbol is currently widely used in the lib. Using that symbol would reduce the workload required to reach that goal. (I do not have the time to look into this in great detail. You two have already looked at it.)

fauxpark:
600x600mil with 40mil pin names:


And 600x400mil, same as THT, with offset power pins:

jkriege2:
Here are some more proposals for the symbol shape (with emphasis on smaller size):
2018-02-14 21_30_19-symbol library editor - led d__kicad_kicad-symbols_led lib
2018-02-14 21_29_58-symbol library editor - led d__kicad_kicad-symbols_led lib
2018-02-14 21_28_43-symbol library editor - led d__kicad_kicad-symbols_led lib
2018-02-14 21_25_16-symbol library editor - led d__kicad_kicad-symbols_led lib

What do you think?

I hope you're not annoyed (too much) by the slow decision-making process here ... I'm just trying to find the best and most suitable symbol for this type of device.

poeschlr:
I assume din and dout are for some kind of daisy-chain-able bus.
In that case having them on opposite sides makes sense as it makes creating chains in the schematic easier. (If not then you can ignore this input)

jkriege2:
yes they are ... but having them on one side would allow to place symbols vertically and daisy-cahin, on opposite side is good for a hor. daisy chain (power-pins always get in the way ;-)

poeschlr:
I did not think about the option of chaining them vertically. It is clearly not my day.

fauxpark:
We also need to consider that at least one variant of this part has two extra pins - a backup DIN and an NC. I would prefer the third 200x200 proposal so that BIN can be placed underneath DIN.

If it helps, here is how I'm currently chaining these LEDs. I was previously using the WS2812B style symbol, which looked much messier.
sch

fauxpark:
The [WS2812](https://cdn-shop.adafruit.com/datasheets/WS2812.pdf) has two VCC pins for the LEDs and IC. The [WS2813](http://www.normandled.com/upload/201605/WS2813%20LED%20Datasheet.pdf) also has two VCC pins, plus a backup DIN.

jkriege2:
fine with me ...

JAN

fauxpark:
Here we have a 400x400 symbol, which is what I think you meant by "200x200"?

sym

jkriege2:
Looks good. I would reduce the pin-label-offset to 10mil or 20mil ... the the text would be separated a bit better (and that's still covered by the KLC, although the scripts will give you a warning)

fauxpark:
Okay, it was 20mil, now 10mil.

jkriege2:
fine with me now (great work BTW!!!) ... should I merge this now and you rework the other neopixels in another PR (you hinted that you wanted to do just that), or should I wait until you reworked all of them?

best,
JAN

fauxpark:
I'm good to merge now, I'll do the rest in the next PR.

jkriege2:
OK, thanks for the work you put in this!

JAN

PR #296, opened 2018-02-11
Add ST M41T62 real time clock

MatthewStephenson:
This is the ST M41T62 low-power I2C real time clock, with alarm interrupt and square wave outputs.

[Datasheet here.](http://www.st.com/content/ccc/resource/technical/document/datasheet/c8/66/38/71/c4/ac/49/9e/CD00019860.pdf/files/CD00019860.pdf/jcr:content/translations/en.CD00019860.pdf)

image

As always, comments and suggestions welcome.


Misca1234:
A slighter shorter link to the device would be 
http://www.st.com/resource/en/datasheet/m41t62.pdf

MatthewStephenson:
Shortened datasheet link, thanks.

I was under the impression that the footprint filter had to be at 0,0 - I'll check this.

poeschlr:
I don't think the position of the footprint field is specified in the KLC.

jkriege2:
Hi!

1. as per datasheet, the name should be `M41T62Q`
2. Would you be willing to also provide the M41T64 and M41T65 devices? Maybe even the LVV-variants? (only if you have time!!!)

The rest looks fine to me ... 

Best,
JAN

evanshultz:
A few more things:
- Pin text offset should be 20mil, not 40mil.
- Pin 11 should be `~IRQ~/OUT` so the "IRQ" has an overbar for logic low.
- The bottom of datasheet page 29 indicates the EP should be tied to VSS, right? Stack those pins.
- The symbol looks excessively big. Can it be shorter? It may also be able to be a bit less wide.

evanshultz:
@MatthewStephenson 
Will you be able to update this PR and fix branch conflicts so it can be merged?

MatthewStephenson:
@evanshultz 
Yes, will do, real life has intruded a little at the moment, but it's on my list.

MatthewStephenson:
Updated as suggested, thanks. Sorry for the delay.

image

Regarding the pin stacking - I did initially have them stacked, but I had thought it violated KLC rule S4.3.3 - "Pins must have the same name" - and I was sure the travis script had previously picked me up on this. It doesn't seem to cause an error this time though, so I suppose it's ok. It's not clear to me from reading that section if the "Special case" rules for power pin stacking invalidate the normal rules - especially as the numbering of the rules continues (6 & 7) rather than restarting at 1.

I don't have time at present to add the other variants - sorry.

jkriege2:
So for any pin stack we require all pins to have the same name. This is violated here. The EP can be a special case. If it MUST be connected to teh Vss, I would rename it from EP to Vss and everything is fine. If this is just an OPTION (the datasheet says "Note: Substrate pad should be tied to VSS."), I would leave it as EP and move it next to the Vss-pin, so the user can connect it, but does not have to do it.

Otherwise this looks good to me and I will merge, when the EP-thing is fixed.

Best,
JAN

MatthewStephenson:
Alright, back to how it was...

jkriege2:
Thanks,
JAN

PR #303, opened 2018-02-14
Add ADS786x series of ADCs

byronhopps:
Added new symbol ADS7866 with aliases ADS7867 and ADS7868

http://www.ti.com/lit/ds/symlink/ads7866.pdf

screen shot 2018-02-14 at 4 49 27 pm

------------


byronhopps:
Looking at the Travis output, it appears that the new symbol I added passes all checks, but when saving the library in the new file format the line order of existing symbols was changed. Since those symbols are not KLC compliant yet, they cause the travis script to fail.

poeschlr:
Yes this is the downside of the fact that different kicad versions use different line ordering in the lib files. (Even different stable versions with the same file format behave differently.)

You can ignore errors for symbols you did not touch.

Note: I did not look at this contribution in any other way. The reason for my comment is to inform the contributor that this (expected) behavior is not a problem.

jkriege2:
The symbol looks good to me ... there's only two things to change:
- please make the SDO-pin tri-state (it's in high-z, when the IC is not selected)
- please change the description to a more standardized format that contains the package at the end, e.g. `Single miniature 12-bit ADC, 100-200ksps, SPI, micro-power, SOT-23-6`

byronhopps:
The requested changes have been made

jkriege2:
Thanks!

JAN

PR #304, opened 2018-02-15
Add TSM102/3

evanshultz:
These symbols are a little tricky. I tried to balance maintaining existing symbols (or pieces of them) while drawing them.

While the voltage reference is like a TL431, I used a zener voltage diode symbol which is unfilled and is our standard diode symbol (which means the TL431 and similar need updating!).

Notes:
- Is there a better way you can think of to show the voltage reference on TSM103 unit A?
- I did not mark which units of the TSM102 are opamps and which are comaparators. Should I?
- TSM102's voltage reference V- pin displays the pin name, like other opamps, so I made the symbol for unit E avoid this text as much as possible. Should I push it right to fully avoid the text?

Datashets:
- http://www.st.com/resource/en/datasheet/tsm102.pdf
- http://www.st.com/resource/en/datasheet/tsm103w.pdf

Image:
image

------------


evanshultz:
@jkriege2 
For TSM103W, since the reference is tied to an input pin of unit 1 and also the common ground, it's a bit hard to separate them nicely. I left that one alone.

Great idea to use a separate power unit for TSM102. I've done that and also added the "COMP" text. The reference can be used totally separately, and often is, so it will need to be separate. New image:
image

jkriege2:
Looks fine to me!

Thanks,
JAN

Issue #187, opened 2018-01-20
Some MCU symbols are very large because of long pin names


poeschlr:
Currently some MCU symbols still list all alternative functions in their pin names. This results in very large symbols. A better approach is to reduce the pin name to the main identifier.

ObKo:
Script that was used to generate stm32 library: https://github.com/KiCad/kicad-library-utils/tree/master/schlib/autogen/stm32

Ratfink:
I'm about to start working on this, but I can't find the XML files for obsolete devices mentioned in [the README for the stm32 autogen script](https://github.com/KiCad/kicad-library-utils/blob/master/schlib/autogen/stm32/README.md).  Seeing as how the obsolete microcontrollers mentioned there aren't available from wholesalers anymore and are barely even mentioned on ST's website, should I just let their symbols disappear from the library?

bobc:
I spent some time working on updating the STM32 scripts, there are several issues that can be improved.

With regard to "obsolete" parts, some F050 parts were renamed to different names (but exactly the same part). I think the missing F313 parts either never went into production or were introduced with different config. I would suggest to copy the obsolete/missing parts to the "obsolete" folder.

Ratfink:
What's the policy on specifying footprints for parts that have exactly one footprint, but that footprint isn't in the library?  A lot of the STM32 chips fall under that category, with weird packages that aren't in the library yet.  I already have footprint filters for all the packages, but I don't know if it's frowned upon to specify a footprint name that doesn't exist yet.

bobc:
I created footprints for many of the missing ST packages - QFN-20, QFN-28, BGA-64, 100 (x3) 132, 169, 176, 216, 240 but I didn't get around to pushing them yet. I can do that over the weekend.

I generated them via a script, so it is relatively easy to add new ones. The full list is 

Package_DFN_QFN:QFN-20-3x3mm_Pitch0.5mm
Package_DFN_QFN:QFN-28-4x4mm_Pitch0.5mm

Package_BGA:BGA-64_5.0x5.0mm_Layout8x8_P0.5mm
Package_BGA:BGA-100_10.0x10.0mm_Layout10x10_P0.8mm
Package_BGA:BGA-100_7.0x7.0mm_Layout12x12_P0.5mm
Package_BGA:BGA-100_8.0x8.0mm_Layout10x10_P0.8mm
Package_BGA:BGA-100_9.0x9.0mm_Layout10x10_P0.8mm
[...]

PR #305, opened 2018-02-15
Major update to STM32 libraries

Ratfink:
Here is the fruit of the work on the STM32 libraries I've been discussing in #187.

For one thing, they are now split by family instead of just being one library that's so large it violates the KLC.  The pins have short names with only their primary functions, and the symbols are a bit shorter than they used to be as well.  A couple pins (namely, VBAT and NC pins) have been moved to better follow KLC.  Quite a few new symbols are added for new microcontrollers.  All symbols have footprints and filters specified, though not all of those footprints exist yet.  I'll file a bug report against KiCad/kicad-footprints for the missing footprints shortly.

The pull request for the updated STM32 symbol generating script can be found here: [coming soon]

Since providing pictures of hundreds of footprints would likely not even be helpful, here's a representative example (STM32F098VCHx):
screenshot from 2018-02-14 21-38-42

There are a few obsolete parts that can't be auto-generated anymore.  I haven't included those in this pull request yet, but I'll try to add them to the obsolete/ directory soon.


strombom:
Idea: SWD, JTAG and Oscillator pins could be marked, including SWO/TDO, SWDIO/TMS, SWDCLK/TCK, TDI, OSC_IN and OSC_OUT?

jcyrax:
Hi, I think creating separate units for parts with large pin count would be nice.

Ratfink:
@jkriege2 @poeschlr @evanshultz Could a librarian please either review this or tell me that it won't be possible to merge in time for 5.0?  The suspense is killing me. 😉

poeschlr:
We are a bit short on manpower right now. I will try to find some time this week.

Ratfink:
Thanks, I really appreciate it.  If y'all are really that short on time though, I would completely understand if it can't be merged for 5.0.  Don't over-work yourself for me!  It's just that two weeks of silence from the people who can review it made me wonder if this PR slipped through the cracks somehow; I'm glad to know now that it hasn't.

poeschlr:
If you want to help out, you could look at some minor additions (addition of one footprint or symbol) and review them. This would free up the more experienced guys for reviewing major changes like yours.

Ratfink:
I see pin stacking is essentially required now, so I'll update these to have stacked power pins shortly.

evanshultz:
@poeschlr 
Are you going to be able to take a look at this as mentioned above? Reviewing this PR is a massive undertaking. maybe we should split it up? Or just rely on a few spot checks and then merge; having scripted symbol generation means fixes are easier.

@Ratfink 
The description is different than what I usually see. For example, I typically see the datasheet description (simplified, if needed), followed by some key parameters, and then the package name at the end (with a hyphen separating the package type from the pin count). All comma separated. That paradigm would mean that `STM32L4S9ZIYx` would change from`Core: ARM Cortex-M4 Package: WLCSP144-DIE470 Flash: 2048KB Ram: 640KB Frequency: 120MHz Voltage: 1.71..3.6V IO-pins: 110` to something like `Ultra-low-power ARM® Cortex®-M4 32-bit MCU+FPU, 150DMIPS, up to 2MB Flash, 640KB SRAM, LCD-TFT & MIPI DSI, AES+HASH, 120MHz, 1.71-3.6V, 110 GPIO, WLCSP-144`. I'm not saying the parts need a description change, just stating that this PR uses a description rather different than most libraries. The existing STM32 is like what you have, but check out the STM8 which is more like what I described above and like other libraries.

Ratfink:
@evanshultz Good point!  I didn't touch the description text, and that would be a good thing to improve.

The problem with trying to use the datasheet description is that the information contained there varies from part to part, so some descriptions would have redundancies if I added other information like voltage range and frequency from the XML data.  Also, parsing data from PDFs automatically would mean the descriptions could end up kinda messy, with inconsistent spaces after commas, em dashes instead of hyphens some of the time, etc.  And using the datasheet description means that aliased parts wouldn't have the correct flash and RAM sizes in their descriptions.

Basically, we have to decide between listing fancy features from the datasheet description, or providing well-formatted and accurate information for each part.  Trying to do both would be a massive headache, I'm afraid.  Either way, I can make sure the package name ends up at the end.

poeschlr:
I am extremely short on time this month. This week most of my time has been used up by trying to calm down some conflicts here. I hope that next week will be more time for actually reviewing stuff.

Ratfink:
Okay, I guess I should un-stack the power pins, since the datasheets for these chips explicitly state you're supposed to have a decoupling cap for each pair.  I'll do that in a bit.

Ratfink:
@evanshultz @poeschlr @jkriege2 @Misca1234 
Any further input about the descriptions?  I'd like to make those more normal, but see my question above on whether to use the datasheet description or a script-generated one.  Once that's decided, hopefully someone could look at the script, do some spot-checks, and we could get this merged before the freeze.

jcyrax:
What is the current stance regarding splitting larger parts into separate units? For instance STM32F437IGTx is almost as large as A4 page. 
image

There was discussion about this in user forum some time ago: https://forum.kicad.info/t/various-random-stuff/8929
Is this https://forum.kicad.info/t/various-random-stuff/8929/8?u=jcyrax still valid?

Ratfink:
I love the idea of splitting large MCU symbols into smaller units, but I think there should be a discussion and corresponding KLC change to say precisely when and how to do it.  This should be a discussion for all processor symbols, not just the STM32 libraries.  The change should likewise be made for all processors to avoid making the library very inconsistent.

I also think that we're getting close the 5.0 freeze, and that trying to make such a sweeping change so close to the release is asking for trouble.  We have this PR *now*, and I think with some relatively minor work it can be merged, making a big improvement over the current STM32 library.  Yes, splitting into multiple units would be make it even better, but at this point I'm afraid that will probably have to wait for 6.0.

Ratfink:
Here's what the STM32F437IGTx looks like after the latest commit.  It's the same size, but the text doesn't overlap in the upper-left corner.
screenshot from 2018-03-17 16-44-28

jcyrax:
I did a visual look around the lib. I don't know if this is a real problem, but STM32F469AEYx and few others symbols power pins are a bit offset to side. I guess due to long pin names a bit down.
image

Ratfink:
Could someone (or -ones) please try to review this before the 5.0 freeze takes effect?  I think Evan's earlier suggestion of reviewing with a few spot checks makes sense: since these symbols are script-generated from official data from ST, many aspects of them are guaranteed to be correct.  The biggest things I can think of that need to be reviewed are footprint names and filters (which could probably be done most efficiently by looking at the script first), and that the pin arrangement and electrical types are OK.

I hate to be a bother, but I'd also hate to see the STM32 library remain as it is, where even 48-pin symbols are almost unusably large, for the next couple years until 6.0 comes out.

PR #306, opened 2018-02-15
Add Decawave DWM1001 under RF_module

ouinouin:
Hi, Here is a Decawave DWM1001 lib schematic, This module is basically a cortex ucontroller coupled with a DWM1000 rf Chipset that is used for indoor objects localisation.
basic documentation with footprint and pinout [here](https://www.decawave.com/sites/default/files/dwm1001_datasheet.pdf).
Footprint and 3D models will be pushed asap (missing bandwidth to sync my 3D job with kicad repo.).
Edit : footprint PR : [here](https://github.com/KiCad/kicad-footprints/pull/376) for the moment Travis CI wants to see the footprint first as I already made the association in the schematic

capture d ecran de 2018-02-15 13-04-13


Misca1234:
A tip to make it easier to review, switch on the pin description when you take the screen shoot

bild

ouinouin:
Hi, 
unofrtunately my screen is a bit small (thanks to 1366x768 that still comes in standard with many office laptops...)
dwm1001

Misca1234:
I suggest that you change the text of the RESETn pin text to the following 
~RESETn~
In this way you get a small line above he text indicating that this pin is active low.


I usually, but this is a personal comment, always add the VCC range and interface to the description
That is, in my case the description would be 
Ultra wide band RF module With ranging location capabilities, I2C, UART, SPI, +2.8V to +3.6V VCC

evanshultz:
A tilde is only required before the pin name, not before and after. You can use before and after if the pin has multiple function, such as `~RST~/INT` to only add an overbar on "RST". But in this case, you can use `~RESET` without a second tilde.

ouinouin:
ok, I added the ~RESET i didnt see it on the KLC despite being mentionned. it may be helpful to make it  figure on the "check list" message when you fill a PR ?
Thanks for your careful review.

Misca1234:
The people who do the merge to the official master repository is doing this on the their spare time.
Therefore the time it takes to get it merged will vary from days or even weeks.

If it takes way to long time, like a month, then either you have missed additional comments in your thread push to fix stuff, merge conflicts raised (see below) or it just been overlooked, if overlooked you can just poke the thread.

Remember, periodically check the "continuous-integration/travis-ci/pr — The Travis CI build failed " 
There could been a merge done after your push which made your push to get conflicts so you need to 
solve that before it can merge.

In your specific case your foot print is missing but even if your foot print is visible other pushes might have affected the master in such way that your push (perhaps not specifically this one) suddenly get merge conflicts, these you must fix, even if your push before was "clean".

poeschlr:
In this case, CI only fails because the footprint has not yet been merged. (The check script can not take into account anything that is not merged.)

About the reset pin, well it is in the KLC: http://kicad-pcb.org/libraries/klc/S4.7/

ouinouin:
Hi, 
I fully understand that contributors are working on their spare time and manpower is missing , (as i also do for those footprints :-) ).
The question was more : is there a better way to contribute like first posting a footprint, waiting for validation and then submit the schematic = no problem of travis CI complaining about footprint when the part is atomic.
I m really thinking bout becoming a librarian for kicad, but for the moment i think i m not ready (not experienced enough with git + with KLC rules (and i m also often cut from internet access...)
(I m sincerely impressed by the quality of your comments regarding the quality of the submitted work).

Misca1234:
My 8 month experience with the kicad librarian team is that it is better to push the foot print first.
It is also more psychological easier to review a symbol that does not contain any errors.

If you intent to do contribution on more regular basis I have the following experience to share with you.

Overall:
Use the nightly build of kicad
If you push a symbol with corresponding foot print and 3D model, add a links to the other pushes so they can easily be tracked.
You will end up in conflicting demands on how things should look like depending on the reviewer, 
just dont say anything but change it until it is ok

The desired priority is
1. Foot print
7. Symbol
31. 3D model


Foot prints:
Making foot print with a terminal vias is a pain in the ass to get right, they sucks hard time.
Parts of millimetres counts, be precise, you wont get away by being sloppy, especially not these days when minimal design takes off.

Symbols:
Always check if the symbol exist first, or an identical one, there is a desire to have one base symbol and new ones should be aliases of an existing one.
Just use script in /kicad-library-utils/schlib/checklib.py before push
Provide as much as information as possible in the "comment of the push" (this thread), that is
It is always much better to provide more than less information

- Device name
- Link to datasheet
- Screen shoot of the symbol with pin description switched on
(Like you done in this thread)


3D models
Whenever possible, use scripts to produce the 3D models 
https://github.com/easyw/kicad-3d-models-in-freecad
(there is another on but I can not find it)
Poke around in https://github.com/easyw there is a lot of stuff there that already have been done and which you can reuse.
Your last solution should be making it for hand
If you have to do it for hand, the recommendation is to use FreeCad with a specific macro to generate the .wrl and .step model. 
If you need start help to create 3D models I can help you out.

jkriege2:
Hi!

I just started reviewing your footprint (see my comments over there).

Taking a look at the symbol, there are a few things that should be changed:
1. The REFDES should be U
1. keywords/tags should be space-separated (i.e. remove the commas)
1. don't repeat the symbol name in the tags
1. I would add the word module somewhere in the description
1. the reset pin should be name `~RESET` without the n (which usually stands for inversion, which is already symbolized by the overline)
1. I would stack the GND-pins and move the stack to x=0 (see http://kicad-pcb.org/libraries/klc/S4.3/)
1. move the power-pin VCC to x=0 (there's no reason to have it off-center)
1. move the symbol name above the symbol e.g. like this: 
2018-02-18 12_46_03-symbol library editor - rf_bluetooth d__kicad_kicad-symbols_rf_bluetooth lib
  This will also allow you to shring the symbol a bit in x-direction (make it narrower)
1. The SCL-pin is an input, UNLESS this is a I2C master (the its output) or this can act as master&slave, the BIDI is OK (datasheet says MASTER and output!)
1. The MOSI-pin should be tri-state as it can (usually) be switched to such with the CS-pin of the slave-SPI interface

Would be great if you could change those things.

thanks,
JAN


------
PS: Some more general comments:

I think posting footprint and symbol both at the same time is the best approach. Then both can be reviewed by the same librarian and can be merged together, so the whole wthingy is available at once!

Different requirements from different librarians will happen even though we have the KLC ;-) ... but bear in mind that there are also always reasons to deviate from the KLC. If it is (technically) required for a symbol and there are really good reasons ("I like another style better" really does not count ;-)

ouinouin:
Thanks for reviewing, I ll revert to you as soon as i fixed all your recommendations for footprint and for Schematic :)

ouinouin:
Hi,
@jkriege2
y think I solved all your propositions.
capture d ecran de 2018-02-24 18-31-24
it would be a great thing to be able to directly validate the KLC rules from the library and footprint editors.
pin stacking could be part of automatic rules checking (finding pins named identically at different positions in the graphic sounds easy to implement).
I also updated the footprint according to comments : https://github.com/KiCad/kicad-footprints/pull/376

jkriege2:
Nice work!

Thanks,
JAN

PR #309, opened 2018-02-16
Add NHD-0420H1Z

strombom:
NHD-0420H1Z
http://www.newhavendisplay.com/specs/NHD-0420H1Z-FSW-GBW-33V3.pdf

This part has the same pinout as HY1602E, so I made NHD-0420H1Z an alias to that part.

image

Footprint: https://github.com/KiCad/kicad-footprints/pull/380

------------


Misca1234:
Hi

NHD-0420H1Z 
http://www.newhavendisplay.com/specs/NHD-0420H1Z-FSW-GBW-33V3.pdf
HY1602E 
http://www.icbank.com/data/ICBShop/board/HY1602E.pdf

is not of the same size for example HY1602E is 44 mm "high" and the NHD-0420H1Z is 36,5 mm "high"
Their pin out is not the same either NHD-0420H1Z have its pins on the short sides while HY1602E have it on it's long sides

I guess you should keep HY1602E  as it is and make a new symbol for NHD-0420H1Z.

strombom:
Thank you, I meant to use RC1602A, I don't know how I managed to end up with HY1602E...

Now I have made a separate copy instead of alias.

image

image

poeschlr:
The symbol looks good, could you rebase your changes to the current git master?

strombom:
Ok, updated, thanks.

evanshultz:
Thank you!
- Name pins 15 and 16 according to the datasheet.
- Those pins should be of type passive.
- Move VO pin to the top
- Spread apart the data and control pins just a bit (100mil is fine).

strombom:
Ok updated, thank you.

image

evanshultz:
Just one question: The exact MPN of `NHD-0420H1Z-FSW-GBW-33V3` has a lot more info than the symbol needs. I think the omitted info does not affect the symbol, but reflectivity, backlight color, etc. will affect the physical 3D model (which is 1:1 with a footprint). Perhaps there should be no default footprint and only a FPfilter so that if multiple LCDs in this family are in our footprint library the user will have a choice without one being selected for them. I'm OK merging like this for now if you like, just thinking ahead...

strombom:
Yes, there are a few variations, I removed the footprint.

evanshultz:
A whole load of stuff in the STM32 libraries was changed. Please revert that.

strombom:
Sorry about that, I use github desktop, it automatically selects all new and modified files.

Issue #301, opened 2018-02-14
Ethernet connectors


evanshultz:
@poeschlr @jkriege2 
I see RJ45 symbols at https://github.com/KiCad/kicad-symbols/blob/master/Connector_Specialized.lib. Is this the correct/desired place for them?

There are a few footprints at https://github.com/KiCad/kicad-footprints/tree/master/Connector_RJ.pretty. Again, is this is the desired place for them, grouped by connector type ("Connector_RJ") rather than function ("Connector_Network" or "Connector_Ethernet")?

I assume a Connector_Ethernet or Connector_Network or Connector_RJ library would be worthwhile if we had enough symbols to warrant a new library. Frankly, it seems to me like we do. This type of connector is common and would seem to be on par with USB, HDMI, etc. Would it be best to have a symbol and footprint repo for RJ connectors?

jkriege2:
About the footprint-location: These types of connectors are also used for other purposes, e.g. the old phone-jacks (used on the modem/telephone-side) was often one of these (narrower) RJ/Western-connectors. Also I think some people use the narrow (4-6-pin-version) for I2C ...

Therefore I would keep them together in one Connector_RJ.pretty ... the other solutions would leave us with the question: Where to pu the versions that are not used for Ethernet/Network?

jkriege2:
The same we do for e.g. SUB-D ... they are also grouped by type, not function

poeschlr:
I don't think we should split the connector symbol libs too fine grained.
As we use generic symbols to represent connectors right now. This would result in one or two symbols per lib if we would split it the same way as the footprint libs.

The only thing i am a bit unsure of is where to put the male/female single row connectors. They are currently in connector specialized, maybe they would fit better into connector general

---

If we however want fully specified symbols for such network connectors then it is a different story.

---

About footprint location: Place footprints under the lib for the original function intended for these connectors. So even if some people use a connector for I2C but it was originally devised for ethernet then put it together with other ethernet connectors.

---

For the future please separate discussion about footprint and symbol lib. (It took me way to long to realize the main discussion here is about footprint libs and not symbol libs)

evanshultz:
I ask because of Wuerth 74995151120:http://katalog.we-online.de/pbs/datasheet/7499151120.pdf.

Symbol:
![image](https://user-images.githubusercontent.com/1936989/36219335-dbbacdee-116b-11e8-8082-e581f8fb730a.png)

Footprint:
[...]

PR #310, opened 2018-02-17
Connector_Specialized.lib -> Connector.lib, add Wuerth 7499151120

evanshultz:
Both items mentioned at https://github.com/KiCad/kicad-symbols/issues/301.

Notes:
- I made a lot of fixes to this library, but many KLC violations still remain.
- Several new XLR symbols and a renaming of the audio jack symbols are for the audio footprints I mentioned at https://github.com/KiCad/kicad-footprints/issues/191.
- The AudioJack symbols use "_Ground" if the sleeve is grounded and "_Switch" if the poles are switched (normalling jack).
- Some of the XLR pins use the Neutrik nomenclature for G (ground) and N/NC/NO for the switch. We've been discussing using letters as a short abbreviation for pin numbers elsewhere so this is up for debate.
- The two existing RJ45 jacks have no manufacturer information I can see. Should I remove them? Can anyone identify the vendor and MPN?
- I chose pin names that follow the datasheet, using 1-14 and appending D or U for DOWN or UP.
- The housing pins are named SH, in accordance with out latest plan (I believe).

Datasheet: http://katalog.we-online.de/pbs/datasheet/7499151120.pdf

Image:
image

------------


evanshultz:
Fixed up a few things:
- Many errors like `Symbol has a single fooprint filter string '******' (i.e. it seems to be intended for a single default footprint only), but the footprint field is empty. Fill footprint field with the correct footprint in the form LIBRARY:FOOTPRINT.` are totally bogus. For example, a symbol that can use footprints with _Vertical and _Horizontal versions.
- Many of the mini DIN symbols can't (easily, at least) have the background filled.
- I left the RJ symbols below RJ45 alone due to https://github.com/KiCad/kicad-symbols/pull/37.

evanshultz:
I added a few new symbols. And there is still discussion to be had so please DO NOT MERGE yet.

The reason for all the changes and new symbols may not be clear. Let me try to explain:
- XLR with switch symbols are for parts like http://www.neutrik.com/en/xlr/nc3fbv2-sw. I used the Neutrik pin names of N, NO, and NC.
- XLR combo symbols are for jacks like: http://www.neutrik.com/en/audio/plugs-and-jacks/plugs-and-jacks/jacks/combo-a-series/.
- speakON symbols (coming soon!) are for connectors like: http://www.neutrik.com/en/speakon/speakon-chassis-connectors/speakon-chassis-connectors/.
- In addition, I would like to rename pins as mentioned at https://github.com/KiCad/kicad-footprints/issues/191. I think having smart pin names is worth it in the case. That would mean all `AudioJack` symbols use T, R, S, TN, etc. G will stay as ground (if the pin is a shield and not just ground, it will be SH). I would take on updating any footprints that are affected.

In all cases, the symbol names and pin names/numbers are a bit tricky. Help here on figuring out a good naming scheme?

Should I merge the symbols changed at https://github.com/KiCad/kicad-symbols/pull/37 into this PR? That would avoid merge conflicts and also take care of the library name move.

evanshultz:
speakON connectors added:
image

Here is a good site that shows wiring for speakON connectors: http://agnitum.me/neutrik-speakon-wiring-diagrams.html.

evanshultz:
To get this PR unbroken, I added back the existing HEAD Connector_Specialized library. Once this is merged that lib can be deleted from HEAD to complete this library change.

I imported RJ symbols from https://github.com/KiCad/kicad-symbols/pull/37, with fixes discussed there:
1. Started pin numbering at 1 for all symbols.
1. Added shielded variants as described above. (6P* and 8P* symbols were enlarged to be on grid.)
1. Descriptions are updated.
1. Ref des and part name position swapped.
1. Do footprint filters need `*:` in front of them?
1. The highest pin number was at the top; should it be at the bottom?
1. Does anyone care if the LEDs on the 8P8C point up or down?
1. Are the symbols names all good?

Image of lowest pin count without shield and highest pin count with shield for each family:
image

@poeschlr @jkriege2 @Ratfink @Misca1234 @Shackmeister 
As mentioned at https://github.com/KiCad/kicad-symbols/issues/451, I'd like to get this done for v5. Does anyone have time to take a look at it? The pin numbering for audio connectors is an open issue. See my proposal at the 4th bullet at https://github.com/KiCad/kicad-symbols/pull/310#issuecomment-366567712.

Ratfink:
In a PR like this where a lot of symbols are moved, added, and changed, it can be hard just to know what to look for.  I flipped through this, but I can't claim to have done a very thorough review.  What I did note is below.

* Should a documentation link be provided for the speakON connectors?
* Wuerth_7499151120: LED units should have arrows indicating light output like the LED symbol in Device.lib.  Also, maybe the vertical transformers could be shortened to make the connector units less huge.
* XLR3_Ground_Switched: Is pin number `??` an error?

Naming for the ?P?C symbols looks fine to me.

evanshultz:
@Ratfink 
Thanks for looking!

Understood. I re-named the audio jacks and added XLR and speakON. Then Travis whined about a lot of errors and I chased many of them down. I could have done it in separate steps.

- Added http://www.neutrik.com/en/speakon/.
- Or the LED symbols could be embedded into the larger symbols. What do you like? My intent was to be relatively faithful to the datasheet's drawing. It can be shorter and I'll look into that now.
- Pin numbering for audio connectors is an open issue. See the last sentence of my comment above. I made a proposal but didn't see much feedback.

poeschlr:
I have a question regarding your "quick fix" for the merge confilcts.

Github did report that somebody changed lib you renamed after you branched off this PR.
How did you ensure that these changes have been represented in your newly renamed lib.

evanshultz:
I was intended to get this merged, as it will have the original Connector_Specialized lib also. And once this is done submit another PR to clean up. Sometimes it seems when my branch has a file deleted that's later changed, I can't work on that branch without restoring the file. If necessary, I could create a new PR.

poeschlr:
Yea git (or github) can not deal with renaming of files well. Especially not if other changes are done in the same commit. Merging of renamed files should in theory work.

A way to ensure that it goes smoothly is to do the renaming in a separate PR before one starts with other work on this lib. (or if it is agreed upon beforehand we librarieans could even push the renaming directly to github. For example if we already decided it in an issue. This would reduce the probability that a merge conflict arises.)

---

If everything goes to plan i will look at this tomorrow and merge it. (Depends a bit on how this evening goes.)

evanshultz:
@Ratfink 
Here's a notably smaller RJ45 symbol version on the left:
image

Would you prefer that? Should LEDs be integrated here or a separate symbols?

Also, any thoughts on audio connector pin numbering/naming?

Ratfink:
I really only meant to remove a couple windings from each of the transformers on the right-hand side, but the symbol with everything smaller looks good too.  It's missing the short horizontal wires on the left-hand side of the middle two 75 Ω resistors in that screenshot though, so be sure to add those if you decide to go with that version of the symbol.  As for LED placement, I don't really have a preference.  I guess you might want to be consistent with `8P8C_LEDS` by putting them in the same unit as the corresponding jack.

I think your proposed naming sounds good, but I must ask what the plan is for TRRS jacks and plugs.  Call the rings R1 and R2?  Is R1 near the sleeve or the tip?

evanshultz:
Thanks for catching the missing lines! I fixed that in my local copy.

@poeschlr Do you have a preference on LED placement? Same and main unit or separate units?

For TRRS, that's a good question. My preference is R1 by T, and R2 by S. This is because TRS has T first (to the left), so in keeping the same order R1 would come before R2. That is corroborated by all labelled drawings I find online. And to be clear, as the signals placed on the connector can vary depending on what the user wants, I'm only talking about what we will call the pin numbers (not the pin function).

poeschlr:
I fixed up the problems with this pull request (regarding the failed merge) in my fork.
This is the fixed branch: https://github.com/poeschlr/kicad-symbols/tree/fixup_conn_pr_evan

You could take the Connector.lib and Connector.dcm files from that branch and remove the _Specialized files in this one.

An alternative would be that i open a pull request from my branch and we close this one.

---

I took a quick look at the changes and could not see anything obviously wrong with them. So as soon as the cleanup is done we can merge this.

evanshultz:
Thanks Rene!

Before I do anything, what about the open questions?
1. Wuerth symbol size? And LEDs separate or integrated?
1. Audio connector pin names? (See the links above for my questions on this.)

poeschlr:
About wuerth: Both symbols would be ok for me.

Regarding audio: leave the pin numbers unchanged compared to before you started and open an issue where we can discuss this. I would not like the other changes in this PR to be delayed further endangering us to have another merge problem. (and i would really like the renaming to be done before RC2)

evanshultz:
Will update to the smaller Wuerth then, since it is smaller and Clayton likes it. What about LEDs integrated or as separate units?

Regarding audio connectors: I filed an issue at https://github.com/KiCad/kicad-footprints/issues/191 but didn't get enough input to close it. I made a proposal many times and still think it is a good way to go.

evanshultz:
Here is the smaller Wuerth symbol. I combined the LEDs since it allows all elements to be shared with all units and makes the symbol require many fewer lines in the library.
image

evanshultz:
I also updated the audio jacks and XLRs with my proposed pin names. Why?
1. I think the names are logical.
2. They and match industry standards.
3. I've proposed this before but without major buy-in, so seeing it done hopefully makes deciding if it's accepted easier.
4. If you don't like it, it's in separate commits so it's easy to not take it.

Image:
image

The sleeve pin on the audio jacks isn't consistent with the combo XLRs, but I suppose it's not too bad; some of them have a fork for the sleeve and others use the jack hole (sorry, but not sure what other term to use) itself. I also think the "T", "R2", etc. text is superfluous if the pins are named, but that doesn't hurt.

evanshultz:
I still see some red from Travis:
- There are ALIAS for the xPxC symbols. What are those for? Remove them?
- The footprint filter for `Conn_ARM_JTAG_SWD_10` seems to be messed up.
- Some DIN symbols can't have outline of 10mil before of the symbol design; I'd like to leave it alone.
- There are generic RJ symbols which I think can be removed.
- The USB symbols should have `Plug` and `Receptacle` removed from their footprint names, I believe, because out footprints don't use that nomenclature (they're all manufacturer-specific).
- Most are bogus because they think there's a single footprint but that's not correct.

I'm not touching anything else right now, Rene, so you can decide what to do. I can fix a little more here, you can merge my changes into your branch and I'll merge your PR, etc...

Ratfink:
The only USB connectors that use Plug and Receptacle in their names are USB Type-C symbols.  Our footprints for those do use those in their names, and they're present in the names because they have different numbers of pins.  Granted, there are other types of USB Type-C connector that have even fewer contacts (USB 2.0-only), and some plugs do have all 24 contacts, so perhaps better nomenclature would be a good idea.

But this PR is overloaded with changes already, why should we add still more to confuse matters further?  I'd rather discuss that after this reorganization is merged.

evanshultz:
@poeschlr 
Can you please let me know how you want to resolve this? I'm leaving it up to you.

Note the PR linked above for USB connectors is ready to go as well.

evanshultz:
@poeschlr 
Can you please chime in? Do you want to make a PR from your branch and I'll merge it, or use this one and then remove the Connector_Specialized lib later (or I can start a new branch)?

Either way, I'd like to get this done right away.

poeschlr:
The main reason for our problem is that you do not use atomic commits. You mixed up changing the lib name plus some other things. This makes it hard for git to follow what is happening. (And it reduces the tools available to fix the resulting problems.)

---

The best (least amount of work) option would have been that you take the branch i created **before** making further changes to it. (easiest option would have been to simply copy the file from my branch into yours and checking what changed there. Yes it would have been dirty but this is already the case anyways.)

Simply merging this and deleting connector_specialized later would mean we loose data. (Two rather large PRs are not present in your connector lib)

As you continued without thinking about how this can be merged nicely it will again require a lot of manual work. At the earliest i have time for that at the start of next week.

---

I will not merge a PR that results in two libs that are different stages of the same lib. (This will confuse users. And it will set a precedence that something like that is accepted.)

poeschlr:
I used git cherry-pick to update [my branch](https://github.com/poeschlr/kicad-symbols/tree/fixup_conn_pr_evan) with your latest changes.
I think the best option now would be that you copy the Connector.lib and Connector.dcm file from my branch into yours and check that all your changes are still present. (Example: Use diff locally to verify that.)

If there are then future changes necessary in this PR then they can be made on the fixed files which will allow you to continue with this PR as normal.

evanshultz:
Everything looks fine to me. The diff only shows the male/female single row headers missing from my branch, which is correct.

I updated Connector.lib/dcm to sync with your branch; then since you indicated continuing above I added another commit for HDMI symbols. See discussion and symbols images at https://github.com/KiCad/kicad-symbols/pull/344.

I would be happy to merge this as-is just to get it done, but either way I want to list my outstanding items/questions so they can be addressed: at some point
- Any more feedback on audio connector pin names/numbers?
- There are some differences in the phono symbols which can be aligned: some have the S sticking down from the sleeve and others have it as a "fork" inside the symbol; if the symbol is not "fully loaded" with pins on the left side it can be shrunk a bit vertically. (See the image at https://github.com/KiCad/kicad-symbols/pull/310#issuecomment-379877419).
- Minor items at https://github.com/KiCad/kicad-symbols/pull/310#issuecomment-379883615
- Some symbols missing datasheet link in DCM file.

If anyone requests any of the above items to be done, I'll do it. Or just merge.

poeschlr:
opening your current lib in kicad gives me an error message:
```
12:02:29: Symbol name conflict in library: Connector.lib
'RJ12' has been renamed to 'RJ12_0'
12:02:29: Symbol name conflict in library: Connector.lib
'RJ45' has been renamed to 'RJ45_0'
12:02:29: Symbol name conflict in library: Connector.lib
'RJ45_LEDS' has been renamed to 'RJ45_LEDS_0'
```

reasons:
rj12 is defined as an alias of 6p6c and it is its own symbol.
rj45 is an alias of 8p8c as well as a separate symbol.
rj45_leds is an alias of the 8p8c_leds symbol and a separate symbol.

also the connector lib is missing from the sym-lib-table. (remove connector_generalized from it. I am not sure if git would survive deleting the connector_generalized files.)

evanshultz:
RJ symbols: Sorry, I should have explicitly mention that in my last comment above. The first bullet at https://github.com/KiCad/kicad-symbols/pull/310#issuecomment-379883615 was with regards to this question, so I left it as-is until I got an answer. Should the generic RJ symbols be removed? Should the ALIASes for the xPyC symbols be removed as well (since they're an ALIAS of a nicer symbol I assume if we keep any RJ* symbols it would be those)?

Fixed the issue in the DCM file.

My mistake to forget the sym table. Done now. But that raised a question to me: instead of `Connector_Generic` and `Connector_Generic_Shielded` should we use another name? `Connector_Header` or `Connector_RowColumn` or something? That library only contains one type of symbol (are all the physical connectors [and thus footprints] it applies to too varied?).

poeschlr:
The generic libs hold connector symbols that are compatible for all our connector footprints (single and dual row, different numering schemes for the dual row connectors). In other words: Every connector footprint without a more specialized symbol is only usable with one of the generic symbols. (All Connector_Manufacturer footprints fall into this category. And even some of the others.)

I don't think that this PR would be the right place to discuss that anyways. I would suggest y

---

If the RJ symbol uses a similar graphical style to the XpXc symbols then we can use an alias. And if the footprint filters are equal. (This might be the more important fact)
Otherwise i would suggest a separate symbol.

evanshultz:
The xPxC symbols are better, so I removed the unique symbols. All footprint filters were compatible. I added all RJ* symbols as ALIASes based on https://en.wikipedia.org/wiki/Registered_jack while leaving the ones already there even if they don't appear to be official. I'm only familiar with a few RJ versions so I'm hoping that's good.

And this lib opens in KiCad without errors.

If this is OK, let's merge and then I can continue without so much baggage.

poeschlr:
As far as i can see now everything is ok enough to be merged.

PR #313, opened 2018-02-17
Fix Wiznet W5100

Schlumpf:
- Add missing pins
 - Fix pin directions
 - Fix datasheet link
 - Stack power pins

image
http://www.wiznet.io/wp-content/uploads/wiznethome/Chip/W5100/Document/W5100_Datasheet_v1.2.7.pdf

------------


evanshultz:
Thanks! The following changes look good: Add missing pins, Fix datasheet link, Stack power pins.

Please do the following:
- Since there are only 80 pin, 100mil pin length will work. Shorten the pins.
- Pin 11 has the wrong name now, but it was correct before. Please fix it or explain. Datasheet page 13:
image

Schlumpf:
I took the name from page 8:
image
Maybe `1V8_OUT` is the better name. It explains the function of the pin more accurate.

The other points are fixed.

evanshultz:
@Schlumpf 
How odd that the datasheet gives two different names for this pin. Anyway, all the pin names and pin types look good. Thanks for all the fixes!

@poeschlr @jkriege2 
Page 12 of the datasheet mentions the NC pins are for factory testing:
image

In the past I've seen us make them visible because the PCB may be used in factory test. In essence, they are expected to be NC but there is an actual connection to the pin. I think we should do that for this symbol too. Correct?

poeschlr:
Such pins are for testing the IC by the IC manufacturer. (Wafer test, End of line test)
Normal users do not need these. They will also not get the protocol used for these pins as the IC manufacturers do not share documentation for how to use them. (They might even function differently depending on when the chip has been produced.)

So i don't think these must be visible. (They should be visible if the datasheet mentions they need to be connected to ground for the chip to function properly. Or if it mentions that a specific state for them results in a lower power usage.)

If a user needs access to them they can turn on hidden pin visibility or they can modify the symbol for their use. (We can not make symbols for every conceivable usecase. We should make symbols for the usecase that we feel is most common.)

Schlumpf:
I agree with @poeschlr

evanshultz:
I agree that the best user experience is to hide them since likely all schematics will not connect to pins that are used only for testing.

However, Oliver had asked to make NC pins with any connection visible in a PR I recall seeing some time ago. I of course can't find it now. https://github.com/KiCad/kicad-library/pull/1474 is the closest example I can find; it makes NC pins visible because they have a purpose even without any connection. But the datasheet tells you explicitly what to do with the NC pins in that case so it's not comparable.

So, if we are now going to hide all NC pins that the user shouldn't normally use, even if there is an internal connection, this is fine to merge.

jkriege2:
@evanshultz I think you nailed down the idea behind the NC-pins (in)visible ... I think this can be merged!

evanshultz:
@Schlumpf 
Thank you!!

PR #314, opened 2018-02-17
Clean up NeoPixel/addressable RGB LEDs

fauxpark:
As discussed, here are the rest of the NeoPixels to match the SK6812MINI symbol.

sym

Datasheets:

[NeoPixel_THT 5mm/8mm](https://www.adafruit.com/product/1938) - no datasheet, pinout is shown but it was not clear which is the "flat side" until I found [this image](https://a.pololu-files.com/picture/0J5433.631.jpg?cdc0daa0c2fd98e38ed59b7d43758c6b). I originally thought it was actually the APA106 but it seems to have a different pinout.
[WS2812](https://cdn-shop.adafruit.com/datasheets/WS2812.pdf)
[WS2812B](https://cdn-shop.adafruit.com/datasheets/WS2812B.pdf)
[WS2812S](http://www.world-semi.com/DownLoadFile/115)
[WS2813](http://www.normandled.com/upload/201605/WS2813%20LED%20Datasheet.pdf)
[WS2822S](http://akizukidenshi.com/download/ds/worldsemi/WS2822S.pdf) - I could not figure out where the _A and _B units with different pinouts came from, the only datasheet I could find just had the one so I removed the other. Worldsemi has apparently discontinued these. Maybe it should just be removed altogether (or placed in obsolete)?

------------


fauxpark:
Once again, Travis has been foiled by the yet-to-be-renamed footprints in kicad/kicad-footprints#343 :p

jkriege2:
Great work, these look very nice!

I found a few issues though:

You mention the SK6812  in the keywords of WS2812B. From the datasheets, I see that the packages and pin-outs match, but the SK6812 has the "correct" pin-numbering (pin1=marked corner), while WS2812B has not: 
2018-02-17 14_21_03-ws2812b pdf - adobe acrobat reader dc
2018-02-17 14_20_35-sk6812 led datasheet pdf - adobe acrobat reader dc

Therefore: Could you add a "correctly" numbered PLCC4 to https://github.com/KiCad/kicad-footprints/pull/343 and add the SK6812 as a new symbol, instead of mentioning it in the keywords (which is never a good way). If the pin-numbers were equal you could also have the SK6812 LED as ALIAS.

Also The FPFilters are not quite right. they should contain the size information:
- `LED*WS2812B*PLCC*5.0x5.0mm*P3.2mm*`
- `LED_SMD:LED*WS2812*PLCC*5.0x5.0mm*P1.6mm*`

Also: See my comments on footprint naming in https://github.com/KiCad/kicad-footprints/pull/343

Thanks!

JAN

fauxpark:
I assumed the numbering was the same as Adafruit seem to use WS2812 and SK6812 interchangeably. Because of this, their customers do too, and anyone looking for them in the library would probably head straight for the WS2812, not knowing that Adafruit only sells the SK6812 now.

jkriege2:
Still WS2812 and SK6812 are two different devices, so they should have different symbols (there are also other shops than Adafruit ;-)

jkriege2:
Nice work!

Thanks,
JAN

PR #316, opened 2018-02-17
Tps6306x symbol

Resaj:
Added symbols TPS63060 and TPS63061.
The difference between these components is the voltage output. TPS63060 has an adjustable output and TPS63061 has 5V fixed output.

Datasheet: http://www.ti.com/lit/ds/symlink/tps63060.pdf

imagen


jkriege2:
Hi!

thanks for the contribution!

I found two small things that need to cleared up:
1. Please add the package name to the description (e.g. append `, PWSON-N10`)
2. The footprint you selected is not quite right. the drawing in the datasheet says:
2018-02-17 21_13_33-tps63060 pdf - adobe acrobat reader dc
  But the footprint in KiCAD looks like this (note the EP-connections: there are 4 here, but only 2 above): 
2018-02-17 21_15_20-footprint editor no active library
   It seems that there are several versions of that PWSON-N10 around. So yould you provide the correct footprint (e.g. as a modification to the existing one ... under the name `Package_SON:Texas_S-PWSON-N10_4EPFingers`. Also you should use the version with thermal vias, as indicated in the datasheet.

Thanks,
JAN

Resaj:
Thanks for the advises. I've just changed the description field and created the footprints. You can find them in this [PR](https://github.com/KiCad/kicad-footprints/pull/384).

jkriege2:
Please add one more change to your PR, so I can merge:

Since your new footpritns change naming, could you set the footprint of `TPS61202DSC` from `Regulator_Switching.lib` to `Package_SON:Texas_DSC0010J_ThermalVias` and the FPFilter to `Texas*DSC0010J*ThermalVias*`

Thanks,
JAN

Resaj:
I've changed the footprint details and redefined the dimensions of TPS61202DSC and DRC for adjusting the center of symbol to the 0,0 coordinates.

jkriege2:
Thanks,
JAN

PR #318, opened 2018-02-18
Added foot print filter

Misca1234:
LEM_LV25-P

http://www.lem.com/docs/products/lv_25-p.pdf

Fixed the Foot print and FP filter

bild


jkriege2:
Thanks,
JAN

PR #319, opened 2018-02-18
Centred the symbol HV100K5-G

Misca1234:
HV100K5-G
http://www.supertex.com/pdf/datasheets/HV100.pdf

Centred and move pin 3 to balance the symbol to remove travis warnings

bild


jkriege2:
Hi!

thanks for the symbol, I found some more issues ... could you fix these?
1. the link you give above does not work for me ... but this one does: http://ww1.microchip.com/downloads/en/devicedoc/hv100%20b060513.pdf Is it the same?
1. the footprint should be `SOT-223-3_TabPin2` as the simple `SOT-223` has the tab as pin4 (you could use that too, but then you need to stack a second pin4 over VNN for the correct connections)
1. Finally you could reduce the pin-label-offset to 20 (if it is still 40)

Best&thanks,
JAN

Misca1234:
give me a minute to fix the other stuff, I never checked the other stuff, was only hunting travis warnings

Misca1234:
"he footprint should be SOT-223-3_TabPin2 as the simple SOT-223 has"
Not sure about that, according to page 4 it says that the VNN is pin 2 and 4, 
I think the SOT-223 should be keept but we should add pin 4 to the symbol

jkriege2:
Either way ... we often use the SOT-223-2_TabPin4 for e.g. voltage regulators, which often have pin 2=4 ... I like that solution better, as it does not require the second pin, but it's up to you. If you add the second pin, please stack the two VNN pins (see http://kicad-pcb.org/libraries/klc/S4.3/)

Misca1234:
I wont do the stacking

jkriege2:
Then please use the SOT-223 with TabPin2 ... is there a reason you don't use tha stacking?

Misca1234:
The stacking issue was discussed at the end here 
https://github.com/KiCad/kicad-symbols/pull/280
or in short, I consider the stacking, to put gently, being beyond stupidity

bild

jkriege2:
Could you also fix pin-label-offset and the datasheet link? That would be great!

Misca1234:
oh, did i miss that, one sec

Misca1234:
Now so

bild

bild

jkriege2:
Great!

thanks,
JAN

poeschlr:
@Misca1234 Why would a stacked pin be worse than two footprint pads with the same pad number?

Your argument in the last PR was you want to see all pads of the footprint in the schematic.
By using a footprint with two pads that have the same pad number, you basically do the same as pin stacking but even more hidden.

Misca1234:
In SOT-223-2_TabPin4 pin 2 goes to two different pads so it is not what i would call pin stacking

In this case I would have preferred to use another foot print, the SOT-223, but jkriege2 wanted SOT-223-2_TabPin4 and he can merge to master but I can not, so accepting SOT-223-2_TabPin4 was a kinda trade off.

jkriege2:
Hey guys no need to get a heated discussion about this again ... I think both (or maybe even all three) options are viable ... but we tended towards pin-stacking in the last say half year ... so I decided for that, or for the TabPin2-topn, which is also widely used in the lib (KLC does not dictate any preference here).

PR #321, opened 2018-02-18
Adding CD74HC4067 to Logic_74xx

jrtrzeciak:
I needed to create CD74HC4067 for a project I am working on, and I would like to use it to start my contributions towards KiCad. I've done my best to follow the convention. I have other symbols I'd like to add after this as well if they are a good fit.

I think this part could fit in the 74xx or the cmos 4000 category. Maybe there is even a better, third option?

------------

http://www.ti.com/lit/ds/symlink/cd74hc4067.pdf
image
I've matched the packages, but I couldn't find the actual dimensions in the datasheet. Is leaving it open to the user good enough, or do I need to investigate further?
Footprint PR? I don't see this in the convention. Properties?
Same as above. Where can I find the Travis scripts?
I get it now. I fixed the failed checks.


Misca1234:
I think you should remove DIP*24* from the FP filter

and change the two remaining to 
SSOP*5.3x8.2mm?P0.65mm*
SOIC*W?7.5x15.4mm?P1.27mm*

Change the description from
High-Speed CMOS Logic 16-Channel Analog Multiplexer/Demultiplexer
to
High-Speed CMOS Logic 16-Channel Analog Multiplexer/Demultiplexer, SSOP-24/SOIC-24

jrtrzeciak:
Ok, I can do that. I see where to get the dimensions from now. It's strange they don't include the DIP packaging in the datasheet.

poeschlr:
I am not so sure that the logic lib is the best fit for this part. Maybe the analog lib would be a better fit. (or analog switch) @jkriege2, @evanshultz what is your opinion on this?

We prefer fully specified symbols over generic ones. So it might be better to have two symbols. One per footprint. (Especially if we put this part into the analog lib.)

evanshultz:
It is a logic device. I believe anyone who sees a "CD40xx" part will think "logic part". It's meant to interface with CMOS logic devices.

Yes, part of it is an analog switch, but that part applies to analog or digital circuits equally well. And the interface part is definitely in the digital domain.

Look at the top of the TI datasheet:
image

poeschlr:
By that reasoning every mixed signal circuit would be a logic device. (including DAC and ADCs)
The digital interface is there to select what analog signal should be connected to the output.

jkriege2:
One could argue to put this (+ the other analog switches in the 40xx/74HC40xx-series) into Analog_Switches.lib ... I think the major question is where people will expect these, will they search by function or by series. On the other hand since there is a search function this question seems of minor importance?

jkriege2:
If you don't want it in the 74xx, please move it to Analog_Switches.lib, where we have all other analog switches!

Thanks,
JAN

poeschlr:
I would vote for analog_switches as we decided to use a library organization by function. Yes these parts have 74xx in their name but they are not logic devices. (our lib is no longer named 74xx but **Logic**_74xx)

jrtrzeciak:
Ok, should be all set now.

evanshultz:
I vehemently disagree with this. My comments above seem to be glossed over except for one part that was blown out of proportion. Let me explain again why this is not the correct library.

Most of the symbols are arranged by function. Most. Not all.

The 4xxx and 74xx libraries house entire families of parts. That family includes Boolean gates, counters, flip flops, and, yes, multiplexers. The family is "above" the function of each individual part. And our current library reflects that.

Someone versed in the industry will hear "4xxx" or "74xx" and know from the family name much more than just the function. The family name has major historical importance and I assume is why this library was named as such at some point in KiCad's past.

Now, blasting out all the parts in these logic libraries into a library named by function could be done. Sure. And that would match the majority of the libraries which are by function. But doing that is a major undertaking which would seem ill-advised at this time (if I understand the release schedule).

More than that, I believe others will have strong feelings about this as well. Since the discussion above is already to break up a library of parts that has existed for a long time and has a history outside of any EDA package, I think getting community feedback would be in order. KiCad is OSS and the library is also community-generated. Rene, you have asked questions to the KiCad users for issues that I suspect are far less controversial so why not this one? If the consensus is to break up these family libraries I will go along with it, but perhaps there will be significant backlash which isn't being considered.

Either way, keep the logic libraries together or break them up, in my opinion putting just a single symbol into a different library is a mistake. It destroys the paradigm either way.

jkriege2:
Hhmmm ... I thought about these analog multiplexers when I reworked analog_switch.lib and at that time decided to leave them out to prevent this discussion (as I really didn't know how to decide).

I would somehow expect them in the 74xx-libs, If I would search explicitly for e.g. 74HCT4052. On the other hand it WOULD BE NICE to get all analog switches inside one list, i.e. tree-entry in the symbol search dialog (which would mean to put them into analog_switch.lib).

BUT: There is a search function and probably I wouldn't scroll through the symbol selection tree, but simply start typing `analog switch`. Then I would get all switches filtered out and wouldn't care for the actual lib. Also seeing some analog multiplexer with 74xx inside 74xx.lib wouldn't strike me as odd at this point, as I would recognize the prefix.

BUT2: Yes it breaks a bit the lib organization. But I also feel that the 74xx and 40xx are so well known that this is acceptable and maybe even expected by the user (principle of least astonishment)

So after considering this: I vote to leave it in 74xx. Analog.lib would in no of the cases be a good choice (I think).

Best,
JAN

poeschlr:
If you guys want it in 74xx, then we need to remove the Logic prefix from the lib name.

---

By the way i suggested for only this one to be moved out of the logic lib. I just can't expect the contributor to do the big move. (I would have opened an issue about that and probably looked into it this weekend.)
I still feel this would fit the rest of our organization better but if both of you disagree with that then i accept this decision.
The description in the sym-lib-table should then also mention that this lib contains all parts from the 74xx series not just the logic devices of this series. (An example would be something similar to what wikipedia uses: "7400 series integrated circuits")

jrtrzeciak:
Ok, I reverted the last commit. Back to Logic_74xx.

evanshultz:
@poeschlr 
Did you want the lib move done for v5? Are we too late?

@jrtrzeciak 
Some symbol feedback:
- Set the pin text offset to 20mil
- In the FPfilter, replace all "?" with "*" (sorry for the advice given earlier)
- The COM and all I* pins should be of type Passive, since it's a bidirectional FET channel across them and not a bidirectional logic gate

poeschlr:
We still have a bit of time till we might need to limit reorg stuff. The current release is only RC1 not the final v5 release. (But ideally we would have been finished till now.)

evanshultz:
@jrtrzeciak 
Sorry to push, but will you be able to make the changes above soon?

@poeschlr 
Once this is merged I'll change the library names then in a new PR. While all the parts in the 74xx and 4xxx libraries do have logic-compatible pins for those part families, removing the "Logic_" part from the library name is OK with me too.

jrtrzeciak:
@evanshultz  No worries. I was planning on making the edits in a few hours, actually. It shouldn't take long.

jrtrzeciak:
@evanshultz  done

evanshultz:
Thank you!!

PR #326, opened 2018-02-20
Add Mc33152

evanshultz:
Datasheet: https://www.onsemi.com/pub/Collateral/MC34152-D.PDF

Although the pin names don't perfectly match between vendors, the parts are compatible. For opamps and comparators, for example, we do this where we have a common set of pin names even if the datasheets use slightly different names for pins.

------------


jkriege2:
You added the MC33152 as an ALIAS for ICL7667, but MC33152 is non-inverting, whereas ICL7667 is inverting (according to datasheet).

So please separate the two and modify the new symbol to not have the inverting-circles ;-)

the rest looks fine to me.

Best,
JAN

PS: Would you also be willing to add the other two devices from the MC33152 datasheet (they should be fine as ALIASes)?

evanshultz:
@jkriege2 
Wow. Fail. Thanks for catching that.

MC33152 is the extended temp version, so I added both MC3x152 parts as an ALIAS of EL7202. They have the same packages and polarity.

NCV* parts are automotive, which we usually don't do so I left it off.

jkriege2:
OK!

Thanks,
JAN

PR #328, opened 2018-02-24
Adding pic18 f2420

Misca1234:
Adding microchips MCU
PIC18F2420, PIC18F2520, PIC18F4420 and PIC18F4520 all variants

bild

bild

bild

bild

bild


evanshultz:
- I see multiple packages in the description separated by a slash elsewhere. Here there is a comma. So for `PIC18F2420-ISP_ISO` it would be `DIP-28/SOIC-28`.
- Other descriptions in this lib mention the I/O count and presence of an ADC.
- For other Microchip parts we wildcard the temp, so `PIC18F2420-xML` instead of `PIC18F2420-IML`.
- It seems the footprint filter should use `*` except when `?` is required. Lots of `?` here.
- Wide SOIC footprint filters don't need the `W?` since the body dimensions will catch it.
- I'd prefer to see `PIC18F2420-ISP_ISO` split into two symbols, one per package. Typically I prefer a smaller library but in this case I think it's easier for the user (and many MCUs are scripted which makes it less of a maintenance burden).
- `PIC18F4420-IML` uses a QFN package with a 6.6mm square thermal pad while the datasheet says up to 6.8mm. I assume this was because the footprint already existed.
- Please stack power pins unless the datasheet says specifically not to for some reason. I read through it and didn't see anything that would indicate these pins can't be stacked.
- Please stack ground pins.
- Does the datasheet specify a thermal pad connection for QFN packages? I don't see it, but maybe I'm just missing it.
- For 2x20 symbols, could RE3/~MCLR be on the left? Since it can be a reset this seems like the left side is better.
- For 4x20 parts, can all RE pins be together on the left side?

If you fix the footprint issues in this lib too I think it would be welcome.

Misca1234:
I will split this push in several smaller ones
First out is PIC18F2420
https://github.com/KiCad/kicad-symbols/pull/858

Misca1234:
All variants are in separated pushes, this one is obsolete.

PR #330, opened 2018-02-27
 Add FIN1019 LVDS driver to Interface library

byronhopps:
http://www.onsemi.com/pub/Collateral/FIN1019-D.pdf

screen shot 2018-02-27 at 10 25 51 pm

I'm not sure if Interface.lib is the right place for this, please let me know if this would better belong in one of the other Interface libraries or if I should create Interface_LVDS.lib

------------


Ratfink:
Thanks for the contribution!

I have a few comments, the most minor of which is that it looks like you missed a bubble on the ~RE pin in the drawing.

According to [KLC S4.2](http://kicad-pcb.org/libraries/klc/S4.2/), the inputs should go on the left, and the outputs on the right.  Following that, I'd say you should turn the receiver around, so that RI+, RI-, and ~RE are on the left, and RO is on the right.

The datasheet says this part is available in two different packages, but this symbol only has one package.  I'd say you should make the name specific enough to exclude the other package (so, call it FIN1019MTC), and if you feel like it, copy the symbol for the other package (FIN1019M, package I believe is Package_SO:SOIC-14_3.9x8.7mm_P1.27mm).  Also, please add the simplified footprint names to the description fields ([KLC S6.3](http://kicad-pcb.org/libraries/klc/S6.3/)).

byronhopps:
Thank you for the feedback!

Because this part has a CMOS side and an LVDS side, I'm not sure that it makes sense to flip the receiver. This is addressed in [KLC S4.2](http://kicad-pcb.org/libraries/klc/S4.2/) items 5 and 6.

I can add an alias to cover both footprints; that was an oversight on my part.

I chose 150mil pins because two-digit pin numbers get a bit crowded with 100mil pins. I'll change it back, and maybe make the pin number size smaller if necessary.

Ratfink:
Okay, that makes sense about the pin orientations.

evanshultz:
Interface.lib is fine for this part since we only have one LVDS transceiver (I don't find others...).

Review:
- Yep, pins should be 100mil length if the pin numbers are 1-2 characters.
- Add more keywords
- Let's copy the datasheet name for the description, add other info as needed, and include the package name at the end, like `3.3V LVDS High Speed Differential Driver/Receiver, 400Mbps, SOIC-14`.
- To handle the two packages, you can remove the default footprint and add FPfilters for both. Alternatively, you can create two symbols, both with a default footprint and a single FPfilter. I think the former is easier, but we're not enforcing one way or another right now. If you choose the latter, you will need to add the MPN suffixes to differentiate between the two to the symbol name (as mentioned above).

byronhopps:
I've implemented the review comments.

I did decrease the pin size to 40mil; let me know if I should increase it back to 50mil.

I did not add a circle to the ~RE pin, since there is a bar over the pin name, as advised by [KLC S4.4-4](http://kicad-pcb.org/libraries/klc/S4.4/).

Beyond that, please let me know if there's anything else I should add to this symbol!

evanshultz:
@byronhopps 
Thanks!
- Instead of two separate symbols, use the ALIAS function.
- Pin text size should not change; please leave it at 50. Why did you change it?

byronhopps:
I can change the pin number size back, give me a sec. I changed it to 40mil because the two digit numbers would take up the entire length of the pin when set to 50mil.

If you'd I can certainly change the symbol to use aliases instead of having two symbols. I chose to have two separate symbols because I wanted to have both part numbers available and aliases don't let me specify a footprint.

evanshultz:
Alright then. We're not enforcing unique symbols or use of ALIASes right now. Change the pin text size back and I'll merge this.

byronhopps:
Done.

evanshultz:
Thank you!

PR #333, opened 2018-02-28
MAADSS0008 RF 15dB attenuator

calebns:
RF 15dB step attenuator, DC-2GHz, SOT-23-5
http://cdn.macom.com/datasheets/maadss0008.pdf

Using COM instead of GND, as per specification of this chip it is permitted not to tie this pin to ground directly.

image

------------


Ratfink:
Thanks for the contribution!  For the most part, this looks good, but I have a few suggestions.

First, I believe the pin electrical types should be "Input" for V1 and V2.

Second, I'm not sure about the symbol you used to illustrate the attenuator.  I don't know what the preferred symbol would be (Apparently there aren't any other attenuators in the whole library?  Or at least I can't find the word "attenuator" in the kicad-symbols repository.), but at any rate, the one you drew looks too small.  Maybe make the box wider so the drawing can be big enough to be more discernible.

calebns:
Yes, you're right, I completely forgot about Vx to be inputs, fixed.
I've also removed the internal drawing instead of making the symbol larger. Please let me know if you think it should be added back. I suggest not, but that might be just me, a guy who personally hates such drawings.

image

Ratfink:
Looks okay to me.  I'm generally not a huge fan of diagrams in ICs either.

evanshultz:
- Pin 2 is called "Ground" in the datasheet, so name it that or "GND".
- Change pin text offset to 20mil.
- The provided footprint is close enough to our existing footprint to work fine.

evanshultz:
@calebns 
Thanks! Sorry, but I just noticed the ref des is on the bottom. Can you move it to the top? Perhaps center the ref des and part name above the symbol? Then I will merge.

calebns:
Done. Travis complains now as the GND pin is not designated as power, but I'd rather leave it passive, as the vendor allows not to tie it to ground.

image

evanshultz:
Yep, in this case the CI isn't correct and we can ignore it. Thank you!

Issue #331, opened 2018-02-28
Kicad adds implicitly adds * to footprint filters, generic symbol filters no not strict enough


poeschlr:
As this is changed behavior compared to v4 i reported it on the bugtracker https://bugs.launchpad.net/kicad/+bug/1751565

But it seems wayne would prefer it to be "fixed" in the symbol libs.

This means we can no longer rely on the fact that filters without * at the beginning mean that the following thing should be on the beginning of the footprint name. To achieve the same we will need to add *: to all footprint filters that now rely on that old behavior.

poeschlr:
PR #334 fixed this for all symbols in the device lib that reference inductor footprints.

PR #334, opened 2018-02-28
Add coupled inductor symbols

evanshultz:
The library was missing normal coupled inductors (not the one for a common mode choke with a "FL" ref des prefix. Relevant datasheets might be:
- http://www.coilcraft-cps.com/pdf/ST526PND.pdf
- https://www.vishay.com/docs/34357/ihcl-4040dz-5a10.pdf

Image:
image

------------


poeschlr:
These look good.
Could you add *: in front of L_* in the footprint filters? See https://github.com/KiCad/kicad-symbols/issues/331

Otherwise these look good.

evanshultz:
What a pain! This is one reason why a connection between the client and GH libs is important. Maybe not always active, but an "Update Library from GitHub" button. There are many uses...

Anyway, I did just that change to all inductor symbols. Thanks for taking a look!

poeschlr:
Thanks

PR #338, opened 2018-03-01
Atmel MCU QFP footprints/filters

fauxpark:
Part 4 of 5? fixing footprint stuff for Atmel MCUs (#337). Here are the QFP packages. Next and last will be the QFNs.

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #340, opened 2018-03-02
Fix Atmel MCU QFN footprints/filters

fauxpark:
Finally, the Atmel QFNs (#338). I would appreciate an extra pair of eyes here as I have some confusion about a few things.

- Not sure how the EP sizes work, looks like they all completely fill up the interior of the footprint. Is this intended or should there be variants for the smaller sizes that the Atmel datasheets spec?
- Should the EP size be in the footprint filter?
- There is a missing 20-pin, 3x3mm, P0.45mm footprint ([example, p.212](http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-8235-8-bit-avr-microcontroller-attiny20_datasheet.pdf#page=212)). I've blanked out the footprint field for all of the symbols using it. `UQFN-20-1EP_3x3mm_P0.4mm_EP1.85x1.85mm` might work, but I dunno how much difference that 0.05mm would make. Plus the EP size also doesn't match.

------------


evanshultz:
@fauxpark 
- We need drawings of the recommended footprint from Atmel. (See below.) The EP size in the footprint will be the copper pad size, not the size of the exposed pad on the bottom of the package.
- Yes. Absolutely. If this is too wrong we will end up with poor thermal performance or soldering bridging, depending on the extent of the mismatch.
- Can you please add the correct FP anyway? Perhaps you could make the exact one needed? Unfortunately, I do not see this package at https://www.microchip.com/quality/packaging-specifications. Would you be willing to check with Atmel to see if we can get that info? But even without that, we can still enter the footprint info so the symbol will be ready when the footprint is submitted. Are any other packages missing the footprint info?

I'll take a look at the PR now.

evanshultz:
[ATMEGA128-16MU](http://ww1.microchip.com/downloads/en/DeviceDoc/doc2467.pdf) uses a 9x9mm QFN with a 5.4mm square EP. The datasheet doesn't show the footprint. http://ww1.microchip.com/downloads/en/DeviceDoc/64L_QFN_9x9_with_5_4x5_4_EP_MR_C04-0154A.pdf has all the same basic dimensions, but the drawing looks different and I don't see any designation/name in common (datasheet says 64M1 while the drawing is C04-154).

The footprint you picked has a 7.35mm square EP while the drawing I found is 5.5mm max. However, maybe this drawing doesn't go with this package?? Do you know?

I've asked how we can find the recommended footprint for an Atmel part that doesn't show it in the datasheet. Let's hold off on merging this PR until we get more info.

fauxpark:
Over the course of these PRs I've compiled a partial list of Atmel drawing/package numbers by looking at the Ordering Information section of each datasheet and matching them up with a footprint after confirming the drawing, eg. `44A` -> `TQFP-44_10x10mm_P0.8mm`. There are some drawings that map to the same footprint. But I can't see these package names anywhere in that link - I think they are Microchip-specific. Looking up Atmel package specs just leads me to that page too. Did they ever have any equivalent?

As for the mega128, I just paste the device name (minus the suffix) into Google and unless it's a really old chip it usually gives me a Microchip page with links to the complete and summary datasheets. I can see the QFN drawing in both.

poeschlr:
Just a heads up: you don't always need a landing pattern from the manufacturer to design a footprint. Having the package dimensions (Which is on page 369 of the linked datasheet) plus the IPC specifications should be enough.

fauxpark:
Do you mean that the current footprints are wrong and the EP size for eg. `QFN-64-1EP_9x9mm_P0.5mm_EP7.35x7.35mm` should actually be 5.4x5.4mm? Are there any parts which actually have such a large thermal pad?

A quick grep through the repo for "QFN-64-1EP" returns only three other symbols, all with different sizes, but none the exact size.
[DSPIC33FJ64GP306A](http://ww1.microchip.com/downloads/en/DeviceDoc/70593d.pdf), also specifies a 5.4x5.4mm thermal pad.
[TUSB8041](http://www.ti.com/lit/ds/symlink/tusb8041.pdf), appears to be 6x6mm.
[ATSAML21J16B](http://ww1.microchip.com/downloads/en/DeviceDoc/60001477A.pdf), with 4.7x4.7mm.

It seems like whoever added these footprints just made them as big as possible to maximise compatibility with same-dimension parts.

fauxpark:
The underside of the 3D model, for reference:
3d

poeschlr:
A lot of symbols have been added in the past without a fitting footprint so i would not be surprised if some atmel devices do not have a fitting footprint.

I would not restrict the search for a fitting footprint to what we have right now.
If there is no fitting footprint for a component you should still connect it to the correct footprint name. Maybe add an issue where you document missing footprints. (That way the user will be notified of the missing footprint instead of having a possibly wrong one added to the pcb.)

---

In this case there would be two options.
- Option 1 is the largest possible copper pad with an reduced mask cutout. (Name would be `QFN-64-1EP_9x9mm_P0.5mm_EP7.35x7.35mm_Mask5.4x5.4mm`)
- Option 2 would be to use a copper pad just big enough for this component(Name would be `QFN-64-1EP_9x9mm_P0.5mm_EP5.4x5.4mm`)

As these footprints are intended for a MCU, i don't think one would require the maximum copper pad size. So i personally would go for option 2 as it gives the user more space for the fanout.

fauxpark:
Okay, I've revised the EP sizes and Travis tells me these are the missing footprints:

```
Package_DFN_QFN:DFN-10-1EP_3x3mm_P0.5mm_EP1.64x2.6mm
Package_DFN_QFN:QFN-20-1EP_3x3mm_P0.45mm_EP1.55x1.55mm
Package_DFN_QFN:QFN-20-1EP_4x4mm_P0.5mm_EP2.6x2.6mm
Package_DFN_QFN:QFN-28-1EP_4x4mm_P0.45mm_EP2.4x2.4mm
Package_DFN_QFN:QFN-32-1EP_5x5mm_P0.5mm_EP3.1x3.1mm
Package_DFN_QFN:QFN-32-1EP_4x4mm_P0.4mm_EP2.8x2.8mm
Package_DFN_QFN:QFN-32-1EP_7x7mm_P0.65mm_EP4.5x4.5mm
Package_DFN_QFN:QFN-44-1EP_7x7mm_P0.5mm_EP5.2x5.2mm
Package_DFN_QFN:QFN-64-1EP_9x9mm_P0.5mm_EP5.4x5.4mm
Package_DFN_QFN:QFN-64-1EP_9x9mm_P0.5mm_EP7.65x7.65mm
```

evanshultz:
The copper pad size is typically larger than the exposed pad on the package. It seems like you've made the list above using exposed pad size of the package. Our footprints show the copper pad size, so the sizes above are too small.

I'm concerned because we don't know what footprints Microchip would recommend, so we don't know what footprints we're missing. Sure, we could use IPC info but if Microchip tells us that would be best. I think KiCad users would be most secure if the official libraries use the recommended footprints.

Reminder to please add the UQFN-20 footprint back. I don't see it. Even if a footprint is missing now, we can assign it to the sysmbol and generate them as a second step. Or hold off merging this until footprints are made. Either way, we can get the symbols correct now.

fauxpark:
Yeah, I didn't realise the package drawings != land patterns...

This post over on AVR Freaks claims Atmel has never provided footprints and instead point to JEDEC and IPC: http://www.avrfreaks.net/forum/atmels-footprint-free-policy

> Reminder to please add the UQFN-20 footprint back

Not sure what you mean by this - I didn't remove any UQFN footprint?

evanshultz:
You mentioned in the first post you removed the UQFN footprints. Please add the correct default one, even if the footprint doesn't yet exist.

fauxpark:
I said I blanked out the footprint fields for the symbols using the missing footprint (they were invalid anyway), and suggested the UQFN footprint as a candidate, but I don't think it fits. In my latest commit I filled out all the footprint fields and filters with what I *thought* were the correct values, and compiled the list of "missing" footprints.

I don't know if I can do much more on this without access to the IPC standards. Apparently there is a program that displays all the recommended footprints but I can't figure out how to get it. I guess it is also behind the paywall with the actual standard.

poeschlr:
I have found an older version of the ipc standard on some russian server. (http://pcbget.ru/Files/Standarts/IPC_7351.pdf)

One thing it states is that you need to use up all tolerances. So the starting shape is:
L_n + sqrt(E_p^2+ E_m^2 + L_t^2)

In words: Nominal lead size plus square of the sum of all tolerances (in this case the placement tolerance, manufacturing tolerance and the tolerance of the thermal pad of the chip.)

To that you then add the heal, toe and side stuff. For the exposed pad we have 4 sides so i would simply add the side value in both x and y direction. (These values are provided as tables for different lead shapes.)

L_n + sqrt(E_p^2+ E_m^2 + L_t^2) + 2*JS

For the placement tolerance i used 0.05 and for the manufacturing tolerance i used 0.1 in my script for generating smd resistor and similar footprints. (I did some research what might be typical here but i did not really document it.)

---

To finish the 5.4x5.4 example:
5.4+sqrt(0.2^2+0.05^2+0.1^2)+2*JS = 5.629 + 2JS

If we use the flat no lead table for the side requirement we get JS = -0.03
-> size should be about 5.57x5.57mm (I think 5.6x5.6 would be ok as well. Which incidentally is the maximum size of the pad.)

---

In conclusion: i think it would be ok if you use the maximum size given for the thermal tap of the package.

fauxpark:
Now the EP sizes are set to their max values and I've updated the above list of missing footprints.

evanshultz:
@poeschlr 
Can you codify this in KLC? This is one of the bigger questions I see, especially as we've added the EP size to footprint names, and if the vendor doesn't give us all details for the footprint I think we need an "in house" standard where we can point contributors.

poeschlr:
Will do but it might take me a while. (I am currently a bit short on time)

jcyrax:
@poeschlr I'm reading the linked document and I think there is error in your formula. You mention using nominal length

> In words: Nominal lead size plus square of the sum of all tolerances

but on page 20 formula is Xmax=Wmin + 2Js + sqrt(Cw^2+F^2+P^2)
And page 24 flat, no lead Js = -0.04 (table 3-13)

Should minimal size be used or I missed something?

poeschlr:
you are right.

CW would for all other landing pattern features again be a root mean square therm. (As the tolerance for most dimensions would be a combination of tolerances of different parts.)
I this case CW is simply ep_max - ep_min

So 

Xmax=Wmin + 2Js + sqrt(Cw^2+F^2+P^2) = 5.2+sqrt(0.4^2+0.05^2+0.1^2)+2*JS = 5.615 + 2JS

This will result in a ep size smaller by about 0.015 compared to what i previously calculated.
So we arrive at basically the same conclusion that the use of the maximum ep size might be a fitting estimate.

fauxpark:
@evanshultz are you waiting for me to create the footprints or can this be merged as is? My next move I think will be shortening the pin names and shrinking the boxes.

evanshultz:
Behold! http://ww1.microchip.com/downloads/en/AppNotes/Atmel-8826-SEEPROM-PCB-Mounting-Guidelines-Surface-Mount-Packages-ApplicationNote.pdf. It seems this doc covers all the packages used by these MCUs, but please take a second look in case my bleary eyes missed one. If you can choose the footprints from this doc, and then assign the footprints to these symbols, that's good with me. Maybe we're lucky and can use an existing footprint, but otherwise please update the list of needed footprints here so we can track it.

fauxpark:
Excellent! I'll get right on that.

fauxpark:
Done, and updated the list above again. Only the ATxmega8E5-M4U was missing from the table; I have left it as datasheet max.

evanshultz:
There are so many errors and symbols here that I'm inclined to just merge. A few spot checks look fine and you've done great work. Let me know if that's OK and I'll merge.

Could I please ask you to create a new issue in this repo for missing Atmel footprints? That way we can track it once this PR is merged.

fauxpark:
Do you mean in kicad-footprints? An issue about missing footprints I think would be more suited to that repo.

evanshultz:
@fauxpark 
So you're ready to merge? Are there any other Atmel libraries that need this work, or does this cover all footprint types?

fauxpark:
Yep, good to go :) This is the last of them.

evanshultz:
Thanks so much! Your contributions have been huge!

PR #341, opened 2018-03-02
ESDA6V1BC6 symbol added

Resaj:
[Datasheet](http://www.st.com/content/ccc/resource/technical/document/datasheet/21/07/21/e3/a8/df/42/a2/CD00001906.pdf/files/CD00001906.pdf/jcr:content/translations/en.CD00001906.pdf)
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/401)

imagen


Ratfink:
Thanks for the contribution!

One note: I'm not sure that the pin electrical types are right.  You have pins 1, 3, and 4 marked as input, but pin 6 marked as passive.  Surely they should all be the same type, but I'm not sure which type that would be.

Maintainers: what would be the preferred pin types for this sort of part?  Symbols already in the library seem mixed as to whether they're passive or input.

Resaj:
Yes, the type is wrong.
I'm going to specify all pins as passive, included pins 2 and 5. If later you want I changed them as input, only tell me and I'll change them.

jkriege2:
I only noted a few minor things (after the footprint-topic is resolved):
1. remove the word package from the description
2. The FPFilter should be `*:SOT?23*`: THe `*:` is required by newer KiCAD versions (it seems) and the pin-count is not required, because you use all pins and the correct package can be matched by the pin-count filter.
3. Please draw the diodes in the IEC-default unfilled style with a line going through, as done in device.lib (combine as required, the symbol in the datasheet looks to me like a US/IEEE-style back-to-back zener diode): 
2018-03-03 11_08_46-symbol library editor - device d__kicad_kicad-symbols_device lib
2018-03-03 11_08_21-symbol library editor - device d__kicad_kicad-symbols_device lib
2018-03-03 11_08_07-symbol library editor - device d__kicad_kicad-symbols_device lib

Otherwise this looks good!

Thanks,
JAN

Resaj:
Done! This is how it looks now:
imagen

jkriege2:
Nice work!

Thanks,
JAN

Issue #166, opened 2018-01-13
KLC complient symbols for HDMI connectors would be a nice addition.


poeschlr:
Discovered the problem while checking https://github.com/KiCad/kicad-footprints/pull/252.

Ratfink:
Created in #344, merged via #310.

PR #344, opened 2018-03-03
Add HDMI connectors

jcassette:
This a draft proposal for issue #166 

I used the pinout from https://en.wikipedia.org/wiki/HDMI

A few questions:
- Which style is preferred? "Graphical" or "functional"? I remember there was a discussion somewhere on this but I cannot find it.
- How to handle pins 14 and 19? In HDMI 1.0 - 1.3 they are respectively "Reserved" and "Hot plug detect", but in HDMI 1.4+ they are "HEAC+" and "HEAC-".
- Which pin types should be specified?

image image


Ratfink:
Thanks for working on this!

As for which style is preferred, [pins should be grouped by function](http://kicad-pcb.org/libraries/klc/S4.2/).  This goes for connectors as well as anything else.

I'd suggest going with something more like your second screenshot, with a drawing of the connector inside the body rectangle.  Take a look at the USB connector symbols for examples.

evanshultz:
@jcassette 
Thank you for contributing to KiCad!

Please make sure you sign he CLA above so your PR can be merged.

- Yes, functional is a better symbol design. @Ratfink's suggestion is good.
- You will need two symbols, one for HDMI <1.4 and one for HDMI 1.4+.
- Pin types:
-- Power and ground are always Power Input.
-- CEC is Bidirectional
-- The other data pins can be Passive so they can be a source or sink without generating an ERC error (like our USB symbols)

Thanks again!

Misca1234:
Well, my comments are some minor issues but 
Dont forget to make pin 18 and 17 be of the type "power input"
You could place pin 2,5,8 and 11 on the right (or left) side, 
normally top and bottom side is reserved for power input.

A suggestion could be 
group: 13, 14 and 19 on left side
group: 15, 16 on right side
group: 2,5,8 and 11 on right side

In this way you get your symbol a little smaller

calebns:
I'm just wondering...  HDMI +5V power pin can be either input or output, depending on if the connector is going to be used in host or device. What pin type should be chosen to guarantee ERC check to pass in both cases?

Ratfink:
@calebns It should be "power input", just like @Misca1234 said.  If users need it to act like a power output pin for ERC, they can add a PWR_FLAG symbol.

jkriege2:
For style ... we have graphic symbols and we have symbols like USB-connectors: 
2018-03-04 15_39_11-symbol library editor - connector_specialized d__kicad_kicad-symbols_connector_
2018-03-04 15_38_36-symbol library editor - connector_specialized d__kicad_kicad-symbols_connector_

Since this is not a general purpose connector, but used for one specific bus, I would vote for the box-style (like USB) with pin-names. 
Also: Please make the pins 100mil long, I think that's enough.

Thanks,
JAN

jcassette:
Thank you all for your suggestions. Here are the updated symbols.
I have grouped the pins by their respective bus (TMDS / CEC / I2C / HEAC).
For HDMI < 1.4 the pin 14 is Not Connected and hidden, for HDMI 1.4+ it is Passive and named HEAC+.
@Misca1234 Pins 2, 5, 8 and 11 are shield pins and I believe they should be at the bottom.

image

image

evanshultz:
@jcassette 
- Let's change the names a bit (based on our USB symbols which I think are relevant here):
-- HDMI_A
-- HDMI_14_A
- Can you shift down the graphics a bit and then shrink the symbols vertically?
- Set the line width of the outside box to 10mil.
- Set the footprint filter to `*:HDMI*`.
- The SDA pin is also bidirectional (sorry I missed that earlier).
- Is there any case where the user will not ground the shield for each diff pair? In other words, would grouping the shield and data pins be better in any situation rather than putting all shields at the bottom?
- Add a tilde (`~`) as the URL or use a link to the HDMI spec/standard.
- Please add the relevant HDMI versions to both symbols.
- This is really nitpicking, but the actual logo has the bevel on the "M" in "HDMI" on the trailing edge; yours is on the leading edge.

Since this is your first time, there are a few things I should mention:
1. Click on the "Details" link next to the Travis CI text above. This will bring you to the results of the auto-check scripts where you can find some errors/warnings. If you can correct these items, or justify a deviation, it will help speed up review and merging. Thanks!
2. These rules the check scripts use are governed by KLC, which you can find at http://kicad-pcb.org/libraries/klc/.

@poeschlr @jkriege2 @Shackmeister 
I hope this isn't opening a can of worms... should these symbols have a pin 20 to match with pin 20 on the footprints? This would allow connection, if the user desired, to avoid cutting up a plane. Similar to what we discussed when we made generic connector symbols with mounting holes, except HDMI jacks will almost certainly be at a board edge and because of the high-speed signals will likely want a stitched plane around the board edge. I think we do want a separate pin 20 for the user to connect to, but I want to loop you all in now instead of you getting surprised later.

poeschlr:
HDMI connectors normally come as shielded units. The "mounting pins" serve the purpose of connecting the shield. They must be connectable. (Typically the shield is connected to GND.)

About naming, these should get pin number 20 if datasheets call it that way. If the datasheet does not give a pin number for them, then they should get the pin number "SH" to fit with other shield pins.
If they need to always be connected to the other GND pin then it might make sense to stack that pin with the other GND pin. If not then it should get the pin name "Shield".

---

The discussion you reference was never about shield pins. For these it has always been clear that they need to be in the symbol. The discussion was if mounting pins without any electrical purpose should also be connectable via the symbol. 

Pins like the mounting pins for a molex picoblade connector. These pins are completely encased in plastic. They do not surround the connector and can therefore not be used for shielding in any way. There is no way the cable can connect to them.
In contrast to this, the shield of the cable is typically connected (via a large area) to the shield around the hdmi connector.

Ratfink:
@evanshultz I don't think a direct link to the HDMI spec is possible.  [The HDMI website](https://www.hdmi.org/manufacturer/specification.aspx) says that you have to register to download the HDMI 1.3a spec, and newer specs require you to become an adopter, which involves paying licensing fees.  Sheesh.

evanshultz:
@Ratfink 
Yes, I know. But there are suitable places that we could link. For example, some symbols in this lib use Wikipedia links (not saying this is good or appropriate but I'd approve it if the relevant info was there).

@jcassette 
Please add a pin named SH on the bottom. Here is the PR to update the footprints: https://github.com/KiCad/kicad-footprints/pull/410.

jcassette:
@evanshultz 

> Let's change the names a bit (based on our USB symbols which I think are relevant here):
> -- HDMI_A
> -- HDMI_14_A

As I understand it, HEAC was introduced with HDMI 1.4 but is optional, i.e. there may be HDMI 1.4b devices which use pin 19 as Hot Plug Detect and do not use pin 14. That's why I used "HEAC" and not "1.4". I suggest HDMI_A and HDMI_A_HEAC. What do you think?

> Can you shift down the graphics a bit and then shrink the symbols vertically?
> Set the line width of the outside box to 10mil.
> Set the footprint filter to *:HDMI*.
> The SDA pin is also bidirectional (sorry I missed that earlier).

Okay, I will do that sometime this week.

> Is there any case where the user will not ground the shield for each diff pair? In other words, would grouping the shield and data pins be better in any situation rather than putting all shields at the bottom?

I have no idea.

> Add a tilde (~) as the URL or use a link to the HDMI spec/standard.

Agreed with @Ratfink. We could use http://pinouts.ru/Video/hdmi_pinout.shtml

> Please add the relevant HDMI versions to both symbols.

I will leave it as it is unless you disagree with my comment on HEAC above.

> This is really nitpicking, but the actual logo has the bevel on the "M" in "HDMI" on the trailing edge; yours is on the leading edge.

I did my best on the logo but I did not draw a bevel. I think it's just a glitch in the screenshot.

> Please add a pin named SH on the bottom.

Okay, as pin number 20, right?

evanshultz:
IMO, having the version number is easier for the using building a schematic. Can you use the names I suggested above but on the 1.4 version name pin 14 `RSVD/HEAC+` and pin 19 `HPD/HEAC-`? And the 1.0-1.3 version symbol pin 19 to `HPD` since everything else is abbreviations? That also allows putting the version numbers (with relevant info) in the description. What do you think?

The Russian website works. It only covers the pinout and not other related info about HDMI, like Wikipedia, but I'll accept it.

Yes, the pin name and number are both SH. Yes, SH is not a number. :)

jcassette:
I have made the changes and added type B, C, D connectors

evanshultz:
Looking good! Thanks.
- The Russian website does not have information for Type C or Type D. Can you find a link for those symbols that gives information for them?
- And it's wrong anyway! For Type A, HEAC+ is pin 14 and HEAC- is pin 19. It shows them reversed.
- If we assume all diff pair shield pins should be grounded, which is what I see searching online, they can all the of type Power Input.
- Instead of RSVD, can you please call the pin UTILITY?
- And the UTILITY pin should be visible and of Passive type, I suppose, except Type B. On only Type B are pins named RESERVED and specifically shown as Not Connected on the the device. So for Type B name those two pins NC and keep them invisible. Does that sound OK to you?
- For Type B, please have one pin space between CEC and SCL like on the other symbols.
- Confirm pins of 13 vs 17 on Type C. Are they swapped?
- It appears HEAC as also an option for Type C but the pins don't show that.
- Would you be interested in adding a Type E symbol?

I found HDMI specs online up to version 1.4. I won't link to them here, but if you can also find them it should explain more about my comments above.

evanshultz:
@jcassette 
Are you able to finish off this PR?

jcassette:
Sorry for the delay, I hope to have some time for this during the weekend.

jcassette:
> The Russian website does not have information for Type C or Type D. Can you find a link for those symbols that gives information for them?
> And it's wrong anyway! For Type A, HEAC+ is pin 14 and HEAC- is pin 19. It shows them reversed.

You are right, http://pinouts.ru/Video/hdmi_pinout.shtml is incorrect. I will link to Wikipedia instead.
For type C I have http://pinoutguide.com/PortableDevices/mini_hdmi_pinout.shtml
For type D I had http://pinoutguide.com/PortableDevices/micro_hdmi_type_d_pinout.shtml

> I found HDMI specs online up to version 1.4. I won't link to them here, but if you can also find them it should explain more about my comments above.

I have found it. It seems to be the only reliable source of information regarding the pinout. It would be good if we could link the symbols to it.

> If we assume all diff pair shield pins should be grounded, which is what I see searching online, they can all the of type Power Input.

Okay

> Instead of RSVD, can you please call the pin UTILITY?
> And the UTILITY pin should be visible and of Passive type, I suppose, except Type B.
> On only Type B are pins named RESERVED and specifically shown as Not Connected on the the device. So for Type B name those two pins NC and keep them invisible. Does that sound OK to you?
> For Type B, please have one pin space between CEC and SCL like on the other symbols.

Okay, done it

> Confirm pins of 13 vs 17 on Type C. Are they swapped?

Sorry the type C symbol was incorrect.

> It appears HEAC as also an option for Type C but the pins don't show that.

Type C was introduced with HDMI 1.3 so I am making a symbol for 1.3 without HEAC and one for 1.4 with HEAC.

> Would you be interested in adding a Type E symbol?

Done

evanshultz:
Thanks! Would you mind posting new screenshots of new and changed symbols? I'll take a closer look tomorrow.

One thing I notice can be improved: the footprints filters. They're all `*:HDMI*` right now. But due to our footprint naming convention, they could be `*:HDMI*A*` for type A, `*:HDMI*B*` for type B, etc. Could you change that?

BTW, all Travis CI errors present now are OK.

evanshultz:
I looked at all the symbols and they look good to me with two exceptions. Thanks so much!
1. The Type B connector is proportionally taller than the others. Can you shrink it vertically just a bit? The NC pins won't actually be visible so they can be placed elsewhere (like between other pins).
2. Is http://pinoutguide.com/PortableDevices/mini_hdmi_pinout.shtml incorrect about the Type C pinout? Please confirm. It looks some pins don't match the official spec.

And there is one naming change that I will propose: To me it seems that the HDMI type is more important than the version, so I think the type should come first. That would mean changing `HDMI_1.4_D` to `HDMI_D_1.4`, for example. Is that OK for you?

If you can do that let me take one final look. This is a great improvement and I really appreciate you taking the time to add all these missing symbols!

jcassette:
I am trying to do this but I do not have much time and KiCad keeps modifying all the library whenever I save it...

evanshultz:
KiCad will reorder pins and even symbols when you save. Different versions have different behaviors. The pins do not need to be in order to merge. Let me take a look...

evanshultz:
That link looks OK to me now, also.

All but type B could be shrunk just a little bit so the relationship to the top pin on the left side and the upper edge is closer, but that's so minor. If you want to do that, it would be fine. But I think these symbols are good now as they are. Thanks so much!

I'm going to wait on merging to see what happens with https://github.com/KiCad/kicad-symbols/pull/310, but either way your work will be in KiCad soon! It may be easiest to copy your symbols into another PR to get everything to merge easiest, in which case someone will close this PR.

Ratfink:
Merged via #310.

PR #349, opened 2018-03-04
ADLxx/SPFxxx RF gain blocks and RF_Aplifier library

calebns:
http://www.analog.com/media/en/technical-documentation/data-sheets/ADL5541.pdf
http://www.analog.com/media/en/technical-documentation/data-sheets/ADL5542.pdf
http://www.analog.com/media/en/technical-documentation/data-sheets/ADL5610.pdf
www.qorvo.com/products/d/da001910

Qorvo does not provide package designation, but the chip seems to be identical to SOT-89.

image

image

Footprint PR:
https://github.com/KiCad/kicad-footprints/pull/405

------------


evanshultz:
- Could you base these all off the standard opamp triangle symbol? That will leave room for control pins on the top and bottom and look more standardized. This will also help you place the pins all on grid since they aren't now. The input and output pins will not need pin names.
- Add the package name to the end of the description for SPF5189Z, since you've chosen it.
- Update the package for ADL554x to the ADI-specific one at https://github.com/KiCad/kicad-footprints/pull/405. (Even if we haven't finished the footprint we know the name.)
- Use all lowercase letters for the keywords.

@poeschlr @jkriege2 
This is adding a new symbol library. Any problems or suggestions? Since there's only a couple symbols do we dump this into RF.lib (at least for now)?

If so, can we turn on Travis before any symbols are merged? Maybe create a blank repo and then close/reopen this PR?

evanshultz:
@calebns 
It seems the check scripts are working here now (maybe I just needed to refresh...). Please fix pin lengths and locations. Thanks!

calebns:
I vote for having a separate RF_Amplifier library. As for now there will be only 4 sybols, but I believe that in the near future I'll add some more. Also, this is how Digikey/Mouser categorizes this kind of chips - I just find it easier to look for them in RF_Amp rather than generic RF bunch-of-stuff.

image
image

evanshultz:
- The FPfilter for ADL5542 should have a "*" between "Analog" and "LFCSP".
- Can you position the text for ASDl5542 closer to the symbol?

This is adding a new symbol library. Any problems or suggestions? Since there's only a couple symbols do we dump this into RF.lib (at least for now)?

poeschlr:
I don't see a problem with that but i would have a request. If there are symbols in another lib that would fit better in the new one they should also be moved.
So if there are there should be at least an issue opened to alert us about that. (The actual moving could be done at a later time.)

Ratfink:
If we don't want tiny RF_*category* libraries, there are quite a few already in the library that should be dumped back into RF.lib too.  The solution that's most consistent and the least work would be to allow the new library.

I really don't like this as an argument in favor of it, but because symbol libraries are kept together in two files per library, the more we chop up the symbol libraries for now, the fewer merge conflicts there could be.  So pending the .sweet format, tiny libraries could ease our maintenance burden.  But that's favoring technical advantages over purity, so again I really don't like this argument, but it might be worth taking into consideration.

evanshultz:
@poeschlr @Ratfink 
Agreed. Yes, assuming the library will grow I'm in favor of the new library too. I've added onto the issue at https://github.com/KiCad/kicad-symbols/issues/352 to track the purging of parts from RF.lib.

@calebns 
Just to be clear, you won't have to do anything here. Your PR can be merged in the current library.

Misca1234:
As a principle, I prefer many small parts, and the only arguments I have for that at the moment, beside Ratfink and poeschlr ones is that I compare it with software.
In software it is always better from an overview and maintenance point of view to have many modules than a monolithic.

calebns:
@evanshultz , fixed
image

calebns:
By the way, CI now complains that the SPF5189 is not centered. Most likely it's because it takes min/max of X,Y coordinates of pins and calculates the center as an average of their values. In this case the symbol is not symmetrical (there is no pin going upwards) so travis expects to have it moved by +150mils in Y direction.
However, if I move it, the pin ends will not be aligned to 100mils. What to do? Should I try to tweak the symbol (and most likely enlarge it) to keep the CI happy or can we ignore this error?

image

Ratfink:
That looks like a perfectly centered symbol.  The warning about the symbol's center is something we often have to ignore.  It's a fairly vague KLC rule that's hard to express in code, so the code for checking it often says things that are unhelpful. 🙂

evanshultz:
@calebns 
@Ratfink said it all above. No need to work. Thanks for adding these symbols and I will take a look at the footprint now.

PR #350, opened 2018-03-04
Modified the footprint filter and set the specific footprint

Misca1234:
This push fixes the footprint warnings in the  Sensor_Touch.lib

CY8CMBR3002, CY8CMBR3102 CY8CMBR3106S, CY8CMBR3108, CY8CMBR3110, CY8CMBR3116
Ignored the VCC pin warnings.
Fixed the footprint filter and it specific footprint


evanshultz:
Two things:
- These pins have the wrong pin type for VCC. Would you mind fixing that too while you're at it?
- Did you check the datasheets to verify that the packages you picked are correct?

Misca1234:
I changed the VCC to power output

bild

bild

bild

bild

bild

bild

evanshultz:
VCC pins should be Power _Input_. Not Power Output.

The QFN24 EP size is 2.6mm square, but the slug on the part is 2.65mm square. We do not have a QFN24 with a bigger EP but otherwise the same. Please make an issue that we need a better footprint for this part and I can merge. (The QFN16 footprint is OK for this package.)

Misca1234:
> VCC pins should be Power Input. Not Power Output.
Well, it is kinda output, it is the chip's internal voltage regulator output but I have changed all of them to power_input

evanshultz:
They must be Power Input for the ERC tool to work. It's based on nets with power on them, not the actual source of the power (that's controlled by the power flag). Yeah, it seems backwards. And in this case, since current can't be drawn from VCC (as far as I can see), they could even be Passive.

I reviewed this thoroughly since I have more time now:
- On CY8CMBR3110, the last function of pin 14 has only one "Z".
- On 3002 and 3102, why not put CMOD on the right side for consistency?
- And for 3002 all the GPO could be on the left for the same reason.
- Only 3102 has the GPO pins separated instead of next to each other.
- Perhaps all CS* pins should be Passive? I think the ERC tool will be happier than if the pin is Input, Output, or Birdirectional if the pin is used as a touch sensor.

You didn't include the datasheet in the body of the PR. It is at http://www.cypress.com/file/46236/download.

Can you please make an issue that there is a missing footprint, or else contribute the footprint and use it in the relevant symbols here?

poeschlr:
@evanshultz wrote:
>They must be Power Input for the ERC tool to work. It's based on nets with power on them, not the actual source of the power (that's controlled by the power flag

The conclusion you reach is right in this case. (See details below.)

But i disagree with the quoted statement. A pin is power input iff it is a consumer of power. It is power output iff it is the source for power. (outputs of a dc/dc converter, of a ldo, ...)
The power flag is used to tell ERC that a pin that is not a power output is the source of power. This is needed if the supply is via a generic connector, or if a passive component is in series to the power supply (a fuse, the inductor of a filter, ...)

Just to make it clear: The direction the current flows alone is not what determines if it is a power input. You also need to take into account the potential of the pin compared to the potential of its counterpart. 
Now letslook at the pin with the more positive potential. It is power out iff the current flows out and power in otherwise.
For the more negative pin the current direction is exactly opposite. (The current flows out of the positive pin into the negative pin -> both are power output.)

---

In this case it is a bit confusing. 
Making it power output would negate ERC. The reason is that the datasheet states the pin should be connected to VDD if the voltage of the VDD pin is within a specific range. (ERC would then think the component is powered even though you still need to get power from a different source.)

So in this case i would really set it to passive. (power input will be problematic if the voltage is not within the range and the pin is only connected to a capacitor.)

Misca1234:
@evanshultz 
> You didn't include the datasheet in the body of the PR. It is at http://www.cypress.com/file/46236/download.

I did not create them in the first place so I did not give the layout much time, I was only hunting down libraries with errors, but I fix those points you pointed out

@poeschlr 
I will make them passive

Misca1234:
The following have happend

Changed the data pin to the short name, for example pin "CS1/PS1/GPO0/SH" on CY8CMBR3102 became CS1
put the datapin together
Made VCC passive
Made SCL bidirectional, clock stretching
Move VCC to the top
Placed all data pins on left side

bild

bild

bild

bild

bild

bild

evanshultz:
This mostly looks good to me. Thanks for making further changes. But still two questions:
- Should all CS* pins be passive, in the case they're used as touch inputs? In that case they aren't sending or receiving a logic signal and the pin will be hooked to only passive components.
- I don't interpret I2C clock stretching to mean the SCL bus is bidirectional. Only the master generates a clock so I think it should still be an Input on these symbols. The slave devices will hold down the clock, but they're not putting anything back on the bus so I don't think these pins should be Bidirectional. This was brought at up https://github.com/KiCad/kicad-symbols/pull/233 and after I made the same statement I didn't see any objections. Have I misunderstood the behavior of I2C clock stretching?

Misca1234:
I will change the CS pins to passive


> I don't interpret I2C clock stretching to mean the SCL bus is bidirectional"

But is not the pin type, such as input, output, bidirectional or open collector an indication of how a pin can (or can not) manipulate the line

Even if the pin type is some sort of "information direction indicator", then when the slave is holding down the pin it is "sending" information to the master that i2C communication should be put on hold.

Misca1234:
CY8CMBR3102
Pin 6 CS1 is now passive

CY8CMBR3106S
Pin 20 CS4 is now passive

CY8CMBR3108
Pin  9 CS5  is now passive
Pin 10 CS6  is now passive
Pin 11 CS7  is now passive

CY8CMBR3110
Pin  9 CS5  is now passive
Pin 10 CS6  is now passive
Pin 11 CS7  is now passive
Pin 12 CS8  is now passive
Pin 14 CS9  is now passive

CY8CMBR3116
Pin 16 CS8   is now passive
Pin 15 CS9   is now passive
Pin 14 CS10  is now passive
Pin 13 CS11 is now passive
Pin 12 CS12 is now passive
Pin 11 CS13 is now passive
Pin 10 CS14 is now passive
Pin  9 CS15 is now passive

evanshultz:
Thanks!

Hmm.. I don't see it that way. By holding down the bus, the slave device is turning on a transistor (Rds(on) probably <100ohm) between SCL and ground. The amplitude of the clock signal then drops dramatically during the logic high state, as it's pulled up by a passive resistor (I've used 2.2k-3.3k) and the slave device's transistor acts as a voltage divider. The master then sees that the logic high threshold is not reached by the next expected edge transition and understands the slave device is asking to go slower.

I do see your point that the master is getting some type of information from the slave device, but to me the clock signal is still only being generated from the master device as it takes that information and then adjusts the clock speed (assuming the slave device releases the transistor).

My main point is that while the slave can inhibit the clock, only the master generates/controls the clock.

If this is where we stand, let's get more input. @jkriege2 and @Shackmeister contributed to the original conversation about I2C SCL pin types. Any further input from you or anyone else?

So what happens with ERC? It doesn't generate any error or warning, by default, if a B pin is connected to I or O (which intuitively sounds correct). That the only user perspective I can think of we should check.

Misca1234:
What does the pin type mean for you then?

evanshultz:
@poeschlr @jkriege2 @Ratfink @Shackmeister 
Any thoughts on SCL pin type? It's been brought up a couple times and it would be nice to make a collective decision to eliminate this open issue.

Ratfink:
I see the question as "is open-drain a kind of output?", which I would certainly answer in the affirmative.  So I'd vote for bidirectional.

poeschlr:
There was a PR by @jkriege2 that fixed other I2C devices. https://github.com/KiCad/kicad-symbols/pull/233

I would go with the  same solution that has been used there. (At least one of the devices had pin-type Bidirectional for the SCL pin.)

evanshultz:
Yes, one is Bidirectional. One is Open Collector. And 5 are Input. I didn't check untouched SCL pins in that PR, just changes, and I didn't notice all SCL pins in that library. So, this still seems unresolved...

Misca1234:
Because the I2C bus is pulled up externally I guess both SCL and SDA pins are open collector on both the "master" and "slave" sides.

It really comes down to how people look at the term "pin type" means.
Is it pure electrical construction or does it indicate in which direction "information" flows.
Kicad mixes both terms in the "pin typ" drop down box.

If we say that the pin should reflect electrical construction then both SCL and SDA should be open collector.

If we say the pin should reflect some sort of information direction then both SCL and SDA should be bidirectional (because the master looks at the SCL and gain information from the slave via clock stretching (if you for example bitbanging I2C it is more than obvious that the SCL pin in a master is treated as an "input" from the slave and you consider that the slave is "sending" information)

And of course, the information direction approached is messed up for chips where you a have a open collector LED driver stage.

Or we just define for I2C that (mix)
SDA is bidirectional and SCL is open collector for a slave and 
SDA is bidirectional and SCL is output for a master

or we just define for I2C that (information approach)
SDA is bidirectional and SCL is bidirectional  for a slave and 
SDA is bidirectional and SCL is bidirectional for a master

or we just define for I2C that (electrical approach) 
SDA is bidirectional and SCL is open collector for a slave and 
SDA is bidirectional and SCL is open collector for a master

evanshultz:
Let me try another approach to this discussion...

What is the reason for pin types to exist? It is for ERC, so the user can catch errors and avoid designing in problems. A net should not have two outputs, for example. A net should also not have only Inputs. Nonetheless, even with proper pin types and no ERC errors there could still be problems. So how can the pin types be set up to help the KiCad user as much as possible?

SDA should not present any problems because it is an IO pin for any device. You've made it all Bidi above and that's fine. No need to spend any more time talking about SDA.

For SCL, your 2nd and 3rd approaches prevent the ERC engine from helping because open collector and bidi pins do not generate any ERCs when connected together in any fashion. The first approach doesn't work either, but that's because an open collector and an output pin generate a warning. This would be false because the SCL pins should be tied together.

There is only one combination that makes sense (to me, at least), and you didn't give it above: when **SCL is an Output on the master and an Input on the slave**. That will generate an error if there are multiple Output pins on the same net. It will also generate an error if there are only Input pins on the net. If the I2C master has configurable pins, like an MCU, then it's up to the user to configure them but because there is a Bidirectional device no ERC errors will be created.

Is there any other reasonable way to look at this?

Ratfink:
Evan, I understand your point there, that it would be good to give an error if there's an I<sup>2</sup>C bus with a non-one number of masters, even if it means fudging pin types a bit.  Practicality beats purity, and all that.  Okay, your suggestion gets my vote.

evanshultz:
> even if it means fudging pin types a bit

Depends on perspective. :)

Misca1234:
If the pin type is only for the ERC then it is ok.
Which means that the pin type assignments really does not matter as long as ERC is happy.

evanshultz:
@Misca1234 
Wrong. It does matter. There are many choices of pin types that will never result in a ERC error. That would be unhelpful and make the ERC engine useless. There are a far smaller number of choices, and possibly only one choice, which fit the pin function and also result in ERC errors being created with improper wiring. For the I2C bus, I believe it should be clear why I'm advocating for the best choice.

Unless someone else has additional information, can we agree how to configure I2C pins and close this topic? Updating KLC to cover common busses (I2C, SPI, I2S, CAN, etc.) would be nice.

You mentioned above that CS* pins are now all passive, so if the I2C pin types are correct I think this can be merged.

Misca1234:
This discussions goes around in circles.
I asked you what your view of what the pin type is for and you answered it was for ERC.
Now you imply that is not the case, by saying there are pin types that never will result in ERC errors.

Which means that there are pin types that are excluded from ERC, which means that the pin type is not for ERC but a mixup of some for ERC, some for convenient and some for the beauty of the eye. 

If pin type is a mixup of some stuff "nice to have", some ERC etc. then there is impossibly to have a discussion of what is right or wrong or even make a reasonable decision but all comes down to personal opinion, apply to authority, secret information or other useless conventions.

evanshultz:
Here are the default pin-to-pin ERC settings:
image

Pin types are for ERC and I think that explains it without any words.

evanshultz:
@Misca1234 
If you change the SCL pins to Input type I'll merge.

Misca1234:
I don't have the branch any more, I have to recreate it

PR #351, opened 2018-03-05
FPGA_Xilinx: Removed slashes from symbol names

orsonmmz:


orsonmmz:
I see there are failing checks, but I cannot really fix all the components in the library now. The patch is about making the symbols compatible with KiCad, as '/' and ':' characters are not allowed in symbol names. I suggest reviewing the raw patch for a clearer view.

Ratfink:
Thanks for the contribution!

I see that you're trying to fix just one specific problem with the symbols, not every KLC problem they have.  While it would be great if you could do more, I don't see a problem with a PR that's just a small step in the right direction. 🙂

One request though: there is an existing entry in FPGA_Xilinx.dcm for the XC3S200AN/FT256 which isn't updated by your patch.  If you change the name there to match the new name of the symbol, I think this would be ready to merge.

orsonmmz:
Ratfink, thank you for verification. I applied the requested change to FPGA_Xilinx.dcm.

Ratfink:
Thanks for the good work!

PR #355, opened 2018-03-07
Fix Atmel MCU DCMs

fauxpark:
Some of the datasheets were 404'd or otherwise not quite right (pointing to summary, wrong device, etc).

Also filled out some missing entries. I could not find any information on the "ATmega1261" though - it is defined as an alias of the 1281 but I don't think it actually exists. Perhaps a typo that was missed when someone actually added the 1281?

------------

### DIFF
--- a/MCU_Atmel_ATMEGA.dcm
+++ b/MCU_Atmel_ATMEGA.dcm
@@ -69,13 +69,13 @@ $ENDCMP
 $CMP ATMEGA128A-AU
 D TQFP64, 128k Flash, 4k SRAM, 4k EEPROM
 K AVR 8bit Microcontroller MegaAVR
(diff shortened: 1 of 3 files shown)


Ratfink:
Thanks for the continued work on these libraries!

Overall these changes look good.  I see you used datasheets from DigiKey for a few of the parts which appear to be obsolete and no longer have information available from Microchip directly.  I suggest moving those parts to the `obsolete/` directory in the future.  Let me know if you want to do that; otherwise I'll create an issue so it isn't forgotten.  For now though, I'll merge this PR.

fauxpark:
Will do. I'm still waiting on #340 to be resolved, then I've got some pin reordering (by KiCad) and name shortening (by me) to go.

PR #359, opened 2018-03-07
Stack pins of FUSB302B, add FUSB307B

Ratfink:
Associated footprint PR can be found here: https://github.com/KiCad/kicad-footprints/pull/416

Updated the FUSB302B USB PD PHY symbol with stacked pins and a smaller body.  Updated the documentation for ON Semi's acquisition of Fairchild.  I left the VCONN pins unstacked because it's perfectly valid in some cases to leave them NC, and adding an NC mark to stacked pins makes KiCad throw an ERC error.
Datasheet: http://www.onsemi.com/pub/Collateral/FUSB302B-D.PDF
screenshot from 2018-03-07 16-08-36

Added a symbol for the new FUSB307B USB TCPM.  
Datasheet: http://www.onsemi.com/pub/Collateral/FUSB307B-D.PDF
screenshot from 2018-03-07 16-08-49

------------


evanshultz:
- The *_N pins should be open collector. See below.
- S4.2 is saying input on left, output on right. So I'm asking for these symbols which side (USB or I2C) we should define as input and which as output so that it works better for users that are working with this symbol. Both are bidirectional, but the major function of the part seems to be a USB to I2C bridge. However, I'd merge this either way.

All else is fine and I'll wait for an update.

Pin types. They should be selected according to what the pin is (or can do), and not to make ERC quiet. We want ERC to catch actual errors. Anyway, this won't cover everything but here's a quick discourse:
-Output-type pins:
--Output: A pin that is driven both high and low inside by the IC. The low state is provided by a pulldown transistor, while the high state can be by a second transistor (in a totem pole) or an internal pullup resistor. These states are undefined and depend on the exact part or part family, but by far are most often ground (0V) as the low state and some positive voltage for the high state. Between high and low there is some voltage which is disallowed and where the output should not linger because an input of that output signal cannot be guaranteed to deliver a logic high or a low.
--Tri-state: This is like an output pin, that is actively driven both high and low, but it also can be disconnected internally (high-impedance, floating) so the pin is not in any logic level. This is used when the output pins of the IC may be connected to something else which will control the net (for example, two masters which operate mutually exclusively, when the pin will startup and/or shutdown in a high-impedance state and external circuitry will pull the net high or low, or many other reasons). The disconnected state is the third state of the, hence the "tri" in tri-state.
--Open Collector (or open drain, for a MOS device): A internal pulldown transistor, like above, is present. But no internal mechanism pulls the pin high. This is done external to the IC. This pin type allows multiple ICs to be connected together, with any of them asserting the net low. A logical OR.
--Open Emitter (or open source, for a MOS device): Like the above but flipped around. A internal pullup transistor, is present. But no internal mechanism pulls the pin low. This is done external to the IC.

-Input-type pins:
--Input: The pin is high-impedance and accepts a signal from a low-impedance source.

-Bidirectional-type pins:
--Bidirectional: The pin can be placed into two modes, as an input or as an output. There will be circuitry for an input stage and additional circuitry for the output stage, but only one of the two is active at either time (some pins can use both so they can output a signal and monitor it at the same time). The exact type of output, from the list above, is unspecified when the pin is configured as an output.

-Passive-type pins:
--Passive: The pin is not intended to connect to another device which has any input or output requirements, but only connect to a passive component (resistor, capacitor, inductor, etc.). Passive type will not generate ERC errors when connected to the pin types above because there is no known information about the direction and type of signal on a passive pin. An IC may have a mode where it should connect to passive circuit and also logic circuitry (like a combo touch sensor and GPIO pin of an MCU), and in that case the decision needs to be made about what pin type is best when the symbol is implemented and checked in a schematic.

By the way, all of this circuit inside an IC was first developed externally before ICs were invented. Still today, many of these input and output circuits are built with individual parts when interfacing between an IC and discrete circuitry. There is a lot more information online about this, but I tried to write briefly above and from the perspective of reading a datasheet and selecting the correct pin type in KiCad. If you have more questions, read up a bit online and then ask. Seraching around real quick for a couple links that may help, see https://en.wikipedia.org/wiki/Logic_level, https://en.wikipedia.org/wiki/Open_collector, https://www.electronics-tutorials.ws/logic/pull-up-resistor.html, and http://www.ti.com/lit/an/sdya009c/sdya009c.pdf.

Ratfink:
Thanks for the explanation of KiCad's pin types!  In particular, it's really helpful to know that Tri-state is meant as Tri-state *Output* (maybe KiCad's interface should call it that to avoid confusion?).

With regard to S4.2, it also says that logic-level signals go on the left, and driver pins go on the right.  USB Power Delivery's BMC signaling uses funny non-logic levels and the chips have special drivers to work with those levels, which is why I put the CC pins on the right and the I<sup>2</sup>C pins on the left.

evanshultz:
Inputs must be only high-impedance, and there aren't other states for input pins, so tri-state being an output is implied. But yeah, it could just as easily say "Tri-state output" in the interface.

That specific KLC clause, I understand, is with regards to something like http://www.ti.com/lit/ds/symlink/uln2003lv.pdf where you have input signals at a logic level on the left, and driver outputs on the right. Or an opamp with inputs pins on the left and output on the right. Either way, it's saying that signal flow should be left-to-right. For this symbol, I'm just asking if it's best to put the USB jack on the left or right of the schematic page and flow signals from the USB jack to this IC left or right. Either way works for me, but I wanted to bring it up so it wasn't a discussion only post-merge.

I'm going to close and reopen to Travis will see the footprint that was merged.

evanshultz:
These symbols are all good now. Do you want to do any additional fixing of the Travis errors just for fun, or make any further changes, or merge as-is? Let me know.

Ratfink:
Oh, I still haven't finished the changes we discussed earlier.  I'll commit those changes, then I think it'll be ready to merge.

evanshultz:
Thanks! When you change symbols visibly/substantially within a PR, it's helpful to post new screenshots for quicker reviewing.

PR #364, opened 2018-03-10
Add SN74CBTLV3257 Mux/Demux

suzizecat:
Hello

I would like to submit the SN74CBTLV3257 Mux/Demux from TI.
You may found the DS here : [SN74CBTLV3257](http://www.ti.com/lit/ds/symlink/sn74cbtlv3257.pdf)

I have few question regarding this merge : 
 - Is there an appropriate library to put this in ? (I put it into a "Multiplexer" library but I would rather put it in an already existing lib)
 - Is my way of labeling is good (I want to keep something to know the output names, but the power/command block isn't good looking..

I will add aliases and UQFN version later.

image

------------


Misca1234:
Well some minor details

Change Vcc into VCC
(it looks like the data sheet uses small 'c' but it is actually 'C' with a gravity centre slightly lower than the 'V'

I assume you are using a very small font to be able to make the symbol very small
but try to avoid to deliberately create travis errors.
I would recommend to use a slightly bigger font so travis does not complain.
If needed a symbol can go outside travis requirements if needed, but it is nice to avoid the travis complains.

I understand you want to use the text 'SEL' to make it more clear but 
stick to the data sheet and use S

suzizecat:
I now understand why there was those Travis errors...
Don't know why the default size of my texts went into that small size...

I don't quite understand why Travis complain about "Pin name position offset" :thinking: but I will look into my design to see if I find something wrong...

Any though about the library ? I would really appreciate not having to move my symbol later, since KiCAD cannot handle that kind change properly ATM (it break my design anytime a change in libs is made... ).

OK for other names changes.

poeschlr:
The other 74xx parts are in the 74xx lib. There was a discussion about splitting it but it was decided to keep it as one large lib. (At least for kicad 5)

suzizecat:
The problem here, is that my Mux doesn't really fit into 74x(g)x libs since it's not a standard 74xx series component... (as [Wikipedia](https://en.wikipedia.org/wiki/List_of_7400_series_integrated_circuits) says)

I don't mind putting it into 74xx lib, but I prefer to be sure :wink:

jkriege2:
why is it not a member of the series?

jkriege2:
Also: Does it make sense to have the OE and SEL-pins in the dedicated POWER-symbol? Wouldn't it be better to not separate the four MUXes, but combine them into one box-shapes unit with internal drawing (+1 power-unit), as they are not really independent?

poeschlr:
Even if it is not officially considered part of the series, i doubt that users will realize that.

And yes after the v5 release is done we might need to revisit the logic series libs and find a better more general organization scheme. (We just don't have enough time left before the release to do a proper job.)

suzizecat:
> Wouldn't it be better to not separate the four MUXes, but combine them into one box-shapes unit with internal drawing (+1 power-unit), as they are not really independent?

If I have to redraw that with all muxes in one block, it would IMO be better to add the power pins on the same block like some other muxes/demuxes.

I just find that having muxes split like logic gates would be pretty nice, but I don't mind since it imply some problems (pin names inside/outside symbols, size...)

evanshultz:
@jkriege2 
I agree with separate switch symbols. I often find myself placing them in different areas of the schematic but showing the signal flow without net names is easier.

I like having ~OE and S on the power symbol.

@suzizecat 
@jkriege2 updated analog_switches and separated individual switch symbols at https://github.com/KiCad/kicad-library/pull/1773. I think those would be an excellent style to copy for this symbol. For example, the switch of ADG419 would work.

suzizecat:
Changed symbols, moved to 74xx and fixed sneaky KLC issues

image

evanshultz:
Here are my notes:
- Remove the "SN" from the symbol name. This will make it similar to other parts in the lib, give the part number available from many vendors, and make it start with "74".
- Your symbol only applies to SSOP, TSSOP, SOIC, and TVSOP packages. VQFN has an exposed pad and UQFN has a different pinout. Remove those footprint filters.
- Did you check which footprints already exist in the KiCad library? Sometimes a footprint named MSOP or TSSOP would fit the SSOP in the datasheet, for example.
- The TVSOP dimensions should be `4.4x3.6mm`. You have them backwards.
- Add pin pitch to your footprint filters.
- Pin lengths on unit E can be 100mil and the box slightly wider. Alternatively, pin 15 could perhaps move up 100mil to give a bit more distance from the GND text. Do you like either option?
- Perhaps the dashed horizontal line on units A-D should be removed since there's a toggle input (pin 1) for all switches. I think the dashed line makes more sense when there's a toggle input in each unit.

suzizecat:
- Didn't saw that this part was available from other vendors, changing that.
 - Filters updated
 - Changed the TVSOP size (after taking the screenshot)
 - Pin pitch OK
 - I'd rather let 150mil pins when dealing with 2-digits numbers. I've slightly increased the symbol size and moved VCC/GND pins. The origin isn't centered bus the symbol isn't symmetrical soooooo... it should be OK, shouldn't it ? :innocent:  

image

image

evanshultz:
100mil pins with 2 character pin numbers is common in many libs. But it seems the 74xx library, which is script-generated, is using 200mil pin lengths. So I suppose it doesn't matter too much. And it meets KLC.

I wrote horizontal instead of vertical in my last comment, but you figured it out anyway. ;)

For the record, commit https://github.com/KiCad/kicad-symbols/pull/364/commits/5487d46a53a5d39087b6a2a6289df51af4db3da9 changed the 74xx lib from version 2.3 to 2.4. It created a lot of noise but I trust you didn't touch any other symbols.

suzizecat:
I wouldn't dare !
Thanks for the review and the hard work ! :+1:

PR #365, opened 2018-03-10
Add SK6805

fauxpark:
This is the "NeoPixel Nano", another addressable RGB LED that I must have missed.

Datasheet: https://cdn-shop.adafruit.com/product-files/3484/3484_Datasheet.pdf
Footprint PR: KiCad/kicad-footprints#418

sym

------------


Ratfink:
Looks good!  Since the other addressable RGB LED symbols have the same pin name offset as this one, I think we can ignore that Travis message, and that's the only possible problem I see.

Thanks for the contribution!

PR #367, opened 2018-03-10
Added symbol for TI BQ25601

Duckle29:
Added symbol for TI BQ25601

------------

Datasheet: http://www.ti.com/lit/ds/symlink/bq25601.pdf
image

Matches PR: https://github.com/KiCad/kicad-footprints/pull/419


Duckle29:
Oh and I accidentally used an old email for the first commits / pushes, so all contributors have signed

Misca1234:
You should change your FP filter to
WQFN*1EP*4x4mm*P0.5mm*EP2.6x2.6mm*ThermalVias*

A suggestion would be to place pin 13,14 ontop of the symbol

Fill the symbol with foreground colour

Pin 7 should be open collector
I suggest you change Pin 5 to be an input to avoid discussion
I suggest you change Pin 6 to bidirectional

Pin 19 and 20, make them passive 
Pin 15 and 16 make them passive 
I suggest you rename pin 25 to EP1

The two "not connected pin" ontop, move them to the side so they dont interfere with other pins, make is more easily to read

Ratfink:
@Misca1234 The footprint PR is https://github.com/KiCad/kicad-footprints/pull/419.

Duckle29:
@Misca1234 The idea with having pin 13 and 14 where they are is to keep with the datasheet: 
image

I have though moved REGN down as it needs to have a cap to ground, and didn't need to be betwewen BTST and SYS

As for the two NC, I've moved them to the side, I initially put them where I did, as I was unsure if invisible pins could accidentally connect to a trace passing over them.

I also assume you meant don't fill with any color, as filling with foreground makes the symbol full red.

I've changed the fp filter to `WQFN*1EP*4x4mm*P0.5mm*EP2.7x2.7mm*ThermalVias*` as the footprint is now called that (2.7mm instead of 2.6)

The new symbol looks like this: 
image

evanshultz:
- The symbol should be filled with background color, not foreground. It should be yellow.
- @Misca1234 is asking you to stack pins that are the same net. Pins 13/14, 15/16, 17/18, and 19/20. You should also stack pin 25 on the other ground pins as the datasheet says to do, which means you'll need to rename it to GND.
- There is a typo in the description and it's missing some good info from the datasheet title. It's also missing the package. You could use `I2C Controlled 3A Single-Cell Battery Charger for High Input Voltage and Narrow Voltage DC Power Path Management, WQFN-32`.
- Pin types:
-- BAT: Power Input
-- BTST: ERC will be easier if this is Input
-- PMID: Passive (all that goes here is a cap)
-- REGN: Passive
-- SYS: Input (I disagree with the above; Input is best to me since it's a high-impedance input)
-- TS: Suggest passive since it's an analog input with resistors and/or a TC connected to it
- NC pins will connect to a trace over them. Either position should be fine since it's within the symbol body.
- TS and REGN aren't related, so you may want to add an extra pin space between them.
- The symbol is very tall. You could shrink it a bit, but it's not required. Not so much space is needed between SYS and ~PG and also the bottom edge can move up once you've stacked the ground pins.

Duckle29:
Thanks for the tips. This is what it looks like now:
image

I also replicated the typical application diagram, and it looks like this: 
image

evanshultz:
@Duckle29 
Looking good now! Thanks for making changes.

Please see the Travis CI output above by clicking on the Details link. There is a bit more work to be done.
- Many pins are off grid. Please place them on a 100mil grid.
- While you're at it, please shorten the symbol. It looks like it could be 400mil shorter.
- Pin 16 needs to be Passive type to stack it on pin 15.

Also: Isn't SYS the wrong pin type? Now that I look at it again, SYS accepts the power converter's output for charging the battery. It also has sense circuitry. So Power Input or Passive or even Input should be OK, but Power Output seems wrong to me.

evanshultz:
Finally, the FPfilter can remove the `ThermalVias*` at the end. If the user doesn't want a footprint with vias, or another footprint with a different suffix is available, we allow easier selection of otherwise matching footprints by not forcing a match on a footprint with vias.

Duckle29:
I've made most of the requested changes. This is now what the symbol looks like in a circuit: 
image

I don't think I can shrink it any more (shrunk by 200mil), as it'd hurt the ease of schematic layout.
Yes the sys pin should be a power input, not output. Regardless I've set it to passive. I've fixed the grid allignment, and the default package. 

This is how the package looks on it's own: 
image

evanshultz:
Argh. Travis isn't checking. Sometimes it gets stuck. Let me close and re-open (@Ratfink I don't know the login).

Ratfink:
@evanshultz It integrates with Github, so once you log in as yourself using your Github account, you automatically get the permissions to re-run builds for the KiCad repositories.

evanshultz:
@Duckle29 
Can you please check if all the emails you've used for GitHub are part of this account? That should resolve the CLA issue above where one account has signed but the other hasn't.

 See http://kicad-pcb.org/libraries/klc/S3.1/. This symbol is offset. The square defining it is `S -400 -700 400 1200 0 1 10 f`. You can see it's at -700mil and +1200mil vertically, and it could just as easily be -900 and +1000. If you can fix that I'm ready to merge.

@Ratfink @Misca1234 
Do you guys find any other outstanding issues?

Duckle29:
@evanshultz Done, as for the other email, I can't really add that. It's an old github no-reply, from before I changed my github nick.

evanshultz:
@Duckle29 
Thanks! Everything looks good to me now.

@Ratfink @Misca1234 
Bring it up if you find anything else to mention. If not, I'll merge this within the next day.

PR #370, opened 2018-03-10
Set Travis to use Python 2.7

Ratfink:
The default language is Ruby, which causes Travis to waste about four seconds setting up a Ruby environment we don't need every time CI is run.  This PR changes the language to the one we actually use, Python 2.7.

I suggest squash-merging this PR, as the commit history is a bit of a mess.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,8 @@
+branches:
+  only:
+    - master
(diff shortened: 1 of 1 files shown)


poeschlr:
thanks

Issue #27, opened 2017-12-05
Issues with logic libs


evanshultz:
See https://github.com/KiCad/kicad-symbols/pull/11#issuecomment-349285725.

Items noted in that PR:
Here are some things that could be fixed.
- There are no packages or package types (just "SSOP" without pin count) listed in part descriptions
- No supply voltages given in descriptions
- Some hyphens in FPfilter, such as 74LVC1G02
- Some descriptions use all lowercase letters
- 74LS09 is describes as "Open Collect" instead of "Open Collector"
- The datasheet links in Logic_74xx don't work for me
- At least in the IEEE libs there are some lowercase letters for pins names; do we allow that?

There was a recent discussion about having a separate power symbol for even single-gate logic. I still support this so I want to resolve that discussion. I believe @jkriege2 was with me on this point.

jcyrax:
Hi

I can help with this issue if nobody has started working on them.

jcyrax:
@SchrodingersGat @jkriege2 @poeschlr I wan't to inform just in case that I started working on this problem so that nobody is doing it at the same time.

jkriege2:
Hi!

note that @bobc scripted some of these symbols: https://github.com/bobc/kicad-symgen and it might be a good idea to use/extend that generator!

Best,
JAN

Issue #190, opened 2018-01-20
KLC-modifications/additions: OPAMPS/logic gates and power-units


jkriege2:
transfer of old issue https://github.com/KiCad/kicad-library/issues/1839

# original post:

I would like to suggest two additions/modifications to KLC, in part inspired by the discussion https://github.com/KiCad/kicad-library/pull/1828:
1. On S3.8 (http://kicad-pcb.org/libraries/klc/S3.8/). For logic and OPAMPs that's a bit inconsistent. I would suggest to ALSO separate the Power-ports for two special cases, namely logic and OPAMPs (I think these are also the most common devices that come as multi-unit-symbols). My rationale would be: If I design a big logic-circuit, it's good to collect all the power-units at one location and connect them to the appropriate power-plabes (+caps and what not). This tidies up the schematic significantly. If we discourage extra power-symbols for single-unit logic-symbols, this will lead to a mixture, as some logic ICs have their power-pins inside the schematic (may be annoying if you change  power-suuply to them, as you have to do it for every device separately) and some are collected somewhere. Also this would allow us to make unify the single-gate-logic families. Maybe (maybe!) it is possible to design the power-symbols in a way that they can be positioned over the actual symbol (at least for gates), so they can (if wanted) be "combined" to a single device. I think @bobc has some examples of that as alternative symbol style.
2. In addition I would like to see the same exception for OPAMPs for the same reason and with the same style requirement of beeing combineable (I think that ios already common for many multi-unit OPAMPs.
![2017-11-17 20_18_36-eeschema test_addlin _ d__kicad_test_addlin](https://user-images.githubusercontent.com/11967523/32964759-885671b2-cbd4-11e7-93ae-b03fe6bde4ed.png)
3. _[item 3 removed from old post as it does not match thematically and was not generally agreed on]_

------

# Discussion

I think this was mainly agreed on in the old issue. So if nonone has any additions/problems with that I will propose some wording soon. Please leave your comments here, so they can be incorporated in the final wording.

bobc:
I'm not sure what exactly is being proposed here. Overlaying the power pins looks rather ugly, and only works for certain symbols like AND.

Power units were introduced to fix a specific problem which had no other reasonable solution, and are still not ideal. I'm not convinced we need to change the graphical style again, there seems to be very minimal gain. I think we need to be careful whether we are fixing real problems, or tinkering with the style for the sake of it. 

[...]

PR #371, opened 2018-03-10
Rework of 74xGxx lib

jcyrax:
I completely reworked 74xGxx library (part of issue https://github.com/KiCad/kicad-symbols/issues/27)
They are created using script and https://github.com/bobc/kicad-symgen

I used wikipedia as source for descriptions and types.
Symbols are for nexperia part numbers made as generic as possible. For instance, if specific symbol has multiple footprint types with different pinout variants I added full name for that specific pinout.
I didn't create all pinout variants where there was more than one, only single per logic gate type. 

There are some footprints that will be needed with this. I'm currently making them.

How this will be tested? There are 96 symbols and a lot have multiple units? I don't want to make that much screenshots. 

------------


evanshultz:
Thank you!! This is such a huge improvement to the library!

Nexperia (from NXP; from Philips) uses their own package naming system. We need to use industry-standard package/footprint names. For example, for `74LVC1G126GV` you selected `Package_TO_SOT_SMD:SOT-753_1.5x2.9x1.0_P0.95` as the footprint. However, we should name our footprint something with an industry standard, if possible, which in this case is `SC-74A`. You can see this in the [datasheet](https://assets.nexperia.com/documents/data-sheet/74LVC1G126.pdf) on page 2:
image

Also see https://www.nxp.com/docs/en/package-information/SOT753.pdf, where NXP also lists an industry-standard name of `TSOP5`.

While SC-74A may not exactly be the same as any footprint we have, our [`SOT-23-5`](https://github.com/KiCad/kicad-footprints/blob/master/Package_TO_SOT_SMD.pretty/SOT-23-5.kicad_mod) should work fine.

If you do create new footprints, use industry-standard names.

Furthermore, to provide a 3rd party reference for footprint names, when I search for "74LVC1G126" at Digi-Key I see the following footprints available:
image

Testing by librarians is best done by pulling your patch and then we can review locally. Agree it's too much work to do it another way. Any issues can be added as they're found and then you can re-run the script to fix. Sound OK? (I haven't look at any symbols yet but I will in the next day or two.)

jcyrax:
Yes, I was a bit confused. I saw SOT in kicad lib and decided to use that naming.

I plan on creating all footprints referenced in this lib. I'll update used footprints and filters to use industry names.

evanshultz:
The first batch of notes from just arrow-keying through the symbols:
- I favor floating pins without a box for power symbols, as that allows placing the power pins on the gate if desired. We do something very similar for opamps and comparator symbols.
- 74AUP1G0832 uses a box symbol that hides the gates. Same for 74lvc1G57 and all 74lvc1G9x.
- Do we want to use pin types with shapes or just generic lines? For example, 74AUP1G373 uses the clock type pin for pin 1. Do we want that or just a line with no triangle? Similarly, the 74lvc1G125 has a circle on the ~OE pin.
- Perhaps pins 6-7 of 74lvc1G123* could be moved on the left side and it could be more square?
- A few symbols are using 40mil pin text offset.
- Since text can't be placed per-unit, maybe shoving the text to the right would clear both the main gate and the power pins for all symbols?
- I believe you should use the generic package type as a suffix for the symbol name. Such as `74LVC1G126_SC74` instead of `74LVC1G126GV`.

Take the above as points of discussion; there's a lot that changed and so probably there's a bit of talking about it before we settle on what to do.

@poeschlr @jkriege2 
See the last bullet point. I think this is where we want to add the package type as a suffix to the symbol name? Right? Multiple manufacturers make these parts, and they're interchangeable, so it seems like the right place.

poeschlr:
@evanshultz i suspect these are made by many manufacturers and there is no guarantee that all of them use the same coding to identify which package the part comes in. I would therefore agree with using the package suffix.

jcyrax:
> I favor floating pins without a box for power symbols, as that allows placing the power pins on the gate if desired. We do something very similar for opamps and comparator symbols.

I sort of don't like that style. If you place over symbol it is ok, but I find it weird if I have floating pins somewhere in schematic. Also the size of gates wouldn't allow power pins so close to overlay. Then you have gates with enable signal that would prevent this. And the least thing is, I don't think that symgen can generate unit without outline currently.

> 74AUP1G0832 uses a box symbol that hides the gates. Same for 74lvc1G57 and all 74lvc1G9x.

I thought about making them, but previous lib had only boxes for them. Can we leave this as boxes for now and after this is merged I could find some time to make those changes?

> Do we want to use pin types with shapes or just generic lines? For example, 74AUP1G373 uses the clock type pin for pin 1. Do we want that or just a line with no triangle? Similarly, the 74lvc1G125 has a circle on the ~OE pin.

I think you always need circles for gates that don't use pin names. For latches and flip flops and symbols with pin names, we can probably remove them. So only pins with straight line.

> Perhaps pins 6-7 of 74lvc1G123* could be moved on the left side and it could be more square?

And offset power pins? Is that what you mean? Or you mean to move them because of function and just make height larger?

> A few symbols are using 40mil pin text offset.

Will fix that.

> Since text can't be placed per-unit, maybe shoving the text to the right would clear both the main gate and the power pins for all symbols?

I'll check if I can control this.

> I believe you should use the generic package type as a suffix for the symbol name. Such as 74LVC1G126_SC74 instead of 74LVC1G126GV.

@evanshultz @poeschlr My thought was to use exact manufacturer part number so it wouldn't be generic, but exactly opposite. These would be exact nexperia parts. And you could add others manufacturers as aliases with their names or create new symbol if pinout differs so names wouldn't conflict.
I'm not against doing with package names, but is there a guarantee that pinouts are same from all manufacturers?

poeschlr:
For now i think this lib should stay generic. We can think about a way how to get these symbols more specialized at a later time. (And if we do we should definitely add more then one manufacturer at a time.)
I think it is already a great improvement if we have symbols with already assigned footprints identified via a suffix. (You can add the NXP part number to the description as an example of a compatible part.)

---

I think for v6 we should then reorganize the logic libs.

One option would be to have a lib per manufacturer 74xx_NXP, 74xx_TI, ... (Would reduce the risk of multiple manufacturers using the same part number to mean something entirely different.)

Or we split it more along the lines of function. Examples: Logic_And, Logic_Mux, ... Each of them holding fully specified symbols with name [Man]_[MPN]. Each of these would hold gates from every logic family. (This might be too fine grained.)

We could already start a discussion in a separate about all of this if you are interested.

jcyrax:
OK, I'm fine with using package notation for now.

For v6 changes, we can postpone the discussion until then.

evanshultz:
> I favor floating pins without a box for power symbols, as that allows placing the power pins on the gate if desired. We do something very similar for opamps and comparator symbols.
> 
> I sort of don't like that style. If you place over symbol it is ok, but I find it weird if I have floating pins somewhere in schematic. Also the size of gates wouldn't allow power pins so close to overlay. Then you have gates with enable signal that would prevent this. And the least thing is, I don't think that symgen can generate unit without outline currently.

Yeah, style is subjective so there's often no obvious rule and consensus wins. But it does seem possible. I simply deleted the box around the second unit (the line is `S -200 300 200 -300 2 0 10 f`) and end up with this (also see some opamp symbols which have a similar approach):
image

It might not be possible using the editor, but it is possible if you're writing the symbols from a script.

> 74AUP1G0832 uses a box symbol that hides the gates. Same for 74lvc1G57 and all 74lvc1G9x.

> I thought about making them, but previous lib had only boxes for them. Can we leave this as boxes for now and after this is merged I could find some time to make those changes?

Of course you can leave them as just boxes. But I noticed that 74AUP1G3208 shows the logic gates inside a box, and so it caught my eye when other similar parts didn't show the gates inside the box.

> Perhaps pins 6-7 of 74lvc1G123* could be moved on the left side and it could be more square?
> 
> And offset power pins? Is that what you mean? Or you mean to move them because of function and just make height larger?

I was thinking moving the pins to make the part more square. The power pins could be offset or not. It just looks ugly to me when it's to short and wide. We could also consider to change pin 7's name to `R/C` (not saying we should or it's a good idea, but that doesn't lose intelligibility to do it).

> One option would be to have a lib per manufacturer 74xx_NXP, 74xx_TI, ... (Would reduce the risk of multiple manufacturers using the same part number to mean something entirely different.)

I've never seen this. The PNs in each logic family were selected with a specific function that is common to anyone using that PN in their part. Maybe there are some conflicts, but I've never seen or heard of them.

Here are family documents from a few vendors:
- https://www.nxp.com/docs/en/brochure/Logic-MicroPak.pdf
- http://www.ti.com/lit/sg/scyt129f/scyt129f.pdf
- https://www.fairchildsemi.com/collateral/TinyLogic.pdf

A few spot checks all match. And this might be good reference info later for this library.

Isn't using a generator script a good reason for keeping all 74xx libs together instead of separating them by function? Not only is it "logical" (pun intended), but then the generator can write out all members of of a single library. The user experience is more important, but this still seems a good reason to me.

jcyrax:
> Yeah, style is subjective so there's often no obvious rule and consensus wins. But it does seem possible. I simply deleted the box around the second unit (the line is S -200 300 200 -300 2 0 10 f) and end up with this (also see some opamp symbols which have a similar approach):

I was under the impression that its sort of under debate about this power unit usage (see https://github.com/KiCad/kicad-symbols/issues/190). If you say that its decided to use this way, then I'll move on and try to implement it like that.

However, I have questions:
Is this how it would be used in schematic? Not touching symbol outline?
image

evanshultz:
Consensus is people agreeing, and since there is an open discussion there isn't agreement. :) I'm just stating that I favor one particular style but eventually a decision has to be made and one style wins. The library does need closure on this point so let me poke the other issue. Thanks for bringing it up!

thomasonw:
Ref to issue:  #756 ,  Good to see you had been working on this lib.  Just took a look  and see one issue (Which seems to also have been talked about above):  How to deal with power pins.  And the issue I see is specifically power-pins in single entity logic devices. 

Looking for example at the 74LVC1G06_TSOP-5, this is a single logic entity and is present coded with separate power-pins in a 'B' unit, However  KLC specifically calls out this should only be done with multi-device units:  [ttp://kicad-pcb.org/libraries/klc/S3.8/ ](http://kicad-pcb.org/libraries/klc/S3.8/)

In part " _... This approach should not be taken for single-unit symbols. Power pins in this case must be drawn on the same symbol as the logic pins._"

jcyrax:
Decision to put power pins on separate unit was because of few reasons. One is that logic with enable pins would look bad in my opinion. Second that kicad-symgen that I use to generate symbols isn't capable of creating that kind of style.

Anyway I would totally understand if these problems prohibit inclusion in official libs.

Misca1234:
@evanshultz 
I guess you are deepest involved in this one

evanshultz:
@jcyrax @bobc 
I'm keen to see all the logic libs move to script generation, visible power pins, and the other modernization improvements.

To allow this to happen, I propose new libraries with `_v6` are merged (like `74xGxx_v6.lib`) for all logic libs to shake out issues in the generator script and symbols without affecting the existing libs. These scripted libraries can then potentially be used for the new v6 file format testing as well.

Are either of you up for continuing your contributions so far? If not, do you mind me taking over?

thomasonw:
I would say go for it!   I have no insights into script generated libs, and do thing the cleanup is a key part of letting KiCAD be more widely accepted.

PR #294, opened 2018-02-09
Add symbols for connectors with mounting pins.

poeschlr:
As discussed in https://github.com/KiCad/kicad-footprints/issues/117

Script: https://github.com/KiCad/kicad-library-utils/pull/210

@Shackmeister here the symbols i promised you over on the forum.

I am not sure if my symbolic for mounting pins is any good.

screenshot from 2018-02-10 00-36-04

PR #372, opened 2018-03-10
Adding symbols for connectors with mechanical mounting pin.

poeschlr:
Replacement for https://github.com/KiCad/kicad-symbols/pull/294 (As i screwed up by making a wrong change in it that i could not revert nicely.)
Script: https://github.com/KiCad/kicad-library-utils/pull/210

---

As discussed in https://github.com/KiCad/kicad-footprints/issues/117

The intention is to have symbols that allows connecting mechanical
mounting pins to a net. This allows connecting the pins to a zone
for increased mechanical strenghts or to allow a zone to better
connect to the area around or under the connector.

---

Footprint filter (single row): `Connector*:*_1x??-1MP*` 
Footprint filter (dual row): `Connector*:*_2x??-1MP*`
This assumes the footprints to be in libs starting with Connector and with a footprint name of the format `..._[num rows]x[num pins per rows]-[num uniquely numbered mounting pins]MP_...` (And yes i will add this to the KLC as soon as @Shackmeister adds the first footprints that are compatible with this.)

---

Screenshot:

screenshot from 2018-02-21 23-43-46


poeschlr:
Screenshots for testing of footprint filters with the footprints of https://github.com/pointhi/kicad-footprint-generator/pull/118

---

Normal connector symbols do not show footprints that include a pin number for the mounting pads.

screenshot from 2018-03-11 03-19-43

---

Symbols with mounting pads show the footprints with the mounting pads and none without them:

screenshot from 2018-03-11 03-19-36

---

Symbols for shielded connectors still show only the one footprint with a shield pin:

screenshot from 2018-03-11 03-22-26

evanshultz:
Everything above looks fine.

PR #373, opened 2018-03-11
Atmel MCU fixes

fauxpark:
Fixing the outstanding issues with the Atmel MCUs. Don't merge yet!

So far I have:
 - Let KiCad rearrange the .lib files first (to avoid noisy diffs)
 - Reset the pin names and numbers to 50mil
 - Centred all symbols (this does not mean they're all on the 100mil grid yet!)
 - Centred the Footprint and (unused) Datasheet fields on (0,0)
 - Fixed some miscellaneous bits missed from previous PRs because the script outputs were just so long

Still to do:
 - [x] Align the pins properly
 - [x] Move power pins to the top and bottom
 - [x] Add NC pins
 - [x] Shorten pin names. I think I'm going to throw out most of the stuff in parentheses and just use their "main" name eg. `PB0`, except where there is an "important" alternate function also on that pin: `PB0/XTAL1` or `~RESET~/PE6` etc. The main names will be closest to the edges of the boxes. Would appreciate some feedback on this if anyone is opinionated.
 - [ ] Shrink and align the boxes after the name chopping is done
 - [x] Rename libs and symbols from ATTINY, ATMEGA to ATtiny, ATmega
 - [ ] Move obsolete symbols to obsolete/
 - [ ] Other stuff?


evanshultz:
Thanks for continuing to fix these libraries which were so out of date!

1. After pin alignment and moving power pins, let's merge. This will reduce a ton of noise in the Travis output and also make it more clear about what to do next (as you said). That should be all pin movements. Please stack all pins that are connected together except in the case where there is a clear reason to not stack them (like the datasheet instructs decoupling capacitors on individual power pins). This would be a good time to show some sample symbols for quicker reviewing.

1. NC pins should be added if there are a handful of them, but if there are a lot we can do other things. But I suspect there are only a few so adding them is OK.

1. Is there any reason to delay or speed up or cancel the proposed lib re-naming, @poeschlr?

poeschlr:
If we want to rename the libs we should do that as soon as possible. I think we should set ourselves a deadline (Some RCx tag) when we introduce something similar to a "feature freeze". (At least no lib renaming after some point. As renaming libs will result in error messages for the user after lib update. We could even go further and also disallow changes that would break user designs but that i think this should be limited to a few weeks before the final release. If we want to do that at all.)

Yes i know this will most likely mean that there will be some timeframe where we will not be able to do everything we would like to do. But i think it will be for the better overall.

fauxpark:
@evanshultz - now that I have reset the pin name sizes to 50 mil, the text overlaps and sometimes even comes out the other side:
sym

This is not too much of a problem since they are going to be shortened anyway, but I think it should at least be done before merging.

evanshultz:
That is an enormous reduction in the Travis outputs. Thank you!

But yikes.... At this point the symbols are unpleasant and I would not like to merge. Can you either: shorten the pin names (my preference) or widen the symbols as the next step? That way we can get back to nice-looking symbols if we're to merge. We will also want to re-name the libraries right away if we're going to have a chance to make it for KiCad v5.

Please advise what you can get done in the next day or two. If we have more time, that's nice, but we'll only want to merge usable symbols and merging frequently will be the best chance to get more fixes into the official libs for the upcoming release.

On the other hand, if you want to do all the changes before merging go ahead and work. When it's ready we can merge regardless of the KiCad release schedule. Up to you. :)

fauxpark:
I have added the EP and NC pins now, if the missing footprints were in the library all the symbols should be associated and filtered properly. Some have quite a few NC pins (ATtiny13-20MU has 12 out of 20!) - what is the best way to handle this? I see stacking is not allowed, does pcbnew draw airwires if they are?

> At this point the symbols are unpleasant and I would not like to merge.

My point exactly :) I will do the pin names next.

poeschlr:
For other similar symbols we shortened the pin names by not including detailed information about the possible alternative functions of a pin. We only left the identifier. (And for some special pins the important function. Example for XTAL and reset pins we kept that information.)

With luck the new file format will give us more options here.

fauxpark:
Okay, pin renaming is done. Some samples:
at90usb1286
attiny28
atxmega8e5

Still not quite there yet, but usable I think, and mergeable.

evanshultz:
Looking good! So that means the 4th checkbox above is done? And are you saying you'd like to merge now?

Can you re-name the libraries before the first merge so that's done? Note that you need to edit the sym-lib-table file when you do.

And one question: I looked at one datasheet (http://ww1.microchip.com/downloads/en/DeviceDoc/doc7679.pdf) and AREF is either an input or output. It seems some symbols make it Bidirectional and some Input. The AREF pins should at least be all the same type, so can you add homogenizing those pins to you list?

fauxpark:
I was wondering what to do about them. I have changed all the AREF pins to input, except where it is an alternate function of an IO pin, and on the ATmega16M1 which can apparently be configured as a current source output.

The pin types in general need a closer look I think, but they should be OK for now.

evanshultz:
@fauxpark 
Thanks for your continued work!

One quick thing: Because Microchip bought Atmel please rename the libraries accordingly. I believe all parts are now Microchip.

I'll review this with an intent to merge tomorrow.

evanshultz:
Pin stacking is allowed, and preferred. We've only identified one specific case where pins should be unstacked: when the datasheet calls out specific decoupling on individual pins. If there is another reason then please point it out. Otherwise, all pins of the same net can (and should) be stacked.

If the qty of NC pins is huge, like your example above, we can handle that another way by putting the pin count in the footprint filter. But since you've already added all NC pins that's OK too. When you shrink symbols we can revisit this.

ATmega1281-16AU and ATmega1281-16MU are weirdly missing an entry in the DCM file. Maybe it's a typo of the MPN or they really are missing? Can you look into that also?

Here are other things to look at for the next steps:
- Pin types for all pins, especially the special-function ones that aren't power or GPIO.
- Pin placement when you shrink symbols. Should all GPIO be on the left side, right side, distributed as evenly as possible, etc.
- Do all XTAL pins go on the same side? How far apart are they so a crystal symbol fits there nicely.
- Place power and ground pins next to each other or leave a space between them (like the GND and EP pins, for example).
- Pin length should be 100mil if pin numbers have <3 characters (usually 99 pins or less). All NC pins can be 100mil.
- Place refdes and MPN inside the symbol outline if it's large enough?

I took at a look at all symbols in the symbol viewer. This looks fine to merge now. If there are any issues we will catch it later. At least we can set a new benchmark. I'm OK to merge.

I'd appreciate if somebody else can check this. If there are any major issues let's address them now, otherwise I will merge in a day and we can catch any outstanding work in the next PR.

Ratfink:
I'll take a look this evening.

fauxpark:
I meant NC pin stacking - S4.3 says they shouldn't be stacked. I see the logic behind that, and figured it was because pcbnew would draw airwires between NC pins.

The ATmega12**6**1 symbols I could not find any information on, and I think they probably don't exist. They're just aliases of the 1281 though, and I'll remove them unless someone manages to dig up a datasheet.

XTAL pins should probably always be on the same side (left), except when they share a pin with a GPIO line. My personal preference is to preserve the ordering and grouping of the IO pins.

I opted for symmetry with the power pins - if there is only VCC, it goes in the middle; if there are VCC and AVCC, they are centred 200mil apart, etc. But perhaps VCC should always be in the middle. The groups of GPIO pins should also be 200mil apart.

I kept all pin lengths at 150mil for consistency (and OCD 😉) because all of the symbols are somewhat related. There *are* only six with 100 pins though...

Ratfink:
I just looked through this PR, and I didn't see any issues that haven't already been discussed.  This is quite an improvement over the old Atmel libraries, thank you for all the care you're putting into them!

evanshultz:
Great! Merging then.

I moved the open items to https://github.com/KiCad/kicad-symbols/issues/387.

ioch:
There are some popular Atmega MCUs with incorect electrical pin types.
PC[0-7] pin electrical type Output, should be Bidirctional and
PF[0-7] pin electrical type Input, should be Bidirctional in following symbols:
- ATmega1281V-8AU
- ATmega1281V-8MU
- ATmega640V-8AU
- ATmega640V-8CU
- ATmega64L-8AU
- ATmega64L-8MU

PE[0-2] pins on Atmega161 are also called bidirectional in datasheet.
Sorry, for not doing proper pull request thing. I don't fell comfortable editing symbols in bulk.

PR #374, opened 2018-03-11
 Changed pin numbers for T0-92 package of 78L05

christianlupus:
The pin numbers in the different packages (TO-92 and SOT-89) are not identical because input and output are interchanged. This might lead to shortcircuits in finished designs.

In this PR I simply interchanged the pin numbers of input and output so that the current footprints match.

See
http://www.st.com/content/ccc/resource/technical/document/datasheet/15/55/e5/aa/23/5b/43/fd/CD00000446.pdf/files/CD00000446.pdf/jcr:content/translations/en.CD00000446.pdf


poeschlr:
Does the affected symbol have any alias? If so might this now break the aliased parts?

christianlupus:
I have to admit that I did not check all aliases.
Now after doing so, I found that the pinning for TO-92 footprint is not consistent over all suppliers. Some make the input left other make the input right. One idea was to separate the symbol and make two out of it (IN-GND-OUT and OUT-GND-IN) with corresponding aliases attached. Was this a valiable solution?

christianlupus:
You are right. Plase do not merge. I will close ASAP.

PR #375, opened 2018-03-12
Added mcp2021a symbols Interface_CAN_LIN library.

sebastianbasierski:
Added symbols for mcp2021a Microchip LIN interface.

Datasheet: [http://ww1.microchip.com/downloads/en/DeviceDoc/20002298C.pdf](http://ww1.microchip.com/downloads/en/DeviceDoc/20002298C.pdf)

Screenshots of symbols.
mcp2021a
mcp2021a

mcp2021a-ep
mcp2021a-ep
Matching footprint: [https://github.com/KiCad/kicad-footprints/blob/master/Package_DFN_QFN.pretty/DFN-8-1EP_4x4mm_P0.8mm_EP2.5x3.6mm.kicad_mod](https://github.com/KiCad/kicad-footprints/blob/master/Package_DFN_QFN.pretty/DFN-8-1EP_4x4mm_P0.8mm_EP2.5x3.6mm.kicad_mod)

Signed-off-by: Sebastian Basierski <sebastian@basierski.pl>

------------


Misca1234:
I suggest you 
Convert all letters to capital, that is Txe to TXE

mcp2021a-ep:
I think you have forgot to add the EP pin to the symbol


$CMP MCP2021A-330
+D LIN Transceiver with Voltage Regulator, 3.3V, 3.3V
into
$CMP MCP2021A-330
LIN Transceiver with Voltage Regulator, 3.3V, DIP-8, SOIC-8

$CMP MCP2021A-330-EP
LIN Transceiver with Voltage Regulator, 3.3V, DFN-8

Why do you have the suffix '-330' to the symbols ?
you could just call them
MCP2021A
MCP2021A-DFN

I am also sceptical to your chosen foot print for MCP2021A-330
I guess it should be 
DIP*W7.62mm*
SOIC*3.9x4.9mm*P1.27mm*


Do you have the ambition to add MCP2022A as well ?

sebastianbasierski:
Thank You for review. 

Of course, I forget about EP:/
330 was taken from supply voltage, i suggested it by similar symbols.
Right, symbols needs to be changed as You suggested.

I will try to fix that today.
MCP2022A will be done also.

Misca1234:
Would not TSSOP-14_4.4x5mm_P0.65mm in Package_SO match what you need ?
(In case you want to change your name)

sebastianbasierski:
It would be fine, i'm wondering why i didn't notice it?
Anyway, i'll provide symbols today evening.
Thank you for help.

sebastianbasierski:
I know why i didn't notice this footprint.
This will be offtop, but i'm not sure if it is a bug or proper behavior.

Let's assume i'm looking for TSSOP-14_4.4x5mm_P0.65mm footprint.


When i click on 'Search by Keyword' i see only those footprints


But when i click 'List All' i see more footprints and in fact i see the one i'm looking for.


Sorry for offtop, but i think this is not proper behavior, but maybe i'm mistaken.
Can You, @Misca1234 or someone else look at it? Or should go to the developers mailing list?

evanshultz:
It's fine to go off topic if needed. :) I see a similar behavior as you. I'm not sure why, but I usually just browse to the library and rarely use the search feature. This is a question for the user community or developers, since here we're only talking about library issues.

sebastianbasierski:
Ok, i'll ask there then. 
Good i'm not alone with this one;)

Misca1234:
If you post it into the develop community, they should check kicad on windows platform as well.
For me, the search does not work at all, it does not list anything (I have anyway always used "List all")

Test to use 
*TSSOP
or
?TSSOP
as search criteria, that might work on your platform.

evanshultz:
@Misca1234 
The package marking is not the MPN, so page 26 should not be used to set the symbol name.

@sebastianbasierski 
- Your datasheet link in the first post is broken.
- Can you post new screenshots of the symbols?
- I'm not too familiar with a LIN bus, but I suppose open collector is the right type for LBUS?

MCP2021A:
- The symbol names should be unique for each package type since they cannot all use the same symbol. You may use a wildcard on the voltage regulator output if you like. Temperature should be a wildcard. Slashes are not allowed. This means `MCP2021A-xxx-xMD` for DFN, `MCP2021A-xxxxP` for DIP, and `MCP2021A-xxxxSN` for SOIC. Or you can make 6 symbols and replace the `xxxx` with `330x` and `500x`, similar to what you started with. Or, if you like, you may only submit a subset of the total symbols for this family. If you want to do all of them that's great but not required.
- Pin 4 of the DFN symbol has a typo in the pin name.
- Pin 8 on both symbols should be named `~FAULT~/TXE` so the overbar isn't on the forward slash.
- Does the datasheet mention the connection of the exposed pad in the DFN package? Often it does, but I don't see it here.

MCP2022A:
- Pin 5 is missing the overbar.
- Same fix for pin 8 above for pin 14.

evanshultz:
The datasheet is at http://ww1.microchip.com/downloads/en/DeviceDoc/20002298C.pdf.

@sebastianbasierski 
I had a feeling I needed to review the footprint filters, and I did. MCP2021A looks fine for both symbols, but for MCP2022A remove the `14*` for all 3 of them. Since all 14 pins are required and present, the pin count isn't needed here.

sebastianbasierski:
Sorry for not answering, i was absent for a few days. 
I'm not sure about removing ``14*`` from filter. I'm thinking about KLC, S4.5, point 3. In MCP2022A symbol there are a hidden pins, so number of pins on symbol is different than number of pins in matching footprints.
>If the pin-count in the schematic does not match the pad-count on the footprint, the footprint filter must include the pad-count of the footprint (see also requirements for footprint filters)

The rest of issues i'll try to address tomorrow.
Thank You for review.

Ratfink:
Where [S4.5](http://kicad-pcb.org/libraries/klc/S4.5/) refers to omitting pins, it means omitting them from the symbol entirely, not merely setting them as invisible.  Invisible pins are still counted when matching symbols to footprints, they just aren't shown on the schematic.  So you should be safe to remove the `14*` from the footprint filter.

sebastianbasierski:
Hi, 
I've just uploaded changes.
I hope there will not be much to fix:)

@evanshultz 
> * Your datasheet link in the first post is broken.

Fixed.

> *  Can you post new screenshots of the symbols?

Sure, screenshots are included at the end of the post.


> *  I'm not too familiar with a LIN bus, but I suppose open collector is the right type for LBUS?

Yes, it is.


>    The symbol names should be unique for each package type since they cannot all use the same symbol. You may use a wildcard on the voltage regulator output if you like. Temperature should be a wildcard. Slashes are not allowed. This means MCP2021A-xxx-xMD for DFN, MCP2021A-xxxxP for DIP, and MCP2021A-xxxxSN for SOIC. Or you can make 6 symbols and replace the xxxx with 330x and 500x, similar to what you started with. Or, if you like, you may only submit a subset of the total symbols for this family. If you want to do all of them that's great but not required.

I used wildcard for temperature.


MCP2021A:

>  * Pin text offset should be 20mil.

Fixed.

> * Pin 4 of the DFN symbol has a typo in the pin name.

Fixed.

> *    Pin 8 on both symbols should be named ~FAULT~/TXE so the overbar isn't on the forward slash.

Fixed.

> *    Does the datasheet mention the connection of the exposed pad in the DFN package? Often it does, but I don't see it here.

I didn't find this information either.




MCP2022A:

> * Pin 5 is missing the overbar.

Fixed.

> * Same fix for pin 8 above for pin 14.

Fixed.


Updated screenshots.
mcp2021a-xxx-xmd
mcp2021a-xxx-xmd

mcp2021a-xxx-xp
mcp2021a-xxx-xp

mcp2021a-xxx-xsn
mcp2021a-xxx-xsn

mcp2022a-xxx-xp
mcp2022a-xxx-xp

mcp2022a-xxx-xsl
mcp2022a-xxx-xsl

mcp2022a-xxx-xst
mcp2022a-xxx-xst

evanshultz:
Above I wrote:

> The symbol names should be unique for each package type since they cannot all use the same symbol. You may use a wildcard on the voltage regulator output if you like. Temperature should be a wildcard. Slashes are not allowed. This means MCP2021A-xxx-xMD for DFN, MCP2021A-xxxxP for DIP, and MCP2021A-xxxxSN for SOIC. Or you can make 6 symbols and replace the xxxx with 330x and 500x, similar to what you started with. Or, if you like, you may only submit a subset of the total symbols for this family. If you want to do all of them that's great but not required.

- Notice only one hyphen in the MPN. While the top left of the table on page 44 of [the MCP2021A datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/20002298C.pdf) would seem like there should be two hyphens, the right side of the table, and [Digi-Key](https://www.digikey.com/products/en?keywords=mcp2021a), and the [Microchip product page](http://www.microchip.com/wwwproducts/en/MCP2021A) only show a single hyphen. You can see I only used a single hyphen above. Please remove the second hyphen from your symbol names.
- Please center VBB on the top above VSS.
- Move the ref des to above the top left of the symbol and the PN above the top right. (NC pins can be anywhere inside the symbol body where they're unlikely to be connected.)

That looks like everything to me. Thanks!

evanshultz:
Oops. Clicked wrong button. Sorry.

sebastianbasierski:
@evanshultz 
> Notice only one hyphen in the MPN. While the top left of the table on page 44 of the MCP2021A datasheet would seem like there should be two hyphens, the right side of the table, and Digi-Key, and the Microchip product page only show a single hyphen. You can see I only used a single hyphen above. Please remove the second hyphen from your symbol names.

```MCP2021A-330E/MD```
There is slash between temperature and package symbol, and KLC states that slash is forbidden. 
So i thought i should replaced it with hyphen symbol.

> Please center VBB on the top above VSS.

Fixed.

> Move the ref des to above the top left of the symbol and the PN above the top right. (NC pins can be anywhere inside the symbol body where they're unlikely to be connected.)

Fixed.

Updated screenshots of symbols.
mcp2021a-xxxxMD
mcp2021a-xxxxmd

mcp2021a-xxxxP
mcp2021a-xxxxp

mcp2021a-xxxxSN
mcp2021a-xxxxsn

mcp2022a-xxxxP
mcp2022a-xxxxp

mcp2022a-xxxxSL
mcp2022a-xxxxsl

mcp2022a-xxxxST
mcp2022a-xxxxst

evanshultz:
I see. For Microchip parts in the library we have ignored the slash. You got it now. Thanks!

The VBB pin looks nice, but now the PN is laying on top of it and the VREG pin. I can't merge with that overlapping stuff. Perhaps move the VREG pin to the right of VBB and then push the text slightly right (and ensure it is left-justified)? You may do something a bit different but please correct the overlap.

sebastianbasierski:
Hi, 
Right, that looked messy:/
Now should be better.

mcp2021a-xxxxmd
mcp2021a-xxxxp
mcp2021a-xxxxsn
mcp2022a-xxxxp
mcp2022a-xxxxsl
mcp2022a-xxxxst

evanshultz:
I just realized that the EP for the DFN package of MCP2021A is Unspecified type and we didn't figure out how to connect that pad. Let me ask Microchip and see if we can get a quick response.

evanshultz:
@sebastianbasierski 
Microchip says to tie the EP to VSS. Please stack the pins. Can you make that last change?

sebastianbasierski:
Hi, sure.
I'm attaching only updated symbols

mcp2022a-xxxxp
mcp2022a-xxxxp

mcp2022a-xxxxsl
mcp2022a-xxxxsl

mcp2022a-xxxxst
mcp2022a-xxxxst

sebastianbasierski:
Hi, sure.
I'm attaching only updated symbols

mcp2022a-xxxxp
mcp2022a-xxxxp

mcp2022a-xxxxsl
mcp2022a-xxxxsl

mcp2022a-xxxxst
mcp2022a-xxxxst

evanshultz:
There was some confusion This only applies to the DFN package of MCP2021A. The EP pin can be named VSS and stacked on the existing VSS pin.

No MCP2022A symbols of MCP2021A symbols in other packages needed to be touched. Can you revert those changes?

sebastianbasierski:
Sorry:/

mcp2021a-xxxxmd
mcp2021a-xxxxmd

I see Travis throws failure on pin name, but in KLC S4.3 there is 

>  Special Case: Pins of electrical type [Output, Power Output, Power Input] are special cases. Connecting these pins together causes ERC (electrical rule check) errors and must be avoided.
For pin stacks using these pin types:
    Visible pin must be set to correct electrical type
    All invisible pins must have electrical type set to Passive

So, following this point forbids me to follow S4.4. Am i right?

evanshultz:
You can place the two VSS pins on top of each other. The pin types look correct so it should be OK and resolve the error (when power pins of the same name are stacked all but one can be Passive type).

sebastianbasierski:
Ok, i moved pin one on another.

mcp2021a-xxxxmd
mcp2021a-xxxxmd

evanshultz:
Thanks for this nice contribution! I hope you're willing to submit more symbols and footprints in the future!

It looks like you've reverted some commits but I'm trusting nothing was un-done and I don't need to check everything all over again.

sebastianbasierski:
I reverted only changes from last night - those NC pins which i changed by mistake.
I'll add some more symbols, maybe tomorrow even;)
Thank You for your time, it was good lesson for me!

PR #376, opened 2018-03-12
added mmbzxx diode

sebastianbasierski:
Added double zener diode MMBZXX.

Datasheet: [http://www.onsemi.com/PowerSolutions/product.do?id=MMBZ27V](http://www.onsemi.com/PowerSolutions/product.do?id=MMBZ27V)

Screenshot
mmbzxx

Matching footprint: [https://github.com/KiCad/kicad-footprints/blob/master/Package_TO_SOT_SMD.pretty/SOT-23.kicad_mod](https://github.com/KiCad/kicad-footprints/blob/master/Package_TO_SOT_SMD.pretty/SOT-23.kicad_mod)


Misca1234:
This is a dead minor detail, but I read elsewhere in a thread that the wild card should always be '*' and never '?'

poeschlr:
@Misca1234 it is not as clear cut as that. But you are right most other symbols use '*' instead of '?' everywhere. I am not sure what other symbols similar to this one use. (Or more general: What other symbols that reference sot packages use.) If they use '?' then we can accept it for this one as well.

Normally footprint filters do not include the field after the package type. As the next field is the pin count. This is why in a symbol referencing a SOIC package there is a '*' after SOIC. 

This is another thing where we could use some more thinking. (So after the v5 release we might want to look into it. Maybe we should discuss this when it is clearer what the new file format will bring. Maybe we will get more powerful filter options.)

jkriege2:
Sometimes it make sense to use a `?` (matches one or no character) instead of `*` (matches any umber of characters or none). E.g. for `SOT-23` using `SOT?23*` will match any SOT-23-variant, but will NOT match `SOT-223`! I think that's the reason why it's used here. I used that scheme for many other SOT-23-symbols!

sebastianbasierski:
I've seen filters like this used in various parts.
But there is another question, if '?' exists why not use it?

evanshultz:
We have added fields, or wanted to add fields, and using an asterisk gives the most flexibility. So a question mark is used if required, but otherwise an asterisk. At least that's what I've seen done.

sebastianbasierski:
I think using '?' here in this case(filter) is better solution than '*'. @jkriege2 is on my side;)

sebastianbasierski:
Hi.
Can anyone make a decision, please?
Should i fix something in this patch or is it ready to be merged?

evanshultz:
I agree. Just a couple small changes, please.
1. Could you change the description to `Dual Zener Diode, Common Anode, 225mW, SOT-23`?
1. Please add `dual` as a keyword.
1. Use `xx` as a wildcard instead of `XX`.

sebastianbasierski:
Done.

evanshultz:
Thank you!

Issue #352, opened 2018-03-05
Reorganization of RF libraries


evanshultz:
I find symbols at https://github.com/KiCad/kicad-symbols/blob/master/RF_Bluetooth.lib. Another symbol that hasn't been merged is at https://github.com/KiCad/kicad-symbols/pull/75.

However, they should be in an MCU library. I do not see one for this manufacturer present, but we should create MCU_Nordic.lib to hold them once the un-merged PR above is resolved.

Ratfink:
If we're moving general-purpose MCUs with built-in radios to the MCU libraries, there's more to move than just Nordic Semi parts.  Off the top of my head:

* RF_WiFi.lib: ESP8266EX → new library MCU_Espressif.lib

And possibly:

* RF_Wifi.lib: ESP-12E → MCU_Module.lib
* RF_Wifi.lib: Particle_P1 → MCU_Module.lib

evanshultz:
Right. Even if there are BT and/or WiFi and/or other peripherals, the core function of the part is an MCU so it should be in a library that is obviously for MCUs.

The above all sounds fine to me, and thanks for checking it out. Normally I'd say we should have a reasonable number of symbols to support a new library. But seeing as how we don't have a vanilla MCU.lib, perhaps all MCU are to be segregated by manufacturer regardless of the number of symbols?

evanshultz:
A few more inspired by https://github.com/KiCad/kicad-symbols/pull/349.

Move to RF_Module:
- MM002 (and update URL)
- TD1205
- TD1208
- UL865 (and update URL)
- iM880A
- iM880B

Move to RF_Amplifier (amplification and attenuation are essentially the same):
- MAADSS0008

evanshultz:
PR submitted at https://github.com/KiCad/kicad-symbols/pull/379.

PR #379, opened 2018-03-13
Move parts within RF libs and create new MCU libs

evanshultz:
Moved symbols around and created new libs as discussed at https://github.com/KiCad/kicad-symbols/issues/352. I believe I got everything at that issue, but if I missed something or there is more to do please let me know.

While I was at it, I polished up the old Nordic symbol and the Espressif one visually (shortened pin length to 100mil, pin text offset to 20mil, arranged pins like other parts in the same lib, etc.). I didn't make any functional changes. Here they are now (not that it really matters, but the text is actually in the middle of the symbol but it gets auto-placed when the symbol is dropped into a schematic):
image

I didn't move the MAADS part because of another outstanding PR that is creating the RF_Amplifier library.

There are honestly so many of these RF libs that it's hard to figure out what goes where. Most of the things I found are modules, but they weren't in the module library. Maybe this is enough for now and we take another look at this after v5 is released?

I also corrected the name of the Nordic part in RF_Bluetooth.lib to have a lowercase "n".

@Ratfink 
Would you mind looking at this since you had some input into it? Anybody is welcome, though, and I'd like to get this merged quickly because all these changes are likely to result in branch conflicts fast.

------------


Ratfink:
I haven't really looked at this yet, but I just noticed a mistake from your screenshot.  Pin 26 of the ESP8266EX is 200 mil, please fix its length.

evanshultz:
@Ratfink 
My... I must be going blind. Fixed now. Thanks for taking a look!

Ratfink:
Looks good.  I see there are still a few problems with some symbols, but since this is almost entirely just a reorganization I don't think those have to be fixed in this PR.  I'll merge this in a few hours unless anyone else has something to say.

PR #386, opened 2018-03-16
Power supervisor

daryl-t:
**MAX6355**
_{S4.4, S5.2, S6.3}_

[Datasheet](http://datasheets.maximintegrated.com/en/ds/MAX6351-MAX6360.pdf)

max6355

**MAX691**
_{S3.1, Off-Center, S3.2, S4.1, S4.2, S4.4, S5.2, S6.3}_

[Datasheet](https://datasheets.maximintegrated.com/en/ds/MAX690-MAX695.pdf)

max691

**TPS3831**
_{S3.1, Off-Center, S3.6}_

[Datasheet](http://www.ti.com/lit/ds/sbvs193d/sbvs193d.pdf)

tps3831

**TPS3839DBZ**
_{S3.1, S3.6}_

[Datasheet](http://www.ti.com/lit/ds/sbvs193d/sbvs193d.pdf)

tps3839dbz

**TPS3839DQN**
_{S3.1, S3.6}_

[Datasheet](http://www.ti.com/lit/ds/sbvs193d/sbvs193d.pdf)

tps3839dqn

**MC33064D & MC34064D**
_{S4.1, S5.2, S6.3 & Incorrect Aliases Separated}_

[Datasheet](https://www.onsemi.com/pub/Collateral/MC34064-D.PDF)

mc33064d
_NOTE: MC33064D is -40C to +85C SOIC-8_

mc34064d
_NOTE: MC34064D is 0C to +70C SOIC-8_

mc33064p
_NOTE: MC33064P is -40C to +85C TO-92_

mc34064p
_NOTE: MC34064P is 0C to +70C TO-92_

------------


evanshultz:
@daryl-t 
Thanks for making a contribution to KiCad!

General:
- Your datasheet links above are all broken.
- The package type should be at the end of the description. For example, MAX6355's description should end with `SOT-23-6`.
- Footprint filters don't need pin numbers if all pins are present (including NC pins). So for MAX6355, it should be just `SOT?23*`.
- If a pin has only one function, and is active low, only the leading tilde is needed and not the trailing one. For example, pins 1 and 3 of MAX6355.

MAX6355:
- You can remove the `/Triple` from the description because this specific part is only a dual.
- ~RST should be Open Collector type (we don't have open drain).
- Because it goes directly to a power supply output, and it also the power supply for the IC, VCC1 and VCC2 might best be Power Input type.

MAX691:
- This part comes in two packages. The footprint filter should include the SO package. Add `SOIC-16_7.5x10.3mm*P1.27mm*` and remove the default footprint. Also change the packages in the description to `DIP-16/SOIC-16`. Or add a second symbol, one for each footprint, and then give them the MPN suffix to differentiate them.
- Please put inputs on the left and outputs on the right (pin 2 on the right side, pin 11 on the left side).
- Spell out RESET in both pins 15 and 16, as shown in the datasheet.
- It looks to me like all outputs are push-pull. Please check and confirm.

TPS3831:
- The footprint prefix should be `Texas_`. This one is wrong. Please fix it hee and I'll fix the footprint.
- Footprint filter should be `Texas*X2SON*1x1mm*P0.65mm*`. Change the question mark to an asterisk.

TPS3839DQN:
- The footprint prefix should be `Texas_`. This one is wrong. Please fix it hee and I'll fix the footprint.
- Footprint filter should be `Texas*X2SON*1x1mm*P0.65mm*`. Change the question mark to an asterisk.
- Move the NC pin inside the part body. Typically the pin will be on the edge of the part.
- Move pin 5 to the bottom of the symbol.

MC3x064:
- Why did you separate these D and P symbols? Since they share the same SOIC and TO92 footprint, and are pin-compatible, they should use an ALIAS.
- Pin numbers on SOIC are wrong, NC pins are missing, etc. Answer the item above before reviewing that more.

@poeschlr @jkriege2 @Ratfink 
Are we allowing PRs that inflate symbols to eliminate off-center warnings? Some of the above only or predominantly enlarge the symbol. But we have many symbols that are off center and it doesn't seem to be important. For these small ICs, like linear regs, they are often off center when they're built nicely. I don't find off center symbols, especially small one like these, to be a problem and I'd rather keep them smaller even if there's a warning. Besides, the symbols were working before and this will disrupt any schematics where they were placed. (If there was a true improvement I wouldn't concern myself with this but I don't see the value in making these bigger to avoid a warning.)

Ratfink:
Personally, I'm opposed to enlarging existing symbols simply to silence the slightly-off-center warning, for the same reasons you seem to be.  I wouldn't find e.g. shifting the existing symbol for TPS3839DBZ to the left by 100 mil nearly as distasteful, but your point about disrupting schematics would still stand (though, if people aren't distributing their projects with the cache-lib, I suppose that's their fault and not ours).

daryl-t:
Sorry for the error with the broken links - now fixed. I'm new to this got the formatting wrong. I will also amend the other recommendations.

evanshultz:
Just a question: What happened here? There were a ton of commits that were apparently reverted??

We're requesting above that you don't enlarge symbols just to meet that warning. And I posted a number of requests. Will you be able to make those fixes so this PR can be merged?

evanshultz:
@daryl-t 
This PR is now in conflict with master. And we discussed above not wanting to make changes that simply enlarge symbols. Please advise if you will be fixing and updating this PR.

Misca1234:
@daryl-t 
Are you still interested in working with this push ?

Misca1234:
This push have been overtaken, see
https://github.com/KiCad/kicad-symbols/pull/1468

Issue #387, opened 2018-03-16
Fixing Atmel (now Microchip) MCU libs


evanshultz:
@fauxpark has done a ton of work already. Here is what is left to consider.

 - [ ] Shrink and align the boxes
 - [ ] Move obsolete symbols to obsolete/
- [ ] Pin types for all pins, especially the special-function ones that aren't power or GPIO.
- [ ] Pin placement when you shrink symbols. Should all GPIO be on the left side, right side, distributed as evenly as possible, etc.
- [ ] Do all XTAL pins go on the same side? How far apart are they so a crystal symbol fits there nicely.
- [ ] Pin text offset should be 20mil.
- [ ] Place power and ground pins next to each other or leave a space between them (like the GND and EP pins, for example).
- [ ] Pin length should be 100mil if pin numbers have <3 characters (usually 99 pins or less). All NC pins can be 100mil.
- [ ] Place refdes and MPN inside the symbol outline if it's large enough?
- [ ] Renam AT89S2051/S4051-PU and AT89S2051/S4051-SU symbols to have a "24" in them (datasheet indicates that this is the clock speed i.e. the AT89Sxxxx versions run at 24MHz). So rename as AT89S2051/S4051-24PU and AT89S2051/S4051-24SU.
- [ ] Review/fix description, keywords, and datasheet for all symbols.

See https://github.com/KiCad/kicad-symbols/pull/373 and https://github.com/KiCad/kicad-symbols/pull/382 for some previous discussion on those points. To keep it all together, though, let's move all discussion here.

fauxpark:
The pin name offsets should already be done.

RE renaming the 8051 symbols  - should we just remove that bit from all symbols instead, or add more aliases for clock speed? Right now it seems there is mostly just one set of symbols defined for each device.

evanshultz:
@fauxpark 
Now that https://github.com/KiCad/kicad-symbols/pull/389 has been merged, I checked off a few more items from the list above. Please confirm that the checked off items are correct, and also if there are any more items to add. If you're willing to keep going it would be great!

fauxpark:
Second point (obsolete symbols) is finally done 👍  thanks @poeschlr.

I'll look at pin types next.

evanshultz:
Just merged https://github.com/KiCad/kicad-symbols/pull/702, which counts towards the 4th item above.

Elijahg:
[...]

PR #389, opened 2018-03-17
More Atmelchip fixes

fauxpark:
Relates to #387

For all symbols with < 100 pins (and NC pins on 100-pin symbols), the pin length has been reduced to the default 100mil, and body sizes adjusted accordingly. VCC and GND are now always centered on 0, with other power pins like AVCC, VBAT etc. 100mil to the sides. Also, I removed the ATmega1261 aliases of the 1281 symbols and ATtiny10-P/S -- Google turns up absolutely nothing on them.


evanshultz:
Needed DFN, QFN, and BGA footprints should now be added.

fauxpark:
BGA symbols added, and now we have the nice green checkmark :) I think so far I've ticked off a good number of things from #387.

fauxpark:
All symbols now have their clock speed in the descriptions, and I've added some missing aliases for different speeds.

@evanshultz I'm up to 16 commits here now. I'll hold off on any more so you can review when possible.

evanshultz:
Sure. Let me take a look soon. Hopefully today.

evanshultz:
I want to let you know I haven't forgotten about you. I'm about halfway through these commits and it will probably take a couple days to finish this.

Just for clarity about what has happened with [ATtiny10](https://www.microchip.com/wwwproducts/en/ATtiny10), only SO8 and DIP8 symbols have been removed as those packages don't exist (now, anyway).

fauxpark:
No worries.

I Googled the names of all the symbols and the ATtiny10-P/S and ATmega1261 were the only ones that didn't turn up a datasheet. The existing ATtiny10 datasheet also doesn't mention -P or -S variants, and from looking at the revision history, it seems it never did.

evanshultz:
Here are the other things I noted when reviewing. There's so much that I couldn't check everything manually. I took a look at each commit, spot checked a few symbols in depth, opened the libs in KiCad to see if any errors were created, and finally took a cursory look at symbols and descriptions in KiCad. For all items I may only list a single PN but please look at all parts in the datasheet and in the family for each item. Also, consider these as questions and suggestions and not directives:
- I don't see ATtiny2313V, ATtiny24V/44V/84V, ATmega128L, ATmega1280V at all.
- ATmega163 seems to be obsolete.
- I don't see ATmega165 with A, PA, V, and PV suffixes.
- Can you use Microchip datasheets instead of Digi-Key? Is there a reason not to use an official datasheet? Did you only do that if a part is obsolete?
- Many descriptions have no noun. For example, `ATmega1281-8AU` is `8MHz, 128kB Flash, 8kB SRAM, 4kB EEPROM, JTAG, TQFP-64`. There's not the datasheet description to `8-bit Atmel Microcontroller with 16/32/64KB In-System Programmable Flash` or even just `MCU`. But, I suppose these library names are descriptive enough and it's just me seeing a string of verbs and no noun that bothers me. So it's OK.
- I don't see ATtiny48/88 available in BGA package but they're in your library.
- ATxmega128A1, for example, is available in 2 different BGA packages but the description doesn't indicate this. I wonder if the exact type of BGA package should be shown, at least if there are multiple?
- Some of the BGA pad numbers don't exactly match the datasheet. For example, the `ATxmega16D4-CU` symbol has pad A4 as `PR1/XTAL1` but the datasheet shows just `PR1`.
- Many of the symbols are overly large (and some just big in general!). That may be because the pin names are now shorter. Is this something you're planning to address?

I can't find anything else except what's above. I think we're close!

By the way, it seems you did all this by hand. You're a machine! Using a symbol generator script would be much, much easier for library overhauls of this magnitude in the future.

fauxpark:
- The V and L suffixes are a mystery to me. I was under the impression that they didn't matter (like the R suffix for tape & reel), but I just Googled and didn't get Digikey and Mouser results until I added the V to `ATmega1280-8AU`. I'll get on that.
 - I haven't gotten around to picking out the obsolete symbols and moving them yet :)
 - I've got ATmega165/A/P/PA here, but see the first point.
 - I used Microchip datasheets as much as possible - I don't see a page or list of legacy ICs on their site.
 - Wasn't sure what to do with the description format so I left it as is. But I see what you mean. Perhaps something like "AVR MCU with ..."? I don't know if the length is worth caring about.
 - http://ww1.microchip.com/downloads/en/DeviceDoc/doc8008.pdf#page=283 lists -CCU (BGA-32) packages for both ATtiny48/88.
 - I can add "7x7/9x9mm BGA-100" to the descriptions of those particular symbols if you'd prefer.
 - Some of the more "important" alternate pin functions like RESET and AREF are kept in the pin name. In this case the datasheet (and some others) specifies PR0/1 as also being XTAL1/2.
 - Yeah, I still haven't shrunk them yet. I'm unsure what the threshold would be for placing refdes/PN in the centre would be, but I'm leaning towards 100 pin symbols only.

evanshultz:
- I see both ATmega1280 and 1280V at http://www.microchip.com/wwwproducts/en/atmega1280. I also see Digi-Key carries both at https://www.digikey.com/products/en/integrated-circuits-ics/embedded-microcontrollers/685?k=atmega1280. And both sites also have ATmega128 and 128L. From that it seems that with and without "V" and "L" are valid PNs that can be included in KiCad.
- Datasheets: Ahh... alright. if Microchip doesn't have one then it's OK. Hopefully that's only the case for parts that will eventually move to the obsolete folder.
- I see most MCU libraries use the word "Microcontroller" in them, but some don't. It varies. So leaving it like this is OK. (We can standardize them all together, if desired, at a later date.)
- ATtiny48/88 BGA: The datasheet hasn't been updated in 7 years, and the product page at https://www.microchip.com/wwwproducts/en/ATtiny48 doesn't show BGA packages. Digi-Key doesn't carry BGA packages for these parts either. I suspect that package has been deprecated.
- Yes, let's identify them uniquely. Could you perhaps use the BGA type (TFBGA, UBGA, etc.) so it matches the type of BGA in the footprint name? Do you think the BGA size is better?
- Alright. It would be good to identify that "important" means for all MCU libs, but we don't need to figure that out to get these symbols merged.

Obsoleting and shrinking symbols is fine to wait. Just checking on your schedule.

fauxpark:
It appears the V and L have something to do with clock speed and/or operating voltage: the ATmega1280's are all 16MHz and run at 2.7-5.5V, and the ATmega1280**V**'s are all 8MHz, at 1.8-5.5V. So my latest batch of commits should cover this.

I'll remove the tiny48/88 BGAs and change the BGA descriptions.

evanshultz:
This all looks fine to me. Thanks so much!

I'll leave this open for a little while longer in case you want to take a look. If I don't see any new comments in a day or two, I'll merge.

evanshultz:
Thanks again!

Issue #232, opened 2018-01-28
SPI interface pins of a lot of symbols have the wrong electrical type.


poeschlr:
The following is only valid for symbols of devices that do not use the same pins for other functions.

For master devices the electrical types should be as follows:
- CLK = output
- Chip select (CS or ~CS) = output
- MOSI = output
- MISO = input

For slave devices:
- CLK = input
- Chip select (CS or ~CS) = input
- MOSI = input
- MISO = tri-state (Currently a lot of devices have the MISO pin as output. This would result in erroneous ERC messages if more than one slave is connected to the same master interface.)

jkriege2:
Two comments on that:
1. When we correct, we should take care to check for the existence of a CS-pin ... if it's not there it might well be that MISO is a true output
2. I checked the SDA/SCL pins of  I2C and made a PR to fix a few bugs: https://github.com/KiCad/kicad-symbols/pull/233

bobc:
I have a PR in progress for Memory_EEPROM (#222), I will it update for MISO pin types.

I would be a bit wary of doing global edits without checking datasheets, it could lead to more errors then we solve like that (well, I already made that mistake once!).

In all the libs I have reworked so far, they all have parts with pin types that are just completely wrong (e.g. power pins set to power out instead of power in, outputs set as inputs etc). I guess it might help to do some global edits, but most of the bad parts need a complete overhaul anyway.

jkriege2:
Yes ... For the I2C it was rather quick (with a text-editor global search you quickly find the ones with another type than the expected one and can check them ine-by-one then ...), but it's going to be checking each one separately in the end ... I fear

PR #392, opened 2018-03-17
RF.lib: Add CC2500 schematic component

woo2:
This adds the schematic component for the TI CC2500 transceiver.

See: [cc2500.pdf](http://www.ti.com/lit/ds/symlink/cc2500.pdf)
cc2500

------------


Ratfink:
Thank you for the contribution!

It looks like you created this using the stable 4.0.7 release.  Unfortunately, RF.lib has already been converted to the new format used by the KiCad 5 nightlies, and it seems KiCad 4 didn't like that and clobbered the .dcm file when you saved it.  It gives me no pleasure to do this, but I have to ask that you recreate this symbol with a recent nightly build of KiCad 5 before this can be accepted.

evanshultz:
It would also be possible to manually modify RF.lib and RF.dcm. This can be done directly in GitHub or on your local computer and then committing.

woo2:
5 looks really nice! I figured something was off with that dcm diff. To fix the problem, I exported the schematic in 4.0.7, reverted my commit, imported the part in 5 and committed again. Thanks for the help!
cc2500

Ratfink:
Thanks for fixing the file format problems!  I have a few notes:

* Please set the pin name position offset to 20 mil.
* Unless I misunderstand something, SO should be electrical type Tri-state, since it can be disabled using the ~CS input.  See #232.
* As far as I can tell, the datasheet refers to pin 18 only as DGUARD, not VDD_GUARD.  Please set the pin name to match the datasheet.
* Ideally I'd like to see the SO and SI pins next to one another, unless you have a good reason to do otherwise.
* If you move RBIAS to the right side next to DCOUPL, you could move XOSC_Q2 down the old location of RBIAS so that the two XOSC pins have the same spacing as a crystal symbol.  This would make drawing schematics with this chip a bit easier.
* The datasheet says that GDO2 is just a digital output, so please set its electrical type to Output.

I'm not sure what to do about the footprint name and filter.  What you currently have set doesn't match the datasheet's recommended footprint.  The datasheet says that this part is made in TI's QFN-20 "RGP" package, which they have a drawing of [here](http://www.ti.com/lit/ml/mpqf126f/mpqf126f.pdf).  We have a footprint for another part that uses an RGP package, but the footprint (Texas_S-PVQFN-N20_EP2.7x2.7mm, corresponding drawing [here (page 25)](http://www.ti.com/lit/ds/symlink/drv8662.pdf)) doesn't match this part's recommended footprint either, in EP size or pad dimensions.  I'm fairly certain this part needs its own footprint, but I don't know how to name it.  @evanshultz @poeschlr @jkriege2 any input on this?

evanshultz:
Does the datasheet mention SO being tri-state? Generally it's true that MISO should be tri-state in standard, vanilla SPI, but I agree with the comments in the linked post that it may not always be. Or the interface for a wireless product may use the SPI protocol but only support a single slave. Do we know???

Yeah, TI likes to use the same name for multiple footprints. Annoying. I asked them about this one.

Just for confirmation, http://www.ti.com/product/CC2500/samplebuy appears to show that the RTK package is obsolete, leaving only the RGP one. The datasheet apparently didn't get the RTK page removed.

woo2:
I'll work on the footprint when I have some free time this weekend. I found [this](https://e2e.ti.com/support/wireless_connectivity/low_power_rf_tools/f/155/t/584901), RTK has the same footprint as RGP, it seems. I plan to use Figure 33.1 on page 81 for the footprint. In terms of naming, Texas_S-PVQFN-N20_EP2.4x2.4mm seems right, no?

With respect to tri-state on SO, page 61 of the datasheet does say that the default for GDO1 is 3-state. I'll fix that and make the other DO pins proper outputs. Only ATEST should be bidirectional.

xl0:
@woo2 , I think I've got the right footprint here, from CC1101, see footprint PR #453 .

woo2:
I'm not really sure what to do about this. The CC1101 and CC2500 both use the S-PVQFN-N20 spec, with a 2.4mm thermal pad, but they have different pad lengths. the 2500 is .76mm and the 1101 is .95mm. This isn't expressed anywhere in the footprint name, and it is the only difference in the footprints as far as I can tell.

evanshultz:
Can you clarify where you got those numbers? I see page 103 of http://www.ti.com/lit/ds/symlink/cc1101.pdf shows 0.95mm pads but I don't see a footprint drawing (just a package drawing) in http://www.ti.com/lit/ds/symlink/cc2500.pdf.

woo2:
See page 85 of the CC2500 doc. Section 33.1.

Misca1234:
@woo2 
Hi
1. Could you change your foot print to Texas_S-PVQFN-N20_EP2.4x2.4mm
2. Solve the merge conflict ?

Misca1234:
Thx

PR #394, opened 2018-03-18
MAX2679 & BGA28xx RF Amplifier

calebns:
WLP-4 GPS/GNSS gain block
https://datasheets.maximintegrated.com/en/ds/MAX2679-MAX2679B.pdf

Footprint:
https://github.com/KiCad/kicad-footprints/pull/437

image

------------


calebns:
Added BGA28xx RF amplifier family
https://www.nxp.com/docs/en/data-sheet/BGA2800.pdf

image

evanshultz:
MAX2679:
- Can you put the amount of gain in the description?
- Can you tweak your description to match the datasheet description of `GPS/GNSS Ultra-Low Current
Low-Noise Amplifier`?
- Is this footprint MAXIM-specific? And is it described enough? For example, https://datasheets.maximintegrated.com/en/ds/MAX40002-MAX40005.pdf mentions a WLP-4 but in a different size. And https://www.ablicinc.com/en/doc/datasheet/temperature_sensor/S58LM20A_E.pdf is also a WLP-4 in a slightly different size. I'll re-post this in the footprint PR, but it's here too so this doesn't get merged without knowing the footprint is correct.

BGA280x:
- Can you start the description with `MMIC wideband amplifier`?
- `SOT-363` is a 6 pin package so you can remove the trailing `-6`.
- Can you format the gain part of the description like `+20.2dB @ 2150MHz` for BGA2800, since the gain varies a bit over frequency? Perhaps use the rating on page 1 of the datasheet?
- BGA2801 is rated up to 2.2GHz, not 0.75GHz.
- BGA2815 is rated for DC-2.2GHZ operation, not 0.25-2.15GHz.
- BGA2817 says 3.3V on page 1 of the datasheet and not a voltage range.
- You're missing BGA2850, BGA2865, BGA2867, and BGA2874, if you want to add them.
- BGA2870 mentions a voltage of 2.5V, not a range.
- It seems that pins 2, 4, and 5 must always be connected together. Why they have two different pin names is a mystery to me. Please confirm just to make sure I understand correctly.

calebns:
All above should be fixed.

Yes, both GND1 and GND2 must be tied together. I have no clue why the vendor decided to separate them on the diagram.

evanshultz:
Excellent! This looks good. I'd like to wait for the footprint to be merged but this is otherwise ready to go!

Misca1234:
`Yes, both GND1 and GND2 must be tied together. I have no clue why the vendor decided to separate them on the diagram.`

We would never know without the internal schematic, one reason could be that the manufacture wants the user to connect all legs so an untied leg would not act like an antenna and channel disturbance into the silicon.

Another reason could be that some customers gets more information about the chip and therefore can do things that the unaware customer can not do.

evanshultz:
Thanks for pinging me about this and also for submitting the symbols!

Issue #393, opened 2018-03-18
MCP 23016 has wrong pins in the library


un23:
Hello,

in legacy/microchip.lib, the pinout of the 23016 is not completely correct. Pin 10 is mentioned twice, but the ground pin 8 is not connected at all. This is also wrong in the 4.0 repository.

Regards
  un23

diff --git legacy/microchip.lib legacy/microchip.lib
index ed1053e..bcc2d01 100644
--- legacy/microchip.lib
+++ legacy/microchip.lib
@@ -162,8 +162,8 @@ X GPB2 4 -500 600 100 R 50 50 1 1 B
 X GPB3 5 -500 500 100 R 50 50 1 1 B
 X ~INT 6 500 300 100 L 50 50 1 1 O
 X GPB4 7 -500 400 100 R 50 50 1 1 B
+X GND 8 0 -1000 100 U 50 50 1 1 W
 X CLK 9 500 800 100 L 50 50 1 1 I
-X GND 10 0 -1000 100 U 50 50 1 1 W
 X TP 10 500 700 100 L 50 50 1 1 O N
 X VCC 20 0 1000 100 D 50 50 1 1 W
 X GPB5 11 -500 300 100 R 50 50 1 1 B

Ratfink:
Fixed in #396.

PR #396, opened 2018-03-19
Fix and update MCP23016

evanshultz:
From https://github.com/KiCad/kicad-symbols/issues/393.

Changes:
- Removed duplicate pin 10.
- Add pin 8.
- Change pin name `GND` to `VSS` to match datasheet.
- Change pin name `VCC` to `VDD` to match datasheet.
- Corrected footprint filters.

I did not change the GPIO pin names to match the datasheet. I did not stack ground pins.

Lastly, I'm not sure this part should be in the legacy folder. It's NRND, but it's still available from Microchip and on DigiKey. I don't think it's a big deal, though, so I left it.

------------


Ratfink:
Looks good, thanks for fixing this bug!

PR #397, opened 2018-03-19
move AS334x VHFT to input pin group

robkam:
Move Input type pin VHFT on AS334x from Passive type group to Input type group.


robkam:
capture

evanshultz:
@robkam 
What was the reason to move pins? Did you find when building the schematic that the new pin locations are better? If so, can you share a snippet of the schematic to demonstrate?

I agree VS should be Passive type. What about SCALE1 and SCALE2?

robkam:
@evanshultz no schematic yet .. it's that the 3340 and 3345 pin groups currently don't match each other and that I've tried to keep the other AS33xx IC pins in numerical order. 

I'd guess SCALE1 and SCALE2 are passive, does that makes VHFT passive as well?

evanshultz:
I think it's fine the way it is.

robkam:
Thanks

PR #400, opened 2018-03-21
Fixed description and pin name of the KSZ9031RNXCA Ethernet PHY

vlunyov:
Fixed two issues:
1. Description (easy to check);
2. Pin name of the pin 8 was doubled with pin 7. The correct pin name can be found on page 5: http://ww1.microchip.com/downloads/en/DeviceDoc/00002117C.pdf


Ratfink:
Duplicate of #401.

PR #404, opened 2018-03-21
Added PIC24FJ64GB004

pa0los:
Product description:
https://www.microchip.com/wwwproducts/en/PIC24FJ64GB004

Datasheet:
http://ww1.microchip.com/downloads/en/DeviceDoc/39940d.pdf

image


Misca1234:
Hi

As a general rule the pin names should be the same as described in the datasheet but for CPU's, such as this one, it is common style to use the shortest possible, 
look at this push as an example
that is, just use RB8, RB9 etc, unless the pin have some very specific meaning that is important ot point out, such as the MCLR pin

I also suggest to use a shorter pin length than the one you use, otherwise the pins "look so long"

Misca1234:
@pa0los 
See above comment

antoniovazquezblanco:
@pa0los ping! :)

antoniovazquezblanco:
Please don't close the PR. We will take over it when we find a moment if you don't want to continue with it. Thanks

Misca1234:
When 
https://github.com/KiCad/kicad-symbols/pull/1266

Is in I can re make this

Misca1234:
@antoniovazquezblanco 
I overtake this

PR #409, opened 2018-03-22
Fixed broken install prefix setting

metacollin:
This does not work for macOS users.  It will try to install to /usr/local/share/kicad (which does not exist) and fails without installing anything.

This is caused by how CMAKE_INSTALL_PREFIX is being handled.  Without using FORCE because the project ( ) command, whose job is to setup the basic variables needed, will have already set CMAKE_INSTALL_PREFIX.  Some older versions of cmake with an incomplete implementation of project don't set the install prefix, allowing this to accidentally work.  That is not the intended behavior however, nor does it work with recent versions of cmake.

This commit fixes the problem by using the canonical way to handle the CMAKE_INSTALL_PREFIX according to the cmake developers: https://public.kitware.com/pipermail/cmake/2010-December/041135.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,4 +57,3 @@ install(FILES ${lib_files}
 install(FILES sym-lib-table
         DESTINATION ${KICAD_TEMPLATE}
         COMPONENT resources)
-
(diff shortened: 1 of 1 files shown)


poeschlr:
Same change for footprint repo: https://github.com/KiCad/kicad-footprints/pull/450

antoniovazquezblanco:
Following https://github.com/KiCad/kicad-footprints/pull/450 this PR does not solve the issue it intended. I will close the current PR for this reason.

If the issue is still present please note it here or open a new issue so that we will keep track of it.

PR #332, opened 2018-02-28
Various symbol fixes

evanshultz:
Replaces https://github.com/KiCad/kicad-symbols/pull/325.

Update [ADC128D818 ](http://www.ti.com/lit/ds/symlink/adc128d818.pdf)and several fixes in the FET library. New part is [FDS2734](http://www.onsemi.com/pub/Collateral/FDS2734-D.pdf).

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #410, opened 2018-03-22
Transistor_FET: Fix 200mil pins, add new symbols

evanshultz:
It seems at https://github.com/KiCad/kicad-symbols/pull/332 there are still a few FET symbols with 100mil pin length. So I fixed those.

I also found that the IRF7324 symbol had some weird graphical artifacts and many elements were for each unit instead of being in common. So I fixed/improved that.

And finally I added the following symbols:
- http://www.ti.com/lit/ds/symlink/lmg5200.pdf
- http://www.ti.com/lit/ds/snosd10c/snosd10c.pdf (note on page 15 that pin 17 is tied to SOURCE)
- All products at https://www.wolfspeed.com/power/products/sic-mosfets/table (example of each symbol below):
-- https://www.wolfspeed.com/media/downloads/145/C3M0065090J.pdf
-- https://www.wolfspeed.com/media/downloads/1185/C3M0030090K.pdf
-- https://www.wolfspeed.com/media/downloads/176/C3M0065090D.pdf
- Infineon half-bridge FETs:
-- https://www.infineon.com/dgdl/irfi4019h-117p.pdf?fileId=5546d462533600a401535623d74d1f6f
-- https://www.infineon.com/dgdl/irfi4020h-117p.pdf?fileId=5546d462533600a401535623e7271f73
-- https://www.infineon.com/dgdl/irfi4212h-117p.pdf?fileId=5546d462533600a401535623fc841f7a

Image:
image

------------


evanshultz:
Oops. Forgot the push the DCM file updates. Some issues fixed, too.

Missing footprints are already submitted at https://github.com/KiCad/kicad-footprints/pull/447. That's the only Travis error.

evanshultz:
I made the LMG3410 look a bit better (I think) and noticed the LMG5200 was missing a dot. Image:
image

evanshultz:
Does any librarian have the time to take a look at this? The only Travis errors are missing footprints which have their PR referenced above.

From https://github.com/KiCad/kicad-symbols/pull/445, two things:
1. I fixed one error with a bad footprint filter for a symbol in Power_Management.
2. Should these GaN parts also be in Power_Management since they're not discrete transistors?

Ratfink:
Clicking through the Transistor_FET library from this PR, I see quite a few with 200 mil gate pins, quite a few with 100 mil gate pins, and at least one (BSP129) where the gate pin is visually disconnected from the rest of the symbol.  I'd prefer to see fixes and additions in separate PRs anyway, so don't go fixing everything here please, I'm just pointing out that there's still work to be done.

It looks like a new footprint is needed for IRFI4019H, as our TO-220-5 with the correct stagger type has the wrong Y pitch.  Also, C3M0065090D needs a new footprint association because the TO-247's name changed recently.

That's just from a quick look.  I'll try to do a more thorough review tomorrow.

evanshultz:
Thanks!

Fixed a couple footprints you mentioned. I calculate IRFI40xxH footprints to have 2.055mm of offset, using the middle of the min/max range, which is awfully close to 2.06mm. So it looks like the right footprint is at https://github.com/KiCad/kicad-footprints/blob/master/Package_TO_SOT_THT.pretty/TO-220F-5_Vertical_StaggeredType2_Py2.06mm.kicad_mod. Have I blown it?

Ratfink:
Ah yes, that's the correct footprint all right!  I blew it for not noticing that the package is a TO-220F rather than a regular old TO-220.

* For the Wolfspeed FETs, I'd like to see a junction dot where pin 2 connects to 3-7.  That might require moving a line so the dot doesn't merge with the pin number.
* C3M0030090K has the wrong datasheet link (404)
* LMG5200
  * Either re-arrange the drawing so the pin names can be inside without overlapping things, or set the pin names to be outside again.  Right now it's a mess.
  * Maybe do the gate driver the same way you did for LMG3410?  Right now it looks like the GaN driver box might also be wires for the HB and LI pins.
* LMG3410
  * Pin 17 is NC.  The datasheet says it can be connected to source or left floating, so it should be visible and next to the source pin, not stacked with it.
* I'd feel happier putting the GaN power stages in Power_Management.

evanshultz:
- FETs with source sense pin: Moved line and added dot.
- LMG5200: Moved pins inside and adjusted.
- LMG3410: As noted above I followed the TI design on it. It would be more work to not connect this pin.
- Moved GaN symbols as mentioned.

Image:
image

Ratfink:
This is almost ready: it looks like the only thing that needs to be fixed is the footprint for IRFI4019H, which was renamed in https://github.com/KiCad/kicad-footprints/pull/574.

evanshultz:
Yep. Now we're good!

Ratfink:
Thanks for all the work on this library!

PR #411, opened 2018-03-22
Add different packages to AS33xx and V33xx

robkam:
Add different packages to AS33xx and V33xx 
Do the footprints suit the data sheet dimensions?
Some pins on AS334x moved around for consistency, (mainly VS). 
DIP V33xx are pin compatible with AS33xx but on the data sheet the pin types differ.

http://www.alfarzpp.lv/eng/sc/AS3310.pdf
http://www.alfarzpp.lv/eng/sc/AS3320.pdf
http://www.alfarzpp.lv/eng/sc/AS3330.pdf
http://www.alfarzpp.lv/eng/sc/AS3340.pdf
http://www.alfarzpp.lv/eng/sc/AS3360.pdf
http://coolaudio.com/docs/Cool%20Audio_Data_Sheet_V3320_VCF.pdf
http://coolaudio.com/docs/COOLAUDIO_V3340_DATASHEETS.pdf

Including AS33xx DIP for comparison:
capture


suzizecat:
Hi !

 1. You should add a screenshot of your symbols so it's easier to have a look at your symbols.
 2. Then you should have a look at Travis output, in example, it says that your VCC/VSS pins aren't of the right electrical type. They should both be "Power Input" (yes, VSS is a power input :stuck_out_tongue: )
 3. Your footprint filter isn't right. If you want to match `Package_SO:SOP-16_4.0x10.0mm_P1.27mm` (wich, by the way, doesn't seems to exists) you should use the following patern : 
`SOP*4.0x10.0mm*P1.27mm*`(don't forget the joker at the end of the patern). You may look into the footprint library to find out the exact name of the footprint you want to match.

Also, if you need more informations, you should go read the [KiCAD Convention library](http://kicad-pcb.org/libraries/klc/)

Regards,
julien

robkam:
@suzizecat thanks for your help. 
I'm just about to add the screenshot etc.
I've been looking at the Travis output that's why there are the multiple commits, unfortunately I was also copying changes from the wrong .lib file, nearly fixed now.
I've used footprints from the library but not very familiar with these so I'm not 100% confident that they're as specified in the data sheet.

suzizecat:
@robkam Could you add a link the the datasheet ? 

Also, if you intend to push more components, you should consider to download a local version of travis check scripts. That will save you a lot of time (and commits). [KLC Helpers Scripts](https://github.com/kicad/kicad-library-utils)

robkam:
Done.

suzizecat:
I just had a look to your footprint filters. You should :

 - [ ] Check that you got the dimension of the package (SOIC and SOP)
 - [ ] Check the size of your components (SOP) (Cf V3320 DS)
 - [ ] Check that the units of size are presents (QFN)
 - [ ] Check the exposed pad and his size are present (QFN)
 - [ ] Check that the pitch is present (QFN)

In example : 
 - `SOIC*P1.27mm*` -> `SOIC*3.9x9.9mm*P1.27mm*`
 - `QFN*4x4*0.5mm*` -> `QFN*1EP*4x4mm*P0.5mm*EP2.6x2.6mm`
 - `SOP*P1.27mm*` -> `SOP*7.5x11.45mm*P1.27mm`

I just quickly overviewed your symbols, make those checks for all your components and you should be fine :ok_hand:

suzizecat:
The V3320 SOP footprint doesn't seems to exist ATM. You may create it, but that's something else...

robkam:
@suzizecat the footprints are as close as I can find in the libraries to match the data sheets. They're common packages so I shouldn't need to create any footprints. The dimensions in the data sheets are slightly different. I think this is because the footprints are for the solder pads and not the package pins.
e.g.  V3320 I've selected 
Package_SO:SOP-18_7.0x12.5mm_P1.27mm
the [data sheet](http://coolaudio.com/docs/Cool%20Audio_Data_Sheet_V3320_VCF.pdf) has 
7.30 to 7.70 x 11.25 to 11.65, (also swapping length and width).

robkam:
Still fixing V3320, V3340D and V3340M pins.

robkam:
Now ready for review and merging.

evanshultz:
@robkam 
Because there is so much noise here, do you mind closing this PR and starting a new one? It's now extremely difficult to track down changes from the existing library to your changes. Please link to this PR when you submit the new one.

PR #414, opened 2018-03-24
Add WS2811

adamheinrich:
3-channel 8-Bit PWM LED driver by WorldSemi (https://cdn-shop.adafruit.com/datasheets/WS2811.pdf):

ws2811


jkriege2:
nice work!

thanks,
JAN

PR #422, opened 2018-03-26
Add TI CC11011 Low-Power Sub-1 GHz RF Transceiver

xl0:
Datasheet: http://www.ti.com/lit/ds/symlink/cc1101.pdf
cc1101
Footprint PR #453

------------


evanshultz:
Thanks for making your first contribution to KiCad! We have check scripts which can catch many issues, but in this case all that is sees is that the footprint is missing. That's great! You can see this at the Travis CI "Details" link above. The rules are set by KLC, which you can find at http://kicad-pcb.org/libraries/klc/.

Here are my review notes:
- Please add `QFN-20` to the end of the description.
- Can you add more keywords? See other symbols in this lib.
- Footprint filter can be simplified to `Texas*PVQFN*EP2.4x2.4mm*`.
- Please stack AVDD and GND pins. Then you can shrink the symbol a bit, at least horizontally.
- You need to add a pin for the thermal pad. Typically this will be the next unused pin, or 21 here. And it must be tied to ground so you can name it GND.
- Pin 2 is SO. not MISO. Also, please check the datasheet if this pin is tri-state. Similar issue with pin 20.
- Pin 5 should be Passive type (it cannot supply power and only connects to a capacitor). And put it on the side of the symbol. Both items apply to pin 17 as well.
- Pin 6 is missing the "ATEST" part of the pin name.
- Pin 8 should be type Input.
- Can you flip the locations of pins 8 and 10? It might also be nice to spread them out so a crystal symbol can fit nicely there if the user doesn't drive in an external clock.
- Pin 10 should be Passive type.
- Perhaps swap positions of pins 12 and 13 and move them to the center of the symbol body.

Please take a look at the above and provide feedback. If you agree, please make changes. If you have a different idea or have any questions please comment. Thanks!

FYI, I'll wait to sort out the footprint naming before merging this. Sometimes TI parts are wacky.

xl0:
Thanks, will fix!

xl0:
- [x] Please add QFN-20 to the end of the description.
- [x] Can you add more keywords? See other symbols in this lib.
- [x] Set pin text offset to 20mil.
- [x] Footprint filter can be simplified to Texas*PVQFN*EP2.4x2.4mm*.
- [x] Pin length should be 100mil.
- [x] Please stack AVDD and GND pins. Then you can shrink the symbol a bit, at least horizontally.
- [x] You need to add a pin for the thermal pad. Typically this will be the next unused pin, or 21 here. And it must be tied to ground so you can name it GND.
- It was there. I left it on top of the stack, TI says it's the main GND.
- [x] Pin 2 is SO. not MISO. Also, please check the datasheet if this pin is tri-state. Similar issue with pin 20.
- Done, nothing about tri-state in the datasheet.

- [x] Pin 5 should be Passive type (it cannot supply power and only connects to a capacitor). And put it on the side of the symbol. Both items apply to pin 17 as well.
- [x] Pin 6 is missing the "ATEST" part of the pin name.
- [x] Pin 8 should be type Input.
- [x] Can you flip the locations of pins 8 and 10? It might also be nice to spread them out so a crystal symbol can fit nicely there if the user doesn't drive in an external clock.
- [x] Pin 10 should be Passive type.
- [ ] Perhaps swap positions of pins 12 and 13 and move them to the center of the symbol body.
- That's how TI draws it, I'd leave it that way, to avoid mistakes when copying the reference schematics.
ti

Update the branch. Please check.
cc1101

I'll wait to sort out the footprint naming before merging this. Sometimes TI parts are wacky.
- indeed.

xl0:
Fixed the commit message.

evanshultz:
Thanks for making the changes!

Could you do a couple more things?
- Move the crystal pins to the left side. They're really an input.
- Move the GDO pins to the right side since they're outputs?
- Put an empty pin space around pins of a particular function and other pins? For example, around the SPI bus pins and RBIAS?
- Perhaps RBIAS can go at the bottom of the symbol so it doesn't cross over other pins.

xl0:
The reference design has different value decoupling caps for different AVDD lines (220pF and 10nF). Should I de-stack them to make the relationship between the caps and pins easier to track? I assume it's marginally important.

evanshultz:
Good thought! Which reference design? I see several at http://www.ti.com/product/CC1101/toolssoftware.

http://www.ti.com/lit/df/tidrbf6/tidrbf6.pdf, for example, has 1uF, 100nF, and 100pF. I assume they're chosen to provide a desired Z vs freq characteristic and it doesn't have anything to do with specific AVDD pins.

xl0:
I meant this one,
http://www.ti.com/tool/cc1101em433_refdes

and the capacitor values are completely different from the one you referenced, so I think it's safe to assume the cap-pin relationship has no consequence. I also did not find mentions of the decoupling cap selection in the main datasheet - they just suggest to copy the reference design.

Will include the changes you suggested above in a couple days.

evanshultz:
Thanks. I didn't find anything in the datasheet either. I'll be happy to merge this in a couple days then! :)

Misca1234:
@evanshultz 
How about that merge ;) ?

evanshultz:
@xl0 
Are you able to make the changes discussed above? I don't see a commit to include them.

evanshultz:
@xl0 
Still planning to get to this? 5.1 is approaching fast so you will need to wrap this up in the next couples days if you want it to be included.

myfreescalewebpage:
@xl0 ping! :)

PR #423, opened 2018-03-26
Add TI BQ2501x 500mA Li-Ion Charger w/Integrated 150mA Synchronous Bu…

xl0:
…ck Converter

Datasheet: http://www.ti.com/lit/ds/symlink/bq25010.pdf
bq2501x
Footprint PR #454


Misca1234:
I suspect your branch is messed up, your push "includes" a lot of deleted items in the dcm file.


If there is no reason for keeping separate "power pins" like the GND pins, there is a strong desire to have them "stacked". But if the data sheet require a decoupling capacitor for each of them.

Usually the pad is nicknamed EP1, but if it should be stacked with GND you have to rename it anyway

You should change the description from
500mA Li-Ion Charger w/Integrated 150mA Synchronous Buck Converter
to
500mA Li-Ion Charger w/Integrated 150mA Synchronous Buck Converter, QFN-14

Rename pin 14 from PE to ~PE then you will get a little line above the PE
Rename pin 15 from /CE to ~CE then you will get a little line above the CE

I suspect that pin 16 and 17 should be of the type "Power output" (they charge the battery pack)

evanshultz:
Not just the DCM.... symbols have been removed from the LIB file too!

Other comments:
- GND pins should always be stacked unless there is a specific reason (and no examples have been given that I can find). Power supplies should be unstacked only if the datasheet has a special per-pin requirement, such as adding a decoupling cap on that pin.
- For this part, the BAT/OUT pins should be stacked too.
- Pin length should be 100mil if pin names are visible and the number of characters in the pin names are less than 3. This request applies to this symbol.

Misca1234:
I disagree to the stacking of BAT/OUT pins, see page 8, these two pins are not of the same type and they are internally coupled to different circuits within the chip.

evanshultz:
That's true. Though I don't think the pin type different will matter here since pins 16 and 17 aren't intended to connect to other circuits. It would appear to me that the "circuits" inside the IC are separate but that may be for die layout reasons as those pins seem to go together to form a single function.

I was noting the application circuits always show pins 16 and 17 tied together, plus all references I find in the datasheet use a singular word. For example, see page 11:
image

Either way, then. Up to you.

xl0:
Thank you for the suggestions. I will make sure the branch is in proper condition.

I have to disagree regarding BAT/OUT stacking. One of them is power output to charge the battery. (The datasheet lists it as I/O because it's used to monitor the battery voltage, but I will set it to power out, makes more sense to me). The other one is power input to feed the buck converter. In theory nothing prevents you from disconnecting the input one, or having it connected to a different source. Not saying that it's a good idea, but maybe I lack imagination.

For the other comments - agreed, will try to update within a few days.

Misca1234:
@xl0 
When you come back, I suggest you remove this branch and create a new due to the huge changes to the .dcm and .lib files

poeschlr:
I took this over in https://github.com/KiCad/kicad-symbols/pull/1349
Hope nobody is mad at me for it.

PR #426, opened 2018-03-26
Add further packages for AS33xx and V33xx

robkam:
Was [Add different packages to AS33xx and V33xx #411](https://github.com/KiCad/kicad-symbols/pull/411)

Add the other packages available for AS33xx and V33xx 
Some pins on AS334x moved around for consistency, (mainly VS). 
DIP V33xx are pin compatible with AS33xx but on the data sheet the pin types differ.

http://www.alfarzpp.lv/eng/sc/AS3310.pdf
http://www.alfarzpp.lv/eng/sc/AS3320.pdf
http://www.alfarzpp.lv/eng/sc/AS3330.pdf
http://www.alfarzpp.lv/eng/sc/AS3340.pdf
http://www.alfarzpp.lv/eng/sc/AS3360.pdf
http://coolaudio.com/docs/Cool%20Audio_Data_Sheet_V3320_VCF.pdf
http://coolaudio.com/docs/COOLAUDIO_V3340_DATASHEETS.pdf

Including AS33xx DIP for comparison:
capture


evanshultz:
- Remove CoolAudio ALIASes.
- For 3310, 3340, and 3360 there are DIP and SO variants with the same pinout. Perhaps an easier way would be to share a symbol, unfill the default footprint, and add footprint filters for both packages. That way a single symbol will work for either package, and the user will select the footprint in CvPcb instead of having a default one pre-selected. But the possible selections will only be the couple valid footprints. What do you think?
- The QFN package has a 2.7mm pad. The footprint you picked has a 2.6mm pad. The footprint should always have a larger pad than the package. I don't see any other potential footprints that already exist. Would you be willing to create a new footprint?
- I assume that pins not mention in the datasheet for QFN packages are NC. For example, pins 1 and 6 of AS3320F. If so, please add those pins to the schematic, name them NC, make them invisible, and place their end along the symbol border so they're not easily connected.
- Add the body side to the footprint filter. So AS3340D would be `SOIC*3.9x9.9mm*P1.27mm*`.
- Also add all qualifiers, so the QFN would be `QFN*4x4mm*P0.5mm*EP2.6x2.6mm*`. (This is assuming a 2.6mm pad is correct, which is discussed above.)
- 3345F: Would you mind making pin 20 visible and 22 hidden? Typically the lowest number is visible.

robkam:
@evanshultz the footprint for `WQFN*1EP*4x4mm*P0.5mm*EP2.7x2.7mm*` looks like a match?

robkam:
* CoolAudio aliases removed
* 3310, 3340 and 3360 are now a single symbol with footprint filters for either DIP or SOIC variant
* Footprint filter SOIC*3.9x9.9mm*P1.27mm* for AS3310D, AS3340D and AS3360D 
* Added NC pins (1,6,10,13,18,21) to QFN packages AS3320F, AS3330F and AS3345F 
* 3345F pin 20 visible power in and 22 hidden passive
* QFN now 2.7x2.7mm

capture

evanshultz:
These changes look good. Thanks! You will need to update the QFN package name as the footprint has changed; I'd like to wait to merge this until the footprint is merged.

robkam:
In the Symbol Library Editor the ref des and value fields are above the symbols ... I'm not sure why they're not in the schematic.

as33xx symbols

robkam:
With text auto place unchecked, in the schematic the symbol's text is placed as intended.
as33xxx

robkam:
@evanshultz please merge

evanshultz:
As mentioned above, I'll wait for the footprint to be merged before merging this.

poeschlr:
For the qfn-24-1EP versions the correct footprint name is `QFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm` or if thermal transfer is important one can also use `QFN-24-1EP_4x4mm_P0.5mm_EP2.7x2.7mm_ThermalVias`.

PR adding the scripted footprint: https://github.com/KiCad/kicad-footprints/pull/700

evanshultz:
Thanks for your patience.

robkam:
Hooray! :-)

Issue #154, opened 2018-01-09
Missing symbols for TRACO TMLM-05 family


antoniovazquezblanco:
The footprint exists (`Converter_ACDC_TRACO_TMLM-05_THT.kicad_mod`) but no symbol for it is present.

antoniovazquezblanco:
Fixed by https://github.com/KiCad/kicad-symbols/pull/427

Issue #155, opened 2018-01-09
 Missing symbols for TRACO TMLM10 and TMLM20 family 


antoniovazquezblanco:
The footprint exists (`Converter_ACDC_TRACO_TMLM-10-20_THT.kicad_mod`) but no symbol for it is present.

antoniovazquezblanco:
Fixed by https://github.com/KiCad/kicad-symbols/pull/427

PR #427, opened 2018-03-27
update TMLM series and add missing symbols

lorem-ipsum:
- remove redundant aliases in favour of TMLM041xx
- add missing symbols with higher output power
- add dual output variants TMLM042xx
- slightly update description

[datasheet](https://www.tracopower.com/products/tmlm.pdf), footprints are already exsting
* TMLM041xx (unchanged)
tmlm041xx
* Dual output variant
tmlm042xx
* Higher power variants (AC connections are not distinguished)
tmlm051xx


lorem-ipsum:
This fixes issues #154 and #155

Misca1234:
If these are galvanic isolated they should perhaps have a division line to mark that ?
(like some of the others ACDC, DCDC)

lorem-ipsum:
so far none of the ACDC converters have that line. I am not really sure if it helps adding it. If it should be added then it should most likely be added to all other ACDC converts as well (would need to check the datasheets for that).

Misca1234:
It is not stated in the guidelines so I guess it is more of a matter of taste among the heavyweighters within the library admin team.

lorem-ipsum:
As there seems to be some agreement by now I reworked the footprints. I also increased the height of the boxes a bit as I found them a bit to crowded.

tmlm042xx
tmlm041xx

Misca1234:
@lorem-ipsum 
You got a branch conflict in this push, it needs to be solved

lorem-ipsum:
I did the rebase and I think I didn't break anything

Misca1234:
I have to dissagree a little with you about your clean up


Device TMLM041xx should be renamed into TMLM 04103
with all the other TMLM 041XX as aliases.
With aliases you also specify the output voltage seperatly

Same goes for TMLM042xx, TMLM051xx , TMLM101xx and TMLM201xx

What do you think about that ?

lorem-ipsum:
I don't really mind. It seems to be not really consistent throughout the library.

Misca1234:
I guess people have different view on this, mine is from a documentation point of view.

You can see on the schematic directly what component it is and it makes the BOM generation more "automatic". Now, the BOM list generation could be better in kicad, but in the best of all worlds you would only need the schematic after 10 years when the PCB need a make over.

lorem-ipsum:
Makes sense. Is it fine as is? I have changed the description from `xW` to `yV zzzmA` to make it more explicit. The power rating is still in the keywords.

Misca1234:
Looks good, thx

PR #431, opened 2018-03-29
Add V13700 alias to LM13700

robkam:
Same pinouts and "functionally compatible to LM13700", data sheet at http://coolaudio.com/features-page.php?product=V13700.


robkam:
duplicate of #432

PR #432, opened 2018-03-29
Add V13700 alias to LM13700

robkam:
Same pinouts and "functionally compatible to LM13700", data sheet at http://coolaudio.com/features-page.php?product=V13700.


Ratfink:
Thanks for the contribution!

I notice that this and #431 are effectively two parts of the same PR (one adds to the .lib, the other adds to the .dcm).  Please merge the changes from one of these two PRs into the other one so we can merge them into the library as one unit.

robkam:
#431 now includes both changes.

evanshultz:
I'm not inclined to merge this. LM13700 is the original PN from a major vendor while Cool Audio is making a knock off for reasons unknown to me. I don't see a reason to have both parts (if anybody knows about Cool Audio they should know about the original part). This would seem to only inflate the library with identical parts (even an ALIAS and not a full symbol), and is similar to why temp variations and component packaging is not included in the official library.

What is the motivation for adding this to the official KiCad library other than a personal desire to include Cool Audio ICs? I'm not saying no or trying to put anybody off, but rather trying to understand.

evanshultz:
When I read "That they're produced by whoever isn't a factor" it sounds like the vendor isn't important but the symbol is. And that certainly makes sense! If no compatible symbol is present then nobody can use it from the official lib. I certainly would, in general, support adding new unique symbols.

Would searching for `13700` work in the existing library without adding any ALIASes?

@Ratfink 
You're usually around during this time. What do you think? Is adding an apparent knock off IC valuable when the major vendor is already there? I'm sure TI (formerly National Semi) has better brand recognition and support, so going the other value would be a different story.

robkam:
A search for 13700 works fine. V13700 is pin compatible so it won't matter if the more familiar symbol is used in the schematic.

robkam:
[Analog Renaissance? The rebirth of the impossible chips](https://electricdruid.net/analog-renaissance/) gives some background info to the ICs I've been adding lately (see PR #426). There's still a couple more to add, SSI2144 and SSI2164. Also here CoolAudio has produced identical copies of the 2164.

evanshultz:
With the original still in production, I lean towards closing this. Please rebut if you like. If there is no discussion I'll close in a day or so. If there are new and original parts that you would like to add that would be great!

robkam:
Okay with the original still in production, I'll close this.

PR #433, opened 2018-03-29
Add FDS6930B

mike7b4:
Add FDS6930B
------------


mike7b4:
screenshot_20180329_165707

Ratfink:
Thanks for volunteering to add this part to the KiCad library!

Unfortunately, I don't think this symbol can be merged in its current state.  All the symbols in the Transistor_FET library use normal FET graphical symbols, with multiple units for parts like this one that have several transistors in one package.  Therefore, this part should be made that way as well.

The best way to do this would be to add this part as an alias of an existing, similar symbol.  A quick search through the library turned up the FDS6890A, which appears to be another dual N-channel MOSFET with the same package and pinout.  If you'd add FDS6930B as an alias of that part and set the documentation for the alias correctly, I'd be happy to merge that.

mike7b4:
I f.. my git branches so I had to create a new separate pull request with the alias fix. #434

PR #435, opened 2018-03-30
Add SSI21x4

robkam:
Add SSI2144 and SSI2164 to Audio.lib. These are recent new versions of SSM2144 and SSM2164 ICs.

Data sheets:
http://www.soundsemiconductor.com/downloads/ssi2144datasheet.pdf and package outline dimensions http://soundsemiconductor.com/downloads/pod16lssop.pdf

http://www.soundsemiconductor.com/downloads/ssi2164datasheet.pdf and package outline dimensions http://www.soundsemiconductor.com/downloads/pod16lsop.pdf

Screenshot:
capture


evanshultz:
- Please remove the CoolAudio ALIAS.
- Where did you get package dimensions, pin pitch, etc?
- Change all pin lengths to 100mil.
- Please add an underscore to separate the pin names, such as making pin 15 of SI2144 `FREQ_CTRL`.
- Please adjust the pins to avoid such close text at the bottom of SSI2144. Perhaps swap pins 8 and 9?
- I think SSI2144 would be better if the CxA and CxB pins had a bit of space between them for the capacitor to live. Could you see if that works out better?
- What is the correct name of SSI2164 pin 1? http://www.soundsemiconductor.com/downloads/ssi2164datasheet.pdf shows several different names.

The schematic took, depending on version and settings, will auto-place text.

robkam:
@evanshultz 
* I'd updated the screenshots above, at the start of the conversation.
* I don't recall where I first got the package dimensions, only that it was a struggle to find out. I've now posted links to the correct data sheets for the parts, above, at the start of the conversation.
* The correct name for SSI2164 pin 1 is "Mode", (the bias there selects between **modes** Class A and Class AB, and is called Mode on the pin connections diagram).

robkam:
Data sheets:
SSI2144 package outline dimensions http://soundsemiconductor.com/downloads/pod16lssop.pdf
(http://www.soundsemiconductor.com/downloads/ssi2144datasheet.pdf) 

SSI2164 package outline dimensions http://www.soundsemiconductor.com/downloads/pod16lsop.pdf
(http://www.soundsemiconductor.com/downloads/ssi2164datasheet.pdf) 

Updated symbols:
capture

robkam:
Apologies, I've messed the history up in this PR.

evanshultz:
@robkam 
Can you clarify what "I've messed the history" means? Is this is still OK to review and merge?

robkam:
Yes it's still okay to review and merge. I messed it up by editing the original PR, instead of just tacking on the adjustments down here.

evanshultz:
- Please remove `V3320` and `V3340` from Audio.dcm in your patch-3 branch.
- 2164 package is actually `SOIC-16`, so change the description (they call is `SOP`, I know).
- 2144 package is `Package_SO:SOIC-16_3.9x9.9mm_P1.27mm`; 2164 package is `Package_SO:SSOP-16_3.9x4.9mm_P0.635mm`. Also update the footprint filters.
- 2164: Would it look a bit better to leave an empty pin space between the 4 sections and align the input and output across the sides of the symbol?
- 2144: Please leave an empty pin space between pins 5 and 15.
- The two symbols have V- and GND swapped on the bottom; it's not a big deal but would you like to be consistent?

I reviewed all the pins and everything else looks good. Thanks!

evanshultz:
Thanks! Just two things left:
- Add the body size in the footprint filter; for example, `SSOP*3.9x4.9mm*P0.635mm*`
- Please move the ref des and value fields to the left and right, respectively, above the symbol.

robkam:
@evanshultz thanks for your help, nearly there ...

In the Symbol Library Editor the ref des and value fields are above the symbol, although in the schematic they appear otherwise. I don't know how to fix this?

evanshultz:
Here is the option in Eeschema you're looking for:
image

Footprint filter looks good. The text is floating above the symbols a bit; can you move it just above the symbol and align the refdes to the left side to give more room for a refdes with several characters?

robkam:
Those are the settings Eeschema has.

e.g. in the Symbol Library Editor AS3330 unit A has the ref des and value fields above the symbol:
as3330a
but in the schematic they appear to the right:
as3330asch

I'm unable to get the ref des and value fields to appear in the same place on the symbol in the schematic as in the symbol library.

robkam:
Those are the settings I have in Eeschema.

How do I get the ref des and value fields to appear in the same place on the symbol in the schematic as in the symbol library?

e.g. in the Symbol Library Editor the ref des and value fields look fine and are placed above the symbol:



but in the schematic they appear somewhere else:
bothinsch

poeschlr:
EEschema does auto place the visible fields for symbols. I think this can be turned of somewhere but if you rotate the symbol you will then need to manually adjust the placement.

robkam:
I'd rather leave Eeschema doing auto placing. Here there's no obvious reason for it to move the fields away from where they are in the library. Is it a bug?

Ratfink:
If it does auto-placement, it *always* moves them away from where they are in the library.  You have to choose one or the other.

robkam:
Okay, with the auto-place option deselected the text remains positioned as it is in the library.
ssi21xx

robkam:
please merge

evanshultz:
I asked for a couple changes to the text placement at https://github.com/KiCad/kicad-symbols/pull/435#issuecomment-385527595 but I don't see any later commits.

robkam:
In the Symbol Library Editor the ref des and value fields are placed as you suggest but in Eeschema with automatic placing enabled they're moved randomly elsewhere. I don’t know how to fix this.

Ratfink:
As I wrote earlier:

>If it does auto-placement, it always moves them away from where they are in the library. You have to choose one or the other.

It's quite simple: if you don't want the fields to get moved around by Eeschema, turn off automatic placement.  There's no way to "fix" it in the symbol.  The auto-placement algorithm is doing exactly what it's supposed to.  It could certainly be made smarter, but the current behavior is not a bug.

You were the only one concerned about that, though.  Evan wanted you to move the value and reference fields down 50 mils, that's all.

evanshultz:
@robkam 
https://github.com/KiCad/kicad-symbols/pull/435/commits/aa1df3ba3c08951bd24cc54f9d768802f727336b fixed the footprint filters but https://github.com/KiCad/kicad-symbols/pull/435/commits/bd862f54205be203cad5652413a0bcba5d101386 unfixed it. Please correct this.

Also, the text fields didn't just move down they moved over as well. Their X placement was fine and it would better if that was moved back so there is more room for a longer ref des.

evanshultz:
Thank you!

PR #436, opened 2018-03-30
Add alias for FDS6930A/B to Transistor_FET library.

mike7b4:
Should be correct this time validated twice with the document.

------------


evanshultz:
@mike7b4 
Just one request: Add `SO-8` to the end of the description. The package is at the end.

In the future, if you like, you can push new commits to the same PR instead of making a new one. That keeps things more clean (usually, sometimes it gets messy) and can be easier to track and find later. In this case, it's still simple so we're all good.

For reference, datasheets are at http://www.onsemi.com/pub/Collateral/FDS6930A-D.PDF and http://www.onsemi.com/pub/Collateral/FDS6930B-D.PDF.

Personal note: `IRF7309IPBF` and `FDS6890A` can be ALIASes instead of unique symbols.

Ratfink:
@evanshultz They can't be aliases of the same part: `IRF7309IPBF` has one N-channel and one P-channel MOSFET, `FDS6890A` has two N-channel MOSFETs.

evanshultz:
@mike7b4 
Thanks for making your first contribution to KiCad! I hope you're willing to make more contributions in the future.

PR #437, opened 2018-03-30
LT1301 boost converter

ppelleti:
[Datasheet](http://www.linear.com/docs/3451).

I was unsure what some of the electrical types for the pins should be, so I made my best guess.

lt1301

------------


evanshultz:
@ppelleti 
Thanks for contributing to KiCad!

1. Can you use the datasheet description of `Micropower High Efficiency 5V/12V Step-Up DC/DC Converter for Flash Memory`?
1. Add the packages, `DIP-8/SOIC-8` to the end of the description.
1. Pin text offset should be 20mil.
1. Footprint filters should be `DIP*W7.62mm*` and `SOIC*3.9x4.9mm*P1.27mm*`
1. Perhaps space out SW and SENSE pins just a bit.
1. Put pin 5 below pins 2 and 3 since it may be tied to ground.
1. Pin 4 should be of type Passive or Input.
1. You may stack pins 1 and 8 since they have to be tied together. I'll merge without stacking too.

If you have any questions please post here or see [KLC](http://kicad-pcb.org/libraries/klc/).

ppelleti:
Thanks, I've implemented suggestions 1-7.

> 6. Put pin 5 below pins 2 and 3 since it may be tied to ground.

Pins 2 and/or 3 could also be tied to ground, but I went ahead and did as you asked.

> 7. Pin 4 should be of type Passive or Input.

I did this, but I also changed pin 7 to Power Output.  It seems that one of the pins should be a power output, since outputting power is the purpose of this component.

> 8. You may stack pins 1 and 8 since they have to be tied together. I'll merge without stacking too.

I elected not to stack them; I think stacking makes it less clear what's going on.

evanshultz:
Thanks for making the fixes!
6. Pin 5 is always connected to ground through a resistor or open. So it will only need ground, and thus at the bottom. The other pins may be connected to some other device and have longer horizontal lines coming from them. That's why I suggested it.
7. It can seem weird, but for the ERC engine it's fine without a power output pin. SW certain't isn't that because there must be an inductor between that pin and the output source of the converter. SW can go back to Passive type.

ppelleti:
> 6. Pin 5 is always connected to ground through a resistor or open. So it will only need ground, and thus at the bottom. The other pins may be connected to some other device and have longer horizontal lines coming from them. That's why I suggested it.

Thanks, that makes sense!

> 7. It can seem weird, but for the ERC engine it's fine without a power output pin. SW certain't isn't that because there must be an inductor between that pin and the output source of the converter. SW can go back to Passive type.

OK, I've changed SW back to Passive.

evanshultz:
Thanks again! The symbol itself looks good.

But I just realized it's in the wrong library. Can you move it to `Regulator_Switching`? The lib it's in is for self-contained switching converters, not controllers. Sorry for missing that earlier.

ppelleti:
OK, it's now in `Regulator_Switching`.

ppelleti:
> It can seem weird, but for the ERC engine it's fine without a power output pin.

I'm finding that ERC isn't recognizing the power output automatically.  I have to manually put PWR_FLAG on it in the schematic.

PR #438, opened 2018-03-30
added tc1121x

sebastianbasierski:
I used recently this part and created symbol, so here it is.

Datasheet: [http://ww1.microchip.com/downloads/en/DeviceDoc/20001358D.pdf](http://ww1.microchip.com/downloads/en/DeviceDoc/20001358D.pdf)

tc1121coa
tc1121coa

tc1121cpa
tc1121cpa

tc1121cua
tc1121cua

tc1121eoa
tc1121eoa

tc1121epa
tc1121epa

tc1121eua
tc1121eua


Ratfink:
TC1121 is an obsolete part according to [Microchip's website](https://www.microchip.com/wwwproducts/en/TC1121) and the datasheet you linked above.  It isn't available from distributors anymore, according to Octopart and FindChips.

Taking all this into consideration, I don't think it makes sense to add this part to the KiCad library.  I'll wait a few days (say, until Tuesday afternoon EDT), and unless any other maintainers object, I'll close this PR then.

evanshultz:
Yep. Digi-Key shows it as obsolete too and has zero on hand.

PR #448, opened 2018-04-02
Add LTC3805

vitperov:
Hi!
It's my first merge-request to kicad-symbols. Please let me know if I did something wrong.

Datasheet: http://cds.linear.com/docs/en/datasheet/3805fg.pdf

ltc3805


vitperov:
Datasheet: http://cds.linear.com/docs/en/datasheet/3805fg.pdf

vitperov:
Fixed:
* Pins now located on 100mil
* Added description to aliases

evanshultz:
Thanks for making a first contribution to KiCad! You can find the rules for library parts at http://kicad-pcb.org/libraries/klc. If you have questions or want to make a comment, please do!

Here are my review notes for this symbol:
- This symbol should be in the `Regulator_Controller` library.
- We wildcard temperature and lead free and packaging, so you will have only one symbol named `LTC3805xMSE`.
- In the description, `current` does not start with a big `C`.
- At the end of the description, include the package by adding `, MSOP-10`.
- Pin 10 should be Output type.
- Put pin 6 on the left side and center pin 11. I suggest placing pin 6 between RUN and Ith since it may have a long line going to some clock source while pins 1, 2, and 5 should only go to ground.
- You could move apart the pins that are together. For example, OC and Isense will need a resistor going to them so giving a little room will be useful

One last thing: Would you be willing to add a symbol for the DFN package? It's totally optional.

vitperov:
Thank you for your comments!
I'll fix that.

Just one remark: SYNC pin is usually unused and connected to ground (see datasheet).
I can place it left down corner, but pacing it above Ith and SSFLT would be inconvenient in most cases.

vitperov:
Done!
ltc3805-2

evanshultz:
@vitperov 
Thanks for fixes!

Ith, SSFLT, and FS pins should only go to ground. SYNC may go to ground, or it might go to some other external circuit. So if SYNC is above those, it will never cross over any other wires. Yes, it won't be able to go as directly to ground but it will avoid wires crossing for no detriment I can see. Here is what I was thinking the symbol could look like:
image

What do you think? If SYNC is at the bottom, it will cross wires. Do you other ideas?

A few other items remain to be done:
- In the description, `current` does not start with a big `C`.
- You could move apart the pins that are together. For example, OC and Isense will need a resistor going to them so giving a little room will be useful
- Would you be willing to add a symbol for the DFN package? It's totally optional.

vitperov:
Hmm... Yes, I understand your logic about SYNC. But in my personal opinion it's less convenient. In such rare cases, that you need SYNC, you can use labeled wire.
But, ok, I will place SYNC above  Ith.

"You could move apart the pins that are together. For example, OC and Isense will need a resistor going to them so giving a little room will be useful"
As I understood, now it means Ith, SSFLT and FS pins?

evanshultz:
It will be OK to leave the pins as you prefer. Are you ready for this to be merged?

vitperov:
Thank you for your help! I think, it can be merged.

About DFN package... Sorry, but not now. I would better add another 2-3 chips, which I am using and my pin assingments can be checked in real PCB.

evanshultz:
Thank you!!

PR #453, opened 2018-04-04
Add LD39015 linear regulator

tagunil:
Add LD39015 low quiescent current low dropout voltage regulator from ST.
Datasheet: http://www.st.com/resource/en/datasheet/ld39015.pdf
screenshot_20180404_184042


evanshultz:
@tagunil 
Thanks for making this first contribution to KiCad! Everything looks great here and I'll merge your symbol. Thanks so much and I hope you will submit more to the library in the future!

PR #457, opened 2018-04-06
Lattice iCE40UP5K-SG48ITR FPGA Symbol

awygle:
Add Lattice iCE40UP5K-SG48ITR FPGA.

This is my first Kicad symbol PR, please let me know of any errors I've committed. I decided to break this symbol up into multiple parts, I hope that's OK.

Documentation is found at http://www.latticesemi.com/Products/FPGAandCPLD/iCE40Ultra, datasheet is http://www.latticesemi.com/view_document?document_id=51968 .

Images:

image

image

image

image


whitequark:
@Misca1234 The three I/O banks get their own parts of the symbol, as seems common for FPGAs.

awygle:
So there are four units in this symbol. VCC and GND are at the top and bottom of the final unit. The first unit is centered (or was, possibly), but because the refdes and description must be in the same place on every unit, I couldn't center the other units without moving them far away from the refdes.

Happy to resolve this however you like, just not sure what is the right thing to do. Unit support seems... Somewhat half baked, but it's very useful for FPGAs.

Misca1234:
ahhh, my bad, i was screening the push to fast

awygle:
I made some cleanup changes based on feedback I've gotten on my other symbol PRs. I'd still like some pointers on how to properly handle multi-unit symbols, though.

evanshultz:
There's shouldn't be any intrinsic issues caused by multi-unit symbols. The centering thing is probably our algorithm. But VCC and GND pins on the last unit can be moved top and bottom to resolve one warning.

whitequark:
There is an issue with PLL supply pins, GNDPLL should not be connected to GND.
screenshot_20180414_124629

evanshultz:
Thanks! I'll let @Misca1234 finish reviewing this.

whitequark:
@Misca1234 Is there anything that needs to be done to get this symbol merged?

evanshultz:
Let's see how https://github.com/KiCad/kicad-symbols/issues/550 shakes out and then I'll jump in and do another review so we can get this merged.

whitequark:
@evanshultz As per https://github.com/KiCad/kicad-symbols/issues/550#issuecomment-387234856, is it fine if the symbol is altered to have its power pins on top and bottom? We need another FPGA symbol now and I'd rather have a decision before trying to contribute that, too.

whitequark:
@evanshultz Is there anything else that needs to be changed for this symbol to be merged?

evanshultz:
@whitequark 
Sorry, I missed this. Yes, power and ground pins should be top and bottom. If you'd update this symbol, split or not, and post a screenshot we can move forward with review.

awygle:
Sorry for the delay, everyone - I made this change six weeks ago but apparently didn't update the screenshots. You can see the current state below:

image

image

image

image

evanshultz:
Thanks for new screenshots! Full review:
- Can all power pins be on the top of the last unit? All pins on the last unit but also the VCCIO pins on other units.
- Should pin 8 be `~CRESET_B`?
- I think pin 7 is the wrong type. Page 42 of the datasheet says CDONE is for I/O with an internal pullup and on the QFN it's a dedicated pin. So should it be Bidirectional type instead of Input?
- Would it make more logical sense for the RGB pins to increase going down?
- VCC pins are numbers 5 and 0. The 0 should be 30.
- Is the package drawing on page 31 or 32 of http://www.latticesemi.com/view_document?document_id=213? I think page 31, which means the thermal pad has an obtuse range of 3x3mm to 5.8x5.8mm! Even page 32 is 5.4x5.4mm nominal, so either way the default package you picked with a 5.15mm square pad needs to be changed. We have one footprint with a 5.6mm square thermal pad but even that might be too small. And the outer pads could be wrong as well. Can you look into what the correct footprint should be?

FYI, pinout is at http://www.latticesemi.com/view_document?document_id=51971.

evanshultz:
@poeschlr 
Would you please comment on splitting symbols, either here or the dedicated thread linked above? I'm inclined to merge this when it's ready, but your voice is important here.

evanshultz:
So that hopefully every notices: This PR is adding a library.

@awygle 
You will need to add this library to the sym-lib-table file before it can be merged.

mashu:
Any update on this ? Thumbs up

awygle:
> Thanks for new screenshots! Full review:
> 
>     Can all power pins be on the top of the last unit? All pins on the last unit but also the VCCIO pins on other units.

I don't like the idea of moving the VCCIO pins to the last unit because they're the IO supply to the bank. The way we currently have them, you can see at a glance what the IO voltage of a pin will be. I'd be OK with moving them to the top of their current units.

>     Should pin 8 be ~CRESET_B?

The _B indicates the active low nature of the signal.  So it should be ~CRESET or CRESET_B. I'll go with ~CRESET since it matches KLC (as opposed to CRESET_B which matches the pinout document).

>     I think pin 7 is the wrong type. Page 42 of the datasheet says CDONE is for I/O with an internal pullup and on the QFN it's a dedicated pin. So should it be Bidirectional type instead of Input?

On the QFN48, it's an open drain output which is low during configuration and then floating (weakly pulled up) afterwards. The pin can be used as a user I/O in the WLCSP30 package because it's double-booked with IOB_12a_G4. I agree the datasheet is confusing.

>     Would it make more logical sense for the RGB pins to increase going down?

Makes no difference to me. I guess that does match the pin ordering. I'll change it.

>     VCC pins are numbers 5 and 0. The 0 should be 30.

Thanks.

>     Is the package drawing on page 31 or 32 of http://www.latticesemi.com/view_document?document_id=213? I think page 31, which means the thermal pad has an obtuse range of 3x3mm to 5.8x5.8mm! Even page 32 is 5.4x5.4mm nominal, so either way the default package you picked with a 5.15mm square pad needs to be changed. We have one footprint with a 5.6mm square thermal pad but even that might be too small. And the outer pads could be wrong as well. Can you look into what the correct footprint should be?

Page 32 specifically lists the iCE40 UltraPlus as one of the parts using it, so I'm pretty sure it's page 32.

> You will need to add this library to the sym-lib-table file before it can be merged.

Sure.

There's no guidance on a footprint from Lattice, and as far as I can tell, no guidance from the KLC on thermal pad size relative to the pad on the part. I can switch to the 5.6mm pad if you like. That will be larger than even the maximum extent, if that is the goal. The external pads are, similarly, larger in every dimension than the pads on the part, and pitched appropriately.

> FYI, pinout is at http://www.latticesemi.com/view_document?document_id=51971.

evanshultz:
Sorry for my delay in getting back to this.

- I would argue that CDONE is not an open collector if there's an internal pullup. It should then be Output type.
- And since it's only an output in this QFN package, would it make sense to put it on the right side of the unit? The package is forcing this pin to likely be placed differently anyway.
- Please move the VCCIO pins to the top of each unit.
- The last unit can be a bit smaller.

Can you post a new screenshot when you make notable visual changes? Here is the current commit:
image

awygle:
> I would argue that CDONE is not an open collector if there's an internal pullup. It should then be Output type.

I disagree - it can still be connected with other open-collector/open-drain outputs, which a true push-pull output could not.

> And since it's only an output in this QFN package, would it make sense to put it on the right side of the unit? The package is forcing this pin to likely be placed differently anyway.

Sure.

> Please move the VCCIO pins to the top of each unit.

Sure. I think this looks weird without a corresponding ground, though.

> The last unit can be a bit smaller.

Sure.

> Can you post a new screenshot when you make notable visual changes? Here is the current commit:

Sure:

image

image

image

image

evanshultz:
You're right about the CDONE pin; open collector is the best way to make the ERC catch issues.

I didn't see many comments about splitting up packages but nothing against my proposal for how to work KLC. Even though it does look a bit weird without ground pins, I think this is the best interpretation of the current and proposed rules. I'm ready to merge. OK?

awygle:
Works for me! Thanks for your help and guidance on this extended process.

evanshultz:
Sure! The libraries have improved dramatically since the last stable release, but there are many more loopholes to find. Thanks for contributing such great symbols and footprints for all users!

PR #459, opened 2018-04-06
Added symbol for Analog Devices AD7616 16-channel data acquisition system

andrasfuchs:
I added the Analog Devices AD7616 ADC/DAS to the library.

I used this datasheet: http://www.analog.com/media/en/technical-documentation/data-sheets/AD7616.pdf

image

------------


poeschlr:
You have the old footprint lib. The symbols in this library must be made using the new one found under https://github.com/KiCad/kicad-footprints

andrasfuchs:
@poeschlr Although I'm not 100% sure, but I think I have the most recent one. I cloned the kicad-footprints repo and I worked with that.

poeschlr:
There is no lib called Housings_QFP in the new lib. It is now called Package_QFP.
So if you still have a lib called housings then you definetly are not using the latest libs.

andrasfuchs:
Hmm, you're right. It's strange: my %APPDATA%\Roaming\kicad\fp-lib-table didn't get updated when I switched from KiCad 4.0 to the 5.0 nightly build, I guess.

So, despite the fact that I had the latest KiCad repos cloned in my local library, modules and packages3d folders, KiCad didn't use them, but used the older ones from the old GitHub repo instead.

I opened the PCB Library Tables dialog from the Footprint Editor, removed all existing libraries, and added the ones in my local modules folder.

Misca1234:
> Hmm, you're right. It's strange: my %APPDATA%\Roaming\kicad\fp-lib-table didn't get updated when I switched from KiCad 4.0 to the 5.0 nightly build, I guess.

Yes I have noticed that as well, now days, i always open up the dialog box "configure paths", both for the start window and the foot print editor and just press OK, it seems that kicad forgott/not update its search pathes correctly when you start the program but remember the default installed ones.

andrasfuchs:
I've moved the multi-pin power and GND pins to overlap each other, I made all but one of them passive and disabled. Is it the right way to do it?

Is there anything else I need to modify to get the pull request accepted?

image

evanshultz:
Thanks for adding this symbol. Sorry for the delay in review as we're getting ready for the v5 release of KiCad and it's busy for the librarians.
- If the datasheet says to specifically put caps on a power pin, it should not be stacked.
- GND pins should stay stacked.
- Use 20mil pin name offset.
- All pin lengths should be 100mil.
- You don't need an ALIAS here; either name the symbol with the suffix or without will be fine (since there is only one package). Also remove the unused entry from the DCM file.
- Please re-arrange and slightly simplify the description to match the datasheet; to include the info you added it could be `16-Channel DAS with 16-Bit, Bipolar Input, Dual Simultaneous Sampling ADC, 2x1MSPS ADC, Serial and parallel IO, LQFP-80`.
- Perhaps pins 31-32 should be on the left side since they're referenced to AGND?
- Can you space apart pins 33-34 so a cap symbol can fit nicely there? And other similar pin pairs.
- Relatedly, since pin 34 goes to ground maybe place it at the bottom of the side. Similar examples for other ground pins exist as well.
- Pins 38-39 could have their locations swapped to stay in order.
- Perhaps move DBxx pins to the top on the right side since some pins above them may connect to DGND. This could avoid wires crossing without any drawback I can see.
- Pin 61 has the names reversed around the forward slash.

andrasfuchs:
I'm still hesitating about the position and arrangement of the decoupling pins, I would appreciate your input.

According to [AD7616's dataset](http://www.analog.com/media/en/technical-documentation/data-sheets/AD7616.pdf) and [AD7616's evaluation kit's schematics](http://www.analog.com/media/en/technical-documentation/evaluation-documentation/EVAL-AD7616SDZ-Schematics.zip), the AD7616SDZ the ideal situation would be if we put 
* one decoupling capacitor between 
  a, REFCAP (31) and REFGND (32)
  b, REFINOUT (33) and REFINOUTGND (34) if the REFSEL (35) is set high
* two decoupling capacitors between
 a, REGGNDD (51) and REGCAPD (52) [with DGND]
 b, REGGND (69) and REGCAP (70) [with AGND]
 c, VDRIVE (49) and DGND (50) [VDRIVE is the 3.3V]
 d, all VCC and AGND pairs (6-5, 15-16, 30-29 and 71-72) [VCC is 5.0V]

If we put the 4 VCC and AGND pins to the opposite side of the symbol, would it not complicate the design significantly?

Here is my current design, but I'm not too satisfied with it, I'm sure it could get more elegant with the proper pin arrangement.

image

image

I will gladly move the pins around as you recommend.

andrasfuchs:
I moved the VCC, GND and CAP pins around according to your recommendations. 
Should I overlap the VCC pins as well?

Here is how the symbol looks at the moment:
image

And here are my decoupling capacitor arrangements:
image

image

What do you think?

evanshultz:
Thanks!

The arrangement of pins looks good to me. I wish I would have mentioned this earlier, but probably the analog gnd pins should be on the left. That is pins 32, 34, and 69. Then this symbol has a proper analog-digital split.

What do you think? If you can do that, or have another idea, I'll review again before merging.

evanshultz:
Yes, I also meant the pins the ground/reference is coupled with.

I understand not wanting to touch the symbol again and having the pin arrangement look balanced. So I checked out a number of ADCs and DACs and all are divided between analog and digital on the left and right halves of the symbol. Therefore, would you please move the analog-referenced pins to the left side? Sorry to only notice this after you made a number of other changes.

andrasfuchs:
No worries, here you go ;)

image

Is there anything else you think I should change?

evanshultz:
The analog inputs run right into the REF* pins and usually we would separate pin groups. The symbol is also a touch taller than it needs to be (100mil on top and bottom) if the pins don't change. But, if you're tired of messing with this symbol I'll merge as-is.

andrasfuchs:
image

I wasn't sure about the space at the upper side of the symbol either, but I checked a few ADCs in the library and it seemed like they intentionally left a space there. Anyway, I moved up the pins.
Should I move the bottom side up a little as well? (The would make the symbol asymmetric, but probably it would be better still.)

I wasn't sure about the other thing you mentioned: did you mean that we should make a space on the left side between the analog inputs and the REFINOUT pin? I made it so, does it look better?

evanshultz:
Thanks! The spacing isn't controlled and so it can vary, as far as I've ever seen and read. Yes, I was thinking moving those pins down at the bottom left like this:
image

andrasfuchs:
image

evanshultz:
Thanks! I appreciate you creating such high-quality symbols and footprints! We look forward to more contributions from you in the future!

andrasfuchs:
@evanshultz Thank you for your work as well!

PR #460, opened 2018-04-06
Updated symbol for Atmel/Microchip SAME70 Cortex-M7 based MCU

andrasfuchs:
Updated a few things about the ATSAME70Q21.

I used these datasheets: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-11296-32-bit-Cortex-M7-Microcontroller-SAM-E70Q-SAM-E70N-SAM-E70J_Datasheet.pdf  and http://ww1.microchip.com/downloads/en/DeviceDoc/60001527A.pdf

image

------------


Misca1234:
Hi

You can use the "short name principle" for CPU's, that is, shortest possible name, for example, just use "PA14" for pin 14.

There is a desire to stack (power) pins with the same electrical functionality ontop of each other, such as the GND pin. An exception is if the datasheet require a decoupling capacitor connected between, let say the GND pins and the VCC pins.

You have only pushed the symbol, have you updated the symbol description (in your push no dcm file is  affected)?

I also object a little to the names you using
for example "ATSAME70Q21A-AN", this combination does not exist in the new "microchip official data sheet", it is named "SAME70Q21" and in the earlier atmel data sheet it is named "ATSAME70Q20A-CN".

andrasfuchs:
Hi,

1,
Ok, so the standard pin naming in KiCad is that none of the secondary functions should be shown? 
I think I keep them in my local library because it speeds up my work, but I will remove that change from this pull request, if you think it's better.

2,
I didn't know that, thank you for explaining, I'll change the VCC and GND pins back to the original form.
I changed their position and I also had to change the size of the rectangle and I changed the pin type from passive to power input. Should I change them back too?

3, 
I agree, it should be just SAME70Q21.

Misca1234:
@andrasfuchs 
Hi

Regarding the ATSAME70Q21
The document file name (data sheet) link in the decsiption is no longer working, i guess Microship organisation have settled a bit now after the purchase of atmel.

The stacking of GND pins is not there

Keep the separation of VDDCORE, they need each a decoupling capacitor, page 2136
http://ww1.microchip.com/downloads/en/DeviceDoc/60001527A.pdf

Regarding the foot print
I guess you made this symbol for the LQFP 144 pin
The only foot print available is the Package_QFP:LQFP-144-1EP_20x20mm_P0.5mm_EP4x4mm

Where are the "translation table" for the pin and its function for the 144 pin, page 30
http://ww1.microchip.com/downloads/en/DeviceDoc/60001527A.pdf
only show the BGA pinout

Do you know if the Atmel chip have thermal pad beneath it?
If so, and the LQFP-144-1EP_20x20mm_P0.5mm_EP4x4mm can be used, then you need to add the EP1 pin (next to the GND) but I can not find a description of any thermal pad.

If the Atmel chip does not have thermal pad beneath, we need to make a new foot print.

andrasfuchs:
Thanks @Misca1234 for taking the time to look into this again.
I haven't used this symbol in the last 5-6 months so I don't think I will make the changes. 
You can discard my pull request, feel free to close it.

PR #462, opened 2018-04-08
Added TPS731 LDO series as aliases of TPS736 series

awygle:
This PR adds the TPS731 series of LDOs as aliases of the TPS736 series.

The only difference between the two is that the TPS731 series only supports 150 mA output current, vs 400 mA from the TPS736 series.

Datasheet: http://www.ti.com/lit/ds/symlink/tps731.pdf

------------


evanshultz:
Thanks for adding these symbols! I appreciate you spending time to improve KiCad's library and hope you will continue to do so!

PR #463, opened 2018-04-08
Added DAC081C081CIMK 8-bit DAC from TI

awygle:
Added 8-bit micropower DAC DAC081C081CIMK.

Datasheet: http://www.ti.com/lit/ds/symlink/dac081c081.pdf

image

------------


Misca1234:
All looks good to me
(beside a dead small detail, the footprint reference is ontop of the symbol name, perhaps the 
footprint reference should be moved so it does not cover the symbol name)


Are you thinking of adding the other two others
DAC081C081CISD, DAC081C085CIMM ?

I bet DAC081C085CIMM can use the already existing footprint
VSSOP-8_3.0x3.0mm_P0.65mm in Package_SON

It is trickier for DAC081C081CISD, it seems that there is no matching foot print so in that case you 
have to a foot print as well if you want add it.

awygle:
I wasn't planning to include the other packages. Do you need me to do that before merging this? How about moving the footprint reference?

Misca1234:
> I wasn't planning to include the other packages. Do you need me to do that before merging this? 

No you dont, it was only a suggestion from my side, I was not sure if you had poked around the footprints or not and might have missed the existing one.
If you want to add them you can do them in separate push.

> How about moving the footprint reference?
It is up to you, it does not affect the functionality or so, it just look better if you bring the symbol up into the editor.

awygle:
I moved the footprint reference.

awygle:
Preemptively changed pin name position offset and made SCL an input on the basis of feedback on other symbols.

awygle:
Anything else you would like me to do for this PR?

whitequark:
I think SDA and SCL should be moved to the right, like this was done in https://github.com/KiCad/kicad-symbols/pull/467.

whitequark:
Updated SDA/SCL pin positions.
screenshot_20180417_222544

evanshultz:
I'll jump in and take this PR.

- Sorry for the bad advice above. I2C pins should be where they were. This is a DAC but the recommendation came from an ADC. Sorry about that.
- Footprint filter should be `SOT?23*`. This will avoid catching SOT-223 footprints.

Otherwise this looks good to me. Thanks!

whitequark:
Done
screenshot_20180420_164454
screenshot_20180420_164445

Do I understand it correctly that the location of the I2C pins correspond to the logical direction of the data flow on the bus? I.e. even though SDA/SCL are bidirectional and thus should be on the right as per KLC, the DAC receives input via I2C and so should have them on the left, whereas the ADC produces output via I2C and so should have them on the right?

As an aside, I find it odd that you require SCL to be specified as an input given that clock stretching is a part of the protocol.

evanshultz:
Thanks!

Correct. http://kicad-pcb.org/libraries/klc/S4.2/ shows inputs on the left, outputs on the right. A DAC converts digital to analog, so I2C (in this case) on the left.

Clock stretching again. This discussion has been around and around lately. SCL should be Input type and if you'd like to see the (my) rationale from the most recent discussion it's at https://github.com/KiCad/kicad-symbols/pull/350#issuecomment-379821677. If you have a stronger counterpoint then please share it. I'll leave this open for a few hours in case you want to comment before I merge.

whitequark:
@evanshultz I agree with your reasoning regarding the SCL pin on I2C slaves. I think SCL being an Output on I2C masters is wrong but it has nothing to do with this PR, and ultimately I think the Kicad ERC engine (as well as most other existing ERC engines in eCAD) is too crude to be a tool you can actually rely on, but the libraries are the wrong place to have this discussion.

So, please go ahead and merge.

evanshultz:
Thanks!

Agreed. The whole constraint system in KiCad (schematic and PCB) is, in my opinion, very weak. But you're right this isn't the place.

PR #468, opened 2018-04-08
Added CAT24M01 EEPROMs from OnSemi.

awygle:
Added CAT24M01 EEPROMs from OnSemi. The DFN package is not included because none of the existing packages match the EP dimensions, but the SOIC, SOIJ, TSSOP and DIP are all included.

Datasheet: https://www.onsemi.com/pub/Collateral/CAT24M01-D.PDF

image

------------


evanshultz:
Thanks!

Please set pin name offset to 20mil.

Because the symbol for the DFN package will need to be unique, I think it's best to identify the packages uniquely. So you would have `CAT24M01L` with ALIASes of `CAT24M01W`, `CAT24M01X`, and `CAT24M01Y`. Can you do that?

awygle:
I can do that, but I'm not sure how footprints work out in that case. Should I leave the footprint field blank and the footprint filters as-is? Or can aliases have unique footprints assigned to them?

evanshultz:
Yes, you can leave off the default footprint and have several footprint filters. An alternative approach is one symbol for each package type. Each ALIAS does not have unique footprints. It's up to you.

evanshultz:
Thanks!

2 things:
- There should only be one package shown in each description.
- Add pin pitch to the footprint filter for all symbols.

awygle:
Made those changes. The DIP footprint doesn't have a pin pitch, so I didn't add one to the footprint filter for that part.

Also I moved the footprint field off of the symbol:

image

evanshultz:
There's no requirement for the footprint field to be anywhere, so it was OK before and it's still OK now. All the changes look good. Thanks for continuing to contribute to KiCad!

PR #469, opened 2018-04-09
Fix MCP7940N

evanshultz:
While posting at , I realized this symbol was wrong. It has 8 pins, but was intended to use with a package that has an EP (and so it needed 9 pins). Even though there's probably only one person who is using this symbol, I wanted to fix it. The best way seemed to be splitting up the symbol into an 8-pin and a 9-pin version and then fixing the footprint filters.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/20005010F.pdf

Image:
image

------------


Timer.dcm
```diff
@@ -120,8 +120,32 @@ K timer
 F https://www.onsemi.com/pub/Collateral/MC1455-D.PDF
 $ENDCMP
 #
-$CMP MCP7940N
-D Real-Time Clock, I2C, Battery Backup
+$CMP MCP7940N-MNY
```
Ratfink: Should say `MCP7940n-xMNY`.  Right now the symbol editor gives an error about this.

evanshultz:
OK. Done now. Stupid text editor (obviously not me!)...

Timer.lib
```diff
@@ -252,6 +252,30 @@ X VCC 8 0 400 100 D 50 50 1 1 W
 ENDDRAW
 ENDDEF
 #
+# MCP7940N-xMNY
+#
+DEF MCP7940N-xMNY U 0 20 Y Y 1 F N
+F0 "U" -250 350 50 H V C CNN
+F1 "MCP7940N-xMNY" 450 350 50 H V C CNN
+F2 "Package_DFN_QFN:DFN-8-1EP_3x2mm_P0.5mm_EP1.75x1.45mm" 0 0 50 H I C CNN
```
Ratfink: The EP size here and in the FPFilter doesn't match what the datasheet shows.

evanshultz:
Right. Forgot to mention that above. 0.01mm less in width but bigger in height. So it should work. And with this PR I wanted to pick a usable footprint in the interest of "fix this silly bug quick so I can spend time on more important things". Is that OK with you?

Ratfink:
I'd rather you set the correct EP size here and create a bug report against kicad-footprints (tagged "help wanted") for the missing footprint.  That way we won't forget that we should make a footprint for this symbol some time, and we won't have to change the symbol when we do.

Ratfink:
Thanks for fixing this symbol, and for making the right footprint!

PR #470, opened 2018-04-09
Added INA260 from TI - High/Low-Side Digital Current/Power Monitor with Integrated 2mΩ Shunt Resistor

rs-development:
I hope its not too bad work, its my first Contribution at all!

Added the INA260 - High/Low-Side Digital Current/Power Monitor with Integrated 2mΩ Shunt Resistor

[Link to datasheet](http://www.ti.com/lit/ds/symlink/ina260.pdf)
Footprint is linked to the the official footprint library.

image

------------


evanshultz:
Thanks for contributing to KiCad! The library can only grow and improve with more people helping!

First, you have added backup files and other things we do not want. The LIB file is for the symbol itself, and when you fill out the description, keywords, and datasheet link that will update the corresponding DCM file. When adding a new symbol, both the LIB and DCM files must be updated.

- Pin name offset should be 20mil.
- I think the VBUS, IN+, and IN- pins should be Passive type.
- Capitalize pin 7's name.
- Pin 7 should be Open Collector type (we don't have open drain).
- Pin 9 should be Input type.
- Add pin 13, making it No Connect type and invisible. The end should be on the border of the symbol so it's unlikely to be connected.
- Put VS and GND pins on top and bottom.
- I wonder if VBUS, IN+, and IN- would be best on top and bottom too. What do you think?
- Add `, TSSOP-16` to the end of the description
- It looks like there's a space at the front of your description

jkriege2:
Note that a comparable symbol has just been merged: https://github.com/KiCad/kicad-symbols/pull/475

rs-development:
So i will adapt my Pin Placement as in the INA226 for easy Part swapping?

evanshultz:
Thanks for making updates!
- Sure, I guess align the pins. (FYI, if the part have an identical pinout they can be ALIASes of each other. But that's not the case here.)
- Stack pins 6 and 11 and put them in the middle of the symbol.
- It looks like pin 10 is inside the symbol.
- Shrink the symbol vertically (this may be affected by item 1 above, so matching symbols as much as possible is OK too.)

rs-development:
So after a lot of time not working on the schematics i finally found time to fix the symbols.

- Stacked the two ground pins
- Moved the pins to match the INA226 footprint

Witch pin type is the correct type for the A0 and A1 pins on these devices, i made them input types, but they have been marked as passive in the INA226 symbol

evanshultz:
- The ALERT pin should not have an overbar. It is shown without it everywhere in the datasheet and page 25 shows that the polarity is user-selectable. (This is wrong on INA226.)
- A0 and A1 should be Input type. That was overlooked on INA226.

Image (the uppercase vs mixed case and pin name differences are correct from the datasheets):
image

@jkriege2 @Ratfink 
INA226 is the `Analog_ADC` library. Both that and this one are intended for sensing current, but they are also ADCs. I think `Sensor_Current` is a better lib since that's the use for this part and it's more specific than just an ADC. Right? Some fixes, noted above, need to be made to INA226 anyway so if you agree it can be moved at the same time.

Ratfink:
From a quick reading of the datasheets, INA226 and INA260 are both current *and* voltage sensors.  I don't think that Analog_ADC is the right library for them, but putting them in Sensor_Current would ignore a major function.  I feel like they belong in Sensor, as the description says that library is for "Multi-function sensors[…]".

jkriege2:
I would agree on the library. Also: Could you move the labels, so they do not interfere with straight-up wiring of the power-pins, e.g.: 
2018-05-01 07_47_53-symbol library editor - analog_switch d__kicad_kicad-symbols_analog_switch lib

Misca1234:
This push have been taken over, see new push
https://github.com/KiCad/kicad-symbols/pull/1080

PR #471, opened 2018-04-10
Add bq27750

vitperov:
Datasheet:
http://www.ti.com/lit/ds/symlink/bq27750.pdf

bq27750


Ratfink:
Thanks for the contribution!  I have a few comments and suggestions.

* The datasheet seems to say that the power pad must always be connected to VSS.  Because of this, it would be good to rename it to VSS and stack pins 1 and 13.
* As best as I can tell, the BAT pin is only used as a power input to the chip.  If this is correct, please change its electrical type to "Power input" (but if I'm wrong about that, please show me).
* I'd suggest moving CHG up so it's right next to DSG.
* I tried building the example schematic from the datasheet, and found a few possible sticky points with the pin layout you chose.  Below is a screenshot with your symbol on the left, and a pin layout I came up with that seems a bit nicer on the right.  I only think the first of these changes is particularly important, the others are just suggestions that you can use as you see fit.
  * I think PACK should be on top because it's a power input to the chip (technically BAT should be too, but that would just make the schematic messier, so I left it by VCELL/INT).
  * I moved TS1 near the battery since it's used to measure the battery temperature.
  * I put PBI near the bottom because although it's technically a power input, it just gets a capacitor to ground, so putting it near the bottom makes the schematic nicer to draw.  The "Passive" electrical type you have is fine for this pin.
  * Moving SRN and SRP to the left makes the schematic a bit neater, more similar to the datasheet's example, and with all the other changes I have, it balances the number of pins on the left and right, allowing the symbol to be a bit smaller.

screenshot from 2018-04-16 11-00-30

The electrical types I've suggested make it so unless PWR_FLAGs are added, there are ERC errors on the PACK and BAT pins in the datasheet's typical application schematic.  @evanshultz @poeschlr @jkriege2 Is this okay, or should the types be something different?

evanshultz:
It would be reasonable to need a PWR_FLAG on the BAT pin because a battery can be a power source.

PACK is an analog input which is used to sense the input voltage. It is also the input source for battery charging. And the power source for the IC. I always see a series resistor between the voltage input and the PACK pin. Based on that, Input or Passive would seem to be an appropriate choice and avoids the need for a flag. If PACK does not always need a series resistor, Power Input would also be a possible choice. I searched around for info on this and find on page 104 of http://www.ti.com/lit/ug/sluubi6/sluubi6.pdf that the default resistance is 0ohm for PACK, and by setting this would allow a proper measurement of the voltage to the FETs even with a series resistor on PACK. So I'm a little conflicted, but avoiding a power flag on the PACK pin seems like a relatively more convenient option for the user without creating any other issues than requiring it. (The source voltage would need a PWR_FLAG unless it comes from a Power Output pin, but because of the resistor that still doesn't help the PACK pin.)

BTW, pin name offset should be 20mil.

Misca1234:
@vitperov 
Will you look into the comments above ?

antoniovazquezblanco:
@vitperov ping!

antoniovazquezblanco:
No problem. It is just a matter of assessing wether people abandons PRs or not. No hurry, we now know you are alive. Thanks! :)

poeschlr:
I finish this in https://github.com/KiCad/kicad-symbols/pull/1350

I hope nobody is mad at me for this.

vitperov:
oh, sorry for inactivity!

@poeschlr Thank you!

PR #472, opened 2018-04-11
Added Micrel MIC5355-S4YMME Dual LDO in MSOP-8 Package.

awygle:
Added Micrel MIC5355-S4YMME Dual LDO in MSOP-8 Package. I intend to add the other variations of the part after getting broad agreement on this symbol.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/mic5355_6.pdf
Screenshot: 

image

------------


evanshultz:
Thanks for submitting more symbols!

- The thermal pad must be connected to ground so stack it with pin 2.
- Then you can shrink the symbol horizontally a bit.
- To me, it might look nice to have EN1/2 and VOUT1/2 across from each other. What do you think?
- Description is missing `Micropower Linear` from the datasheet.
- Remove `*1EP` from the footprint filter (the pin count will filter by footprints with a thermal pad).
- The datasheet says to use as big of a thermal copper pad as possible. Not a very exact description. It also shows vias. Would you be willing to make a new footprint, at least with thermal vias, and possibly with a larger thermal pad?

It's a little annoying that the Ordering Information table on page 2 shows a more descriptive name in the Part Number column, giving the output voltages, but the Manufacturer Ordering Part Number column is obfuscated. Oh well. That's what's shown on Microchip's website and on Digi-Key.

When you add other parts, it seems all MSOP can be ALIASes. Were you also consider to add a symbol for the MLF package?

awygle:
The reason I didn't stack the thermal pad with ground is because of this line from the KLC, section:
"Thus, pins can only be placed in the same location under a very specific set of circumstances: [...] 3. Pins must have the same name". Is that intended to mean "if the pins have different names, don't stack them", or "if you stack pins, name them the same even if the datasheet doesn't"?

I was planning to add the MLF package symbols. I suppose I can do a new footprint as well.

evanshultz:
Good question. If the pins are stackable, they must be named the same. Often the thermal pad pin name isn't given or various between manufacturers, so if it gets stacked we give it the name of the pin on which is stacks.

Thanks!

awygle:
I made the above changes, added the aliases, and added the DFN parts as well. I'll add one more commit once I've made the footprint changes requested.

evanshultz:
- Thermal pad must be numbered as pin 9 on both symbols.
- Update the MSOP footprint to the one with thermal vias once we've settled on a name at https://github.com/KiCad/kicad-footprints/pull/483.
- The datasheet shows the MLF should have a thermal pad of 1.55x2.3mm. The one you picked is 0.1mm bigger. I know it seems pedantic, but would you mind making a new footprint to exactly match the datasheet? And please also make a version of that new footprint with vias like in the datasheet.

evanshultz:
Please update this PR to use the MLF instead of DFN package name and it is good to go. Thanks!

whitequark:
Done

PR #474, opened 2018-04-11
Added PIC18F46K22-I/PT

thomasonw:
Please find addition for  PIC18F46K22-I/PT

CPU reference is: https://www.microchip.com/wwwproducts/en/PIC18F46K22


thomasonw:
OK, erased my local copies of all KiCAD libs, created a new fork and added back in the PIC18F46K22-I/PT.  Still getting lots of Travis errors, but again none related to any changes I made.

How to proceed at this point?

thomasonw:
And still issues with how best to work GitHib.  I have a few other changes to submit (e.g., the INA226 above), but it appears I am not able to make them one at a time until prior pull requests are completed.   Any guidance appreciated.  (Am using the Windows GitHub app)

evanshultz:
Please limit the breadth of the PR. Add one symbol to a branch and push it. Add another symbol to another branch and push it. It seems you're using just the master branch instead of creating new ones and that's going to get messy very quickly.

evanshultz:
If you leave this alone I can review the PIC18 MCU. The existing errors that check scripts report can be left alone. I'll only focus on the MCU.

Then create a new branch for IN226 and push it. I'm sorry that I'm not familiar with the interface, but it looks like https://services.github.com/on-demand/github-desktop/create-branches-github-desktop may be what you should do.

thomasonw:
@evanshultz   Thank you for both points, will see if I can figure out 'Branches' from the desktop app.

evanshultz:
Thanks! It seems you got branches figured out because you submitted more PRs.

- Please always include a screenshot in your post. Perhaps your GH client doesn't support it, but there is a default list of items when a PR is created and a screenshot is one of them.
- Page 5 of http://ww1.microchip.com/downloads/en/DeviceDoc/40001412G.pdf shows just one pin name for each pin. Can you make that the first function in the pin name? We have started using just the short name (`RC7`) instead of the long one (`RC7/RX1/DT1/AN19`) unless there are critical pin functions, but for just adding this one symbol I can allow the longer name.
- We wildcard temperature and packaging from the MPNs, so name the symbol `PIC18F46K22-xPT`.
- NC pins should be named `NC`, not `N/C`.
- Pin 15 has a double slash in the name.
- Pin 18 is missing the `VPP` from the pin name.
- All GPIO pins should be Bidirectional type: wrong ones are 18, 27, 30, and 31.

thomasonw:
Thank you for the feed back, and does seem I have a starter on Branches - but still not Github.  This submission is no longer contained in my local image!  So - with that, and that it was incorrectly started w/o sing a branch I am going to close this submission  - and open a new one using a proper branch as well as with the edits above.

Thank you again for your helping me through this, do want to support KiCAD libs vs. making another massive local to me only files.

PR #475, opened 2018-04-11
Adding INA226

thomasonw:
This pull request is for the addition of the INA226

Datahsheet is located here:
www.ti.com/lit/ds/symlink/ina226.pdf

image


jkriege2:
Hi!

thanks for your contribution. Could you please change a few things, so this matches with other symbols in this lib and follows all the KLC rules?
1. Please mirror the symbol, so the analog inputs are on the left and I2C and other digital-pins on the right. This orientation is used by most ADCs in the lib.
2. The FPFilter should be `*:MSOP*3x3mm*P0.5mm*`
3. Please remove the commas from the tags/keywords
4. Please add `, MSOP-10` to the end of the description 
5. Finally there is a typo in the description analor --> analog. Pleae correct
6. The datasheet link should be http://www.ti.com/lit/ds/symlink/ina226.pdf
7. I think the description should be more aligned with what's in the datasheet-title, i.e. `High-Side or Low-Side Measurement, Bi-Directional Current and Power Monitor (0-36V) with I2C Compatible Interface, MSOP-10`

Thanks,
JAN

thomasonw:
Jan,  Thank you for the coaching.  I have made the corrections and pushed them up.

Ratfink:
I'd like to suggest shortening the pins to 100 mil, and reducing the pin name position offset to 20 mil.

thomasonw:
Like this?   Yes, I never have understood why the default pin length was so long…

jkriege2:
Nice work!

THanks,
JAN

thomasonw:
OK, thanks to all - on to more!

PR #476, opened 2018-04-11
Added TS5A3166DBVR

thomasonw:
Adding TS5A3166DBVR analog switch.

Datasheet is located here:  
[http://www.ti.com/general/docs/lit/getliterature.tsp?genericPartNumber=ts5a3166&fileType=pdf](http://www.ti.com/general/docs/lit/getliterature.tsp?genericPartNumber=ts5a3166&fileType=pdf)

image


jkriege2:
Hi!

thanks for you contribution.
1. Please follow the style set by other devices in the same lib. Generally a closing switch should look like this:
2018-04-12 07_09_09-symbol library editor - analog_switch d__kicad_kicad-symbols_analog_switch lib
2. Also please adhere to the style with separate power-symbol, as e.g. done for `TS5A3159DBV`, which is already in the lib.
3. Please add the package to the description as done for all other analog switches in the lib and maybe also the Ron
4. datasheet link should be http://www.ti.com/lit/ds/symlink/ts5a3166.pdf
5. no commas in tags/keywords
6. Finally would you be willing to also add one or the the two other package variants DCKR (`Package_TO_SOT_SMD:SOT-353_SC-70-5`) and YZPR (the footprint dows not exist yet)? That would be much appreciated (but is just a question ;-)

Thanks,
JAN

thomasonw:
Jan,  Thank you for the review.  I can do the changes but do have a question:  in the guidelines it states packages with only one module should include the power and gnd pins and not use the A B approach, which is why I made this module the way it is; even though it is different from other existing single-switch modules. . . .

ref: [http://kicad-pcb.org/libraries/klc/S3.8/](http://kicad-pcb.org/libraries/klc/S3.8/) "...._This approach should not be taken for single-unit symbols. Power pins in this case must be drawn on the same symbol as the logic pins_."

So my question is:  As i read it, those existing single-switch  modules are wrong.  At this point is it better to follow the existing modules, or the overall guidelines?

And yes, once I get a better handle on how to do the packaging and filter creation I will be happy to  create the other variations of this device.

jkriege2:
Hmmm ... didn't remember that rule (especially when creating the other switches) ...

@evanshultz @SchrodingersGat @Ratfink @poeschlr @pointhi @Misca1234 What do you think? How should we handle this? Stick to KLCor maybe change KLC to have all such things with power-symbols the same (whether it is 1,2 or more units)? I would like that approach, as it would mean having ONE style and makes it much easier to manager different power-signals for different symbols (I think) ...

Ratfink:
I vote to stick to KLC here.  I'd prefer not to force power units onto everything simply because one symbol for that kind of part has them.  If we made the change you're suggesting, that would also require e.g. 8-pin microcontrollers to have separate power units because some 200+-pin microcontrollers have them, and I'm not convinced that splitting up such a tiny part would help anyone understand a schematic.

jkriege2:
@thomasonw a few small thngs before we come to a conclusion on he overall symbol appearance:
1. description should be `Single SPST Analog Switch, 5-V/3.3-V, normally OFF, 0.9Ohm Ron, SOT-23-5`
2. please remove the package type from the tags/keywords

Thanks,
JAN

jkriege2:
@Ratfink @evanshultz @SchrodingersGat @poeschlr @pointhi @Misca1234 Please comment on how we should proceed with with the symbols: Should we have separate power-units only for devices with >=2 analog switches and then have a specialized single-unit power-pins for device with only one switch (note that then we should move the control-pin to the left on these and that they will be significantly bigger than the other symbols for devices with more units)

@Ratfink already stated his preference. Although I think we're not talking about µCs here ... having these power-pins will only help for a very limited set of symbols. Currently I would make them mandatory only for "small" device like logic gates/flip-flops/..., OPAMPs, analog switches ... so if we would reformulate KLC I would propose to limit this requirement and leave it open for other more complex symbols, µCs and othr ICs. The rationaly is that such devices then all have the same style (whether I use single-gate logic or a 7400 and whether I use 4 single-OPAMPs, or an OPAMP in a 4-unit package). Also it make (IMHO) management of different power-planes (especially for OPAMPs and Logic) much easier, as power-supply can be grouped at one location. Finally for the special case of these switches, I think haveing the control-pin at the bottom is quite neat and useful to structure the schematic (control vertical, controlled-lines horizontal) + comparable symbols are used widely.

Awaiting your votes,
JAN

SchrodingersGat:
I personally think that power ports should only be split from a part if the part is **already** split for some good reason, and there is no obvious way to determine *which* unit should house the power pins.

If this were a part with 4 separate switches, which switch gets the power pins? In such a case, a separate unit for the power pins makes sense.

For a single switch, I think adding the power pins to the same symbol is the easiest way forward. Splitting a part into multiple units should be treated carefully. We do **not** want to require this for all parts. Only where it is *useful*

thomasonw:
@jkriege2   Made corrections, also updated snapshot in header of this thread.

jkriege2:
OK, so the consensus seems against splitting ... So please move the control-input to theleft of the symbol ... and if possible make the symbol as close to the other existing switches as possible. When done please also update your screenshot.

Best,
JAN

thomasonw:
OK,  I have edited it to get this as close to the other SPST - NO single switches as I can.  Not sure I understand the move the control to the left of the symbol.  I have it now like this to match the other switches:

image


Are you looking for me to move pin 4 to below pin 1 on the left?

thomasonw:
How does this look...

jkriege2:
looks good. Maybe two very minor things:
1. I think the left end of the bar between the circle should touch the left circle
2. Is there a reason why you don't have a sharp corcner in the dashed line?

thomasonw:
1) Can make it touch, no issue.
2) I just thought it just looked nicer.  Will make it square if you wish.

jkriege2:
OK, one final modification is required: As Travis indicates, pin4 is not on a 100mil grid. Please move pin4 50mil upwards to move it onto the grid.

Best,
JAN

thomasonw:
Just reviewed the Travis report from the last submission.  It has a few notes, though I think most are related to the common look-n-feel of the analog switches.  The one comment about centering:  I considered that - but to get this symbol centered I would have to make it even larger then it is - perhaps grow it 100mil on the top.  Do let me know if any of these need to be addressed.

jkriege2:
The centering is the least important of our rules ... e.g. putting things on the 100mil grid takes precedence!

jkriege2:
nice work!

Thanks,
JAN

thomasonw:
@jkriege2   JAN,  Thanks for your guidance over these past few days.  Have learned a lot and hope to be able to more smoothly contribute to KiCAD going forward.   Have a couple more packages to add, and will also add the other footprint versions of this switch.

Thanks again,

jkriege2:
Looking forward to your PRs ... let me know if you need any help!

JAN

thomasonw:
It looks a bit crowded, and seems to be a bit ambiguous on what is being referenced:

image


Could grow the package some to allow larger names, but am also noting S3.2 allows for names as small as 20mil for smaller parts.

Leave as is, or grow the package seems the two options I would say.  I have no strong feelings one way or another - though will point out the current outline size matches the rest of the Analog switches.

Issue #298, opened 2018-02-12
US-style resistor


jkriege2:
It was discussed over in the forums several times that many US-people wish for a "wiggly" resistor symbol (IEC prefers the box-symbol).

There was the idea to add these as alternative via the deMorgan-alternative-function in the symbol editor (I don't think we really use it for deMorgan-version of the logic gates).

So the question is: 
1. Should we do that at all?
1. And if so:
  a. only the basic resistor ~~+ pots~~
  b. only the basic resistor + pots
  c. basic resistors, pots, sensors (NTC, PTC, LDR ...)
  d. all, including resistor networks
1. If we do it: How should we handle new contributions (also internal schematics of ICs and transistors ...): 
  a. box-required, wiggly as an additional option
  b. either box or wiggly 
  c. both

I would say we start with 2a and require 3a (just my personal preference)

Best,
JAN

PS: The same problem might arise with filled/unfilled diodes and polarized caps (rounded/straight plates vs. filled/unfilled plates)

bobc:
2a and 2b seem to be identical?

The logic gates I generated have de Morgan equivalents, I think other libs such as 74xgxx do not.

We should probably think about how to handle alternative shapes in the proposed SWEET format, and what changes to make in KiCad. For example, as a user, I might want to select a "US style" theme, and thus get the US style as default (instead of having to click each part and select "Converted shape").

Currently we have different libs for ANSI style logic gates and IEC style gates, that might a better way to manage it in the short term.

poeschlr:
I am not sure it is worth our effort. I think there are more important issues to take care of. 
These changes only include artistic improvements for users who prefer the us style. (No user in the thread that prompted this issue complained that they are unable to read a schematic with the box style resistors. They complained that they like the other type more.)

And i fear if we add a us style for some symbols users will start to complain why only for these and not for all others. The next complained will be why the European style is used as the default, ....
(And yes i was more optimistic about the community back when this original discussion took place. That's what being a lib maintainer for a year or two did to me.)

---

Just to clarify: I am not opposed to it, but i think other issues would be more important. (I would give nearly other open issue in any of our repos a higher priority.)

pointhi:
My position on this issue would be: 
1.) Waiting for the new symbol format, and trying to get a feature supporting this usecase into it
2.) Changing the KLC to something like: IEC is mandatory for every symbol, and other styles are optional

poeschlr:
[...]

PR #480, opened 2018-04-12
Res us

Vlad-23:
US Resistor symbols (Networks Varistor Thermistor POT) was added to device lib. POT* was renamed to R_POT*

------------


Ratfink:
I guess I'm not opposed to it, and since we already have alternate styles for e.g. diodes, I suppose there's already some precedent for this.  I would want us to settle all the questions in #298 (that is, we should take all discussion that isn't specifically about this PR's changes over to #298) before merging this though.

evanshultz:
@Vlad-23 
It's requested to attach an image of all symbols or footprints submitted. Could you do that for ease of discussion here? It would be nice to include the box style and US style for comparison in your image.

Vlad-23:
Hi!
Image as you requested. I also included box style for comparison (except for networks), as you mentioned. I think _US prefix is fine because it's commonly called "US-Style" or "American". So it's more obvious than _ALT for most people.
[R_US_SCR.pdf](https://github.com/KiCad/kicad-symbols/files/1909243/R_US_SCR.pdf)

evanshultz:
image

jkriege2:
Ok, I would also be in favour of merging this ... but before we can proceed it would be good to fix some smaller style/implementation issues:
1. please shorten the pins to multiples of 50mil (e.g. there are a log with length 102mil). If you need lines in between, please add them as simple lines.
2. I saw you added `R_POT_US`, but no version `R_POT_TRIM`... is there a specific reason why you didn't add that quite commonly used symbol?
3. You added Thermistor_US and Thermistor_NTC_US, but no PTC variant ... could you please add that one too? Then we have the full basic set of thermistors
4. Please remove the FPFilters `SM0603` and `SM0805` from `Thermistor_US` (and `Thermistor`)

Best,
JAN

Vlad-23:
Hi, JAN

I have fixed issues that you had mentioned. Can you please check it now?

res_scr

jkriege2:
@evanshultz @SchrodingersGat @pointhi @poeschlr @Misca1234 @Ratfink 
I would merge this PR with the US resistors now, because already the limited set of symbols is a good start here. Before I would like to check two things with you guys:

1. Are there any more objections from your side(s)?
2. I think the rename POT_... to R_POT_... in this PR is fine and helps to organize them symbols a bit better in device.lib. Any objections on that?

Best,
JAN

jkriege2:
Thanks for the work you put into this!

JAN

PR #456, opened 2018-04-05
Adding PIC18F46K22-I/PT  and 6P6C Rj12 w/o sheild 

thomasonw:
Please find two additions to the footprints (Sorry, I do not know how to separate them into individual pull requests at this time).

CPU reference is:  https://www.microchip.com/wwwproducts/en/PIC18F46K22

RJ12 (6p6c) is a modification of current RJ12 connected, with the two shield pins removed.

My working knowledge of GitHub is very basic, so there may well be other issues with this request - if I need to make process changes, do let me know.

PR #481, opened 2018-04-13
Adding PIC18F46K22-xPT

thomasonw:
Please find addition for PIC18F46K22-x/PT

CPU reference is: https://www.microchip.com/wwwproducts/en/PIC18F46K22

screenshot from 2018-05-03 04-45-02

I have had issues properly using GitHub and branches, for prior conversations of this submission please refer to Pull Requests:  #474 and #456  - both of which have been closed.


thomasonw:
Last review comment by @evanshultz 

Thanks! It seems you got branches figured out because you submitted more PRs.

- Please always include a screenshot in your post. Perhaps your GH client doesn't support it, but there is a default list of items when a PR is created and a screenshot is one of them.
-     Page 5 of http://ww1.microchip.com/downloads/en/DeviceDoc/40001412G.pdf shows just one pin name for each pin. Can you make that the first function in the pin name? We have started using just the short name (RC7) instead of the long one (RC7/RX1/DT1/AN19) unless there are critical pin functions, but for just adding this one symbol I can allow the longer name.
-     Set pin name offset to 20mil.
-     We wildcard temperature and packaging from the MPNs, so name the symbol PIC18F46K22-xPT.
-     NC pins should be named NC, not N/C.
-     Pin 15 has a double slash in the name.
-     Pin 18 is missing the VPP from the pin name.
-     All 
- Please always include a screenshot in your post. Perhaps your GH client doesn't support it, but there is a default list of items when a PR is created and a screenshot is one of them.
-     Page 5 of http://ww1.microchip.com/downloads/en/DeviceDoc/40001412G.pdf shows just one pin name for each pin. Can you make that the first function in the pin name? We have started using just the short name (RC7) instead of the long one (RC7/RX1/DT1/AN19) unless there are critical pin functions, but for just adding this one symbol I can allow the longer name.
-     Set pin name offset to 20mil.
-     We wildcard temperature and packaging from the MPNs, so name the symbol PIC18F46K22-xPT.
-     NC pins should be named NC, not N/C.
-     Pin 15 has a double slash in the name.
-     Pin 18 is missing the VPP from the pin name.
-     All 

GPIO pins should be Bidirectional type: wrong ones are 18, 27, 30, and 31.

=================================================

Thank you again for your help, esp on branches.

I have made all the edits above except the "...Can you make that the first function in the pin name? "
As I see in the above image, the primary pin name is nearest the edge of the package - is this not correct?  FWIW, I am MORE then happy to remove all the extra information from each pin - leaving just the primary one  (It is, to be honest, one of the things I did not like about the KiCAD libs:  Those LONG descriptions tend to make for some very LARGE CPUs - ugly ones too..)

Do let me know about the pin names, and also note this Pull request was initiated from a Branch!

thomasonw:
I noted Travis picked up on the NC pins and asked for them to be hidden.  I have updated the submission to hide the NC pins.  Is this ready to be merged?

evanshultz:
Thanks!

You can push from master, and in this case since the review was started it would be easier for the reviewer to keep everything together. Now this branch has lost all the history from the old PR and it's not as easy to see incremental changes. But so it goes. This will still be fine.

- The VDD and VSS pins should be stacked unless the datasheet says otherwise. I don't see it in the datasheet, so please either find that info and point it out or stack the pins.
- Shrink the symbol a bit vertically.
- NC pins should have the "X" on the border of the symbol so it's not easy to accidentally connect them.
- You can shorten pin names if you like, but since this lib still has long names and I already reviewed the long names you can leave it alone too.

evanshultz:
- Please see http://kicad-pcb.org/libraries/klc/S4.3/ for info on how to stack pins. Give that a shot and ask if you're having issues.
- NC pins are not on a 100mil grid.
- Can you also post a new screenshot or update the one above?

thomasonw:
@evanshultz Again, sorry for delay - have made feedback edits and pushed up.

evanshultz:
Almost all good (assuming you're leaving the long pin names). Just one thing: I realized that all pins have 150mil length. Please change all to 100mil and post a new screenshot.

thomasonw:
OK, will do.  Wanted to ask:   Do you want me to go ahead and edit this to shorten the pin names?   Right now it matches the rest of the PIC CPUs, but I get the impressing this is an old-way.  Could shorten the names to just the port name, and maybe even move them to one side  (I think layouts work better that way).

 

Ref the new ATmega64M1 symbol for an example.  If you want to get a head-start on the PIC CPUs towards this format am happy to edit this one at this point..

evanshultz:
We will move to script-generated symbols when time allows and shorten pin names then. So shortening by hand now seems like no value add, at least to me. It's really up to you. In general, long pin names and pins on both sides are the majority of symbols in this lib so I'm happy to leave it alone.

thomasonw:
Happy to leave them then – I had simply copied another existing CPU for this submission, changing the few pins who’s functions differ.

evanshultz:
Great. If you can adjust pin length and post a new screenshot this should be good to merge.

thomasonw:
OK, I am not sure what is happening here in Github.  I did shorten the pin names and you can see the it under the Commits tab above and here:  [https://github.com/KiCad/kicad-symbols/pull/481/commits/99ee6bfa2341b2ff6478a6ade881d5a915919616](https://github.com/KiCad/kicad-symbols/pull/481/commits/99ee6bfa2341b2ff6478a6ade881d5a915919616)

But for some reason is it not showing inline as before..

evanshultz:
I see that commit above. I does appear earlier in our comments that it would seem you pushed it, though. This symbol looks fine to me now and there are no errors with it, so I'm good to merge.

Even though you've submitted several symbols, GitHub says this is your first one. Thank you for starting this journey to add to the KiCad library and I hope to see more contributions from you!

thomasonw:
Thank you, yes, funny about Github (bit it is not my skill area).

Yes, I am migrating from Ver 3 to Ver 5 (skipped ver 4 for a few reasons) and have decided to push any additions up vs. our own local lib/footprints.  Takes more work, but paying back as I can for this great tool.

PR #484, opened 2018-04-15
Add 74hc590

lmoelleb:
Size and text positioning based on existing 74HC595 symbol.

Datasheet: [https://assets.nexperia.com/documents/data-sheet/74HC590.pdf](https://assets.nexperia.com/documents/data-sheet/74HC590.pdf).

image

------------


evanshultz:
Thanks for making your first contribution to the KiCad libraries!!

The rules for the library are at http://kicad-pcb.org/libraries/klc/, and will help you understand all the automated checks done above.

You've fixed a lot of the things that needed to be changed, are here are a few more:
- All pin lengths should be 100mil.
- CPC and CPR pins should be type Input.
- RCO pin should be type Input.
- Use the Line style for all pins, and active low pins have a `~` in front of their name to give an overbar.
- It looks like the symbol can be a little shorter (it's OK if this creates a warning because the off-center check isn't too critical and aesthetics are important too.)
- Can you post a new screenshot? The one above does not match your latest commit and having a fresh screenshot whenever there are visible changes helps the librarians.

Lastly, TI makes a 'HC590A. http://www.ti.com/lit/ds/symlink/sn74hc590a.pdf. Would you be willing to add it too? It's not required, but would be nice.

Thanks again!

lmoelleb:
- [x] All pin lengths should be 100mil

- [x] Set pin name offset to 20mil.
- [ ] CPC and CPR pins should be type Input
_Not understood - they already had electric type input?_
- [ ] RCO pin should be type Input.
_Please confirm even output lines should be marked as electric type "Input"?_ 
- [x] Use the Line style for all pins, and active low pins have a ~ in front of their name to give an overbar.
- [x] It looks like the symbol can be a little shorter (it's OK if this creates a warning because the off-center check isn't too critical and aesthetics are important too.)
- [x] Can you post a new screenshot? The one above does not match your latest commit and having a fresh screenshot whenever there are visible changes helps the librarians.
_Screenshot looked up to date to me, so not sure what is going wrong. Do you prefer I post a new screenshot as a comment instead of updating the one in the PR description?_
- [ ] Lastly, TI makes a 'HC590A. http://www.ti.com/lit/ds/symlink/sn74hc590a.pdf. Would you be willing to add it too? It's not required, but would be nice.
_To me it looks like it could be added as an alias? I did notice NXP and TI using different naming conventions on the pins (so Q0-Q7 or QA-QH). Can I add the alias to the existing symbol using the NXP naming convention - or is it better to move everything to TI naming? Personally I have a preference for the numbers used by NXP, but consistency beats preferences at any time of the day._ :)

evanshultz:
- CPC and CPR pins are type Open Collector.
- My mistake. RCO should be an Output. It's good now.
- A new screenshot above is OK. I may have missed it being updated if I kept the tab open.
- Adding an as ALIAS will be OK. I see mostly TI datasheet links in 74xx.dcm, but some NXP. There are 40 pin Q0 vs 4 pin QA. So let's keep the numbers.

lmoelleb:
No idea why I thought I did not change the CRC/CPR pin types - source control shows otherwise. Oh well, I have lost track of worse things :)
I added the alias, so unless I managed to mess that up, this should be OK now.

evanshultz:
Thanks for adding that ALIAS!

Drat! I just noticed something I forgot earlier: we need to do a bit more work with regards to the packages. The description should have the package(s) at the end and the pin pitch should be in the footprint filters. (In this lib, most do not but it is our standard and will eventually get done here.) So that means you should add the following to the end of each description:
- 74HC590: `, SOIC-16/TSSOP-16`
- 74HC590A: `, DIP-16/SOIC-16/SOIC-16W`
- 74LS590: `, DIP-16/SOIC-16/SOIC-16W`

Footprint filters should be:
`SOIC*3.9x9.9mm*P1.27mm*`, `SOIC*5.3x10.2mm*P1.27mm*`,  `SOIC*7.5x10.3mm*P1.27mm*`, and `TSSOP*4.4x5mm*P0.65mm*`. The DIP footprint filter is fine and you can keep it.

All that being said, the packages are rather incompatible for each part but they're all merged together into one symbol. And some packages (like LCC) aren't applicable to this symbol. A better way would be to have unique symbols with only the packages that work for that symbol. But... this isn't any worse than many other symbols in this lib so I'll merge it if you just make the fixes above.

lmoelleb:
OK. Done. Too bad KiCad does not support footprints per alias... put I am too lazy to start digging into a pull request to fix that :)

evanshultz:
Yep. Slowly but surely getting better.

Thanks again for making a first contribution to the libraries! I hope to see more PRs from you in the future!

PR #487, opened 2018-04-15
Fixed orientation of GNDPWR pin

### LINKED ISSUE #486: GNDPWR has text on wire side, the opposite of gnd
a picture probably speaks for itself here. 
image

The GNDPWR should have it's text moved from top (wire side) to bottom

Ratfink:
The orientation of the pin in the GNDPWR symbol was wrong, causing the automatic label placement algorithm to get confused and put the value text in the wrong place.  This commit fixes it, and in turn fixes #486.

Please excuse the 2.3→2.4 line order changing mess.

screenshot from 2018-04-15 16-16-34

------------


evanshultz:
Thanks!

Weird that PWR_FLAG has no doc fields. Do you want to do that or shall I merge anyway?

Ratfink:
It has doc fields, just no datasheet filename.  Most of the other power symbols have no filename set as well, but Travis doesn't complain because the scripts detect that they're power symbols and don't need a datasheet link.  I'm not concerned about it, so I'd say you can merge.

evanshultz:
We can use `F ~` if there's no link. That's done in `Device.lib` for the same reason: to eliminate bogus Travis errors. Sorry to prolong things... just want to ask.

PR #488, opened 2018-04-15
Adding SC-70 version of TS5A3166

thomasonw:
This is a follow on to #476 -- this pull request is adding the SC-70 package version:  TS5A3166DCKR

Datasheet is located here:
http://www.ti.com/general/docs/lit/getliterature.tsp?genericPartNumber=ts5a3166&fileType=pdf

image

Note:  I also looked at adding the BGA version (_____YZPR, but was unable to locate a matching BGA footprint -- perhaps in the future)


Ratfink:
Looks good!  Thanks for adding this package variant.

PR #489, opened 2018-04-17
Bmp280

ChristianHirsch:
Add Bosch Sensortec [BMP280](https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BMP280-DS001-19.pdf) absolute pressure sensor.
image

I know that I'm violating rule _S4.4_, but I set the pin functions according to the datasheet (page 35).


Ratfink:
Thanks for the contribution!  I have just a couple changes I'd like to see before merging this.

* Please set the pin name position offset to 20 mil.
* Please add a short package name to the end of the description, e.g. "Absolute barometric pressure sensor, LGA-8".
* Why did you set the CSB pin apart from the others?  It's part of the SPI interface, so shouldn't it be grouped with the other SPI pins?
* You could stack the ground pins in this symbol, but given the datasheet's configuration of decoupling capacitors, if you'd rather not stack that's fine too.

ChristianHirsch:
I updated/fixed the symbol. Now it is
- slightly offset because I made it smaller (S3.1)
- violating rule S4.4, but the pin types are set based on the datasheet.
image

Ratfink:
Looks good, thanks for creating this symbol!

evanshultz:
Question: could pins 1 and 7 be stacked? I don't see the datasheet saying this explicitly, but the block diagram on page 11 of the datasheet shows only a single GND connection and all connection diagrams at the end of the datasheet show the two pins connected together. Did anybody ask Bosch?

Ratfink:
I didn't ask anyone at Bosch about the ground connection, no.

PR #492, opened 2018-04-17
Add HYM730 and HYM750 RFID reader modules from Winnix

antoniovazquezblanco:
http://www.winnix.net/uploadfile/20130802154623204.pdf

imagen

------------


evanshultz:
Thanks for adding this symbol!

Please see the CI feedback above. If you can fix those basic issues first it will be easier to do a full review.

evanshultz:
But even more than that, what are the pins for? Are they the 2.54mm mounting holes on page 4 of the datasheet? If so, why are there 5 pins instead of just 4? And if they're all ground they can be stacked.

@Ratfink @poeschlr @jkriege2 
I assume this type of symbol is OK, because placing a symbol allows adding it to the BOM. But I can't recall seeing something quite like this before. Do we have a precedent?

Ratfink:
I don't know, this part is weird.  It looks like all the electrical connections are on a connector, so it needs a cable and a second connector on the main board to get those.  The part doesn't even *need* to be mounted on the circuit board, it could be elsewhere in an enclosure.  I really have no suggestion as to how we should handle a part like this.

evanshultz:
@poeschlr 
KiCad can't know the mating connector. Probably it will be the same family as what is on the module, but the user might want vertical, horizontal, etc. And KiCad couldn't know the connector placement, either. Would a template not be as good as a symbol because some of those decisions would already be made in a template, and might not be what the user wants? I understand some Rapberry Pi hat templates, for example, have parts are placed and routed, but this symbol seems so malleable that any template will likely only be the configuration that the template creator wants.

@antoniovazquezblanco 
But we still need answers about the symbol pins, and all fixes to the symbol, because even a template will need to use the symbol.

antoniovazquezblanco:
This symbol was made by @carlosromorgar with whom I work. He will sign CLA, and will try to answer your questions as soon as he finds some time to do so.

Given our current workload it will probably be a long running PR.

Thanks

carlosromorgar:
Thanks everyone, and sorry for my late response.

As you said, the RF module is connected to the board through an FFC cable/connector. Four of the pins in the symbol are the mounting holes, and the fifth is an expose pad in order to dissipate heat (yellow rectangle in the next image).

This image is the footprint that I have used in some projects:
image

This image is the footprint of the connector  (KiCad/kicad-footprints#561) used:

image

The symbol is useful for the BOM and for the footprint that it is necessary to use in the PCB. Maybe it is a good idea if I submit the footprint?

I don't have experience with KiCad symbols and I don't know if a template is better than a symbol. I expect your comments.
Thank you in advance.

BR

evanshultz:
Thanks for the clarification.

I don't think a template is the right solution because there are too many implementation choices that could be made and a template wouldn't capture it. A template is helpful if there is some foundation for the design regarding parts and placement, and then the user can continue from that point. With this, making those decisions is the only open item and it is design-specific.

It's also possible that all 5 pins are floating. Or it may not sit against the PCB and use standoffs for the mounting holes. My perspective is that this is such a malleable part that a dedicated symbol adds very little value for the vast majority of KiCad users. I'm slightly against adding a symbol, but not strongly against it.

A footprint definitely would be useful, IMO.

antoniovazquezblanco:
If a symbol is not added then BOM will not reflect the used part which is a key feature for us.

Offtopic: Also, how do you add a footprint to the pcb without a symbol? If I do so, whenever I want to update my board from the schematics deleting elements I have to redo a lot of things.

Thanks

evanshultz:
It seems like the question about a part without a symbol is answered? Since this part may not sit on the board at all, I'm not so sure about a schematic symbol. It would seem to fall into the category above. (And yes, I would love if KiCad was more powerful with regards to adding "non-schematic" items.)

Off topic: You can add a footprint just by the normal Place > Footprint menu item. If you then lock the footprint, it won't get ripped up during board updating.
image

antoniovazquezblanco:
We will then have the part in our private repo. Thanks for your input

Issue #473, opened 2018-04-11
Specifying EP size in footprint filters?


evanshultz:
Since we started putting the EP size in the footprint names of applicable footprints (QFP, QFN, etc.), I've seen it be included in the footprint filters. This means the filters are valid for a suffix like `_ThermalVias` or without any suffix, but using the footprint filter will always only show footprints with a certain EP size.

Should the EP size be left off the footprint filter?

If the EP size is left off, the user is then shown a list of all parts matching in package type, XY size, and lead pitch. Any footprints matching those criteria would be shown in CvPcb, which would include variations with different EP sizes. If the user wanted to select an oversize EP for thermal reasons, then they could find and select it easily.

Of course there are drawbacks:
1. If the user selected a footprint with a smaller copper pad than the pad on the package it could possibly lead to poor thermals, unwanted solder bridging between the package pad and copper underneath if the mask was missing or scratched, etc.
2. If the EP size selected is much bigger than the copper pad, and the pin lengths of the package were longer than what the footprint was intended to use, the insides of the package pins could touch the copper pad and create a short. The paste applied on the copper pad would make a solder bridge.

So... I'm on the fence. I see possible positives and negatives. If a default footprint is given, which is think is overwhelmingly true, it seems like a slight net positive to me. But I'm not sure. Any other thoughts?

Ratfink:
I believe the only reason you gave for making this change is because it could potentially give improved thermal performance to have more copper.  Are there any other reasons to want a different EP size than the default?  If the only reason is for better thermal characteristics, I'm not sure why the same effect couldn't be achieved by adding copper zone fills and/or vias to the board in Pcbnew, without creating the potential for solder bridges you described.

evanshultz:
Yes, having a bigger EP was the reason why I thought of it. 2 PRs I looked at today were about the EP size and this popped into my head as I was reviewing.

I expect placing a footprint is much easier, and hopefully more safe, than creating a zone manually. But that is an option. Turning off footprint filtering in CvPcb means potentially trolling through a big list of footprints rather than a list that only varies by EP size.

Also, as I mentioned, I expect most or possibly all applicable symbols would have a default footprint. It wouldn't be required to pick a footprint at all if the user doesn't want to. But if the user wanted to select a different EP they'd have a curated list available in CvPcb already if footprint filtering is turned on.

[...]

PR #493, opened 2018-04-17
Temp sensor

antoniovazquezblanco:
https://www.silabs.com/documents/public/data-sheets/Si7050-1-3-4-5-A20.pdf

imagen

------------


Ratfink:
Thanks for the contribution!  There are a few changes I'd like to see before I can merge this.

* My comments from #491 about fixing Travis errors and removing the generic name apply here as well.
* Please make the symbol less wide.  It could certainly be made at least 200 mil narrower without any ill effects.
* The footprint name you've set isn't quite right, as it should also list the EP size.  I believe it should be called `DFN-6-1EP_3x3mm_P1mm_EP1.6x2.5mm`.  Are you planning on creating the footprint as well?
* Please move the reference text to the top left of the symbol, and the symbol name to the top right.

Ratfink:
Thanks for making these changes!  I see a few more things that I'd like changed:

* The symbol is less wide than it was before, but now it's taller than it has to be.  If you could give it the height it had before, and the width it has now, that would be perfect.
* The new footprint name is still incorrect, as the EP size should be of the pad on the landing pattern, not on the chip itself.
* The old footprint filter (without the number of pins or EP size) was correct.  The number of pins gets filtered automatically, and we decided in #473 to leave the EP size out of footprint filters (but keep the number of EPs).

Given the current flurry of activity from #520, I will have to ask that a proper footprint be created for this before the symbol can be merged.

carlosromorgar:
Hi
Thanks for your feedback. I have just pushed the footprint. It is #552 of footprint repository.
Regards

Ratfink:
Sorry to keep telling you conflicting things about this, but since @poeschlr already made a script-generated footprint for this part in https://github.com/KiCad/kicad-footprints/pull/539, it would be best if you could just make this symbol use that footprint (and a matching filter) instead.  I'm really sorry that I wasted your time in creating a redundant footprint.

carlosromorgar:
Upps. Sorry. It is my first time submitting symbols and I did not understand what you told me.

I have seen in KiCad/kicad-footprints#539 that it is not already merged. Should I add  DFN-6-1EP_3x3mm_P1mm_EP1.65x2.55mm reference to my symbol? Or it is necessary to wait until it was merged? Thank you in advance.
BR

Ratfink:
Thanks for making that change, this symbol looks just perfect now.  After the footprint is merged, I'll merge this as well.  Thanks for the contribution!

PR #497, opened 2018-04-17
Added new component NB3N511

jeremysiebers:
Added new component NB3N511 in the Oscillator group.

http://www.onsemi.com/PowerSolutions/product.do?id=NB3N511
http://www.onsemi.com/pub/Collateral/NB3N511-D.PDF

nb3n511

------------


evanshultz:
Thanks for this contribution! I see you started a new PR and that's great.

Review comments:
- Put a comma before the package name in the description.
- The footprint filter needs the body dimensions and should be `SOIC*3.9x4.9mm*P1.27mm*`.
- Can you capitalize all pin names?
- Pin 1 is named `X1/ICLK`; you're missing the letter `I`.
- Pin 3 should be Power Input type.
- I think pin 8 should be Passive type since it will only go to a crystal.
- Can this symbol be shrunk a bit? It seems like a lot of dead space in there, but maybe this is the best.

This is a great first submission to the library. Thanks again for adding to it!

jeremysiebers:
Thank you Evan for the review comments, I updated the component:

- [X] Put a comma before the package name in the description.
- [X] The footprint filter needs the body dimensions and should be SOIC*3.9x4.9mm*P1.27mm*.
- [X] Can you capitalize all pin names?
- [X] All pins should be 100mil long.
- [X] Pin 1 is named X1/ICLK; you're missing the letter I.
- [X] Pin 3 should be Power Input type.
- [ ] I think pin 8 should be Passive type since it will only go to a crystal. --> Shouldn't this be a Ouput type because pin 1 is also a direct clock input, so pin 8 (X2) must be the crystal driver?
- [X] Can this symbol be shrunk a bit? It seems like a lot of dead space in there, but maybe this is the best.

nb3n511

jeremysiebers:
Hello Evan,

- [X] Pin 8 type: Inside the IC, yes, it's probably the output of an inverter. But the pin types drive the ERC engine in KiCad and are set depending on what other pins are connected. Since only a crystal should go here, Passive will be fine. As the output of pin 8 is not intended to drive other pins, Output is probably not the best choice. But in actuality it's rather unlikely to matter what type is used.

- [X] One more thing: can you please move the text closer to the body?

The component looks now like:

nb3n511

evanshultz:
Thank you! I see you have more PRs open so thanks for continuing to contribute!

jeremysiebers:
Hello Evan,

Also thank you for continuing to review these components and give feedback!

PR #498, opened 2018-04-17
Fix TPS736xx EN pin polarity.

whitequark:
The EN pin is active high. The symbol currently in repository lists it as `~EN`.

screenshot_20180417_185157
screenshot_20180417_185211

Also applies to TPS731xx, which is an alias of TPS736xx.


evanshultz:
Good catch. Thanks!

You fixed this on the `TPS73601DBV` symbol. Can you also do it to `TPS73625DBV`?

Also, the fixed output symbol `TPS73625DBV` should have pin 4 named `NR` and not `FB`. Would you mind addressing that too?

It's always nice to have the datasheet handy for verification of these types of things. It's at http://www.ti.com/lit/ds/symlink/tps736.pdf. The image above is a nice except.

whitequark:
Done.
screenshot_20180420_102815

evanshultz:
Thanks for making further changes. There is a problem, though. Now the symbol `TPS73601DBV` still has active low EN, but you changed `TPS76901` to active high and it should not have been touched.

whitequark:
Wasn't me! Git did that when I rebased on master.

Fixed.

evanshultz:
Sorry if that was accusatory. I just meant the head of your branch wasn't correct. :) Thanks for this fix, and the others!

whitequark:
I didn't read it as accusatory! It's just that I was as surprised as you to find that the change went to the wrong symbol.

PR #499, opened 2018-04-17
Max14783E

jeremysiebers:
Added alias device MAX14783E to MAX481E library

https://datasheets.maximintegrated.com/en/ds/MAX14783E.pdf

max14783e

------------


Ratfink:
Thanks for the contribution!

This part can't really be an alias of MAX481E due to the differences in packaging.  The MAX14783E needs a different set of footprint filters, but those have to be the same for all aliases of a given symbol.  So you'll need to copy the symbol and set the appropriate filters for MAX14783E's available packages.  

It looks like we don't have a footprint for the μMAX package ([outline](https://pdfserv.maximintegrated.com/package_dwgs/21-0036.PDF), [footprint](https://pdfserv.maximintegrated.com/land_patterns/90-0092.PDF); we have a couple SSOP footprints that are close but not quite what Maxim recommends), so the best course of action would be to either create the right footprint or skip that package altogether.  Also, the TDFN package can't share a symbol with the others because it needs an extra pin for the EP.

My recommendation: make one symbol for each package you wish to add, with the default footprint set for each.  If you just want to add the SOIC-8 package, the symbol name should be `MAX14783ExS`.

Additional notes:
* The description should have a comma before the package name.
* The keywords shouldn't just be a copy of the description.  Rather, they should be related terms that aren't found exactly in the description, e.g. `rs485 rs422 driver receiver` plus any others you can think of.
* When you make a copy of the symbol, please change the pin name position offset to 20 mil.  You may want to extend the functional drawing wire-lines a bit towards the pins on the left-hand side after you make this change.  If you could fix the same issue in the MAX481E symbol, that would be great too!

jeremysiebers:
Hello Ratfink,

- [X] My recommendation: make one symbol for each package you wish to add, with the default footprint set for each. If you just want to add the SOIC-8 package, the symbol name should be MAX14783ExS.  --> Yes I only want to add the SOIC-8 package.

Additional notes:

- [X] The description should have a comma before the package name.
- [X] The keywords shouldn't just be a copy of the description. Rather, they should be related terms that aren't found exactly in the description, e.g. rs485 rs422 driver receiver plus any others you can think of.
- [X] When you make a copy of the symbol, please change the pin name position offset to 20 mil. 
- [X] You may want to extend the functional drawing wire-lines a bit towards the pins on the left-hand side after you make this change. 
- [X] If you could fix the same issue in the MAX481E symbol, that would be great too!

max14783exs

PR #500, opened 2018-04-17
Fix ADC081C021CIMM ALERT pin type

whitequark:
The ALERT pin is defined as output in the repository, but it is actually open drain.
screenshot_20180417_190539


whitequark:
The violations highlighted by CI are spurious.

PR #501, opened 2018-04-17
Fix TPS73601 symbol grid alignment

whitequark:
The symbol outline was not aligned on 50 mil grid.


whitequark:
Also fix pin name position offset, since CI got angry.

evanshultz:
Thanks for finding this!

For clarity, here is the existing symbol showing the issue:
image

PR #503, opened 2018-04-17
Add ADC101C021CIMM and ADC121C021CIMM aliases of ADC081C021CIMM

whitequark:
Pin and function compatible.
screenshot_20180417_221231


whitequark:
Will do in the future.

PR #504, opened 2018-04-17
Shrink Atmel MCUs

fauxpark:
Another round of Atmel fixes. I decided to make the "base width" 1000mil, and widen any symbols that didn't fit (see ATmega640V-8AU and ATxmega16A4).

Refdes/name is in the centre of all 100-pin symbols.

I also moved some pins around, mostly RESET and XTAL1/2, so that RESET (where present, and not sharing a GPIO) is always at the top left, then the XTAL lines spaced 200mil apart, then AREF, then any other misc pins.


poeschlr:
This PR indeed does exactly what it advertises. (Only changes pin position and the sizes of the graphical rectangle) At least a fast sweep over the diff does not reveal any other changes. (But i guess if only a few things other than the positions changed i might have missed it.)

I made a sweeping look over all changed symbols (using the add symbol preview) and i can't find anything obviously wrong.
Travis also seems to say everything is ok.

So i guess this is ok to merge. Thanks for this great improvement to the lib.

Issue #466, opened 2018-04-08
Footprint filter not working as expected


arildj78:
I'm creating a new part (AD795) while following the guidelines in the KiCad Library Convention on [footprint filter](http://kicad-pcb.org/libraries/klc/S5.2/). Since I'm new here, I'm not sure what to expect from this setting, but as far as I can tell it does not do anything.

Just as a test I entered `DIP*` as the filter (totally wrong footprint btw) but I get access to all footprints anyway.

Is this to be expected or is it a bug?

Ratfink:
If I understand correctly, footprint filters only have any effect in CvPcb (Eeschema, Tools→Assign Footprints), or when placing symbols in Eeschema if the new option "Footprint previews in symbol chooser (experimental)" is enabled in Preferences→General Options→Display.

evanshultz:
@Ratfink is correct, as far as I know.

However, the footprint filter can be toggled in CvPcb. The 4 DIP-6 looking icons with green pins at the control the filtering. Here is a test project I had open showing filtering by pin count and keyword:
![image](https://user-images.githubusercontent.com/1936989/38511969-ab2f80c8-3bde-11e8-8efe-4c6f88ae9071.png)

Here is my KiCad version: (5.0.0-rc2-dev-114-g8fcbb64a4), release build

If you are seeing all footprints in CvPcb when you're using filtering, that is an issue with the core KiCad program and you may want to file a bug at https://bugs.launchpad.net/kicad.

(Notice in the screenshot above that some switches are listed because they contain the text "DIP". It looks a bit weird, but it is correct.)

(Also, this symbol that I happened to have placed is built wrong because there is only one symbol for all packages but some had a thermal pad and others don't. That's why the DFN package isn't shown because all dimensionally-compatible footprints have 9 pins and this symbols erroneously only has 8 pins for the DFN package. Sorry for the aside, but I just noticed this and didn't want to show an example that would confuse further.)

poeschlr:
We could filter out the switches by using `*:DIP*` as filter. (KiCad 5 will have a slightly different behavior with regards to filtering then KiCad 4 had. Because it now allows filtering for libraries as well.)

Ratfink:
Closing this, since as far as I can tell there's no bug, just a misunderstanding about how a feature works.

poeschlr:
[...]

PR #505, opened 2018-04-18
fix for issue #466: fp filter DIP* -> *:DIP*

poeschlr:
This gets the kicad 4 behavior back for `DIP*` filters.
More details about it see https://github.com/KiCad/kicad-symbols/issues/466 and launchpad bug report https://bugs.launchpad.net/kicad/+bug/1751565.

---

Filter result for MCP7940N.
screenshot from 2018-04-18 11-11-06

For reference the original result as reported by @evanshultz:
image


evanshultz:
I see Wayne comment on the launchpad bug since this was posted. Has that discussion finished, or at least reached a point where this can/should begin?

poeschlr:
I still have some hope that there will be a better solution than this PR.

stambaughw:
I'm still looking at this.  There seems to be something amiss with the
CvPcb filtering code.  I will reply as soon as I figure out what is
going on.

poeschlr:
This PR is no longer necessary as the underlying problem has now been fixed on the software side.
(Commit that fixed it: https://github.com/KiCad/kicad-source-mirror/commit/62522ee45015642d2440bf8103e6cf879de57c06)

With today's nightly build (on ubuntu) i already get the correct result without this being merged.

PR #506, opened 2018-04-19
Unconfuse data sheets for LM2904 and LM358

robkam:
LM2904 and LM358 currently point to each others data sheets.
http://www.ti.com/lit/ds/symlink/lm358.pdf
http://www.ti.com/lit/ds/symlink/lm2904-n.pdf


evanshultz:
"Unconfuse". I like that word. Thanks for catching this!

PR #511, opened 2018-04-21
Adding 74hc4051

lmoelleb:
Adding 74HC4051 analog multiplexer.
I did not add aliases to 54HC4051 as I do not see this done elsewhere in this library.
Datasheets:
http://www.ti.com/lit/ds/symlink/cd74hc4051.pdf

Since this is an analog switch I opted for marking the IO pins as passive - as it would be done with a manually controlled switch. The datasheet list them as IO though, so let me know if I should change them.

I chose to have the 8 bit on the right side as it gives a more balanced symbol - for my own use I have a version with two units - one for use as a multiplexer, and one for demultiplexer - then I can simply swap unit to get the direction I want in my schematic - but I did not see any similar trick in the library so I keep it to myself. :)

image

------------


lmoelleb:
I do indeed use the two units to swap A0-7 with A, keeping S0-3 and E on the left side. :)

jkriege2:
- the EP and Q1-variants should be removed (we usually don't have extended-temperature and automotives as separate symbols).
- FPFilter-selection seems fine to me
- please add the HCT-variant as Alias
- MAYBE add the 54HC(T)4051 with DIP-only ... but I'm not sure, whether we have those separately anywhere else ... @Ratfink  what do you think?

Best,
JAN

lmoelleb:
@jkriege2 - do you mean change EP and Q1 to be aliases or remove them completely. I made them as separat symbols as they do not have the same footprint options as the "base" versions. This was based on feedback for my previous PR: [PR 484](https://github.com/KiCad/kicad-symbols/pull/484).

I'll add the HCT later today or tomorrow.

lmoelleb:
Just double checked, and the HCT alias is already present.
image

I am now awaiting feedback on what to do with the Q1/EP symbols:

- Keep as is (separate symbols)
- Change to aliases, reflecting the packaging difference in the description.
- Remove completely.

Ratfink:
W.R.T. the Q1/EP symbols, definitely don't do the second option.  Parts should only be aliases if they differ by just the description/keywords/datasheet.  Now that Jan pointed it out, I do agree that removing those symbols would probably be the best idea.

evanshultz:
54xx is typically high-rel, high temp, or MIL spec. So don't include it either. In this case, it appears all parts are -55 to 125C, but there are always exceptions. So you made the correct choice and there's the reason. :) And there should be no reason to add them if we're making condensed/simplified libraries, because they're function-compatible with 74xx parts.

The Q1 get removed totally. It's automotive-specific.

The "EP" version from TI appears to be only extended rating/testing, but not any different functionally, so it should be removed too.

On a related note, https://assets.nexperia.com/documents/data-sheet/74HC_HCT4051.pdf has a thermal pad which may be abbreviated to EP. That, if you wanted to add it, would warrant a new symbol because it has a greater number of pads. But I just found this when checking around and it's not part of your original submission nor does it need to be done. Just writing it down for completeness.

Ratfink:
I think there's nothing else here that needs to be done.  Thanks for adding this part!

PR #512, opened 2018-04-22
DRV8870

thomasonw:
Adding DRV-8870 H-Bridge controller.

screenshot from 2018-05-03 05-02-40

Datasheet is located at: http://www.ti.com/lit/ds/symlink/drv8870.pdf

====================================================================
Also adding related components  DVR-8871, and DVR8872
Datasheet is located at: http://www.ti.com/lit/ds/symlink/drv8871.pdf
Datasheet is located at: http://www.ti.com/lit/ds/symlink/drv8872.pdf

image

image


evanshultz:
I don't see a footprint that exactly matches the one in the datasheet. We have a few that are close, but not quite as specified. I asked TI for more information.

Note that the TI MPN suffix is only `DDA`, not `DDAR` as I initially wrote above. (The `R` means on a reel which we don't want to include in KiCad.)

thomasonw:
@evanshultz Hello, and thank you for the feeedback.  Apologize for the delay, we had priorities come up this past week.

I have updated the submission with most edits, here is what the narrower device looks like - not sure it is much better, seems kind of squished.  What do you think?

image


In the ISEN pin, this is really intended to be connected through a current sensing resistor, not directly to gnd.  As such I think it is better to leave it on the side vs. move it to the bottom.   Thoughts?

Any feedback from TI re: the footprint?

evanshultz:
- For ISEN, I didn't intend to sound like it is connected directly to ground. But since it goes to ground through a resistor, a placement on the side towards the bottom is good. This is nice.
- Yes, that looks too squished to me too. Is there an intermediate width? If not, wider is OK for me.
- Everything else looks OK.
- TI basically says that their different divisions end up with different footprints and they do their best to align it all but this sometimes happens. So this footprint is OK for now since we can see it at least matches this package in another datasheet (and some cleanup is needed for these footprints anyway). If you'd like to contribute the exact symbol from the DRV8870 datasheet that would be excellent!

thomasonw:
I agree – squished, will go back to the wider package.  Had to get something in-between and keep the 100mil spacing, unless I fiddle with the pin lengths.   If you want could try that.

evanshultz:
Please keep pins at 100mil. Play around until you find the symbol you like best and then push a commit. If the original size is optimal that's fine, and thanks for taking a look at squishing it.

thomasonw:
Added other related components DVR8871 and 8872.

Later this week I will look at creating a footprint (I need to add another one as well).   Do you have a suggested name to use for the 'TI' version of this footprint?  Once I get is added and accepted will come back and update these three components.

poeschlr:
I assume the difference is in the size of the mask opening. Simply add the size of the mask opening to the footprint name. (There are already some footprints in the same lib that use that naming style.)

evanshultz:
- All footprint filters should be `HTSOP*1EP*P1.27mm*` (remove `*:` and add `mm`). Sorry that I forgot the `mm` above and the leading `*:` is not needed because KiCad was fixed.
- 8870: Please make pin 7 Passive type. Sorry I missed that earlier.
- 8871:
-- Rename pin 7 to just GND so it can stack properly.
-- Pin 4 should be type Passive since it connects just to a resistor.
- 8872:
-- Again, let's make pin 7 Passive type.
-- Pin 4 should be Open Collector type. And being an output, please move it to the right side.

If you can, I'd rather merge the footprint first and then the symbols so we don't have symbols without a footprint or with a wrong footprint. Especially since v5 is coming soon I'm leery of this situation.

thomasonw:
OK, have made those edits except one.   Will work on the footprint (sigh, they are a lot less fun I think) and update these when it is in place.

 

On the 8872, I am thinking pin 4 makes more sense to be on the left – as it goes back to the CPU – like pins 2 & 3.  Have left it on the left for now, but can move it if you wish.

 

-al-

evanshultz:
I agree with your assessment of pin 4's location, but KLC says outputs should be on the right. And since none of these drivers are quite interchangeable, I don't see any allowance to bend the rules. Even when we both agree, I would like to defer to KLC for consistency (even though throughout the library there is still a lot of clean up needed!).

evanshultz:
Please move pin 4, or add more thoughts for discussion if you have them, and then I will merge.

thomasonw:
Hello.

I thought I had posted some more thoughts on Pin 4, but do not see them here.  DVR8872: Pin 4 - please consider the following:

S4.2(6) states output/controlled/driver pins should be placed on the RIGHT, and I think this is what you are referring to above.

However, S4.2(5) calls for Input/Control/Logic pins to be placed on the LEFT - and S4.2(1) calls for 'similar function' pins to be grouped together.  I offer that Pin 4 on this device is more of a control pin then an output pin:  its purpose is to communicate with the CPU as part of the overall control/management grouping to the driver chip, while the "output" of the driver chip is driving the motor.  This is reinforced by the TI datasheet illustration on page one:

image

where nFault (Pin 4) is connected to the controller, as opposed to the DC motor.

I might also add that S4.2 talks about the overall objective is to provide "... cleaner schematic routing and symbols which are easier to understand in the schematic".   Having one of the three CPU interface pins located on the opposite side of the symbol seems contrary to this stated objective - and would typically require odd routing to bring the signal back to the left of the schematic so that it may be regrouped with IN1 and IN2 and then routed to the CPU.

And finally, I note the language uses 'Should' as opposed to 'Must' - leaving some room for interpretation. Please let me know if you find my arguments have merit, or if you feel their are insufficient.  I will make changes per your final determination when I update the default footprint name in these three devices.

thomasonw:
I need to update the footprint name, as it was changed before accepting; will create a new pull for that.   Presently we have poor internet connection, so it will be a few days before I can create the new pull.

Issue #237, opened 2018-01-28
Naming for the multi-footprint-approach


jkriege2:
Hi @evanshultz @poeschlr @pointhi @SchrodingersGat!

I'm currently reviewing @bobc's PR on EEPROM/EPROM: https://github.com/KiCad/kicad-symbols/pull/222 . There we have a lot of devices that are available in a plethora of 8-pin-packages without EP (e.g. DIP-8/SOIC-8/TSSOP-8) + a few with EP (e.g. DFN-8). How should we handle these cases in naming?
1. Have a specific symbol for each package.
2. Have a symbol for each pinout/pin-count, i.e. DFN-8 specific and all 8-pin-counts n one without footprint-field filled, but with FPFilters for all matching packages.

If option 2 is allowed, how do we handle naming? e.g. `25LC010A-MC` for DFN, but what to use for the multi-symbol, which does not coper all packages? Still use `25LC010A`?

Best,
JAN

evanshultz:
This is a bit of an aside, but when I reworked the old linear.lib everyone wanted the method where the FPfilter only included valid footprints for the symbol. So the symbols were separated logically and physically. I do not think your question above is this topic.

I like the first approach because it gives smaller file sizes and it's easier (at least for me) to identify the symbol you want when adding ALIASes later. I think that dropping the part of the MPN that identifies the package, as you've done above, is the way to go.

poeschlr:
I vote for one symbol per package. Until the new file format allows inheritance of symbol data this is the best approach as it allows setting of default footprints. (Default footprints reduce the risk that the user selects a wrong footprint.)

bobc:
The KLC http://kicad-pcb.org/libraries/klc/G2.1/ talks about defining generic and atomic parts. It gives some examples, but does not really express a preference for either style or any explicit guidance on which should be used. Therefore it gives the impression that either style is acceptable for submissions.

I again repeat the plea that we need to write in ALL preferences into the KLC, so that both contributors and reviewers know what standard is being worked to. It seems we frequently get held up on "philosophy" issues, when time would better be spent improving the many poor quality parts to a basic level of KLC compliance...

Personally, I would prefer a neutral generic part which can be associated with a suitable footprint and MPN, and atomic parts which associated directly with a footprint and MPN, but I would not insist all  possible atomic variations of MPN and footprint are provided. That would be totally impractical considering for some parts there are thousands of variations. We can't hope to include all of them.
[...]

PR #513, opened 2018-04-22
Add NCS*325 AOP serie from ONsemi

suzizecat:
Hi all !
Here I come with some AOP from ONSemi.

Here's the datasheet : http://www.onsemi.cn/PowerSolutions/document/NCS325-D.PDF

image

------------


poeschlr:
Having the power pins on both units will result in wrong netlists if the user ever connects them both. For this reason please split the powerpins into a third unit and set the "all units are not interchangable" flag. (See other dual opamps in this lib.)

poeschlr:
The new file format will support interchangeability between a subset of units. (If support for it gets implemented in the near future is a different matter.)

And yes for now the only real option we have (other than using only one unit) is to ditch that feature.

suzizecat:
The split is done

evanshultz:
@suzizecat 
I want to touch the opamp lib. Will you be able to make these changes soon? Ideally within the next day or so?

evanshultz:
@suzizecat 
After not hearing anything from you, I submitted https://github.com/KiCad/kicad-symbols/pull/578 which should break this PR when it's merged. To get these symbols into the lib I can add them to my PR or you can fix the conflicts.

suzizecat:
@evanshultz Sorry, i missed your lasts messages. Hadn't lot of time recently.
I can correct the points you raised next week, but if you wanna fix those problems and get those symbols into your PR, please do !

evanshultz:
@suzizecat 
No problem on the short delay. If you're able to make the changes above then these symbols will be ready to merge, either way it gets merged. Can you do that?

suzizecat:
@evanshultz Well... missed some things... was late... But hey ! I solved the conflicts !

I'm editing the points you raised and push another commit

suzizecat:
How should I add the package in the dual version description (multiple packages) ?
I removed the DFN package and may add it in another PR but, unless required, I don't think making it quickly, and other points should be fine by now

evanshultz:
The dual isn't ready because it lists two packages but doesn't include them in the description at all. And since only a subset of the total number of packages are provided they should each use the suffix to make this clear.

But, I'll merge this anyway. I can sort this out and add the DFN package. And also and sort out ALIASes. I have more work to do on these libs anyway.

Thanks! The notes above are to help you in the future.

evanshultz:
Um. Since I merge the line endings fix in another PR this one is broken. Can you please merge from master again? Sorry.

I assume even if you need to touch this you don't want to make more edits.

suzizecat:
I don't mind editing a little bit more the symbols (and descriptions and so on...) I just don't really know what to do (and it make take some time since i'm not really involved in ECAD those times.

So I should add the suffix to the dual versions (and therefore create a full component for each version), add the package in the description and eventually create the right DFN footprint ? (I'll really praise some lord when there will be a followed standard in SMD packages... :sob: )

suzizecat:
I'll try to do that this afternoon since it's really a small work.
Since the DFN component isn't orderable (didn't saw that) shouldn't we NOT release a symbol for it (to avoid stupid design mistake which cost a lot :angel: ) ?

evanshultz:
Yes, that's what I'm saying about the DFN package as well. Thanks for getting this patch ready to merge!

suzizecat:
Well, so I edited the NCS2325 to NCS2325D with the OnSemi  and SOIC footprint, since these have the same pin mapping and name (minimum differentiation).
I also edited the description.

It may be useful to create an alias of the 2325D called 2325DM but I don't think that it's really relevant...

According to that choice and KLC (I believe) I didn't added the default FP.

suzizecat:
I don't get your point...
I won't argue more after your reply to this, but the KLC clearly state
 > A single symbol should be drawn for each footprint variation of an atomic part. Other variations (as listed above) do not require their own symbol alias.

Since MSOP and SOIC have the same pinout and the MPN NCS2325D may describe SOIC as well as MSOP (which is DM) without including the DFN. Why should we create a separate atomic symbol ?

evanshultz:
https://github.com/KiCad/kicad-symbols/issues/237

Let me know if you just want this merged; I can take it from here.

suzizecat:
It's funny how there is always a case which mess up with KLC even with all the hard work you guys put in ^^'
I think that this goes against the original KiCad workflow but I tend to prefer atomic part so I won't complain :wink: 
I do the edition and this will be (finally) ready to go !

suzizecat:
@evanshultz And with that, it SHOULD be ok ;)

suzizecat:
Done

evanshultz:
Thanks for all your effort on this! Sorry you got caught up in the middle of changes to this lib.

suzizecat:
No problem, thanks for the time you took on this (and the time you take for KiCAD :+1: )

PR #515, opened 2018-04-22
Adding MMBT5550 and MMBT5551 Alis 

thomasonw:
This is adding the MMBT5550L and MBT5551L small NPN transistor to the existing BC817 alias.

Datasheet is here:  www.onsemi.com/pub/Collateral/MMBT5550LT1-D.PDF


evanshultz:
Thanks!

I see this is your first contribution to KiCad's official library. You can find the rules for the library, if you haven't already, at http://kicad-pcb.org/libraries/klc/.

I hope you'll be willing to submit more to the library in the future!

PR #517, opened 2018-04-23
Added Quadrature Encoders with Differential Output.

ReeseWang:
Signed-off-by: Ruoxi Wang <thuwrx10@gmail.com>

https://en.wikipedia.org/wiki/Rotary_encoder#Incremental_rotary_encoder

screenshot from 2018-04-23 15-50-03
screenshot from 2018-04-23 15-49-51

------------


evanshultz:
Thanks for contributing!

We already have some nice encoder symbols in the Device library:
image

Can you copy this style? Use the same ref des prefix, symbol design, line lengths, etc.

Also, please post the datasheet for an example encoder using these symbols.

evanshultz:
@ReeseWang 
I'm familiar with encoders having A and B outputs, and sometimes an optional Z output if the encoder provides absolute position. Your symbol has differential outputs for each of those three. In addition, I'm familiar with rotary encoders being passive parts and you have power supply pins. This is why I'm asking for a datasheet because I would like to see the specific part that could use a symbol like this. Thank you.

evanshultz:
Thank you, @Wetmelon. This is very helpful. Page 3 of that link also shows all waveforms together.

I've not seen a differential one before, but I suppose it helps in noisy environments. Like a long cable instead of a PCB mounted encoder.

I'd still like to know the part the OP had in mind. A PCB-mounted part is what requires this type of symbol instead of just wire pads.

ReeseWang:
@evanshultz It's very nice of you to still care about this pull request. 

I don't have any specific part in mind but a collection of parts. @Wetmelon provided a perfect example of this kind of part. Here's some datasheets:
https://www.mitutoyo.co.jp/eng/support/service/catalog/07/E13005.pdf Page 9
https://www.celeramotion.com/sites/default/files/Data_Sheet-Mercury_II_MII5000.pdf Page 9
http://www.ia.omron.com/data_pdf/cat/e6f-c_ds_e_6_2_csm499.pdf Page 3

They have wires coming out of it, maybe with different types of connectors, so only signals (A+ A- B+ B- Z/Index+ Z/Index-) are standardrized across models. I made these symbols for drawing wiring diagrams for automation devices, CNC for example. 

> A PCB-mounted part is what requires this type of symbol instead of just wire pads.

Did you mean that I should draw the symbols differently?

jkriege2:
yes and no ... Having such a generic symbol for tehse devices makes sense IMHO ... but any footprint association is really difficult ... so I'm not sure what to do ...

Wetmelon:
I keep a separate personal library specifically for “drawing” symbols that will never (or very rarely) be associated with a footprint or PCB. Generic squares with 2, 3, or 4 pins. Stuff I would use to make simple schematic drawings instead of autocad or Visio.

I don’t know that KiCad has a similar library but these would be good for it

antoniovazquezblanco:
Sorry to revive an old thread like this but I am interested in cleaning this repo a little bit.

Although it is nice to have a symbol like this, I think that if only a symbol exists without a footprint it would be confusing for users. Imagine opening KiCad and adding this encoder but not having any footprint to go with. Furthermore, given current KiCad workflow, it makes no sense to design a symbol without a footprint in mind given that pin numbering is key (except for rare occasions where common configurations exist).

I would vote to close this PR as invalid unless a part is given to base our review on it.

Sorry to be picky. I acknowledge the value of a component like this but I think it would be more coherent with our current library guidelines to include it based on an actual part or model.

Thanks for your contribution.

antoniovazquezblanco:
> Has there been discussion of simply creating a symbol library that isn't included by default, but has symbols with no footprints?

I am not aware of that. It may be useful for some users, but maintenance right now is difficult enough. Maybe creating an unoficial repo with a less restrictive policies can be an option for you.

Thank you for the interest.

PR #518, opened 2018-04-23
Added component LMD18200

jeremysiebers:
Added component LMD18200

http://www.ti.com/lit/ds/symlink/lmd18200.pdf

lmd18200
------------


jeremysiebers:
- Updated the symbol
- Corrected pin electrical properties

lmd18200

evanshultz:
- Please change the description to `3A, 55V H-Bridge, Motion Control Applications, TO-220-11` to better align with the datasheet
- Can you find a way to allow a capacitor to nicely fit between pins 1/2 and 10/11?
- Change the default footprint to the vertical one in the datasheet unless the prep you selected in available from the vendor.
- Remove the pin count from the footprint filter.
- Pins 8 and 9 need `_OUTPUT` at the end the match the datasheet names (I know this is horribly long).
- Pin 9 is active low so please give it an overbar with a leading `~` to the pin name.
- For this part, I think pins 8 and 9 are better at the bottom since pin 8 has a resistor to ground and both probably go to a controller to the left of this IC.
- Can you put the ref des at the top left and symbol name at the top right?

evanshultz:
Looks good except for pins 8 and 9. I meant at the bottom of the sides, not on the bottom of the symbol, but I wasn't clear. This will allow the pins to wrap around the bottom and get back to any controller without interfering with the power pins. If you can move those to the right side at the bottom this should be good. Thanks again for making the changes!

jeremysiebers:
- [X] You need to update the footprint since they were renamed.

lmd18200

evanshultz:
I don't see that we have exactly this footprint, but `TO-220-11_P3.4x5.08mm_StaggerOdd_Lead4.85mm_Vertical` is awfully close. Please take a look at that.

jeremysiebers:
Set TO-220-11_P3.4x5.08mm_StaggerOdd_Lead4.85mm_Vertical as footprint.

evanshultz:
Thank you!

Issue #520, opened 2018-04-23
Invalid footprint associations in Amplifier_Audio.lib


Ratfink:
As Wayne reported [here](https://lists.launchpad.net/kicad-lib-committers/msg00547.html), there are some invalid footprint associations in Amplifier_Audio.lib which need to be fixed.  They're old references to `Package_TO_SOT_THT:TO-220-5_Vertical_StaggeredType1`, which need to be changed to `Package_TO_SOT_THT:TO-220-5_Vertical_StaggeredType1_Py3.7mm` assuming that is indeed the right Y pitch.

poeschlr:
I did not even know about that mailing list. I think i should join it then. (If i find out how.)

stambaughw:
Below is an updated list with all of the symbol libraries with broken
footprint links as of c3c0a08b28.  In some cases there were so many
broken footprint links that I just included the library name.  Any time
a footprint and/or footprint library are renamed, someone needs to a
search and replace of the changed footprint/library name to all of the
symbol libraries.  I know this is going to be an effort to fix all of
these but we should make an effort to get it done by the stable 5 release.

Battery_Management: LTC4007
Converter_DCDC: All
Diode_Bridge: ABS2, MBL104S
Driver_LED: LT3465
Driver_Motor: L298HN
FPGA_Actel: All
Interface_CAN_LIN: MCP2557FD-xMF, MCP2558FD-xMF
Interface_Optical: Almost all
Interfact_UART: ADM1491EBR
LED: CLV1L-FKB, LED_Cree_XHP50_12V, LED_Cree_XHP50_6V,
LED_Cree_XHP70_12V, LED_Cree_XHP70_6V
MCU_Cypress: CY7C68013A-56LTX, CY8C4xx7LQI-4xx, and many more.
MCU_Nordic: nRF52810-QCxx
MCU_ST_STM32F0: STM32F038E6Yx, STM32F038G6Ux, STM32F042C4Ux,
STM32F042G4Ux, and many more.
MCU_ST_STM32F1: STM32F100R4Hx
MCU_ST_STM32F2
MCU_ST_STM32F3
MCU_ST_STM32F4
MCU_ST_STM32F7
MCU_ST_STM32HY
MCU_ST_STM32L0
MCU_ST_STM32L1
MCU_ST_STM32L4
MCU_ST_STM32L4+
MCU_Texas_MSP430
Memory_Flash
Memory_RAM
Memory_ROM
[...]

PR #521, opened 2018-04-23
Set correct footprint for LTC4007

poeschlr:
Fix for one of the symbols mentioned in https://github.com/KiCad/kicad-symbols/issues/520#issuecomment-383650273


Ratfink:
Heh, oops, we had the same idea.  #522 does the same thing, but also includes a lot of KLC fixes.

poeschlr:
In that case i will close this

PR #524, opened 2018-04-24
Fix footprints in Converter_DCDC.lib

Ratfink:
Trying to work through S5.1 errors caught by the library-utils.  This fixes the footprint associations and filters for TEN20-2411WIN and TEN20-2421WIN.


poeschlr:
well it seems i was blind when i made my PR.

Ratfink:
Hah, that's all right.  I liked your footprint filters better anyway.

PR #525, opened 2018-04-24
Move obsolete Atmel MCUs

fauxpark:
#387

All the ATmega, ATtiny and 90S symbols which are out of production have been moved to `obsolete/`.


fauxpark:
The second commit is because I noticed the datasheet URL for the ATmega48/88/168A/PA and 328/P was redirecting to the summary instead - it looks like Microchip have been working on the Atmel datasheets recently and they have been split up into 48/88/168A, 48/88/168PA, and 328/P.

fauxpark:
@evanshultz just bumping this again - I know you guys are getting busier by the day though ;)

evanshultz:
Thanks for checking in. I was a bit too busy to take a look recently but I can do it now.

How did you come up with the list of parts that are obsolete? Checking each one will take a lot of time so perhaps you find a nice and concise way to make these changes?

fauxpark:
> Checking each one will take a lot of time

Unfortunately this is exactly what I did. I couldn't find a list of legacy products on their website, but [this page](http://www.microchip.com/paramChartSearch/chart.aspx?branchID=30047) can show all of the currently in production AVRs. I have assumed that if it is not on this list, it goes in `obsolete/`.

Generally the datasheets have on the first page "Mature" or a message similar to:

> Not recommended for new designs - Use XMEGA A3U series

evanshultz:
@fauxpark 
Got it. Thanks!

@poeschlr 
I spot checked a few and this seems OK. Are you OK to just go ahead and merge this?

poeschlr:
I don't have a particular opinion on this. If these are indeed obsolete they should really be moved as suggested. If you check a few of them (At least one per obsolete series) and come to the same conclusion as the contributor you should be able to decide if this is indeed the correct way to go.

It might be that this pull request is a bit too large to be honest. Spiting it by series might have made the work of us librarians a lot easier. @fauxpark could you take that into account for your future work?

fauxpark:
I've already been splitting this Atmel stuff into what I'd hoped was more manageable chunks, and trying to avoid any one PR doing too much, but I will keep that in mind.

fauxpark:
@evanshultz I resolved the merge conflicts, it's ready to go :)

poeschlr:
As @evanshultz already spot checked this and you gave a good explanation of how you arrived at the list of obsolete symbols i feel it is ok to be merged. (And the symbols are not out of this world. They can be easily restored if we discover one of them has been declared obsolete by accident.)

Thanks for your great work lately.

PR #526, opened 2018-04-24
Fix footprints in Diode_Bridge.lib

Ratfink:
Trying to work through S5.1 errors caught by the library-utils.  This fixes the footprint associations for Diotec ABS and Vishay MBLS diode bridges.


poeschlr:
Thanks.

PR #528, opened 2018-04-24
Fix footprint fields in can interface lib

poeschlr:
Fixes part of what wayne reported in https://github.com/KiCad/kicad-symbols/issues/520

In addition to fixing the FP file, i moved the NC pin onto the symbol outline.


Ratfink:
Thanks.

PR #529, opened 2018-04-24
Interface optical fix fp field (3pin -> 3Pin)

poeschlr:
Another fix for the missing footprints listed in #520


Ratfink:
Thanks for the fix!

PR #532, opened 2018-04-24
Fixes for footprint fields in the cypress library

poeschlr:
Another series of fixes for #520

This time there will be some fixes on the footprint side as well: https://github.com/KiCad/kicad-footprints/pull/510


Ratfink:
CY8C4xx7LQI-4xx and CYBL10x6x-56LQxx's footprint filters are missing the number of EPs, which I believe we decided should be included in the filter.  Might as well fix that here too (especially the former, since you changed the filter in this PR).

Ratfink:
Thanks for the fixes!

PR #534, opened 2018-04-24
Fixup for actel fpga lib (now microsemi)

poeschlr:
Fixes for #520 plus some minor fixes on the same lib (example renaming to reflect the fact that actel does not exist since 2010)


poeschlr:
This will take me a bit as i will create a IPC qfp generator to create the missing footprints (microsemi does not supply suggested landing patterns only package drawings.)

poeschlr:
This should now be ready.

poeschlr:
Ok now everything that can be done should be done.
It might be that the ACT1020 devices are obsolete as i can not find a datasheet for them.

Ratfink:
It looks like the references to `Package_QFP:VQFP-100_14x14mm_P0.50mm` need to be changed to `Package_QFP:VQFP-100_14x14mm_P0.5mm` (one fewer decimal place for the pitch).  Once that's done I think this will be ready.

Ratfink:
Oh, now that I look closer, should `EX128-TQ100`, `EX256-TQ100`, and `EX64-TQ100` have TQFP footprints instead of VQFP?

poeschlr:
Yes i think you are right. Fixed

Ratfink:
Two out of three, anyway. 🙂

PR #537, opened 2018-04-25
Assign correct footprint to nRF52810 (MCU Nordic)

poeschlr:
Another fix for #520

This also requires a new footprint:


Ratfink:
Thanks for changing this, and making the footprint to match!

PR #538, opened 2018-04-25
Move obsolete parts from memory rom lib

poeschlr:
More "fixes" for #520.

The problematic parts in the memory rom lib seem to be obsolete. I can not even find a datasheet for them.


Ratfink:
That is indeed strange.  I can barely find any mention of those parts ever having existed.  Thanks for moving them.

PR #540, opened 2018-04-25
Added new symbol: PIC16F18854-SO

jeremysiebers:
Added new symbol: PIC16F18854-SO

http://www.microchip.com/wwwproducts/en/PIC16F18854
http://ww1.microchip.com/downloads/en/DeviceDoc/PIC16(L)F18854%20Data%20Sheet_DS40001826C.pdf

pic16f18854-so

Datasheet says on page 3 with "Note 1" that 1 pin is input only. I checked the datasheet itself and via Microchip MPLAB Code Configurator but could not find any pin that is input only.

------------


jeremysiebers:
@evanshultz, @Ratfink 
Hello gents,
Is it possible to merge my addition despite of the errors of the other components?

poeschlr:
The errors reported by travis are for other symbols not for yours. So this should not be a problem.
Note: I did not look at the contribution itself at all. Only answered the question above.

Misca1234:
> Datasheet says on page 3 with "Note 1" that 1 pin is input only. 
> I checked the datasheet itself and via Microchip MPLAB Code Configurator but could not find any pin that is input only.

This have been discussed in other threads, apparently it should be input only

Why have you renamed PIC16F18346-SS intoPIC16F18346-SS_0 ?

You could add all the other variants as aliases perhaps ?

jeremysiebers:
> > Datasheet says on page 3 with "Note 1" that 1 pin is input only.
> > I checked the datasheet itself and via Microchip MPLAB Code Configurator but could not find any pin that is input only.
> 
> This have been discussed in other threads, apparently it should be input only
> 
> Why have you renamed PIC16F18346-SS intoPIC16F18346-SS_0 ?
> 
> You could add all the other variants as aliases perhaps ?

Hello Misca,
I will change pin 1 to input only.
I only have added the PIC16F18854-SO all others I did not touch, if something changed there it was not my intention to do so.

Misca1234:
Thanks for contribution

PR #542, opened 2018-04-25
Fix footprint field of DWM1001 rf module

poeschlr:
And another fix for #520

The two other modules with a wrong footprint field need https://github.com/KiCad/kicad-footprints/pull/520

---

It seems my nightly build version reordered the files. I assume the file has last been modified with a script.


Ratfink:
The symbol that has the wrong footprint is DWM1001, not DWM1000.  [DWM1000](https://www.decawave.com/sites/default/files/resources/dwm1000-datasheet-v1.3.pdf) and [DWM1001](https://www.decawave.com/sites/default/files/dwm1001_datasheet.pdf) have very different footprints, so unfortunately for us, someone has to create a DWM1001 footprint.  I'll get right on that!

poeschlr:
You are right. This means this pr does not fix anything but makes things even worse. Closed for this reason

PR #543, opened 2018-04-25
Fix the package size of TFBGA-216

Ratfink:
As @poeschlr noticed in https://github.com/KiCad/kicad-footprints/issues/374, it's 13x13mm, not 16x16mm.  Thanks for catching that!


poeschlr:
looks good. Thanks

PR #544, opened 2018-04-25
Fix footprint field for TPS2592 (power_management)

poeschlr:
Next fix for #520 There seems to have been (at least) one symbol missed when we added the ep size specifier.


Ratfink:
Thanks for fixing.

PR #545, opened 2018-04-25
Fix footprint field for Bosch sensors

poeschlr:
Another fix for #520

### DIFF
--- a/Sensor.lib
+++ b/Sensor.lib
@@ -99,10 +99,10 @@ ENDDEF
 DEF BME680 U 0 40 Y Y 1 F N
 F0 "U" 300 500 50 H V C CNN
 F1 "BME680" 300 -500 50 H V C CNN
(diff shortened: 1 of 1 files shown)


Ratfink:
I just checked the datasheets, and found that this footprint is correct for [BME280](https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BME280_DS001-12.pdf#page=42), but incorrect for [BME680](https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BME680-DS001-00.pdf#page=44).  It looks like we don't have the correct footprint for BME680 in the library yet.

poeschlr:
Man they both had the same footprint assigned. Both symbols where added shortly before the move. We really can not trust anything.

poeschlr:
Should be fixed now

PR #546, opened 2018-04-25
Rework sensor motion

poeschlr:
Another fix for #520 

L3G4200D is obsolete
Some sensors are no longer recommended for new designs (added a remark to the description. Not sure if we should invest time into creating these footprints)

Some footprints are missing. Some are only slightly wrong (qfn-24-1ep would require ep size 2.6x2.7 but we only have 2.6x2.6)
I assigned the once with the name that the correct footprint should have.


Ratfink:
This needs a few footprint filter and association changes to reflect the final naming, but then it should be ready.

poeschlr:
This should now be done

PR #549, opened 2018-04-26
Fix footprint fields in sensor temp lib

poeschlr:
The next fix for #520 

This requires the footprints added by https://github.com/KiCad/kicad-footprints/pull/527 and https://github.com/KiCad/kicad-footprints/pull/526


jkriege2:
Could you have a look at the Travis otput?
```
Checking symbol 'KTY83':
  Violating S5.1
    filename
    Specified footprint does not exist
     - Footprint file Diode_THT:D_DO-34_SOD68_Horizontal_RM7.62 was not found
  Violating S6.3
    Part meta-data is filled in as appropriate
    Component KTY83 has metadata errors:
     - Symbol name should not be included in description
    ALIAS KTY84 has metadata errors:
     - Symbol name should not be included in description
Changed symbol 'Sensor_Temperature.lib:KTY81'
Checking symbol 'KTY81':
  Violating S3.3
    Symbol outline and fill requirements
    Component background is filled with TRANSPARENT color, recommended is filling with BACKGROUND color
     - exceptions are allowed for small symbols like resistor, transistor, ...
  Violating S5.1
    filename
    Symbol possibly missing default footprint
     - Symbol has a single footprint filter string 'SOD?70*' (i.e. it may be intended for a single default footprint only), but the footprint field is empty.
  Violating S6.3m
    Part meta-data is filled in as appropriate
    Component KTY81 has metadata errors:
     - Symbol name should not be included in description
    ALIAS KT100 has metadata errors:
     - Symbol name should not be included in description
    ALIAS KTY10 has metadata errors:
     - Symbol name should not be included in description
```

PS: The footprints have been merged ...

cu
JAN

jkriege2:
PS: In the SOD68 footprint, the RM and Horizontal-fields need to be swapped so the footprint is matched!

poeschlr:
Done

PR #551, opened 2018-04-27
Added SiA453EDJ MOSFET

SchrodingersGat:
Datasheet: http://www.vishay.com/docs/62864/sia453edj.pdf

image

Footprint: https://github.com/KiCad/kicad-footprints/pull/528


evanshultz:
This looks fine to me but I'll wait for the footprint to be merged.

PR #553, opened 2018-04-27
Fix footprint associations in Video.lib

Ratfink:
Another fix for #520.  We have the correct footprint for these, but the association was lost during a rename.


poeschlr:
Looks good, thanks

PR #554, opened 2018-04-27
Fix footprint associations in Relay.lib

Ratfink:
Another fix for #520.  We have all the right footprints for this library, but after a (partial) rename (completed in https://github.com/KiCad/kicad-footprints/pull/531), some associations were broken.


poeschlr:
Looks good, thanks

poeschlr:
In #520 the relay lib lists "many more". Does this fix all of the problems in the lib or is there still something to be done?

Ratfink:
This fixed all the remaining errors in Relay.lib.

PR #555, opened 2018-04-28
Fix footprint filters in sensor current lib

poeschlr:
The footprint filters did not agree with the footprint fields for these components. (missing * at the end and exchanged "fields")


poeschlr:
Travis errors because https://github.com/KiCad/kicad-footprints/pull/533 has not been merged yet

Ratfink:
Thanks for fixing!

PR #556, opened 2018-04-28
NKK GW12LJPCF SPDT illuminated rocker switch

ppelleti:
[Datasheet](http://www.nkkswitches.com/pdf/gwillum.pdf).

Not sure how to center this symbol.  The KLC script wants me to move it up by 50 mils, but that would put all the pins off the 100 mil grid.

sw_nkk_gw12ljpcf

------------


ppelleti:
Besides complaining about not being centered, Travis also complains that the footprint doesn't exist.  It won't exist until [the PR for the footprint](https://github.com/KiCad/kicad-footprints/pull/534) has been merged.

Ratfink:
Thanks for the contribution!

The not-centered message is only an info message, because we know that symbols sometimes have to be slightly off-center to keep the pins on the grid.  Nothing to worry about.  Thanks for linking the footprint PR.

* Please make the LED symbols match the `Device:LED` symbol.  Yes, this doesn't match the LEDs in the other switch symbols, our switch library could use some love.
* The symbol extends quite a bit farther down than it does up, due to the LED symbol at the bottom.  To better center it, I suggest moving everything 100 mil upwards.

ppelleti:
> The not-centered message is only an info message, because we know that symbols sometimes have to be slightly off-center to keep the pins on the grid. Nothing to worry about.

Thanks, good to know, although Travis CI doesn't seem to understand the distinction, since the commit ends up with a red X next to it.

> Please make the LED symbols match the Device:LED symbol.

Done!

> To better center it, I suggest moving everything 100 mil upwards.

Done!

I also changed the footprint field and footprint filter to reflect the new name of the footprint.

ppelleti:
Here is the new version of the symbol:

sw_nkk_gw12ljpcf-updated

ppelleti:
Thanks, I've made those changes.

I also made an additional change: I renamed the symbol to `SW_NKK_GW12LJxCF`.  If we later have footprints for the Right Angle (H) and Vertical (V) versions of the switch, this same symbol can be used with them.

ppelleti:
Here's what it looks like now:

sw_nkk_gw12ljxcf

ppelleti:
> If we later have footprints for the Right Angle (H) and Vertical (V) versions of the switch, this same symbol can be used with them.

On second thought, I'm wrong about that.  The H and V versions need two additional NC pins.

PR #557, opened 2018-04-28
Fix footpirnt fields in sensor procimity

poeschlr:
Yet another fix for #520 

The knightbright has 2 footprints where the horizontal one is a non standard way to mount the device. So i chose the vertical one where the castelated pads are oriented as they should be.


evanshultz:
Looks good. Thanks!

PR #558, opened 2018-04-29
Fix footprint field in sensor humidity

poeschlr:
Another fix for #520

footprint PR: https://github.com/KiCad/kicad-footprints/pull/539


evanshultz:
I assume we're waiting on the new footprint to be merged here so it doesn't get out of sync again?

PR #560, opened 2018-04-29
Fixing footprint fields and footprint filters for texas mcu msp430 lib

poeschlr:
**Work in progress**

This is another fix for #520 

This lib does not seem to have a valid footprint assigned for any symbol. In addition to this we are also missing some of the footprints.

First run done.


poeschlr:
I think this can be merged for now. The remaining missing footprints can be added at any time in the future.

evanshultz:
There is still a lot of violations in this lib. But better than it was.

PR #562, opened 2018-04-30
Added symbol for ADS1299 24 bit ADC biopotential amplifier

sandhan26:

* Link to the footprint [here]( kicad-footprints/Package_QFP.pretty/TQFP-64-1EP_10x10mm_P0.5mm_EP8x8mm.kicad_mod)

* Screenshot for the symbol
Sample screenshot


evanshultz:
You footprint link above doesn't work and there is no datasheet link. The datasheet is at http://www.ti.com/lit/ds/symlink/ads1299.pdf.

sandhan26:
@evanshultz I have updated the footprint link. Also, you can go from here
https://github.com/KiCad/kicad-footprints/blob/master/Package_QFP.pretty/TQFP-64_10x10mm_P0.5mm.kicad_mod

evanshultz:
1. The first word in your description is "Low-Power" but the datashete is "Low-Noise". Can you align a bit closer to the datasheet description?
1. Footprint filter should be just `TQFP*10x10mm*P0.5mm*` (no library name and no pin count).
1. Do not put text of the part number inside the body.
1. All pins should be of type Line; do not use other pin styles.
1. Stack ground pins and stack power pins unless the datasheet says not to. It looks like pins 59 needs to be separate from the other AVDD pins but the rest can be stacked and also both DVDD can stack.
1. All pin should be 100mil long.
1. After pin stacking, see if the symbol looks nice if it's more narrow.
1. Move the NC pins inside the body so their 'x' is on the border.
1. Move RESV1 to the right side since that's the respective ground but in case it's ever used it won't be on the bottom (this one is a little gray zone but it's not a dedicated ground pin).
1. Even if the pins are out of order, what about going IN1 to IN8 from top to bottom?
1. Pin 35 only has 4 characters in the name.
1. Please put the VCAPx pins on the left side since they reference to AVSS. And maybe put them in order, 1-4, going top to bottom.
1. What do you think about moving pins 24 and 25 apart so a capacitor can fit nicely there?
1. How did you decide on the pin groupings on the right side? Is that from the datasheet somewhere?
1. Each ALIAS should have a matching entry in the DCM file. Can you provide an example so I can take a look? Either copy the files here or even push a commit that I can review and we later update.
1. Also, the base part is 8 channels so you can remove description and keyword for lesser channels.

CLA: I'm not sure sure. I guess you changed emails at https://github.com/settings/emails? Perhaps it only affects in progress PRs? Do new PRs go smoothly?

antoniovazquezblanco:
@sandhan26 Could you please address the comments above?

PR #563, opened 2018-04-30
Fix footprint association for TPS3839DQN

Ratfink:
Another fix for #520.  This one apparently was forgotten in #561; once this is merged we can check Power_Supervisor off the list of libraries with footprint problems.


poeschlr:
Yup i seem to have missed that there is a second symbol using that footpirnt

PR #568, opened 2018-05-02
Revert footprint filter hotfix for implicitly added '*'

poeschlr:
Previosly we needed to add *: to the footprint filters to force
kicad to use that filter as the start of the footprint name.
(Especially noticable for footprints with single letter prefixes
as their filters resulted in a lot of false positives otherwise.)

A fix has now been committed to the source code that makes this
hotfix obsolete. (It is already in the nightly build of ubuntu)

---

The fix to cvpcb has been added in https://github.com/KiCad/kicad-source-mirror/commit/62522ee45015642d2440bf8103e6cf879de57c06

Tested some of these in the current ubuntu nightly and they work as they did before i reverted the hotfix.


evanshultz:
I see a lot of `Footprint filter '*SingleDiode' not correctly formatted
     - Does not end with ('*') character` from Travis.

poeschlr:
I did not use a script but a powerful text editor with regex text replacement support. (atom)

I created a new PR for fixing the problem you noticed with the SingleDiode footprint filters. (They where malformed duplications of the correct footprint filter.)

PR #571, opened 2018-05-03
Fix off-by-one in FXMA108 pin names

whitequark:
Does not affect schematics or pinout, just a cosmetic issue.

Before/after:
screenshot_20180503_182527
screenshot_20180503_182611

Datasheet (URL: http://www.onsemi.com/pub/Collateral/FXMA108-D.pdf):
screenshot_20180503_182721

------------


evanshultz:
Good catch. Thanks!

PR #572, opened 2018-05-04
Added symbol for the Infineon BTS50010-1TAD device

johnfwhitmore:
There's a lot of these PROFET devices. I've added this BTS50010-1TAD device.

Datasheet - https://www.infineon.com/dgdl/Infineon-BTS50010-1TAD-DS-v01_00-EN.pdf?fileId=5546d462576f34750157c38810ca55cd

screenshot from 2018-05-04 18-56-33


evanshultz:
- You do not need to fix the existing symbol, but if you want to here (or in another PR) it would be appreciated.
- It seems you've removed some entries from the DCM file. Can you restore those and add the new ones?
- There are a number of ALIASes which end in a `1`, like `BTS50055-1TMC1`. Those seem to be invalid PNs. Can you please review that?
- Can you match the datasheet description and then put extra stuff after that? And also add electrical parameters to differentiate the various parts. And finally but the package at the end. Something like `Smart High-Side Power Switch, PROFET, Single, 1mOhm, 150A, 18V, TO-263-7`.
- The keywords should be search terms and not the MPN
- The best footprint is `TO-263-7_TabPin4`.
- After changing that footprint, remove pin 8 from your symbol.
- The remaining pin 4 should be named `VS`.
- Stack the OUT pins.
- Then you can shrink the symbol a bit.

johnfwhitmore:
Sorry stalled pcbnew is not working for me. Have to try get it working

evanshultz:
@johnfwhitmore 
Are you planning to come back to this symbol soon?

johnfwhitmore:
Yes sorry back now. I had trouble with a change in Gnu/Linux distro which wouldn't correctly build the latest version of KiCad. So I installed from the distro's repo and then got errors about continuing the work with an older version of KiCad. I think I'm now up and running and will get this finished up.

johnfwhitmore:
Most of the problems are now with that other symbol, not the one I created, but it should not be a big job to resolve its issues.

johnfwhitmore:
So I have to re-start with a question. The footprint used for that symbol, which is in error. I'll link the datasheet:
https://www.infineon.com/dgdl/Infineon-BTS50055-1TMC-DS-v01_00-EN.pdf?fileId=5546d4625a888733015aa9b0007235e9

There is a footprint in the libraries Infinieon-TO-220-7Lead-TabPin8 (Or something like that) but the device shorts pin 4 to the Tab and calls the device the TO-220-7-4 So I wonder should I be creating a new footprint for that device, or using the TabPin8 footprint?

There's also an error in the symbol's alias fields, I've not gotten to the bottom of that as I'm not sure where the original creator got that list of devices from.

evanshultz:
Sorry I wasn't able to get to this until now. Glad you got the issues fixed.

Actually, the footprint can be done in two ways:
1. Use the TabPin8 footprint and have two Vbb pins, 4 and 8.
2. Just a single Vbb pin (4) and use the TabPin4 footprint.

Does that make sense?

Using option 2 makes the symbol file smaller, but stacking Vbb pins with option 1 will look visually the same. I haven't seen any direction for this situation from the other librarians so it's up to you unless someone else chimes in.

The Travis error is because you have one symbol named `BTS50055-1TMA`, and 4 ALIASes of that symbol. The DCM file only has an entry for the symbol but none of the ALIASes. Do you see that? If you add that documentation Travis should be happy.

Please also take a look at the notes I made above. For example the 2nd and 3rd bullets are important.

johnfwhitmore:
Not finished yet but pushing this to see if it clears up some of the Travis errors. Still have to address some issues

johnfwhitmore:
I still have to address your other bullet points but it'll have to wait until tomorrow.

johnfwhitmore:
screenshot from 2018-09-28 15-41-59
Sorry I lost track of this pull request being distracted with other work. Just now getting back to kicad work. I'm not sure where I left it and if I've addressed all the outstanding issues. You'd suggested to use a TabPin4 footprint and I think that's what I've left it at. Perhaps I didn't push it better check... Nope everything up to date.

antoniovazquezblanco:
I think this is pending @evanshultz review :)

Thanks!

evanshultz:
@antoniovazquezblanco 
Thanks for the ping.

@johnfwhitmore 
There are notable cleanup that could be done on MPNs that don't contain "-1TM" in the MPN. However, let's just get this done and merged before starting on anything else so I'll limit my review.
- Thinking about it, let's use our standard footprint for both TO220-7 parts. I don't see why it would need an Infineon-specific one. Then you can eliminate pin 8 and the footprint filter will be `TO*263*TabPin4*`.
- `BTS50055` pin 4 is named Vbb.
- Many descriptions for MPNs containing "-1T" look wrong. For example, BTS50085-1TMA shows 1mR/55A/18V but the datasheet looks like 9mR/44A/58V. (Of course we could choose different datasheet ratings but the point is it looks like an incorrect copy and paste.) Would you mind checking all the "-1TM" parts? In some cases I see the TMA and TMB and TMC are all the same high-level ratings, but in other cases they vary. Is there a consistent methodology you can see?
- A few of the descriptions still contain the MPN in the keywords. Please remove them.

It seems like such a long time ago I'm afraid I missed something. But I match up all the symbols with descriptions so I think we're good. Travis would catch it if there were descriptions missing but another set of eyes to make sure there aren't excessive descriptions would be nice. (I think we're OK since i only count 15 symbols and descriptions starting with "BTS5", but better safe than sorry.)

johnfwhitmore:
@evanshultz Sorry you've confused me a bit with those comments, Oops you're referring to other symbols. Hold on I'll take a look at them, I'd created this Pull Request for the BTS50010 but I'll change some of the other components. You had me confused as there was no pin 8 in the symbol I'd added.

johnfwhitmore:
screenshot from 2018-10-07 17-27-28

johnfwhitmore:
screenshot from 2018-10-07 17-34-02

johnfwhitmore:
screenshot from 2018-10-07 18-09-06
screenshot from 2018-10-07 18-09-27

johnfwhitmore:
screenshot from 2018-10-07 18-57-30

johnfwhitmore:
Right I think that's it? I've gone through all those BTS* Components and updated or corrected the descriptions.

Let me know if I've made a mistake with those changes or any more you'd like done.

johnfwhitmore:
Oh meant to ask would this qualify me for the hacktober T-Shirt ;)

I'm using another PROFET a BTS7004 but I might add that in another pull request, after this one is cleared.

johnfwhitmore:
screenshot from 2018-10-09 16-49-33

johnfwhitmore:
screenshot from 2018-10-09 16-51-41

johnfwhitmore:
screenshot from 2018-10-09 16-53-25

johnfwhitmore:
screenshot from 2018-10-09 16-57-56

johnfwhitmore:
screenshot from 2018-10-09 17-02-02

johnfwhitmore:
BTS724G - OK that'll slow me down.

I recently started a discussion on version control in Kicad on the forum. Got a bit of push back there ;-) However from the point of view of maintainers there has to be a better method. I also contribute to the Linux Kernel project and yes it's very different being a SW project, but the tools are there. So Linux Maintainers only accept the smallest of changes in a commit. So I might change one simple thing and commit it. Add a series of commits, each with a very simple change. The maintainer has the tools where they can see the lines that have changed and what I have changed. Each simple change can be clearly reviewed and either OK'd or rejected and kicked back to me.

The tools make the maintainer's lives much easier then ye have it here, or that's the view I have of it. If Linux developers struggle with something they make a tool to remedy that problem. That's how git was created. I realise that this is just day dreaming, but man it would be an addition, in my limited view.

evanshultz:
Thanks! Any more aligning the symbols so they're size- and pinout-interchangeable, and adding the new part you mentioned, can be done in another PR.

We use similar tools, but probably not as regimented or comprehensive. Well, at least some of us do. This PR is a monster but usually we will separate changes from additions and accept only smaller commits. There wasn't a need to change other symbols in this PR, but since I was tracking what was going on I let it go. Bad on me.

Just need to confirm that merging this is OK and doesn't break anything.

johnfwhitmore:
Deadly, I'll add the BTS7004 in another PR. And I'll perhaps create a PR for the BTS724G, but looking at that symbol and i'm scratching my head a bit at present.

evanshultz:
I'll do a little polish once this PR is merged. 724G is a wacky one.

Eh... Git says only these two files were touched so I'm going for it.

johnfwhitmore:
Thank you for your patience with this one !!!

PR #75, opened 2017-12-19
Nordic nrf52840

ChristianHirsch:
Add symbol for [nRF52840](http://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52%2Fdita%2Fnrf52%2Fchips%2Fnrf52840.html) Multiprotocol Bluetooth 5/BLE/IEEE 802.15.4/2.4GHz Cortex-M4F SoC.
image

PR #573, opened 2018-05-04
Add E73-2G4M04S-52832 BLE5 module

isundaylee:
Add symbol for E73-2G4M04S-52832 BLE5 module based on nRF52832 chip. 

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/555

Datasheet: http://www.cdebyte.com/en/downpdf.aspx?id=243 (page 5)
Screenshot:

screen shot 2018-05-04 at 2 01 20 pm

------------


evanshultz:
Would you mind making this symbol look more like our other Nordic Semi symbols? See https://github.com/KiCad/kicad-symbols/pull/75, https://github.com/KiCad/kicad-symbols/pull/247, and https://github.com/KiCad/kicad-symbols/pull/379. Even though this is a module, it seems to have the same functionality and pinout at a glance.

isundaylee:
Thanks for the review! Sure, to make sure I don't miss anything, it seems that: 

- I should move all GPIO pins except RESET to the right.
- Use VDD/VSS instead of VCC/GND. 
- Move DEC pins to the left of VCC.
- Maybe also label the NFC/AIN functions on the corresponding GPIO pins? Not sure if I should do this since the module data sheet didn't. 

Anything else I should do besides theses?

isundaylee:
I've changed the symbol to match the other nRF symbols' style. Updated screenshot: 

screen shot 2018-05-06 at 4 46 55 pm

evanshultz:
Sorry that I missed your questions above.
- I'm prefer VCC/GND pin names since the module datasheet uses it, but you've matched the position of the MCU symbol and that looks good.
- You can omit the NFC/AIN pin names if you prefer. Again, the pin position is good.
- Compared to the MCU symbol, you rectangle is a bit bigger. Can you shrink it to match?
- Please add the default footprint and footprint filter to match your footprint submission.
- Since this symbol works for both 52832 and 52810 modules, would you mind adding an 52810 ALIAS?

isundaylee:
@evanshultz Made changes as requested. On second thought I feel like keeping the AIN and NFC markings might be handy. Let me know if you'd prefer them removed instead. 

Updated screenshot: 

screen shot 2018-05-18 at 12 46 32 pm

Thanks!

isundaylee:
@evanshultz Thanks! Done. Just to confirm, did you mean `E73*2G4M04S*` or `E73-2G4M04S*` as the footprint filter?

evanshultz:
Either way will be OK with the current filtering method, actually. This should be fine.

PR #575, opened 2018-05-04
LM2594 buck converter

ppelleti:
[Datasheet](http://www.ti.com/lit/ds/symlink/lm2594.pdf).

I modeled the NC pins (1, 2, and 3) as GND pins, because section 11.3 of the datasheet says "For best thermal performance, ... all ground and unused pins must be soldered to generous amounts of PCB copper, such as a ground plane ..."

It seemed like marking these pins as GND pins was the easiest way to make sure they're soldered to the ground plane.

lm2594

------------


Ratfink:
Thanks for the contribution!

All of our other Simple Switchers have symbols that are some variation on this design:
screenshot from 2018-05-11 16-03-00
It would be nice to have this PR match that as well.  They also all use the pin names shown above, regardless of whether the datasheet shows them as complete words or abbreviations.  I'm not sure how I feel about that; what do y'all think @evanshultz @poeschlr @jkriege2?

It also looks like all the other Simple Switcher symbols in the library are broken down so that the symbol specifies an exact footprint.  I also see quite a few aliases for high voltage variants.  So then, I'd like to see:

* Two copies of this symbol, `LM2594M-3.3` for the SOIC package, and `LM2594N-3.3` for the DIP
  * Set the default footprint for each, and make the filters more specific (the SOIC needs body size and pin pitch, and the DIP needs width)
  * Add short package names to the ends of the descriptions, i.e. `SO-8` and `DIP-8`
* Aliases of each for all voltage input and output variants (HV and normal; 3.3, 5.0, 12, and ADJ)
  * Output voltage listed before the rest of the description, like the other Simple Switchers
  * Mention input voltage just before the package name in the description
    * Preferably list the actual maximum input voltage for all aliases, e.g. `Maximum VIN 45V`, rather than just "High Voltage Input" like the ones already in the library.  Or list the whole VIN range, e.g. `VIN 4.5-45V`.

ppelleti:
OK, this time I based the symbol on LM2576HVS-12, although I expanded it to make room for the extra GND pins.  I've done as you requested regarding the footprints, descriptions, and aliases.

lm2594n

Ratfink:
Looks good, thanks for making those changes!

The one thing I'm still unsure of (which I now see I forgot to mention above) is pins 1-3 (gah, that's three things!).  Yes, the datasheet says on page 3 that they must be soldered to the board, and it suggests soldering them specifically to the ground plane on page 31.  But I don't think calling them GND is appropriate, because that isn't what they actually are.  My suggestion would be to name them NC and set their electrical type to *Passive* (setting it to *Not connected* would create an ERC error when they're connected to anything, which we definitely don't want).

I'm not sure if the NC pins should stay on the bottom or move to one side.  The bottom makes sense because it keeps the symbol smaller and they'll probably all be connected to ground in most cases, but since they aren't truly ground pins that might be a KLC violation.  @evanshultz @poeschlr @jkriege2 Any thoughts on this?

ppelleti:
Here is what I came up with:

lm2594m-thermal

lm2594n-thermal

evanshultz:
Yes, calling them NC and Passive type is best to me as well. I don't think the text about those pins is necessary, though.

Can the symbol be smaller? It seems overly tall and wide.

ppelleti:
OK, symbol is now smaller.

lm2594n-smaller

evanshultz:
Looks OK to me. I suppose now it's a bit bigger than the existing symbols above, but I'm OK with that. @Ratfink, would you please merge when you're ready?

Ratfink:
Looks good to me too.  Thanks for all the good work on this PR!

PR #577, opened 2018-05-04
Add new symbols for dual row connectors with odd pin count

poeschlr:
This adds newly scripted connector symbols for dual row connectors with a odd pin count. (example FFC connectors)

The main benefit is that this adds a version with mounting pins and one with shield pins. (So this is the symbol side fix for https://github.com/KiCad/kicad-footprints/issues/562)

Script PR: https://github.com/KiCad/kicad-library-utils/pull/239

---

screenshot from 2018-05-05 01-56-47


evanshultz:
I hope I didn't miss any errors, but I did find that the shielded symbols have the SH pin off grid. Can you fix that?

poeschlr:
This is true for all dual row shielded symbols (and also for the ones with mounting pins)
There is no way to get them onto the grid in x direction without increasing the size of the symbols. (If we want the one pin on the bottom centered on the body.)

Issue #451, opened 2018-04-03
RC2 is near. Is there anything big that anybody wants to be done before that?


poeschlr:
Mailing list announcement: https://lists.launchpad.net/kicad-developers/msg35171.html

I think we might want to introduce something similar to a feature freeze as soon as RC2 is tagged. (So no library or part renaming. No major changes to existing parts -> examples of things that might be a bad idea during that time frame: symbol size changes, changing pin placements, changing the number of units, ....)

So if you have something planned that you really want to be included in version 5, you might need to do it now.

If there is a known major problem that we can not fix till the end of this week we might want to talk to the developers. (It might require the release date to be pushed back.)

poeschlr:
Forgot to tag you guys: @evanshultz, @jkriege2, @Misca1234, @Ratfink, @Shackmeister

poeschlr:
One can turn off automatic text placement. (In v4 this was in the general preferences dialog. Not sure if this is still the case in v5.) There is nothing we can do about it on the library side so it is not really worth worrying about it. (You can file a bug report or add your voice to this one: https://bugs.launchpad.net/kicad/+bug/1682496)

---

About the connector lib. I totally missed your pull request. I am not sure i will have time this week to review it. (I will try my best.) But it will definitely help if you could fix the merge conflicts. 

For the future please try to split up pull requests such that you have one per topic. (This helps with reviewing. Yes i know for symbol libs it might require more merging / rebasing.)

Ratfink:
I'd still like to see #305 make it into 5.0, but I realize that it might be a bit much at this point unless we want to review it with only a few spot checks.

Shackmeister:
My current ongoing projects I would like to finish for V5:
New 3D models for LEDS https://github.com/KiCad/kicad-packages3D/pull/280
New SMD Electrolytic Capacitor footprints https://github.com/KiCad/kicad-footprints/pull/372
Fix SMD Electrolytic Capacitor 3D models, the old are rotated 180 degrees
Renamed picoblade 3D models to include the mounting-pins renaming

My wishlist:
Add JST GH connector models 
Go through all the common packages and make sure the 3D models are up to date

evanshultz:
@poeschlr 
I unbroke my branch so it can be merged. If anybody has time to review please let me know. I do understand there is a lot covered in that PR.

evanshultz:
It isn't big, but I think https://github.com/KiCad/kicad-symbols/pull/454 needs to be merged.

[...]

PR #578, opened 2018-05-05
Remove duplicate power pins (1 of 2)

evanshultz:
@poeschlr @jkriege2 @Ratfink 
At https://github.com/KiCad/kicad-symbols/issues/451#issuecomment-385770534 I noted that many symbols have power pins on multiple units; this can allow nets to be shorted (or other weirdness) and also breaks the pin count filter in CvPcb.

For multi-unit symbols, I moved the power supply pins to a new unit with the highest unit number and set the symbol to be non-homogenous. Except if some power supply pins are shared, like STK433_Sanyo in Amplifier_Audio, where moved the pin to the lowest-number unit. For parts without a triangle symbol where power pins can't nicely stack, like AD8213 in Amplifier_Current, I left power pins only on the first unit.

I also moved all NC pins to only the first unit.

Amplifier_Operational will be next to get any feedback/corrections first and give https://github.com/KiCad/kicad-symbols/pull/513 more time to be finished and merged (hopefully).

Here is the list of what I've touched:

Amplifier_Audio
- LM1876
- LM1877
- LM4752TS
- LM4766
- LM4950*
- OPA1622 (has non-Line pin type for disable pin)
- SSM2120 (should be moved to Audio.lib)
- SSM2122 (should be moved to Audio.lib)
- STK43x_Sanyo (have a shared power pin so moved all to first unit)
- TDA1308
- TDA2005
- TDA7264
- TDA7265
- TDA7266
- TDA7266D
- TDA7266P
- TDA7496

Amplifier_Buffer
(no changes)

Amplifier_Current
- AD8213

Amplifier_Difference
(no changes)

Amplifier_Instrumentation
(no changes)

Amplifier_Operational
- ADA4807-2ACP
- L272
- L272D
- L272M
- LM13700 (this symbol could use a bit of clean up IMO)
- LM2902
- LM2904
- LTC1151CN8
- LTC1151CSW
- LTC6081xDD
- LTC6082xDHC
- LTC6082xGN
- MCP6002-xMC
- OP249GS
- OP279
- OPA2333xxDRB
- OPA2376xxYZD
- OPA2691-14 (should actually be named ` OPA2691x-14`)
- OPA2695xD
- OPA2695xRGT
- TSV912IQ2T

Amplifier_Video
- AD813 (has non-Line pin type for disable pins)

Analog
(no changes)

Audio
(no changes)

Comparator
- LM2901
- LM2903
- LM319
- LM319H
- MCP6569

------------


evanshultz:
Don't merge this yet. I just noticed that some of the differential output (BTL) audio amps have small lines on all units.

evanshultz:
OK. Done. Ready for merge, I believe.

(And as I committed, I realized what commit message was *what* and not *why*. Sorry.)

evanshultz:
Because I didn't see any reply, and I want to get it over with, I fixed the last library where I see parts that need to be updated. If there are more I've missed, please let me know.

evanshultz:
@poeschlr @jkriege2 @Ratfink 
Can somebody look at this and merge it? I'd really like to have this done for v5 since CvPcb is broken now.

poeschlr:
- [ ] LM4766 has an empty c unit
- [ ] OPA1622 has an unnamed pin floating around on the c unit. Seems a bit strange to me.
- [ ] SSM2120 has a strange floating rectangle on unit c
- [ ] TDA7266 one of the lines connecting the upper output pin has been wrongly moved to unit c
- [ ] LM13700 power pins without names
- [ ] OPA2333xxDRB strange floating pin without a name on unit c
- [ ] TSV912IQ2T strange floating pin without a name on unit c

evanshultz:
Thanks for looking!

The floating pins are other power pins which can overlay on the triangle symbol. If it's a control symbol, I moved it to the first unit. But common power pins I put into the power unit. For example, pin 9 of OPA2333xDRB is the thermal pad, which the [datasheet ](http://www.ti.com/lit/ds/symlink/opa333.pdf)says can be floating or connected to V- on page 14. So it needs to be a separate pin and seems to go with the other power pins. And [OPA1622 ](http://www.ti.com/lit/ds/symlink/opa1622.pdf)has V+, GND, and V- pins. Wouldn't all of those three go into the power unit so it can stack on top of unit 1 or 2? The pins appear to be floating because they didn't have a name before due to the size of the triangle causing the text to run over the symbol body. Would you prefer a different solution for those pins? This is items 2, 5, 6, and 7.

Thank you for catching the other items.

evanshultz:
@poeschlr 
Can you take a look at my question above? How do you want to handle symbols that have more power pins than just V+ and V-?

evanshultz:
@poeschlr 
How do you want to handle these power pins so this PR can be finished?

poeschlr:
I would suggest that we discuss this in a separate issue. I think it might be a good idea to add at least some pin name to it as the symbol is otherwise hard to read. (Even if the power unit is stacked on top of any other unit)

Please fix the other issues at least such that we can include this in kicad 5 rc2

evanshultz:
Other issues are fixed. So, if you want, we can merge.
- LM4766: Did not need a 3rd unit since only the Vee pin was in common (went on unit A).
- SSM2120: Weird; I missed it when I reviewed.
- TDA7266*: Fixed.
- LM13700: Power pins never had names because units A/C and B/D are so different. Putting any names on those pins will conflict with one pair of units. Maybe just put them on unit A?

Another option would be to put the triangle symbol on all units. Then it's obvious if many pins are on the power unit that they go together (they're not floating), and it still allows stacking the power unit on other units.
image

poeschlr:
My reasoning behind wanting pin names for all pins is as follows:
Consider ADA4807-2ACP from your screenshot. Without looking at the datasheet, what are pins 5,11,6?
In my mind a schematic should be self explanatory. Meaning you should not need to look at the datasheet to find out what a particular pin is. (You might still need to look at the datasheet to find out how that pin should be used but a name is much easier to understand than a number)

I am prepared to merge this without this being decided and open an issue for it. Right now i don't really have time to look at it so i will take a look tomorrow morning.

evanshultz:
Sounds good. Thanks for taking another look and trying to squeeze this in for v5.

I agree pin naming could use continued work. Once this is merged I will continue on with that, the minor comments I made above, and also separating these symbols by their available packages.

This discussed is summarized and can be continued at https://github.com/KiCad/kicad-symbols/issues/190.

poeschlr:
Thanks.

PR #581, opened 2018-05-06
Add ADS1XXX series of TI ADCs

reportingsjr:
Add the following TI ADC symbols:
  ADS1013
  ADS1014
  ADS1015
  ADS1113
  ADS1114
  ADS1115
  ADS1018

ADS1015 was proven out with a working PCB.
------------

   ADS1013, 1014, 1015: http://www.ti.com/lit/ds/symlink/ads1015.pdf
   ADS1113, 1114, 1115: http://www.ti.com/lit/ds/symlink/ads1115-q1.pdf
   ADS1018: http://www.ti.com/lit/ds/symlink/ads1018-q1.pdf
   ADS1113, 1114, and 1115 symbols are the same as 1013, 1014, and 1015 respectively, just 16 bit instead of 12 bit ADCs so I didn't screenshot them
ads1013
ads1014
ads1015
ads1018
   Matched footprints work


reportingsjr:
I had to move the pins vertically since I accidentally had them on a 50mil grid. I also removed the datasheet link that shows in the screenshots. Too lazy to go back and update the pictures though, haha.

evanshultz:
Thanks for adding these nice parts!

- Use an ALIAS instead of discrete symbols for parts that are pin- and package-compatible. It seems the ADS1013 and ADS113 can be ALIASes, for example.
- Pin name offset should be 20mil. Only ADS1018 is correct here.
- Do not add symbols for automotive versions, so use http://www.ti.com/lit/ds/symlink/ads1113.pdf for ADS111x. And update datasheets to use non-automotive ones.
- ADS101x and ADS111x have quite different packages but are pinout-compatible. You have added a single footprint filter but no default footprint, so please add that and the MPN suffix (at least for parts which have multiple packages). Or else leave off the suffix (as you have now) and add all packages in the footprint filter with no default footprint. Right now your symbols are in the middle of fully specifying the package(s) and being generic (but not covering all packages). Ask if you have questions about this.
- ADS1018 only has one packages, so please add a default footprint and the suffix of ``.
- Use the datasheet description and put the package name at the end. So for ADS1014 it would be `Ultra-Small, Low-Power, I2C-Compatible, 3.3-kSPS, 12-Bit ADCs With Internal Reference, Oscillator, and Programmable Comparator, VSSOP-10` if you're making symbols for each package.
- Use only pins of style Line.
- SCL pins should be of type Input.
- ADS1014, for example, needs pin 2 to name named `ALERT/RDY`.
- Unused pins, like on ADS1013, need to have the unused pins as actual pins. Make them invisible NC pins with the pin end on the symbol border. This is to make the CvPcb pin filter work.
- Can you align the ADS101x symbols? Make them look alike in pin placements and then just remove pins for the ones with less features. That will make it easier for users to swap symbols. And you may be able to shrink the symbols vertically, especially after lowering the pin name offset. But using the full name for pin 2 might make it harder. See what you can do.
- ADS1018 should have DIN on the left side since it's an input. And make that symbol smaller since it's overlay large.

reportingsjr:
- [x] Use an ALIAS instead of discrete symbols for parts that are pin- and package-compatible. It seems the ADS1013 and ADS113 can be ALIASes, for example.

Wow, I never actually looked at what ALIASes were. That would have saved me a bit of time earlier!!
- [x] Pin name offset should be 20mil. Only ADS1018 is correct here.
- [x] Do not add symbols for automotive versions, so use http://www.ti.com/lit/ds/symlink/ads1113.pdf for ADS111x. And update datasheets to use non-automotive ones.
- [x] ADS101x and ADS111x have quite different packages but are pinout-compatible. You have added a single footprint filter but no default footprint, so please add that and the MPN suffix (at least for parts which have multiple packages). Or else leave off the suffix (as you have now) and add all packages in the footprint filter with no default footprint. Right now your symbols are in the middle of fully specifying the package(s) and being generic (but not covering all packages). Ask if you have questions about this.
ADS1018 only has one packages, so please add a default footprint and the suffix of ``.

I went ahead and switched them over to package specific symbols since the kicad doesn't have the other footprint and I'm not going to make it.
- [x] Use the datasheet description and put the package name at the end. So for ADS1014 it would be Ultra-Small, Low-Power, I2C-Compatible, 3.3-kSPS, 12-Bit ADCs With Internal Reference, Oscillator, and Programmable Comparator, VSSOP-10 if you're making symbols for each package.
- [x] Use only pins of style Line.
- [x] SCL pins should be of type Input.

Good catch!
- [x] ADS1014, for example, needs pin 2 to name named ALERT/RDY.
- [ ] Unused pins, like on ADS1013, need to have the unused pins as actual pins. Make them invisible NC pins with the pin end on the symbol border. This is to make the CvPcb pin filter work.
- [x] Can you align the ADS101x symbols? Make them look alike in pin placements and then just remove pins for the ones with less features. That will make it easier for users to swap symbols. And you may be able to shrink the symbols vertically, especially after lowering the pin name offset. But using the full name for pin 2 might make it harder. See what you can do.

I went ahead and aligned them so you can directly swap them. Unfortunately pin 2 is pretty close to VDD. I'll let you decide if the packages needs to be taller. I don't mind them how they are now
- [ ] ADS1018 should have DIN on the left side since it's an input. And make that symbol smaller since it's overlay large.

Are you sure about that? Every other SPI ADC has all of the SPI bus pins on the right side (e.g. LTC1594, ADS1120, etc). It seems weird to split off the DIN pin on its own. For what it's worth the ADS1018 was a copy of the ADS1118I. I just made it an alias and adjusted the ADS1118I symbol. Let me know what you think of it.

I'll squash the commits before you merge if you would like. Thanks for being so incredibly thorough! This is some serious attention to detail!

reportingsjr:
* Remove leading asterisk from footprint filters. This was fixed in KiCad a couple weeks ago.
Done
* Put the package type at the end of the descriptions. For example VSSOP-10.
Done
* ADS1x18I are missing the DGS suffix. If the footprint is not in the lib now, just leave it off.
Those two have two associated footprints, so I didn't add the suffix. (DGS is for the VSSOP-10 footprint)
* For easier sorting and my OCD, would you mind making ADS1018 the symbol with a ADS1118 ALIAS?
Done
* ADS1118I pin name offset is not 20mil.
Done
* VSSOP-10 footprint filter should be TSSOP*3x3mm*P0.5mm* (missing first 2 chars).
Done
* ADS1x13IDGS do not have the programmable comparator; fix description.
Done
* KLC says inputs on the left. It also says to group busses together. So putting DIN on the left is actually in conflict with the library rules either way. The symbols you mention are quite old so I'm not sure they're a good reference. @poeschlr @jkriege2 @Ratfink Keep the SPI pins together or put inputs on the left?
I'll wait to hear back from others. :)

poeschlr:
SPI pins should stay together (Any bus should if feasible have all its pins on the same side.) Unless spi is a secondary function then it is debatable what rule is more important.

reportingsjr:
Updated. I didn't realize that the QFN package for the ADS1x18 parts wasn't in the library. I should have checked!

I also changed the keywords to match the other symbols I added.

evanshultz:
In the future please use incremental changes instead of overwriting everything. This makes it easier to track changes during the review process.

reportingsjr:
Thanks for being so thorough!
I had asked if you wanted the commits squashed before merging. How exactly did you want me to squash them?

evanshultz:
Oh yes. I forgot.

@poeschlr 
Squashing commits is something brought up just recently. Do you have a best practice or desired method?

poeschlr:
The contributors should commit as normal. (No commit --amend, Nothing that would require push --force)

However we the maintainers can use the squash merge feature instead of normal merge to clean up the history of contributions that did not really take care of their git history.

The green button that you normally press to merge has a second button on the right side. (with a downwards arrow on it.) This button allows you to select between 3 options on how to merge a pull request.
Option 1 is the normal merge. This creates merges the history of the master branch with the history of the contributors branch. (Meaning every commit made by the contributor, including merge commits by him are then also present in the history of the master branch)

Option 2 is the squash merge. This creates a separate branch before merging. The commits by the contributor are automatically combined into one commit. This option can help to reduce clutter in the git history of the master branch. (Assuming the person merging enters a sensible commit description into the text fields provided by the github interface)

Option 3 is a rebase instead of merge. I have not yet used this option via the github interface. (I however sometimes use rebase locally as it can avoid merge commits as long as there is no conflict.)

PR #583, opened 2018-05-07
Split 74xx595/74xx596 due to different output pin types.

lmoelleb:
Fixed name of pin 9 (it is not the inverted version of pin 7).

Adjusted datasheet links.

Changed pins to 100mil and remove the line types.

Datasheets:

http://www.ti.com/lit/ds/symlink/sn74hc595.pdf
https://assets.nexperia.com/documents/data-sheet/74HC_HCT595.pdf
http://www.ti.com/lit/gpn/sn74ls595

http://www.ti.com/lit/gpn/sn74ls596
https://assets.nexperia.com/documents/data-sheet/74HC_HCT595.pdf

I was tempted to the pin names using digits from NXP, but did not want to make a change based on personal preference only

image

image

------------


evanshultz:
Good catch! Thank you.

The vast majority of symbols in this lib use TI datasheets, and the symbols have TI pin numbers. Please switch to TI datasheets. If you can do that, I'll merge.

lmoelleb:
@evanshultz I only use NXP for the HTC version as I can't find any TI datasheet for these. LS/HC already links TI datasheet and pins use TI naming,

evanshultz:
Thank you!

PR #584, opened 2018-05-08
Fix CY7C68013A pin polarity

whitequark:
All of PKTEND, SLOE, SLRD, SLWR are active low unless configured otherwise, so they should all be of the same type in KiCAD. Flags are active low too if configured as EF or FF, but if configured as PF they don't have any particular polarity, so I've left them as is.

[Datasheet](http://www.cypress.com/file/138911/download) citation:
screenshot_20180508_000747

New symbols:
screenshot_20180508_000930
screenshot_20180508_000907

------------


evanshultz:
I disagree. Using an overbar implies the pin is active low, and this is not always the case for these pins. If the user would choose active high, then the pin name is definitely in conflict. Do you have other thoughts?

@Ratfink @poeschlr @jkriege2 
Unless one of you librarians disagree, or compelling info come from the OP, I'll close this PR in a couple days.

whitequark:
@evanshultz OK. But then SLRD and SLWR must lose the overbar, since they are also configurable polarity. Do you agree with that?

evanshultz:
I agree with that philosophically. If I was writing the datasheet I agree with your statement fully and I would omit the overbar.

But I don't agree to change it in this case. It sounds wishy-washy, but let me elaborate.

The datasheet names these pins specifically as active low on page 18, and I'd like to use the datasheet pin names where possible for clarity. Here it doesn't really make sense why it's done this way by Cypress, but KLC defers to the datasheet or manufacturer in general.

If the datasheet doesn't provide guidance, then we can get creative. But I vote to use datasheet pin names if given, and if the datasheet is incomplete or unclear or whatever case-by-case issue arises, then deviate. IMO it's more important for the user to match the symbol with the datasheet when using the symbol.

If KiCad adds an overbar that implies a pin is always active low. In this case that's not necessarily true. But I'm inclined to copy Cypress' awkward pin naming (or mistake!) for other pins since the datasheet is canon.

whitequark:
@evanshultz The Cypress datasheet agrees with me (page 17). The only FIFO bus pin that's specified as active low is SLCS#.
screenshot_20180512_203753
I've updated the symbols to match the Cypress datasheet:
screenshot_20180512_203923

Also, as a matter of fact, I think that the reasoning in the comment above is flawed and inconsistent. When we discussed I2C pin types for DRC, you made a good argument for making the pin types strictly speaking inconsistent with the datasheet in order to make it less likely that a designer would make a mistake that goes unnoticed by DRC. Here, removing the overbar would have the same effect (in fact, I *did* make a mistake in my design because of this misleading information--thankfully not a serious one).

whitequark:
> Here you're only changing the pin name, right? I didn't see anything about pin types being changed in your commit, only the overbar being removed. Did I miss something about pin types changing?

You didn't. I was making an analogy to make a point about usefulness of following the datasheet in ambiguous cases.

> Your patch is only removing the overbar from SLRD and SLWR. Page 18, as I noted above, shows them as active low (which seems silly to me since the polarity can be chosen by the user).

I don't believe it does. Page 18 shows SLRD and SLWR as follows:
screenshot_20180512_212620
The star in this case means *programmable polarity*, not *active low*, see note at the bottom of the page:
screenshot_20180512_212627

evanshultz:
> The  Cypress datasheets use the # symbol for denoting active low, not *.

And that's what I missed. Thanks for clarifying and sticking with it until I figured out what you meant. This is patch is only removing the overbar from those two pins, which is fine with me since the active state isn't fixed. Thank you.

PR #588, opened 2018-05-10
 Added symbol for Texas Instruments TPS22993 load switch 

andrasfuchs:
Added symbol for Texas Instruments TPS22993 load switch

image

Datasheet: http://www.ti.com/lit/gpn/tps22993

------------


andrasfuchs:
Thank you for taking the time again for the review!

I think you are right, the exact footprint is not yet in the library. I'm not familiar with the footprint editor, could you guide me what to do? Should I wait until someone creates the footprint this IC has?

evanshultz:
Of course! Thanks for making the updates!

Since we don't prefer to merge symbols when the footprint is missing, or add symbols that have the wrong footprint, I would like to get the right one added before merging this symbol. Creating a footprint isn't tough and it's a fundamental KiCad skill. Are you willing to do it? If so, you're probably best to start with https://github.com/KiCad/kicad-footprints/blob/master/Package_DFN_QFN.pretty/QFN-20-1EP_3x3mm_P0.45mm_EP1.6x1.6mm.kicad_mod and then modify it (we don't really have anything that's very close already as this is a weird footprint) or just start from scratch. Submit the footprint (even if it's incomplete) with any questions at https://github.com/KiCad/kicad-footprints when you're ready. You can tag me specifically if you like but all librarians are capable of helping. Lastly, I suggest checking out the KiCad forum and YouTube for tips on how to make a footprint.

andrasfuchs:
Alright, could you help me find the proper name for the footprint? I think it would be better if I could get its name right for the first time, otherwise I would need to rename it later probably causing some confusion for me and others :)

So, for example, in the QFN-20-1EP_3x3mm_P0.45mm_EP1.6x1.6mm example, what does the "1EP" the "P0.45mm" and the "EP1.6x1.6mm" mean? What would you name my footprint according to the spec?

andrasfuchs:
@evanshultz Thanks!

PR #591, opened 2018-05-10
CUI PJ-063AH barrel jack

ppelleti:
[Datasheet](https://www.cui.com/product/resource/digikeypdf/pj-063ah.pdf).

This barrel jack does not contain an internal switch, so it only has two pins, instead of the three that the standard Barrel_Jack symbol has.

(Actually, since this barrel jack has a metal shield, there ends up being a third pin for that.  But it does not mean the same thing that pin 3 means in the standard Barrel_Jack symbol.)

This symbol keeps the same centering as the original Barrel_Jack symbol, and like the original, it does not have a yellow box around the symbol.  The KLC script is unhappy about both of these things.

barrel_jack_cui_pj-063ah

------------


evanshultz:
Thanks!

I'd like to approach this a bit differently. The existing symbol is generic and so should this one. But, the existing symbol doesn't have a clear name that allows building other variants. In that vein, here is what I think will be best. Please let me know if you can help out with this, and if not you may close this PR and I'll work on it. Up to you!
- Move your pin 2 to align with the existing symbol so that swapping symbols is easier.
- Your pin 3 should be named `MP`, the name we give to mounting pins. (It is not a shield.)
- Place that pin at the bottom of the symbol, perhaps like the symbols in `Connector_Generic_MountingPin`.
- Clear the default footprint.
- The footprint filter should be `BarrelJack*`.

If you want to rename the existing symbol and name yours as below, and also update descriptions, that would be great. Otherwise, I'll do the renaming and description updates later.

This is for my own information. We need at least 4 symbols:
- `Barrel_Jack`: basic 2 pin symbol
- `Barrel_Jack_Switch`: 3 pin symbol (this is the existing symbol)
- `Barrel_Jack_MountingPin`: 3 pin symbol (the one here)
- `Barrel_Jack_Switch_MountingPin`: new 4 pin symbol that is the superset of the above

ppelleti:
OK, I've removed `Barrel_Jack_CUI_PJ-063AH` and I've added two of the symbols from your list, `Barrel_Jack_MountingPin` and `Barrel_Jack_Switch_MountingPin`.

I added a yellow rectangle around the symbol, so that the mounting pin would have something to be below, as in the `Connector_Generic_MountingPin` symbols.

barrel_jack_mountingpin

barrel_jack_switch_mountingpin

evanshultz:
@ppelleti 
Thanks! This looks good to me.

@poeschlr 
Can you take a look? Is this good with you and the style we want to propogate in the library for mounting pins?

ppelleti:
> Your pin 3 should be named MP, the name we give to mounting pins. (It is not a shield.)

For what it's worth, the datasheet does describe the jack as "shielded."

poeschlr:
The mounting pin looks fine to me. (But i might be a bit biased as it looks exactly like the mounting pin used in my generic connector symbols.)

ppelleti:
> But i might be a bit biased as it looks exactly like the mounting pin used in my generic connector symbols.

Indeed, I copied it from there.

evanshultz:
Thank you @ppelleti!

PR #593, opened 2018-05-11
Ltc3892

SchrodingersGat:
This PR adds the LTC3892 series of switching regulator controllers

http://www.analog.com/en/products/power-management/switching-regulators/step-down-buck-regulators/external-power-switch-buck-controllers/ltc3892.html

* Added LTC3892 / LTC3892-1 / LTC3892-2
* Moved existing LTC389x to Regulator_Controller.lib

[Datasheet](http://www.analog.com/media/en/technical-documentation/data-sheets/38921fc.pdf)


SchrodingersGat:
Errors appear to be on symbols I did not touch...

evanshultz:
Common:
- Pin2 is `PLLIN/MODE`.
- Pins 3 and 4 should be type C.
- Pin 8 is `ILIM`.
- Pin 13 should be Input type.
- Pin 14 could be Input type, but Passive is OK too, I suppose.
- Pin 20 should be `DRVcc`.
- Pins 15 and 27 are missing `TRACK/`.
- I'd be tempted to make pins 29 and 12 Passive type.
- Pin 33 should be just `GND`.
- Is the `EP` necessary in the footprint filter since the pin count will catch this?
- What about putting a pin's space between PGOOD1 and PLL/MODE on the left side for clarity?

LTC2892-1:
- Footprint filter is missing `mm*P0.65mm*` at the end.
- The description doesn't mention it's also forced continuous conduction mode.

LTC3890*:
- Since you touched this symbol (just shifted it, actually), would you mind moving the SS* pins from the bottom? Not required, but would be nice.

image

SchrodingersGat:
[x] - Pin2 is PLLIN/MODE.
[ ] - Pins 3 and 4 should be type C.
[x] - Pin 8 is ILIM.
[x] - Pin 13 should be Input type.
[x] - Pin 14 could be Input type, but Passive is OK too, I suppose.
[x] - Pin 20 should be DRVcc.
[x] - Pins 15 and 27 are missing TRACK/.
[x] - I'd be tempted to make pins 29 and 12 Passive type.
[x] - Pin 33 should be just GND.
[x] -  Is the EP necessary in the footprint filter since the pin count will catch this?
[x] - What about putting a pin's space between PGOOD1 and PLL/MODE on the left side for clarity?

SchrodingersGat:
image

evanshultz:
Thanks for the updates!

LTC3890-1:
- ITH* pins are not Passive type

LTC3892-1:
- Please put the package type at the end
- The footprint filter still has `EP*` in it

Ratfink:
@evanshultz We decided in https://github.com/KiCad/kicad-symbols/issues/473#issuecomment-382897958 to include the number of EPs in footprint filters to avoid any possibility of conflicts between similar footprints.  So @SchrodingersGat please change `EP*` to `1EP*` rather than removing it. 🙂

SchrodingersGat:
Alright, think this should be good to go now.

SchrodingersGat:
I have also added the LTC7810.

This is a 150V part which is roughly compatible. The pinout is as close to drop-in replacement as I could make it.

image

http://www.analog.com/media/en/technical-documentation/data-sheets/LTC7810.pdf

Note: The LTC7810 requires a new footprint - https://github.com/KiCad/kicad-footprints/pull/614

SchrodingersGat:
I was not sure about the ITH pins but I think that passive is probably the best choice.

Changes have been made :)

SchrodingersGat:
The footprint for the LTC7810 has now been merged, too

evanshultz:
As mentioned at https://github.com/KiCad/kicad-symbols/pull/593#issuecomment-391564453, the footprint filter for LTC7810 should include the `1EP` section. If you can change that I will merge.

I don't really find a good place for the NDRV pin so I guess leaving it alone is fine.

SchrodingersGat:
The footprint filter for the LTC7810 is `LQFP?48*7x7mm*P0.5mm*EP3.6x3.6mm*` - is this not correct?

SchrodingersGat:
Oops! Fixed :)

PR #594, opened 2018-05-12
Adding Texas Instruments PCM5102 symbol

jeanthom:
Hi, this is my first contribution to the KiCad project !

I'd like to add [Texas Instruments PCM5102](http://www.ti.com/lit/ds/slas764b/slas764b.pdf) audio DAC chip. My symbol looks like this :

PCM5102 symbol


evanshultz:

You will find a lost of rules and guidelines for the KiCad library at http://kicad-pcb.org/libraries/klc/. Please see that webpage for more information about the review items below.

- Please remove the superscript "TM" from the description.
- Add the package type of `TSSOP-20` to the end of the description.
- Place the part number above the top right of the symbol.
- I would consider putting the I2S pins at the top left and analog out at the top right. What do you think?
- I would also put digital power pins left of the analog power pins so this mirrors the D->A signal flow.
- Please put CPVDD on the left or right side. It and VNEG should be Passive type and if they're close to the bottom of the symbol it will be easier to connect them to ground.
- Space apart CAPP and CAPM so a cap symbol can fit nicely there. They should also be type Passive.
- LDOO should probably be close to the bottom of the symbol since it has a cap to ground.
- Would you be willing to add the other parts in this family as an ALIAS? It's not required, but it would be appreciated if you can.

Please post a new screenshot when you make cosmetic changes so it's quicker to review.

Thank you for contributing to KiCad!

antoniovazquezblanco:
@jeanthom could you please address the comments above?

antoniovazquezblanco:
@jeanthom Thanks for your contribution. No need for apologies please :)

Don't worry about git history, we will merge squashing all the history into a single commit.

jeanthom:
I'm currently working on adding aliases for similar parts to the PCM5102. I did aliases from PCM5100 & PCM5101 to PCM5102 ; perhaps I should rename PCM5102 to PCM5100 and point both PCM5101 and PCM5102 to PCM5100 ?

I am also running into an odd issue with the CI tests : I can't fill the datasheet field in the PCM5102 ("Datasheet field must be EMPTY"), but I'm forced to provide the datasheet field for the aliases.

evanshultz:
Thanks for making changes! As stated above, no need to apologize. :)

Follow up to my earlier review:
- Ref de should stay at top left. Only the value moves to the top right.
- CPVDD should be at the top. The CP pins should also be next to the digital power pins. Sorry for some confusion on my part above.
- Make CAPP/CAPM Passive type.
- You may be able to make the symbol less wide, less tall, or both.

Several commits are messy. It appears your branch is rather out of date, but that's because this PR was started in May. Sorry that we're slow at getting to them but hopefully we're getting faster. I appreciate you coming back to this PR and helping to get it finished. Anyway, back to work...

You can make PCM5100 the symbol name and move PCM5102 to an ALIAS. Most of the time we will do that so the lower power number is the symbol and higher part number(s) are ALIASes. But the way you have it is also acceptable.

The datasheets look correct. Either the links you have or http://www.ti.com/lit/ds/slas764b/slas764b.pdf are valid.

jeanthom:
pcm5100

This is what the footprint looks like now.

About the Git history issue, I tried to pull the latest changes in my fork so that it is even with KiCad's master. Perhaps I should create another Git branch once the symbol gets validated, so that the merge will not be a mess ?

evanshultz:
I think things should be OK since Git shows only two files changed. And with a squash merge the Git history will be clean.

I just noticed the pin type of LDOO. I believe it should be Passive since I find no indication this pin can source current for other circuits. Do you see the same thing? If so, please change that pin type and then I think this is ready to merge.

jeanthom:
You are right, while it is the internal regulator output, it isn't meant to be used by the end user, it is only meant to be connected to a decoupling cap.

evanshultz:
Great.

Thanks so much! We appreciate your contributions to the KiCad library and hope you will continue!

PR #597, opened 2018-05-14
(re)add AD9910

lorem-ipsum:
This chip got lost during the library transition. It is a simple copy of
the symbol from the old analog_devices.lib, only the footprint filter was
adjusted to match the new scheme.

It was added to Interface as the other DDS seem to be in interface. It is
not really the library I would expect them in, maybe an independent DDS
lib would be useful.

[datasheet](http://www.analog.com/media/en/technical-documentation/data-sheets/AD9910.pdf)
screenshot

------------


evanshultz:
Thank you!

PR #598, opened 2018-05-14
add LTC6957 dual output logic converter

lorem-ipsum:
This adds the LTC6957 (by now from Analog). [Datasheet](http://www.analog.com/media/en/technical-documentation/data-sheets/6957fb.pdf). I've placed them in the Timer lib, but I am not really sure about this decision.

Attached are the symbols for the DFN variants which have an exposed pad. The MSOP look the same without pin 13.
xdd-3
xdd-1

------------


evanshultz:
Thanks for these nice symbols!
- I'm not sure about that lib either. What would you think about `Interface`, `Interface_LineDriver`, or `Logic_LevelTranslator`?
- Especially since this part seems to cross libraries, a robust description will be nice. Can you replace `Dual Output Logic Converter` with `Low Phase Noise, Dual Output Buffer/Driver/Logic Converter` from the datasheet? Keeping the logic signal type and package at the end is good.
- For the DFN symbols, stack pins 13 and 5.
- Can the symbols be a bit more in common? After stacking the pins as mentioned above, perhaps flip pins 10/11 and 9/8 on the -1/2 version, then place OUT1 and OUT2 on the OUT+ pins for the -3/4 version, and the finally shift VDD and GNDOUT to the right on the -3/4 version. That, I think, will allow a more uniform look between the symbols in this family.

lorem-ipsum:
I did add most of your suggestions and moved the parts to the `Interface` lib. `Interface_LevelTranslator` might be matching, but as it does not exist yet I did not create it. I did not do the pin swapping as it would break the symmetry and changing between these components will always require adjustments to the schematic anyways. But now they are of the same size and the common supply pins are at the same location.

evanshultz:
Sure! The pin moves were to ensure pin names don't run over each other, and they don't with your latest commit. When you make visual changes please post a screenshot so review is easier.
image

Thank you!

PR #599, opened 2018-05-14
Added TC7662A/B symbols

sebastianbasierski:
Signed-off-by: SebastianX Basierski <sebastianx.basierski@intel.com>

TC7662A
Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/21468B.pdf 

tc7662acpa

tc7662aepa

tc7662aija

tc7662amja

TC7662B
Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/21469a.pdf

tc7662bcoa

tc7662bcpa

tc7662beoa

tc7662bepa


evanshultz:
- We have a `LMC7660` symbol in the library. Please copy that symbol design.
- KiCad doesn't have specific CERDIP footprints. And we don't include temp ratings but instead wildcard them out. And I don't see the MPNs you've shown on Microchip's site (probably the datasheet is older and doesn't include any discontinuations. Putting that all together, I see a much smaller number of symbols needed: `TC7662AxPA`, `TC7662BxPA`, and `TC7662AxOA`. (Even if we did include temp, you'd be able to use the ALIAS feature to at least simplify things a little.)
- Eliminate temp in the descriptions.
- Let's simplify `3V to 18V up to 40mA` in the description to `3V to 18V, 40mA` or `3-18V, 40mA`.
- Please add output current to the B version's description (if you can find it).
- Change the CAP+/CAP- and OSC pins to Passive type.
- On pages 8 and 6 of the 7662A datasheet pin 6 is connected to pin 3. I wonder if that pin should be visible on your symbol so that connection can be made?

sebastianbasierski:
Sorry for late response, i was away from computer for a few days.


> -     We have a LMC7660 symbol in the library. Please copy that symbol design.

Done.

> -     KiCad doesn't have specific CERDIP footprints. And we don't include temp ratings but instead wildcard them out. And I don't see the MPNs you've shown on Microchip's site (probably the datasheet is older and doesn't include any discontinuations. Putting that all together, I see a much smaller number of symbols needed: TC7662AxPA, TC7662BxPA, and TC7662AxOA. (Even if we did include temp, you'd be able to use the ALIAS feature to at least simplify things a little.)

Could You explain me, how did You create names for variants? (TC7662AxPA, TC7662BxPA, and TC7662AxOA). I took parts name from datasheet, also i see them on Microchip page (http://www.microchip.com/wwwproducts/en/TC7662A - "RoHS Information").

> -     Eliminate temp in the descriptions.

Done.

> -     Let's simplify 3V to 18V up to 40mA in the description to 3V to 18V, 40mA or 3-18V, 40mA.

Done.

> -     Please add output current to the B version's description (if you can find it).

Found it. It's 100mA.


> -     Change the CAP+/CAP- and OSC pins to Passive type.

Done.

> -     On pages 8 and 6 of the 7662A datasheet pin 6 is connected to pin 3. I wonder if that pin should be visible on your symbol so that connection can be made?

My bad. Will fix.

I'll deliver fixed commit till the end of week.
Thank You for Your time.

sebastianbasierski:
tc7662ax0a
tc7662ax0a

tc7662axpa
tc7662axpa

tc7662bxpa
tc7662bxpa

I've updated patch.
I'm not sure how to solve issue with pin 6 in tc7662a version.
It's called 'NC' in datasheet, but when i set it as visible, i receive S4.6.
Should i change the name of the pin in order to avoid this error?

evanshultz:
-The part names are from the "base" part number (`TC7662`) plus the suffix for the package. And in KiCad we wildcard out the temperature. So that gives the PNs I listed. For example, `TC7662ACPA` and `TC7662AEPA` differ only in the temperature and that's wildcarded with a `x` so they both fold down into the same symbol.
--Note that I made a typo and `TC7662AxOA` should have been `TC7662BxOA`. Sorry for that mistake. Does that explain things? This one's package should say `SOIC-8`.
--Please use a lowercase `x` for the wildcard.
-Where did you find the B version's output current? From the graph on page 8?
-For the A version leave pin 6 named `NC`, give it Passive type, and make it visible. You will get an error but that's OK as this is an unusual use for an NC pin.

-To match KLC, we should have the positive power pin on the top. Please move it there and sorry again. Otherwise matching the existing symbol as you have done is good.
-Each symbol should have a default footprint and a single, matching footprint filter.
-Pin 8 on the A version is `VDD` while on the B version it's `V+`.
-On the B version please put the BOOST pin next to V+ to make that optional connection easier.

sebastianbasierski:
> -The part names are from the "base" part number (TC7662) plus the suffix for the package. And in KiCad we wildcard out the temperature. So that gives the PNs I listed. For example, TC7662ACPA and TC7662AEPA differ only in the temperature and that's wildcarded with a x so they both fold down into the same symbol.
--Note that I made a typo and TC7662AxOA should have been TC7662BxOA. Sorry for that mistake. Does that explain things? This one's package should say SOIC-8.

Everything is clear now.

> -Where did you find the B version's output current? From the graph on page 8?

Sorry, should have gave the source. It's [here](http://www.futureelectronics.com/en/technologies/semiconductors/analog/regulators-reference/switching-regulators/Pages/7057672-TC7662BCOA.aspx?IM=0).

evanshultz:
Thanks. I'll trust the datasheet graph over Future's website, but either way 100mA is OK. I'll look for another commit to address the items above.

sebastianbasierski:
tc7662axpa
tc7662axpa

tc7662bx0a
tc7662bx0a

tc7662bxpa
tc7662bxpa


>--Please use a lowercase x for the wildcard.

Done.

-For the A version leave pin 6 named NC, give it Passive type, and make it visible. You will get an error but that's OK as this is an unusual use for an NC pin.

Done.
There is a S4.6 violation as You mentioned.

>-To match KLC, we should have the positive power pin on the top. Please move it there and sorry again. Otherwise matching the existing symbol as you have done is good.

You mean power pin from A version?

>-Each symbol should have a default footprint and a single, matching footprint filter.

Done.

>-Pin 8 on the A version is VDD while on the B version it's V+.

Done.

>-On the B version please put the BOOST pin next to V+ to make that optional connection easier.

I assume V+ pins should be placed on a left side of symbol, otherwise i can't place BOOST pin next to them.

evanshultz:
Thanks!
- `TC7662Bx0A` need the correct package in the description.
- It's also missing `P1.27mm*` from the end of the footprint filter.
- The check script assumes NC pins are invisible and Not Connected type, so we'll merge with this error.
- I was also thinking put pin 8 on top for all symbols as requested by KLC. The existing *7660 symbols can get fixed once this is merged. If BOOST is then at the top on the left side it's as close as it can be since it's not a dedicated power pin. Do you see a better way?

sebastianbasierski:
>-  TC7662Bx0A need the correct package in the description.

Done.

>-    It's also missing P1.27mm* from the end of the footprint filter.

Done.

>-    The check script assumes NC pins are invisible and Not Connected type, so we'll merge with this error.

Ok.

>-    I was also thinking put pin 8 on top for all symbols as requested by KLC. The existing *7660 symbols can get fixed once this is merged. If BOOST is then at the top on the left side it's as close as it can be since it's not a dedicated power pin. Do you see a better way?

I'll do that. I have no better idea.


I've inserted only changed footprints.
tc7662bx0a
tc7662bx0a

tc7662bxpa
tc7662bxpa

evanshultz:
All looks OK except the V+ pins are upside down! :)

sebastianbasierski:
Wow.
I don't know how i made to do it;)

It's fixed now.

evanshultz:
It happens!

PR #602, opened 2018-05-15
Interface/AD9910: split AVDD_1V8 by datasheet

lorem-ipsum:
the [datasheet](http://www.analog.com/media/en/technical-documentation/data-sheets/AD9910.pdf) recommends different regulators for the different analog 1.8V power supplies (see page 46 of the datasheet). Split the stacked power supply pins as suggested by the datasheet.

Sorry for not spotting while creating yesterdays PR.
ad9910

------------


evanshultz:
Good catch! What about moving pin 6 to the right so it's closer to the digital supply pins?

lorem-ipsum:
Makes sense, I moved the analog pins around, now the 1V8 are in the centre and 3V3 on the left. Pin 6 is now next to digital 1V8.

evanshultz:
Thanks! Can you post a new screenshot for historical reasons?

lorem-ipsum:
screenshot_20180517_231608

evanshultz:
Thank you!

PR #535, opened 2018-04-25
Add Amphenol RJMG1BD3B8K1ANR Magjack.

ppelleti:
Symbol for [Amphenol RJMG1BD3B8K1ANR Magjack](https://www.amphenolcanada.com/ProductSearch/Drawings/AC/RJMG1BD3B8K1ANR.PDF).

Based on existing symbol `RJ45-TRANSFO`.  (Which seems to have disappeared now?)

rj45_amphenol_rjmg1bd3b8k1anr

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #603, opened 2018-05-15
add Würth 7499010121A RJ45 magjack

lorem-ipsum:
based on the existing dual component Wuerth 749151120. Drawings are rather
excessive. [datasheet](http://katalog.we-online.de/pbs/datasheet/7499010121A.pdf)

PR for the footprint will follow. I've seen there is a similar PR for a component from Amphenol (#535). The style is fairly different. As far as I have seen there are only two major types of magjacks. For 100 Base-T (this PR and the Ampenol one) and for 1000 Base-T (the existing dual magjack). A common style for those might be a good idea. Sadly pin numbering varies a lot, so a single generic component will not work.

wuerth

------------


evanshultz:
Looking at this, I only see one issue: Nobody can tell which LED is which. I drew the existing symbol and didn't label those LEDs and that was my fault. If you would like me to merge just let me know and I'll be happy to do that and then add the text for LED colors myself later. I will wait on the footprint before merging this.

lorem-ipsum:
I did add the LED colours to the symbol. The footprint adjustments might take some time to complete as they require some adjustments to my scripting. Maybe I do this one manually

evanshultz:
Thanks! This looks good to me. Once the footprint is merged this is ready to go.

poeschlr:
A common style is always a good idea. I merged the amphenol one as is for now, it can be changed in a separate PR to get it to a unified state if somebody is interested in doing that. (It is good enough to use right now and its footprint has been merged.)

antoniovazquezblanco:
FP has been merged. Should we proceed or is there anything left?

evanshultz:
Yes, the footprint referenced by this symbol is missing the `_Horizontal` suffix. With that added to the default footprint I think this can be merged.

evanshultz:
Yep. This is what I was thinking too, and it should be fine. Thanks!

PR #604, opened 2018-05-16
add line transformer Macom ETC1-1-13

hvraven:
This add the line transformer ETC-1-13 ([datasheet](https://cdn.macom.com/datasheets/ETC1-1-13.pdf)) to the transformer lib. [Footprint PR](https://github.com/KiCad/kicad-footprints/pull/585).
I've decreased the size of the pins to 80 mils to make them fit on the grid. The alternative would be pins with 100mil and connection lines, but in my opinion this makes the symbol to wide.
etc1-1-13

------------


Ratfink:
Thanks for the contribution!

Notes:

* I think the pin length you have is okay for this symbol.
* Please add lines for the core, as in the datasheet.
* It seems like pin 4 should be included in the symbol.  The datasheet recommends grounding it, which can't be done in Eeschema if it's not present in the symbol.  It should be electrical type Passive.  As for where it should be on the symbol, I don't have a strong suggestion.  Maybe including it as a second unit would be an option.  @poeschlr @jkriege2 @evanshultz Any ideas?

evanshultz:
Pin 4 on a second unit seems OK, since there's no guidance if it's a shield it's probably best not to have it connect to any other part of the schematic. It's too bad KiCad can't check if all units are placed, though, which is my only real concern with that. Having it floating below the symbol would work too.

I've never seen something as ridiculous as having two dots on the same winding...

lorem-ipsum:
Finally came around to work on this again. I rebased against the current master and added the pin 4 floating below the transformer. I tried to place it on a different part, but found no satisfying solution to place the names and part number as they seem to be linked between the two parts. Is there a way around it?
screenshot_20181101_140023

poeschlr:
@lorem-ipsum I am not sure what your question means. Could you maybe try to rephrase it?

myfreescalewebpage:
@poeschlr he means the reference and name are at the same coordinates for all units.

@lorem-ipsum do not worry about that, reference and name are auto-placed on schematic unite per unit. Just choose one position which is suitable to not override the symbol in the editor.

Joel

myfreescalewebpage:
This PR remind me this: https://github.com/KiCad/kicad-symbols/pull/1338 (basically the same without the shield pin) in which we decided that the symbol should be generic.

I wonder if the symbol should be generic too.... Shielding is not unusual on this kind of device.... This devcie could then be named "EMI_Filter_LL_1523_4" with the same logic.

@poeschlr and @evanshultz maybe your position ?

Joel

myfreescalewebpage:
@KiCad/librarians what is your opinion about my previous comments on this PR ?

antoniovazquezblanco:
I would not leave pin 4 floating (I would probably add the Mounting or Shield keywords).

Given footprint PAD numbering I would not go for a generic symbol. Little to no other devices will share ping numbering I think.

Just my two cents. :)

myfreescalewebpage:
@antoniovazquezblanco for transformers, LED, etc, we have already several generic symbol with arrangement of the pin name in the symbol name. This is good to avoid multiplication of similar symbol. However, adding atomic part is also very nice, maybe the atomic names can the alias of the generic symbol for example ?

antoniovazquezblanco:
> Maybe the atomic names can the alias of the generic symbol for example ?

That is usually a little bit difficult due to shared footprint infor with the original symbol.

myfreescalewebpage:
@lorem-ipsum I sugegstt he design should now be rebased on new "L_Core_Ferrite_Coupled_xxxx" symbols. Pin 4: I suggest the following placement:

Capture

myfreescalewebpage:
A samll vertical line with "Mounting" is not bad too has done here: https://github.com/KiCad/kicad-symbols/pull/899

myfreescalewebpage:
@lorem-ipsum what do you think of my last comment ?

hvraven:
Hi @myfreescalewebpage ,

sorry for not getting around to fix this. Now rebuilt it on top of `L_Core_Ferrite_Coupled` and added the mounting pin as suggested.

Cheers,
Hendrik

myfreescalewebpage:
Thanks for the update, this is great. Merging.

aris-kimi:
Hi @myfreescalewebpage ,
This is odd. Symbol design is misleading. I have never saw anything like this basicaly. I had sent an email (webmaster@macom.com) to the company, in case they don't respond i will ask for somebody else to reach them by owned domain mail if possible.
I will update here.

hvraven:
Hi @aris-kimi ,

yeah the symbol is weird. If you can get some better information from Macom that would be great.

PR #605, opened 2018-05-17
add Mini-Circuits LFCN/BFCN/HFCN passive 1206 filters

lorem-ipsum:
this series uses two different package styles, but the same pins for all
packages. I've created aliases anyways (94 dfferent variants in total) as
this is not the case for all series from Mini-Circuits (e.g. for the band
pass filters). This allows to be consistent throughout all filters.
[Example datasheet](https://www.minicircuits.com/pdfs/LFCN-80+.pdf) and [footprint PR](https://github.com/KiCad/kicad-footprints/pull/589)
lfcn

------------


evanshultz:
Can you please post screenshots of all symbols?

lorem-ipsum:
Sorry for the delay, here are the other two variants:
bfcn-1445
hfcn-672

antoniovazquezblanco:
Ping @evanshultz :)

evanshultz:
The footprints PR is now https://github.com/KiCad/kicad-footprints/pull/882. Hopefully they will be updated by next week and then I can review.

The symbols look correct and the BP one matches that shown in IEEE 315-1975 section 15.16.2. 19.2.3 would allow a box with only FL-BP, FL-HP or FL-LP text inside the box. I suppose the BP symbol above, and the LP and HP variations, are OK if nobody else objects before the footprints are merged.

My head is spinning after looking at so many parts and their specs, so sorry if I got confused:
- I see all part numbers suffixed with a `+`. Did there used to be non-`+` versions? What does the `+` denote? I think it's fine as-is unless you feel adding a `+` is important.
- The ones using the FV1206 footprint with no suffix will match the others in CvPcb. As a default footprint is selected I guess we leave it since removing the trailing asterisk feels wrong.
- BFCN-152W-75 mentions the impedance in the description but the others don't. Why?
- BFCN-152W-75 description is missing the -7 package suffix.
- LFCN- parts don't match the F2 freq shown at https://www.minicircuits.com/WebStore/RF-Filters.html. Can you please check if they are in error or if I'm looking at the wrong info?
- I'm not sure pin types really matter here. Maybe we should go with Passive since these parts contain only passive components?
- I'm thinking using only asterisks in the footprint filter instead of any question marks doesn't hurt anything and gives us the greatest flexibility. Question marks are only used in some very specific situations where we know they're required as far as I'm aware.
- Can the ref des fit on top? Maybe center the ref des and value with ref des on top? Or maybe there's enough room to do ref des on the left and value on the right?
- The ref des should start with `FL`.

lorem-ipsum:
Thanks for looking at this. Most likely I will do the changes tomorrow.

>     * I see all part numbers suffixed with a `+`. Did there used to be non-`+` versions? What does the `+` denote? I think it's fine as-is unless you feel adding a `+` is important.
`+` is the ROHS compatible specification from Mini-Circuits. It has no other meaning, so there is no benefit in adding this.
>     * The ones using the FV1206 footprint with no suffix will match the others in CvPcb. As a default footprint is selected I guess we leave it since removing the trailing asterisk feels wrong.
>     * BFCN-152W-75 mentions the impedance in the description but the others don't. Why?
All others are 50 Ohm, the other is the exception. It should probably be added to all of them, I will correct this.
>     * BFCN-152W-75 description is missing the -7 package suffix.
Good catch, will fix it.
>     * LFCN- parts don't match the F2 freq shown at https://www.minicircuits.com/WebStore/RF-Filters.html. Can you please check if they are in error or if I'm looking at the wrong info?
They are using the F1 frequency which is the frequency Mini-Circuits uses in their description.
>     * I'm not sure pin types really matter here. Maybe we should go with Passive since these parts contain only passive components?
Makes sense, will change this. 
>     * I'm thinking using only asterisks in the footprint filter instead of any question marks doesn't hurt anything and gives us the greatest flexibility. Question marks are only used in some very specific situations where we know they're required as far as I'm aware.
Ok. I will change it.
>     * Can the ref des fit on top? Maybe center the ref des and value with ref des on top? Or maybe there's enough room to do ref des on the left and value on the right?
I prefer to have them aligned with the rectangular outline, not sticking out on the side. As for some the name is rather long putting them next to each other does not fit for all of them. Stacking them on top of each other will work. If thats preferred to the current solution I can change it to this. 
>     * The ref des should start with `FL`.
Ok.

evanshultz:
Thanks for getting back to this and sorry for my delay!

Interesting about the LP filters not using the F2 freq. The HP filters use the F2 (3dB down) freq and the BP filters are also using the 3dB down freqs. Why are LP filters different?  I'll check all the LP filters tomorrow then.

Stacked ref des and value sounds good.

lorem-ipsum:
I've just realised that they use the D suffix for filter which can accept an applied DC voltage, while those without must the capacatively coupled if DC voltage is present. I would add this to the description to distinguish them, but are currently lacking a proper description for that. Would "DC stable" match?

lorem-ipsum:
I integrated the changes. I mostly used search and replace inside of the .lib and .dcm file, but made sure not to break anything. This also changed the RefDef on two other filters from `F` to `FL`.

evanshultz:
Thanks!!

One question I need to ask about the descriptions again. HFCN-1000 (https://www.minicircuits.com/pdfs/HFCN-1000+.pdf) only shows a usable range from DC-100MHz. HFCN-1000 (https://www.minicircuits.com/pdfs/HFCN-1000+.pdf) similarly shows 1080-4000MHz. So why are the HFCN parts using the 3dB down frequency (F2) in their description but the LFCN aren't (they use the top end of F1)? The BP filters are also using the 3dB down points. I would have expected that the frequency in the KiCad description would be consistent. Why are the LFCN different? Sorry that I'm just not seeing it.

Comparing https://www.minicircuits.com/pdfs/LFCN-320D+.pdf and https://www.minicircuits.com/pdfs/LFCN-320+.pdf I suppose "DC capable" would be an OK term to add to the description. Does that sound OK to you?

You mention `B39162B8813P810` and `SF14-1575F5UUA1` were also changed to have a FL ref des prefix. It's not my field of expertise but from https://global.kyocera.com/prdct/electro/product/pdf/sd18_un.pdf I'm not sure if this is a filter. Can you help me out?

And some comments on the latest commits:
- BFCN-5100 missed getting the OUT pin set to Passive.
- Should all GND pins be Passive type too since they also aren't a current sink but instead are also only passive component terminals? I'm imagining a board which only includes these filter parts or other passive components and no power supply, which I have seen. In that case, the user would need the PWR_FLAG symbol to avoid an ERC error but if the GND pins are Passive type ERC will be fine. Either way would be OK for me, with a slight preference for Passive, so what do you think?

lorem-ipsum:
After looking through some more of the datasheets I think I finally understood your issue with the stated frequency ranges. Looking at it I see two options. I would prefer to leave the low pass and band pass as they are and change the high pass to the F1 value (also adding the upper frequency). This is the range the manufacturer specifies for their products. However this is not consistent on the dB suppression point. The other solution would be to use the 3dB suppression point for all of them, changing the low pass filters.

I've corrected the BFCN-5100, attached the "DC capable" to the descriptions and set the output pins to passive. I was not doing it at first as the library check complains about it, but your argument makes sense.

I am not really sure on the `SF14-1575F5UUA1`. It looks like some dual band pass filter, with different band passes between the Antenna and RX and Antenna and TX. But I am not sure whether this makes it a filter part. At least Digi-Key is listing it as SAW filter.

lorem-ipsum:
Should these actually stay in `Filter`, or should I open up a `RF_Filter` library?

evanshultz:
Sadly, there's a typo of the word "capable". The "p" and "b" are swapped. Can you fix it, please?

I prefer the 3dB point, changing the LP filter descriptions. That would be consistent, as you said, and for me (an audio guy) 3dB down is the common rating for filters. If you know of another common rating method for this type of part then please let me know, but if you're OK with using 3dB for all let's do it.

These symbols must stay in this library for the 5.0.x KiCad cycle. Perhaps we will review this and move them later. It's a good idea to consider.

`SF14-1575F5UUA1` is fine then. Thanks for looking at it.

lorem-ipsum:
Then hopefully everything is set now. I've got the excel table of all parts and took the 3dB point from there. Also fixed the typo.

evanshultz:
Looks good to me. When Rene finishes the footprints and they're merged this can be merged too.

Thanks!!

aewallin:
> Should these actually stay in `Filter`, or should I open up a `RF_Filter` library?

+1 for the RF_Filter library.
I submitted a few PRs with mini-circ passive filters in other form factors, e.g.
#1225 , #1226 (attenuator), #1223 , #1219 , #1215 , #1212 

All of these could use the same visual symbols, for consistency. 
variants cannot use different footprints, I guess? so they have to be separate symbols if we want atomic parts with footprint assigned.

evanshultz:
You are correct that an ALIAS cannot have a unique footprints, so if you want the footprint specified you will have to split the symbols.

aewallin:
OK, the right way to deal with the large number of passive RF-filters would be some kind of script-driven symbol-generator that generates (at least) three types of symbols (such as the ones above):
- low pass
- high pass
- band pass
The script could then read from a CSV, or similar, database the correct pin-numbers, datasheets, and footprints.
That way there would be a consistent visual symbol for all passive RF filters.
Does that workflow and scripting already exist?

evanshultz:
The footprints are now merged and this looks good. However, there are a couple things to wrap up before this gets merged.

1. I assume the visual style is acceptable to all. If not, please chime in in case any changes are desired.

2. Do you want to place these in a RF_Filter lib? Keep in mind other symbols in this lib may be RF parts as well. If so, please update the sym-lib-table. Note that this will mean the table needs to be updated by the user until v5.1 is released to see these symbols.

Regarding the scripting, we've been using YAML as the format for the external data file for footprints. I don't know what syngen uses, though.

lorem-ipsum:
These there created by hand with a bit of manually scripting in form of a proper editor and shell scripting using `awk`. Meaning I created the version requiring a own symbol and then added the aliases manually in the `.lib` and `.dcm` files. Scripting help for those would be nice though. Something simple which would track all the aliases and there differences would already be a great help.

I will move them to the RF_Filter lib, as there a plenty of those they deserve it :-)

lorem-ipsum:
I've moved the filter to the new library (rebasing on top of the current master, hence the force-push). The existing SAW filters could be moved to the RF_Filter as well, leaving the filter lib empty for the time being.

evanshultz:
@KiCad/librarians 
I'm going to merge this PR which creates a new lib named `RF_Filter` unless somebody speaks up in the next day.

myfreescalewebpage:
Ok for me, do you also plan to move all other filters created recently ?

evanshultz:
Not unless they were added since 5.0 was released. Newly-added symbols are OK to go into newly-created libraries without any chance of breaking existing schematics.

PR #608, opened 2018-05-21
Added Texas Instruments CDCVF2505 PLL clock driver

endofexclusive:
Datasheet:
http://www.ti.com/lit/ds/symlink/cdcvf2505.pdf

NOTE: CLKOUT and 1Y* are 3-state outputs according to the datasheet functional block diagram.

cdcvf


evanshultz:

You will find a lost of rules and guidelines for the KiCad library at http://kicad-pcb.org/libraries/klc/. Please see that webpage for more information about the review items below.

This is an excellent symbol! Great first contribution!
- Can you change the description to `Clock Phase-Lock Loop Clock Driver, 24-200MHz, 3.3V, SOIC-8/TSSOP-8`.
- Please move the ref des to the top left and part type to the top right.

endofexclusive:
Sure. The position policy for refdes and name was not fully clear from the KLC illustrations. Here is a second try. :-)

endofexclusive:
up

evanshultz:
Thanks! That text position is one of many things that KLC isn't clear about, but we strive to address issues as they come up.

I hope to see more contributions from you in the future!

By the way, your color scheme above is quite a departure from the default KiCad look.

PR #609, opened 2018-05-21
Add TPS3702x

adamheinrich:
Overvoltage and undervoltage monitor IC by TI, SOT-23-6.

Datasheet: http://www.ti.com/lit/ds/symlink/tps3702.pdf

tps3702


evanshultz:
This is a very nice symbol! Thank you for taking all the feedback in the past into account and continuing to contribute!
- Since there's only one package and no suffix to worry about, please remove the trailing `x` in the symbol name.
- The footprint filter should be `SOT?23*` to ensure only SOT 23s are caught and allow CvPcb's pin count filter to work.

BTW, if you're using KiCad v5 you can see the pin types directly within the pin editor instead of showing the pin table.

evanshultz:
Thanks!

In the future would you please post incremental commits? That makes it easier for the librarians to see what has changed.

PR #612, opened 2018-05-23
Mechanical: Rename mounting holes to be consistent and update FPFilters

antoniovazquezblanco:

------------

### DIFF
--- a/Mechanical.dcm
+++ b/Mechanical.dcm
@@ -5,7 +5,7 @@ D Heatsink
 K thermal heat temperature
 $ENDCMP
 #
-$CMP Heatsink_PAD
+$CMP Heatsink_Pad
 D Heatsink with electrical connection
 K thermal heat temperature
 $ENDCMP
@@ -15,7 +15,7 @@ D Housing
 K housing
 $ENDCMP
 #
-$CMP Housing_PAD
+$CMP Housing_Pad
 D Housing with connection pin
 K housing
 $ENDCMP
--- a/Mechanical.dcm
+++ b/Mechanical.dcm
@@ -23,11 +23,13 @@ $ENDCMP
 $CMP MountingHole
 D Mounting Hole without connection
 K mounting hole
+F ~
 $ENDCMP
 #
 $CMP MountingHole_Pad
 D Mounting Hole with connection
 K mounting hole
+F ~
 $ENDCMP
 #
 #End Doc Library
--- a/Mechanical.dcm
+++ b/Mechanical.dcm
@@ -8,6 +8,7 @@ $ENDCMP
 $CMP Heatsink_Pad
 D Heatsink with electrical connection
 K thermal heat temperature
+F ~
 $ENDCMP
 #
 $CMP Housing
@@ -18,6 +19,7 @@ $ENDCMP
 $CMP Housing_Pad
 D Housing with connection pin
 K housing
+F ~
 $ENDCMP
 #
 $CMP MountingHole
--- a/Mechanical.dcm
+++ b/Mechanical.dcm
@@ -25,7 +25,7 @@ D Mounting Hole without connection
 K mounting hole
 $ENDCMP
 #
-$CMP MountingHole_PAD
+$CMP MountingHole_Pad
 D Mounting Hole with connection
 K mounting hole
 $ENDCMP
--- a/Mechanical.dcm
+++ b/Mechanical.dcm
@@ -20,12 +20,12 @@ D Housing with connection pin
 K housing
 $ENDCMP
 #
-$CMP Mounting_Hole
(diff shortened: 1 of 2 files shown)


evanshultz:
- To match the footprints, the suffix would be `_Pad` and not all uppercase.
- Could you also remove the other footprint filter since it won't work with the footprints at https://github.com/KiCad/kicad-footprints/tree/master/MountingHole.pretty?

I suppose the non-pad symbol is for NPTHs for PCB mounting. Kinda weird since this symbol does not appear in a netlist or BOM and there is a symbol in KiCad, but OK...

@poeschlr 
Do you want to retain these ref des prefix of `MK` for mounting holes? Seems odd. `MP` or `MT` or `W` seem better and more common to me.

antoniovazquezblanco:
@poeschlr what should we do with the reference designator prefix?

poeschlr:
Is there an official standard designator prefix for such parts? (I do not really see where MK comes from.)

My reasoning is that it is best to change it to something defined in an official standard than to change it to something that we define now just to change it to the correct thing in a few months (or even years)

antoniovazquezblanco:
IEEE 315-1975:
- MK: hydrophone, microphone, telephone transmitter
- MP*: brake, clutch, mechanical interlock, mechanical part, miscellaneous mechanical part (bearing, coupling, gear, shaft)
- MT: accelerometer, measuring transducer, mode transducer, motional pickup transducer, primary detector
- W: bus bar, cable, cable assembly (with connectors), coaxial cable, conductor, distribution line, distribution path, Goubau line, strip-type transmission line, transmission line, transmission path, waveguide, wire

antoniovazquezblanco:
From https://blogs.mentor.com/tom-hausherr/blog/tag/reference-designator/, the following image summarizes IPC-2612.

imagen

I think MH would be the way to go but I have no way of accesing IPC-2612.

evanshultz:
Some more food for thought:
- http://www.pcblibraries.com/downloads/temp/Reference_Designators_8825188.pdf
- https://www.ee.iitb.ac.in/~spilab/Tips/ansii_graphic_symbols_for_electrical_and_electronics_daigrams_1993.pdf (probably the source of the copy above that lists MK but not for mounting holes)

I also support `MH` prefix for mounting holes. Seems the most obvious and possibly in alignment with some standard.

Also, the Pad symbol could have a footprint filter of `MountingHole*Pad*` so it only catches the ones with copper. The non-Pad will catch all mounting holes but that's the case now anyway and the user may want copper even without a net connection (clearance, screw hole land, etc.) so I'm OK with that.

evanshultz:
One last thing: you can put a tilde in the datasheet field to avoid any check script errors. We've done that elsewhere and I don't know of any objections.

antoniovazquezblanco:
> http://www.pcblibraries.com/downloads/temp/Reference_Designators_8825188.pdf

404 error

> One last thing: you can put a tilde in the datasheet field to avoid any check script errors. We've done that elsewhere and I don't know of any objections.

Will do with the other changes. Thanks

antoniovazquezblanco:
Done. Any other request for changes?

evanshultz:
@poeschlr @jkriege2 @Ratfink 
I'm good to merge. Any other librarians have input? That was the only discussion raised by anyone but me above, so I'll merge with within 48 hours if there's no more comments.

poeschlr:
As far as i can see (from looking at the files within githubs web interface) this is ok for me.  If you opened the files in kicad then this is ok to merge.

antoniovazquezblanco:
@evanshultz is this a go?

Sorry for the ping and thank you very much!

antoniovazquezblanco:
What do you wan't me to do? Rename to `_Pad`?

evanshultz:
That would be great, if you don't mind. http://kicad-pcb.org/libraries/klc/G1.6/ indicates non-acronyms in symbol names should not be capital. If you don't want to, I can merge as-is and make another PR to fix that. Just asking since you've already found a similar issue and chased it down.

antoniovazquezblanco:
@evanshultz Done!

evanshultz:
That looks good.

@antoniovazquezblanco @poeschlr 
I just noticed the Housing has a ref des of MK. Does that be H or MP?

antoniovazquezblanco:
I have no idea on what refdes a housing should have. Will wait for instructions.

evanshultz:
@antoniovazquezblanco 
v5 is scheduled for tagging/release on 9 Jul. I'll merge this as-is if we don't get input before then. Feel free to ping me if you would like.

Thanks again!

evanshultz:
Nope. I'll merge as is then.

Thank you again!!

antoniovazquezblanco:
Thank you for your work ;)

PR #613, opened 2018-05-23
Added new symbol: R_Kelvin and R_Kelvin_US

jeremysiebers:
@evanshultz 

These map to the new footprint symbol R_Vishay_WSK2512_6332Metric_T2.66mm

r_kelvin

r_kelvin_us

This addition links to [https://github.com/KiCad/kicad-footprints/pull/573](url) with the following comments:

- [x] If you submit a symbol, which I would support, can you add a wiggly US resistor symbol since we're doing that now?
- [X] A new symbol is not required, but to get the benefits of this part the Kelvin connection is required. (It could be done using net ties from pins 1 and 2 but having 4 pins makes it easy and matches the component/footprint design.) It might be nice to make the pin on each side look like a "Y" on its side instead of having the Kelvin pins sticking out weirdly from the resistor body.

------------


jkriege2:
aren't these the same as the existing shunt-symbols?

Ratfink:
The pin numbering is different from R_Shunt.  Evan suggested changing the numbering in https://github.com/KiCad/kicad-footprints/pull/573#issuecomment-389693184.  I don't disagree with his reasoning, using numbering for the main pins consistent with the normal resistor symbol kinda makes sense.  Then again, we've never done that with any other special resistors.  See e.g. potentiometers, which have pins 1 and 3 as the opposite ends and pin 2 as the wiper.

I really don't like the idea of adding a second equivalent generic symbol with a different numbering.  Also, given that we're in 5.0 freeze, I don't like the idea of changing the pin numbering of an existing generic symbol in a non-compatible way.  The pin numbering we have now *works*, and we already have a bunch of footprints that match it, so I think it would make sense to not keep changing things in a non-useful way.  Maybe we can revisit the pin numbering for 6.0, but changing this at any point would just break a lot of users' stuff, which would be bad.  If we could start from scratch, I'd favor the numbering Evan suggested, but since we're not doing that I'd rather not change it.

I *would* be in favor of creating an R_Shunt_US symbol, to help round out the set of R_*_US symbols.

jeremysiebers:
@evanshultz @jkriege2 @Ratfink @poeschlr 
Gentlemen, please decide what we want to have regarding this shunt type so that I can finish this addition.

jeremysiebers:
Removed proposed shunt resistors, created R_Shunt_US symbol with same dimensions as R_Shunt:

Existing shunt symbol:

r_kelvin

New US symbol:

r_kelvin_us

evanshultz:
@jeremysiebers 
Sorry. I was on vacation.

OK. I'm fine with this. Technically a shunt resistor doesn't need/require 4 terminals while a Kelvin one does, but this is probably easiest and understandable for all users. If I don't hear anything else in the next couple days I'll merge this as it seems to address all the items mentioned above.

@Ratfink 
Other resistors, like the pot you mention, have a function different from a regular resistor. Also, a 4 terminal resistor can be used without the sense pins like any other resistor. Conversely, any regular resistor with 2 terminals can (nearly) have "4 terminal" voltage sensing added by designing the PCB to provide Kelvin connections. That's why I thought it would make sense for the 4 terminal resistor to have the same "primary" pin numbers as the 2 terminal resistor symbol, but this is fine in light of the existing symbols and footprints which are functionally the same. Your point above is well taken. :)

Ratfink:
@evanshultz I just applied a couple little fixes to this branch, would you mind taking a look and merging if everything looks good?

evanshultz:
Sure. Looks good. Thanks all!

rs-development:
I would like some to ask if there is any reason why **not** use one side of the shunt for the power connection and the other side for the sense connection. If there is not problem with that approach i think we need another symbol.

Using Pin 2&4 (or 1&3) for the power connection (green arrows) and 1&3 (2&4) for the sense connection (blue arrows). 

Screenshot at 10-57-42

PR #615, opened 2018-05-23
Added new component: LEM_LTSR series

jeremysiebers:
LEM_LTSR series (6/15/25)

https://www.lem.com/sites/default/files/products_datasheets/ltsr_6-np.pdf
https://www.lem.com/sites/default/files/products_datasheets/ltsr_15-np.pdf
https://www.lem.com/sites/default/files/products_datasheets/ltsr_25-np.pdf

lem_ltsr-np

Footprint under construction

------------


jeremysiebers:
Hello Evan,

Yes there is already a footprint pull request: [https://github.com/KiCad/kicad-footprints/pull/600](url)

- [X] Please remove the generic symbol since you have the specific ones.
- [X] All symbols share the same generic description and include the part number. (Yes, some existing symbols have a poor description.) Please use something like Current Transducer, Unipolar, 6A, 5V supply
- [X] The vendor name should not be in the symbol name, so remove the LEM_ part. --> also will do this for the footprint and adjust the footprint filter to: LTSR*NP*
- [X] The IN* and OUT* pins should be Passive type. Ref pin should be Input type for lack of anything better.
- [X] Let's use the datasheet names for the "output" side pins of +5V, 0V, Out, and Ref. (They do use various names but those 4 are both concise and consistent.)
- [X] Add a default footprint. I don't see one in the footprint library already, and we don't like to merge symbols missing footprints, so are you willing to submit the footprint?--> I already did see: [https://github.com/KiCad/kicad-footprints/pull/600](url)
- [X] How did you determine pin numbers? This will relate to the item above. --> Pin numbers are counted from the 1 to 6 numbers and then further from 7 till 10 like any other IC footprint counter clockwise. The only difference is that at LEM they started counting in the bottom-view of the symbol i.s.o. counting like any other IC footprint from top-view. So I kept it at their counting order and finished it for the remaining pins (7 to 10).

lem_ltsr-np

image

image

image

image

pin numbering

ltsr_footprint

evanshultz:
- Your description and keywords are flipped. Also, do not put the MPN in the description. Description should be `LEM Current Transducer, Unipolar, 6A, +5V supply` and keywords can be just `current transducer`.
- Since underscores are used to separate sections of the symbol name, let's avoid using it within the part number and change `LTSR_6-NP` to `LTSR6-NP`.
- The footprint should have the vendor name in it. Use `Sensor_Current:LEM_LTSR-NP` with `LEM*LTSR*NP` for the footprint filter.
- After seeing it again, I think the Ref pin should be Passive type. It may have a voltage on it, could be a buffer, could be a resistor divider, could be ground... it's not a logic input. What do you think?
- Another question for you: having OUT* pins and an OUT pin seems a bit confusing. What about naming the "output" pins in mixed case like `Ref` and `Out`? Would that be more clear?
- I find some other LEM current sensors in the lib already use a vertical layout for the high current path while yours in horizontal. I'd like to have a similar style for all LEM symbols where reasonable. Would you take a look at these symbols and see what you think? Could they all be more similar?
image
- Thanks for clarifying the pin numbers. It's not obvious from this datasheet, although it is from many of their others.

jeremysiebers:
Hello Evan,

- [X] Your description and keywords are flipped. Also, do not put the MPN in the description. Description should be LEM Current Transducer, Unipolar, 6A, +5V supply and keywords can be just current transducer.
- [X] Since underscores are used to separate sections of the symbol name, let's avoid using it within the part number and change LTSR_6-NP to LTSR6-NP.
- [X] The footprint should have the vendor name in it. Use Sensor_Current:LEM_LTSR-NP with LEM*LTSR*NP for the footprint filter.
- [X] After seeing it again, I think the Ref pin should be Passive type. It may have a voltage on it, could be a buffer, could be a resistor divider, could be ground... it's not a logic input. What do you think?--> I agree
- [X] Another question for you: having OUT* pins and an OUT pin seems a bit confusing. What about naming the "output" pins in mixed case like Ref and Out? Would that be more clear? --> I agree, if this is allowed it would make it more clear. I will change this in the symbol.
- [X] I find some other LEM current sensors in the lib already use a vertical layout for the high current path while yours in horizontal. I'd like to have a similar style for all LEM symbols where reasonable. Would you take a look at these symbols and see what you think? Could they all be more similar?

The symbol looks now like this:

lem_ltsr-np

image

The 15A and 25A also have changed description/keywords section.

image

image


image

evanshultz:
Thanks @jeremysiebers and @poeschlr !

PR #618, opened 2018-05-25
Add TI TPD3S0x4 USB protection ICs

whitequark:
Add TPD3S0x4 Current Limit Switch and D+/D– ESD Protection for USB Host Ports.

Screenshot:
screenshot_20180525_041052
Datasheet: http://www.ti.com/lit/ds/symlink/tpd3s014.pdf
Datasheet excerpt: 
screenshot_20180525_041113

------------


evanshultz:
- Remove part number from description
- Change package in description from `SOT-23` to `SOT-23-6`
- I'm thinking pin 4 should not be Power Output type, since this part doesn't source power.
- What about putting the D1/D2 pins on a separate unit? Would that make using this part easier? Since the USB could be facing left or right that's why I wonder if this symbol should be more flexible. Perhaps you could show this connected to a USB connector so we can see the schematic wiring?

whitequark:
> Remove part number from description
> Change package in description from SOT-23 to SOT-23-6

Done

> I'm thinking pin 4 should not be Power Output type, since this part doesn't source power.

Since on the net connected to pin 4 there will be no other Power Output pin, this will result in a ERC error. I don't think this is desirable.

> What about putting the D1/D2 pins on a separate unit? Would that make using this part easier?

Would it? Maybe if it could somehow be made "passthrough" so it can be dropped over the existing D+/D- wires, but I think adding D1/D2 pins on both sides of an unit violates KLC.

> Since the USB could be facing left or right that's why I wonder if this symbol should be more flexible. Perhaps you could show this connected to a USB connector so we can see the schematic wiring?

From a real schematic:

screenshot_20180606_003049

evanshultz:
OK. I'm fine with that D1/D2 placement. Just wondering. Thanks for sharing the schematic.

OUT should not be Power Output type. You will need to add a Power Flag on the net to avoid an ERC violation just like any other net and sources power but doesn't come directly from an IC pin (like a buck converter that has a series inductor between the IC and the supply's output).

Misca1234:
> OUT should not be Power Output type. You will need to add a Power Flag on the 
> net to avoid an ERC violation

You thinking of that he is in his schematic connect pin 4 to a "+5V power symbol"?

evanshultz:
The descriptions look great now. Thanks!

If the VBUS pin of the USB connector is sourcing power directly, it will need a PWR_FLAG since the USB jack pin is Passive type. Pin 4 of U15 is not sourcing power (it's a pass through) so, similarly, another PWR_FLAG will be needed there for all downstream circuitry that sources power. The power pins of all connected devices will be Power Input type so it works.

Here is my test schematic that gives the following error as DG412 is a similar switch with Passive pins on each side of the switch as I'm proposing (see green arrow by U13.2):
image
```
ErrType(3): Pin connected to some others pins but no pin to drive it
    @ (260.35 mm,105.41 mm): Pin 2 (Passive) of component U13 is not driven (Net 69).
```

But when I add the PWR_FLAG, which indicates a net that can source power even without a Power Output pin on it, you see no green ERC mark on the net:
image

Note that this is with the default ERC rules.

whitequark:
I was not aware of this mechanism. Updated, thanks.

evanshultz:
You're welcome. :) The OUT pin should be Passive type, like the DG412 in my screenshot above. It doesn't sink current.

whitequark:
Sorry. Fixed.

evanshultz:
Thanks for another nice symbol!

Issue #164, opened 2018-01-12
Separate former linear library physically and logically


evanshultz:
I reworked the whole lib but separated symbols logically, so some footprints were not available for all ALIASes of a symbol. Each symbol should be unique both logically and also physically.

poeschlr:
This issue does confuse me. I really don't know what you mean by physically.

From another comment of yours (can't find it now) i would guess you want to make these parts more atomic. (Meaning not only separate the symbols from others but also the footprints)

myfreescalewebpage:
@evanshultz is this issue still valid today ?

evanshultz:
Yes

myfreescalewebpage:
Ok, but as indicated by Rene, can you explain more the purpose here, I'm also a little bit lost... Maybe with an example ?

evanshultz:
Maybe it's just the wording.

Logical = the symbol that contains the pins and their functions (which drives the netlist)
Physical = the tangible package that a part is delivered in; there can be many

Real parts have one logical representation but may come in many physical versions. For example, ON Semi sells LM393 in 3 physical variants: https://www.onsemi.com/pub/Collateral/LM393-D.PDF.

In the libraries above, they are mixed which can lead to user problems. I can't recall exactly where a problem was reported, but somebody (I think on the forum) had noted this after I cleaned up the old opamp library. To illustrate this, let's look at https://github.com/KiCad/kicad-symbols/blob/master/Amplifier_Operational.lib and investigate two symbols which don't have a default footprint specified so the package is chosen in CvPcb:
1. `ADA4807-1` has two footprint filters which are both valid packages for this symbol. See the last page of https://www.analog.com/media/en/technical-documentation/data-sheets/ADA4807-1_4807-2_4807-4.pdf. So even though there is no default footprint, the user should only be able to choose a footprint that the part can be bought in. We haven't specified the full MPN, but the user can't pick a logical and physical combination that doesn't exist.
2. `LM2902` is the first symbol that can truly be used generically. This logical symbol applies to many, many MPNs. The datasheet (http://www.ti.com/lit/ds/symlink/lm324-n.pdf) shows three packages available but there are more than those three footprints in the FP filter. Similar problems exist as you review all the ALIASes. It looks like as more ALIASes were added, footprints that applied to the new ALIAS were added but those footprints are _not_ available for all combinations KiCad allows a user to pick.

[...]

PR #620, opened 2018-05-25
Move opamps from Timer to Amplifier_Operational

evanshultz:
For some obscene reason (probably an innocent mistake) one opamp symbol was in `Timer`.

I also updated the description and fixed the footprint filter for the MN310x symbols.

------------


evanshultz:
@jkriege2 @Ratfink 
Would you be able to tackle this? Should be quick and fixes a symbol organization problem we don't need to hear popping up during the KiCad v5 release cycle.

Note to self: Once this and https://github.com/KiCad/kicad-symbols/pull/513 are merged, begin work on https://github.com/KiCad/kicad-symbols/issues/164. Hopefully that can be done before v5 as well since we did have some users ask about it already.

PR #622, opened 2018-05-26
Add Nuvoton W83627DHG-P(T)

docbrown:
This adds the [Nuvoton W83627DHG-P(T) Super I/O Controller](http://www.nuvoton.com/resource-files/DA00-WW83627DHG-P.pdf) commonly found on PC motherboards.

w83627dhg-p

Sample image is from an SVG export because my screen isn't big enough for a legible screenshot.


evanshultz:

You will find a lot of rules and guidelines for the KiCad library at http://kicad-pcb.org/libraries/klc/. Please see that webpage for more information about the review items below

- Remove the vendor name from the description.
- Please put `QFP-128` at the end of the description to show the package.
- We don't include temp variants so please remove the PT version.
- VREF can be Power Output type. Sometimes the check script isn't quite correct and in this case it's OK to merge with a warning/error.
- Please add pin 109 even though it's NC. Make the pin No Connect type and invisible. Place the end on the symbol body so it can't easily be connected. This is so the pin count filter in CvPcb works.
- All pins should have style Line.
- There is a gap between pins 49 and 85 but those pins are in the same group in the datasheet.
- Using your same paradigm for all other pins, pin 83 should be `GP42/IRTX/SOUTB/FAN_SET2`.
- I wonder if pins 19, 2, 118, and 58 should be with the other HW monitor pins on the left side?
- Stack the VSS pins.
- Similarly, if the three 3VCC pins must be connected together then stack them too.
- Pin 105 does not give the full name from the datasheet. Is this OK?
- Is it intentional to put the GP* names to the inside of the symbol on both sides?
- It might be nice to use some lines to show the pin groupings. For example, the top left 11 pins could have a box around it and text inside the box of "LPC" or "LPC Interface". What do you think?

Please note that in the future if notable visual changes are made it's helpful to include a fresh screenshot.

antoniovazquezblanco:
@docbrown could you please address the comments avobe?

docbrown:
Sorry for the delay. I started a new job and haven't been able to work on the project that uses this component. I've addressed at least the first 4 suggestions, because I knew how to fix them in GitHub's editor, but I intend to work on the others when I have a chance to set up KiCad again.

> * There is a gap between pins 49 and 85 but those pins are in the same group in the datasheet.
> * Using your same paradigm for all other pins, pin 83 should be GP42/IRTX/SOUTB/FAN_SET2.
> * I wonder if pins 19, 2, 118, and 58 should be with the other HW monitor pins on the left side?
> * Is it intentional to put the GP* names to the inside of the symbol on both sides?

I'll admit I had a lot of trouble coming up with good groupings for this symbol. I ended up grouping pins by the peripherals needed for my project (serial & parallel ports, SPI flash bridge, and PS/2 controller) and placing the grouped peripheral's pin name on whichever side the pin is on, but I realize this is entirely subjective since pins can be assigned in different ways. I'm open to alternative grouping suggestions.

evanshultz:
@docbrown 
Thanks for jumping back in. Will you intend to set up KiCad and complete this PR shortly or should we consider it abandoned (at least for now) and close this PR?

The first 3 bullets above look fixed nicely. Thanks!

Good groupings are difficult to be sure. I don't have a strong preference other than what I wrote above so do what works best. You could consider using multiple units to break things up.

poeschlr:
@docbrown something very important. Would you accept the license agreement? (Meaning do you accept the license under which the kicad library is published. This is absolutely necessary. We can not even fix remaining problems of this PR for our self if you do not accept the license.)

evanshultz:
@docbrown 
Are you going to finish off this PR?

PR #627, opened 2018-05-26
Cleanup and fix diff amp lib

evanshultz:
As mentioned at https://github.com/KiCad/kicad-symbols/pull/620, there is still a nice chunk of work to be done to clean up the opamp (and opamp-like) symbols. Since the diff amp lib only had a couple parts in it I decided to start there and solicit approval.

Mentioned at https://github.com/KiCad/kicad-symbols/pull/578#issuecomment-390337461 is that all pins should be labelled (I believe that is the request). I had proposed using smaller per-pin text at https://github.com/KiCad/kicad-library/pull/1611#issuecomment-332682896 and a screenshot was shown next. It reads to me like this was accepted by the old head librarian and some solution is requested by the current one (@poeschlr ).

After playing around it seems like 20mil is the smallest pin text size we should go with. Otherwise it's so tiny you have to zoom in mightily and there's no way it can print on a reasonably-sized sheet of paper. And with these small symbols 10mil pin name offset seems to work best. For LM733 it doesn't leave room for the leading "G" in the gain pins but at least this indicates the pin function. I also included AD628, another symbol I need, which shows how this can work nicely if there are fewer pins.

I also fixed a couple footprint filters that were bad or ambiguous (`TO*5*` could pick up TO-52* or TO-251* or TO-247-5* so at least it's better now).

So is this OK? Other thoughts or proposed ways to label pins on opamp symbols? If this is good, I'll clean up the instrumentation amp and current amp libs in a similar way next.

Image:
image

------------


evanshultz:
@poeschlr @jkriege2 @Ratfink 
Can you please take a look? I believe like this was discussed and accepted but not implemented, so I went ahead with a PR.

evanshultz:
Can I get some feedback on this? If this is accepted, I can move on to eliminating unnamed pins and then sorting out opamps by package. If this isn't done for v5 I feel sure we'll hear about it from users.

Misca1234:
I can not see any issues to merge this

evanshultz:
@poeschlr 
Is this the style you're looking for to avoid pins with no name as you requested in a link above?

evanshultz:
@poeschlr 
Will you be able to take a look at this soon? I think we can head off a lot of user questions and/or frustration if this is done for v5, and also I'd like to work more on this lib as mentioned above.

poeschlr:
I did not notice this only affects two symbols. In that case it is good to go. Thanks

evanshultz:
Thanks! I'll move on to the bigger opamp/comparator libraries and do the same for unnamed pins then. This approved use of smaller text should definitely be added to KLC.

PR #628, opened 2018-05-26
Fix footprint associations in Transistor_BJT

Ratfink:
Another fix for #520.  This fixes (I believe) all the broken footprint associations in Transistor_BJT.lib.

MJE13003's datasheet says it has a TO-225 package, but this is a synonym for TO-126 according to http://www.onsemi.com/pub/Collateral/AN1040-D.PDF#page=23, so I changed it to TO-126 because that's the name we use in our library.

The pin numbering of MAT02 was badly broken, so I fixed that as well.

------------


Ratfink:
I suppose it could be safely removed, but looking at KLC I'm not even sure that there should be a warning given in a case like this.  I'll plan to remove the extra filter in the morning (UTC-4) unless someone else has something to say about it by then.

PR #629, opened 2018-05-26
Add ESP-07 symbol

Ratfink:
802.11 b/g/n Wi-Fi Module, similar to ESP-12E.

Datasheet:
http://wiki.ai-thinker.com/_media/esp8266/docs/a007ps01a2_esp-07_product_specification_v1.2.pdf

This is heavily based on the ESP-12E symbol already in the library, but with some changes that I think make it better: the body is smaller and the pins are shorter, plus the text fields are arranged in the normal way.  If these changes are approved, I'll make another PR to make the same changes to the ESP-12E symbol already in the library.

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/609

screenshot from 2018-05-26 14-18-03

------------


evanshultz:
Here are some thoughts, mostly just consider and not firm requests
- Should pin 1 have an overbar?
- What about pin 2 on the left since it's an input?
- And then several things for the IO pins from page 10 of the [user manual](http://wiki.ai-thinker.com/_media/esp8266/esp8266_series_modules_user_manual_v1.1.pdf):
-- Maybe all IO on the right in increasing order, top to bottom?
-- I assume you grouped by function but only pins 15 and 16 are seemingly dedicated to one function instead of primarily just GPIO. So maybe drop the other names and just use the GPIO number?
-- The datasheet shows "GPIO" in a couple places but many more instances of just "IO".
- The user manual I linked above seems like a better doc than the product spec.

Ratfink:
* Yes pin 1 should have an overbar, good catch.
* I just followed the pin layout of the ESP-12E symbol we already have, with the pins not present on the otherwise drop-in replacement ESP-07 removed.  If I rearrange the pins on this symbol, I'd want to rearrange that symbol too, which I'll do in a later pull request.
* I'll change the documentation link to the user manual.

evanshultz:
That looks good. I think rearranging the pins is the best option to design a schematic using this symbol. And while you're at it, perhaps put the GPIOxx on the right side so that it's closest to the pin number?

If you want to optionally adjust the existing symbol I'll look at it too.

Ratfink:
Okay, I've rearranged and renamed the pins of both symbols, so now only GPIO1 and GPIO3 have alternate names (which are listed first, following the documentation).  If you'd rather I undo the changes to ESP-12E here and submit it as another PR, I can do that without too much trouble.

screenshot from 2018-06-08 14-01-31
screenshot from 2018-06-08 14-01-36

evanshultz:
I'm fine with piling on relevant changes. :) Thanks!

ESP-07 looks good to me but when I went to check the ESP-12E I find the datasheet link is for ESP-12F. What document were you using to update the 12E? I don't want to be looking at something contradictory to what you're seeing.

Ratfink:
According to https://github.com/KiCad/kicad-library/pull/1742#issuecomment-338488852, ESP-12E is compatible with ESP-12F.  Which I suppose means that I should add an alias for the -12F.

evanshultz:
Looks nice. Thanks!

Issue #580, opened 2018-05-06
Switch naming is inconsistent


Ratfink:
## Background

The naming of our standard switch symbols is presently quite inconsistent.  Below are some examples:

The suffix `x??` has two different meanings:

* `SW_DIP_x02`: two SPST switches in a single unit
* `SW_DPST_x2`: a single DPST switch split across two units

`SPST` is sometimes implicit, sometimes not:

* `SW_SPST`: number of poles and throws explicitly listed
* `SW_Push`: number of poles and throws is implicit

Format for listing number of poles and throws is inconsistent:

* `SW_DPST`: double pole, single throw switch
* `SW_Push_Dual`: double pole, single throw pushbutton switch
* `SW_Rotary12`: single pole, 12 throw rotary switch
* `SW_Rotary3x4`: 3 pole, 4 throw rotary switch

Naming of normally closed switches is inconsistent and unclear:

* `SW_Reed_Opener`
* `SW_Push_Open`

Reconciling these inconsistencies would make the library easier for users to search, and could make smarter footprint filtering possible if the footprints are renamed accordingly.  This is related to https://github.com/KiCad/kicad-footprints/issues/144.

## Proposal

My proposal for a consistent naming convention for standard switches is as follows:

**Fixed fields**
Mandatory fields
*Optional fields*

**SW**\_*[type]*\_[poles]P[throws]T*x[quantity]*\_*[NO/NC]*\_*[control]*\_*[light]*\_*[options]*

* type: e.g. Push, Reed, DIP
* poles: integer (S=1, D=2)
* throws: integer (S=1, D=2)
* quantity: two-digit integer
* NO/NC: if one switch of a given type uses this, all must to ensure proper filtering
* control: e.g. Temperature
* light: e.g. Lamp, LED
* options: anything else, e.g. MultiUnit

This would directly translate into footprint filters, as footprints could be named the same way with manufacturer name and part number as additional suffixes.

## Effects

This proposal would reconcile all the inconsistencies above:

Before | After
---:|:---
`SW_DIP_x02` | `SW_DIP_SPSTx02`
`SW_DPST_x2` | `SW_DPST_MultiUnit`
`SW_SPST` | `SW_SPST`
`SW_Push` | `SW_Push_SPST_NO`
`SW_DPST` | `SW_DPST`
`SW_Push_Dual` | `SW_Push_DPST_NO`
`SW_Rotary12` | `SW_Rotary_SP12T`
[...]

PR #632, opened 2018-05-28
Coilcraft CST symbols

evanshultz:
Symbols that match with https://github.com/KiCad/kicad-footprints/pull/588.

Datasheets:
- https://www.coilcraft.com/pdfs/cst.pdf
- https://www.coilcraft.com/pdfs/cst2010.pdf

I prefer split symbols because it allows placing the halves of the transformer in different places on the schematic. But some may not prefer that and KiCad doesn't check that all units are placed so I included single-unit symbols as well. Previous discussion at https://github.com/KiCad/kicad-library/pull/837. Any feedback of thoughts about how these symbols should look or be built?

image

------------


Misca1234:
By the risk of hijacking this thread, of curiosity, why do you prefer splitting the symbols, instead of using labels (https://github.com/KiCad/kicad-library/pull/837 was closed for comments), labels is your friend to get ride of "stray lines" that clutter the scheme (like the example you gave of page 27 of the [UCC2893 datasheet](http://www.ti.com/lit/ds/symlink/ucc2893.pdf) ) ?

evanshultz:
So... leave this in limbo or merge it and then update later if/when we ever find a better name?

`MultiUnit` doesn't communicate any better than `Split` to me and it's longer. That's why I chose these names. And it got a good response at https://github.com/KiCad/kicad-symbols/issues/580#issuecomment-396073808. @Ratfink

poeschlr:
If we use Split then it should at least be capitalized normally not all caps.

evanshultz:
All are done.

poeschlr:
Looks good, thanks

PR #633, opened 2018-05-28
SAM-M8Q

jneiva08:
Added symbol for ublox SAM-M8Q.

The datasheet is: https://www.u-blox.com/sites/default/files/SAM-M8Q_DataSheet_(UBX-16012619).pdf

The symbol is the following:

sam-m8q

I took the opportunity to modernize and fix KLC S3.6 error on Linx_RXM-GPS.

The symbol is the following:

linx_rxm-gps

This PR for SAM-M8Q footprint is [#605](https://github.com/KiCad/kicad-footprints/pull/605)

------------


evanshultz:
This is a really nice symbol! Thanks for following the conventions so well!
- Pin 18 should have an overbar (add a leading `~` to the pin name). I suspect 8 should as well if it's ever used, but the datasheet doesn't tell us much.
- Put the ref des on the left above the symbol and part number on the right.
- Can the symbol shrink any? It seems a bit overly big.

I'll wait for the footprint to be merged to merge this.

Thanks for fixing the existing symbol! Since it appears you only changed the pin name offset, shrunk pin length to 100mil, and made it a bit less wide this shouldn't be too bad. Did you make other changes that I missed?

jneiva08:
I didn't made yet the commit, i'm putting the image to see as preview.

m8q

Tried to shrink a little but do not feel right because the text is overlaps. Putting ~ on pin 8 and 18 could remove "_N"

On RXM-GPS-RM moved the rectangle to fit on 2.54mm grid. Moved NC to the rectangle, and fix KLC S3.6.

evanshultz:
Yeah, that does look cramped. What about swap the positions of pins 18 and 8, then move pins 19 and 7 down by 100mil? Also, ref des on the left and part number on the right.

jneiva08:
hope have all right, this time.
the symbol looks like:

symbol

evanshultz:
This looks good to me. Thanks!

However, the ref des and part number are flipped. Please swap their locations.

Lastly, would it be easier if the actual pin name was retained for pins 8 and 18? Adding the overbar is good but also keeping the `_N` means the KiCad symbol has exactly the same names as seen in the datsheet. I realize the user can still match them up, but I like to make it aligned. And it seems adding those two characters will still fit in the symbol nicely. What do you think?

jneiva08:
Switched place of ref and part number.  Put back "_N" on pin 8 and 18 like datasheet.

Now is how it looks:

ublox

evanshultz:
Sorry that I missed this update! The footprint was already merged and this is all good now. Thanks!

PR #634, opened 2018-05-29
Add symbol for SIM800C

jneiva08:
Added symbol for SIMCOM SIM800C.

The datasheet is: http://simcom.ee/documents/SIM800C/SIM800C_Hardware_Design_V1.05.pdf

The symbol is the following:

sim800c_symbol

This PR for SIM800C footprint [#615](https://github.com/KiCad/kicad-footprints/pull/615)

------------


Misca1234:
Another thing (which made me change the last comment) 
On page 13, pin 26 is named USB_DN but on page 14 the same pin is named USB_DM?
This must be a manufacture type, I guess they represent the negative and positive lines, 
which then must mean USB_DP is positive and and therefore pin 26 should be negative and the name USB_DN makes more sense.

What are your thoughts ?

Misca1234:
You should align the bottom line of the symbol to the GND pins as well, now the GND pins "hangs in the air"

jneiva08:
I added "_N" to MIC and SPK to be identical to SIM900. Already fix it.
The USB DP/DN pins should be a typo,  I followed pin description table pag.14 (the Microchip usb hub uses DP/DM).
The SIM pins should be all grouped, the sim_rst don't need '~'. (Can seen SIM900 and UL865 on library)

evanshultz:
@jneiva08 
- Can ADC be on the left side since it's an input?
- And maybe pins 41/42 on the right side?
- We have used Passive type for antenna pins in the past because they only go to passive parts or traces.
- I also think the symbol could be shrunk a bit.
- Flip the symbol ref des and name so ref des is on top.

Just a question for my own information since Google shows loads of DP/DM and DP/DN. P and N are obvious as Positive and Negative; why use M? Do you know? I don't see an explanation of _why_ after a quick online search, even though it is a common nomenclature.

Can you post a new screenshot after all of the changes requested above?

Shackmeister:
DP is Data Plus or D+
DM is Data Minus or D-

jneiva08:
have already made the alterations requested.

Now looks like this:
sim800c

Misca1234:
@Shackmeister 
Yes, if you look in the data sheet the manufacture is naming pin 14 (supposedly minus) in two different ways, on page 13, pin 26 is named USB_DN but on page 14 the same pin is named USB_DM?

So they use "minus" and "negative" for the minus pin so the question was name should be used for pin 14, USB_DN or USB_DM (the "plus" pin have the "same" problem)

jneiva08:
I could make a trade, put back the USB VBUS and grow a little vertically and decrease horizontally?

evanshultz:
Sure. If it looks nice to you then make a commit and screenshot to everyone can quickly see it too. Thanks!

jneiva08:
done, now looks:

sim800c

PR #636, opened 2018-05-30
Added LTC6902

SchrodingersGat:
multi phase clock generator with spread-spectrum outputs

http://www.analog.com/media/en/technical-documentation/data-sheets/6902f.pdf

image


evanshultz:
- How does the keyword "signal" relate?
- Perhaps pins 9 and 10 would be best at the top since they're resistor-coupled to pin 1.
- Then pins 2 and 3 can slide down.
- As discussed in another PR, perhaps pins 9 and 10 should be Passive type since they only go to a resistor in the datasheet? Even if they were "digitally" controlled by an external part, they would likely need a resistor or transistor or something in series to the pin since the IC has a current sink inside.

Can you confirm no other changes took place in the v2.3 to v2.4 conversion? The MCP7940 stuff looks OK to me but I'd like another set of eyes.

SchrodingersGat:
> How does the keyword "signal" relate?

The chip generates a "clock signal" - I think it should reasonably be searchable using this term

> Perhaps pins 9 and 10 would be best at the top since they're resistor-coupled to pin 1.
> Then pins 2 and 3 can slide down.

Fair enough!

> As discussed in another PR, perhaps pins 9 and 10 should be Passive type since they only go to a resistor in the datasheet? Even if they were "digitally" controlled by an external part, they would likely need a resistor or transistor or something in series to the pin since the IC has a current sink inside.

Yep, I'll keep this in mind for future symbols too.

> Can you confirm no other changes took place in the v2.3 to v2.4 conversion? The MCP7940 stuff looks OK to me but I'd like another set of eyes.

I only added this part. Any other changes are a result of KiCad rearranging the file.

PR #637, opened 2018-05-30
Added LTC3638 140V switcher

SchrodingersGat:
140V / 250mA switching voltage regulator

Datasheet: http://www.analog.com/media/en/technical-documentation/data-sheets/3638fa.pdf

image

Copied from existing LTC3630 regulator


evanshultz:
I know you said this is a copied symbol, but here goes:
- It is not synchronous (rectification); it has a freewheeling diode.
- Should pin 1 be Passive type? That's the way many others are and because of the inductor this pin will never touch an external net that will source current to other parts.
- Pin 5 should not be Input type. I suppose Output?
- Pins 6 and 7 should be Input type.
- Should pin 10 be Passive type? It only connects to a capacitor. Maybe for pin 11 too since it's not a logic input and has only a resistor to ground.
- I think the symbol would be far better as I've shown below. It puts inputs mostly on the left and outputs on the right, will make connecting the FBO and Vfb pins easier between multiple symbols, and allows for better/easier external component connections:
image

```
# LTC3638
#
DEF LTC3638 U 0 20 Y Y 1 F N
F0 "U" -310 450 50 H V L CNN
F1 "LTC3638" 40 450 50 H V L CNN
F2 "Package_SO:Linear_MSOP-12-16-1EP_3x4mm_P0.5mm" 50 -450 50 H I L CNN
F3 "" 0 -100 50 H I C CNN
$FPLIST
 Linear*MSOP*12*16*EP*3x4mm*P0.5mm*
$ENDFPLIST
DRAW
S -300 400 300 -400 0 1 10 f
X SW 1 400 200 100 L 50 50 1 1 P
X SS 10 -400 -200 100 R 50 50 1 1 P
X Iset 11 -400 -300 100 R 50 50 1 1 P
X OVLO 12 -400 200 100 R 50 50 1 1 I
X RUN 14 -400 300 100 R 50 50 1 1 I
X GND 16 0 -500 100 U 50 50 1 1 P N
X GND 17 0 -500 100 U 50 50 1 1 P N
X Vin 3 0 500 100 D 50 50 1 1 W
X FBO 5 400 -300 100 L 50 50 1 1 O
X Vprg2 6 -400 0 100 R 50 50 1 1 I
X Vprg1 7 -400 100 100 R 50 50 1 1 I
X GND 8 0 -500 100 U 50 50 1 1 W
X Vfb 9 400 -200 100 L 50 50 1 1 I
ENDDRAW
ENDDEF
```

Lastly, I note that this footprint could use a little polish. The silk is overly big and the pads do not match the recommended footprint. That's just a quick look. And also, according to https://github.com/KiCad/kicad-footprints/issues/436 we decided to defer figuring out how to name footprints that fall into this category where multiple pins are missing. (`Linear_MSE` would work and it's quite simple.) Would you care to fix up the part design (if not the naming) or no?

SchrodingersGat:
Hey @evanshultz I have made the fixes as suggested, and also made the same fixes to the existing LTC3630 parts. Is it ok to merge now (and fix the footprints later?)

evanshultz:
Yep! Thank you.

PR #638, opened 2018-05-30
added symbol for SODIMM 200/DR1/DDR2 socket

herostrat:
This is the symboi for the SODIMM socket footprint of https://github.com/KiCad/kicad-footprints/pull/375

It consits of to parts:
A
grafik

B
grafik

It is gerneric and shaped as the footprint.
Because it is generic I placed it in connector.lib

There are a number of similar sockets, here are some example datasheet:
https://www.mouser.de/datasheet/2/418/NG_CD_1473149_K3-1239525.pdf
https://www.mouser.de/datasheet/2/418/NG_CD_1746530_D1-658815.pdf

TE seems to be the dominant producer.

I  do not think we should split part B any further.
ATM it is splittet into the two rows of the hardware.

Please let me know if there is anything you feel I should change.

------------


evanshultz:
I saw you jumped into conversation about the missing SODIMM footprint earlier, but let me welcome you to the KiCad library!

You will find a lot of rules and guidelines for the KiCad library at http://kicad-pcb.org/libraries/klc/. Please see that webpage for more information about the review items below.

Review feedback:
- I'm not sure this symbol should be split. How big is it in relation to the default A4 schematic sheet in KiCad? Since it's not relevant to use only some of the pins, a single symbol seems better to me unless it's insanely large.
- The description is missing the key word "socket". And I'm not so sure about including the RAM standards in the description (keywords are OK) since this could be used in future memory standards or some non-memory purpose. Perhaps just `SODIMM 200 Pin Socket`?
- Since there's no datasheet just use a tilde for the datasheet link.
- The ref des should be `J`, I think.
- All outline boxes are filled with background color.

When you make visual changes to a symbol, please post new screenshots for easy checking.

herostrat:
Thanks a lot for the welcome!
Yes I read it beforehand, but I am sure I missed a few things.

For your feedback:

 - [ ] It does not fit in one A4 paper, neither with two nor with one units. 
For A4 one solution with two units is:
grafik

TBH I do not think that one can fit 200 pins even with basic wiring on a single A4 page.
I devided it into two units so that it can at least fit on a single A3 page:
grafik

Of course I can change it into one unit. Whatever you think is best.
Please let me know what you think.

 - [X] The new description is:

grafik

 - [X] I updated the datasheet to ~

 - [X] I set the pin name offset to 20mil

 - [X] I changed the ref to J (XU was a wrongly taken from [Wikipedia - Reference Designator](https://en.wikipedia.org/wiki/Reference_designator) )

 - [X] I background-filled alls boxes

evanshultz:
Thanks! This looks good. Let's wait a couple days to get any feedback on splitting the symbols and then I will merge.

We really appreciate your initiative to jump in and help improve the library!

herostrat:
If there are no objections I would really appreciate merging the symbol :)
Thanks a lot

poeschlr:
The splitting this way is ok to me. (It really makes sense as there is a physical separation on the part it self.)

The connector lib is also ok as this is a generic symbol. So thanks for your contribution and also for the reviews you made over the last days (Yes i noticed that ;) You are now on a list.)

PR #639, opened 2018-05-30
Add TBD62783A 8-channel source driver by Toshiba

Schlumpf:
image
Datasheet: http://toshiba.semicon-storage.com/info/docget.jsp?did=30523&prodName=TBD62783APG
SSOP-18 footprint PR: https://github.com/KiCad/kicad-footprints/pull/619

------------


evanshultz:
- I think this is best in a `Driver_` library. Maybe `Driver_Motor` or `Driver_Relay`? This part is similar to DRV8860 which is in the latter one. That being said, just a generic `Driver` library is probably best since some of these are general purpose drivers; I don't know why there isn't a generic one.
- Can you update the description to `8-Channel Source Type Transistor Array, TTL and CMOS compatible, 500mA, 50V, DIP-18/SOP-18/SSOP-18/PSOP-18`?
- Please move the PN to the right so it doesn't hit the VCC pin.
- I'll wait for footprints to be merged before merging this.

Schlumpf:
I fixed description PN and footprints.
About the library, I'm not sure. If we move the symbol, we should move A2982 as well, because it is very similar to TBD62783A.

evanshultz:
OK. Since 3/4 footprints are there and there's no default footprint, I'll go ahead and merge this. Will review the footprint you submitted next.

Thanks!

PR #640, opened 2018-05-30
Audio

sebastianbasierski:
Fixed S3.1 and S3.6 errors in Audio library.


evanshultz:
You moved many pins of CS4245 off grid that were OK before!

sebastianbasierski:
That's the reason:
  Violating S3.1
      Symbol slightly off-center
        Center calculated @ (0, 50)

My attempt to fixing it caused all those errors.

For now, i think i'll just revert patch that caused it and leave only fixes for S3.6.

evanshultz:
Thanks!

To provide salient info for the folks who see this later, and to include one change not mentioned above, here is what happened:
- Pin name offset changed to 20mil
- TDA1022 default footprint added and footprint filter fixed
- File version 2.3 -> 2.4

PR #641, opened 2018-05-31
ADAU170x integrated DSPs

evanshultz:
I'm adding a new library with just a single symbol. However, ADI makes tons of DSPs and it seems illogical that no more will be added during the v5 cycle. So why not go ahead and get this started?

Datasheets:
- http://www.analog.com/media/en/technical-documentation/data-sheets/ADAU1701.pdf
- http://www.analog.com/media/en/technical-documentation/data-sheets/ADAU1702.pdf

The differences between the parts aren't in the datasheets (that I can find) and were taken from the table at http://www.analog.com/en/education/education-library/product-faqs/adau1701.html.

image

------------


poeschlr:
Adding the new lib is a good idea. (I did not check the symbol it self as i am limited on time right now. Just wanted to reassure everyone that i agree with adding this lib even after the freeze.)

poeschlr:
Looks good, thanks

PR #644, opened 2018-06-01
Added simbol for Molex 1041681620

jneiva08:
Added symbol for Molex 1041681620.
Is a combo connector for microSD and microSIM.

The datasheet is: https://www.molex.com/pdm_docs/sd/1041681620_sd.pdf

The symbol is the following:

symbol

This PR for footprint is [#624](https://github.com/KiCad/kicad-footprints/pull/624)

------------


evanshultz:
I think we might be best to use a generic symbol since there are other microSD/microSIM combo connectors. https://www.digikey.com/en/supplier-centers/m/molex/nano-sim-microsd-sim-and-micro-sim-memory-card-connectors shows two Molex parts and TE apparently has some from looking at http://www.te.com/usa-en/products/connectors/card-socket-connectors/memory-card-connectors/sim-card-connectors.html?tab=pgp-story.

jneiva08:
Caminho you tell what i need to change to to be more generic? Changing the name of pins?

evanshultz:
The symbol name is the specific part of it since it mentions the Molex PN. One option would be `MicroSD_MicroSIM` to make it generic. I had already started a discussion about using pin names instead of numbers for these exact connectors at https://github.com/KiCad/kicad-symbols/pull/592, but not much feedback yet so we're missing some consensus and a direction.

A couple other things I quickly note:
- This symbol should have a ref des starting with J.
- The microSD card outline seems a bit disproportional.

So for now let's just put this on hold until we gather more input from the other librarians.

jneiva08:
Already renamed the connector an change ref letter.
Did a shrink a little the symbol. now looks like this:

symbol

I have seeing the links you posted, the equivalent connector (uSD & uSIM) conectors from TE is already obsolete.
All the other are uSD/nanoSIM + nanoSIM or equivalent.

For the renaming the pins, the SIM could be Sx, x is the pin number. For SD/uSD could be C(ard) + function name our pin number.

jneiva08:
I did a list with identical active combo connectors:
combo_conn

On GSM SIM, all connection default is of 8 contacts.
On microSD the order is the same but with a letter  different.

We need to choose witch one to be our default order/letter.

antoniovazquezblanco:
What is the status on the naming issue? Do we have a way fordward?

jneiva08:
@evanshultz @antoniovazquezblanco any news?

evanshultz:
I think it was agreed at https://github.com/KiCad/kicad-symbols/pull/592 to allow named pins. That leaves two options I see:
1. Add a new symbol with names that allows this to be merged.
2. Do nothing and wait until closer to v6's release to replace to existing symbol with one having named pins.

Do you have a proposal for the first item, if you want to pursue it? I haven't looked at this closely enough to see if there's a reasonable naming scheme and would all symbols with pin names to be merged right away. Perhaps I can take a look tomorrow?

jneiva08:
If i understood, the SD card will be C1 to C8 and SIM card will be S1 to S7.
In my symbol need to change on SD cad from Tx to Cx, right?

evanshultz:
That is what I'm proposing. I suspect that once the symbols are submitted there will be further discussion. That's why I was planning to tackle this after 5.1 was tagged.

jneiva08:
@evanshultz have made the modifications on  uSD pins, the SIM pins was already in the new style.

Have fixed the DAT0/1 to be I/O. As you can see on page.14 in following [datasheet](https://www.alliedelec.com/m/d/04db416b291011446889dbd6129e2644.pdf).
The other symbols needs to fix DAT0/1 to be I/O, soon will create a PR.

The symbol looks now:
usd_usim

myfreescalewebpage:
@evanshultz what is the status here ?
@jneiva08 can you solve the branch conflict please ?

jneiva08:
@myfreescalewebpage, hope i didn't miss anything

myfreescalewebpage:
@evanshultz I let you finish with this PR. Personally I really wonder about the generic name of this because the pin number is really specific to the Molex part...

evanshultz:
@myfreescalewebpage 
Can you let me know what you're talking about? I think the pin names are generic according to the standard. If I've missed something then I think it should be more generic.

myfreescalewebpage:
Just an example: sometimes you have no card detect pin. And sometimes you have a SIM detect pin. This make 4 variants of the symbol.

Or maybe we should add both and depending of the footprint this will be connected or not ?

Joel

evanshultz:
@myfreescalewebpage 
Yes. That's exactly my idea. We make a few generic symbols that can be used with all card footprints. If the symbols aren't generic, we would need one symbol for each footprint and that's insane.

myfreescalewebpage:
So here you need to add at least the WP pin for the SD and the DETECT pin for the SIM

evanshultz:
It looks to me like this symbol has the DETECT pin for the uSD card. https://github.com/KiCad/kicad-symbols/pull/592#issuecomment-390761214 calls this pin `CD`. And for the SIM, only the 8 standard pins are present. Right?

myfreescalewebpage:
For the SD i was speaking of WP (Write Protect pin).

I agree with your comment but your goal is to make a standard footprint, isn't it ? So you need to add all possible pins we may found on this kind of device, right ? This one for example has a SIM detect pin called "CD" : https://github.com/KiCad/kicad-symbols/pull/2519

evanshultz:
I don't see a WP pin for the SD card at https://www.molex.com/pdm_docs/sd/1041681620_sd.pdf. Am I missing it?

To me, if the card detect pin is changed to `CD` both in number and name then this is OK. Or maybe we think about naming (see below).

Really, this particular symbol is a bit strange so I guess a manufacturer-specific symbol might be OK. But we don't often do manufacturer-specific symbols that have the manufacturer name. What is your exact proposal? Would it be best if I tried to make a list of all the symbols that would be needed and then pin numbers/names? I'm not sure what's the best way to proceed.

myfreescalewebpage:
> I don't see a WP pin for the SD card at https://www.molex.com/pdm_docs/sd/1041681620_sd.pdf. Am I missing it?

No you're correct. But the name of the symbol is generic. So the symbol should be generic. So the symbol should have WP pin.

We usually don't add the manufacturer in the symbol name yes, but for this one "Molex_Combo_microSIM_microSD_Cards_1041681620" is great for me (maybe a little bit long). It's consistent with pther device we already have in the lib (this one for example: https://github.com/KiCad/kicad-symbols/pull/2630)

evanshultz:
I would support all favors of generic symbols that are needed. If there are 9-pin microSD card footprints, then there should be a symbol for that. And symbols with `WP` and `CD` and all other pins as well. Including a `_Shield` symbol with a `SH` pin for ground. Or something like that. I didn't list out every permutation but I think of it like the USB symbols or 3.5mm audio jack symbols; once we have the common symbols in the lib we're pretty much done and ready to go for any footprint imaginable.

On this one, I'd merge it since a combo footprint might be fairly uncommon. And it could always be replaced by a slate of generic symbols between now and v6, if there's times to think about this and make all those symbols.

So I think re-numbering and re-naming the SW pin to `CD`, and possibly moving the `SH` pin on the bottom like we do with shielded connector symbols, ought to be fine. The symbol could shrink a bit, but why mess with that now? Not that I mean to merge something unfinished, but this is a great place to build off of to complete the set of card symbols.

myfreescalewebpage:
I agree with you. This is a great subject we need to progress on.

PR #646, opened 2018-06-02
Modify Sensor_Temperature to conform to S4.3

sethhillbrand:
TMP20 sensors had hidden power pins that created global GND labels.

This commit makes the hidden pins passive while leaving the stacked visible pins as power input.


poeschlr:
Thanks

PR #649, opened 2018-06-03
Add DMP3013SFV P-channel power MOSFET

Ratfink:
Datasheet: https://www.diodes.com/assets/Datasheets/DMP3013SFV.pdf

screenshot from 2018-06-03 14-19-49

------------


evanshultz:
Nice!

Just one comment: should the `*8` in the footprint filter be removed since we usually take out pin count? Since this is a manufacturer-specific footprint maybe we have more flexibility here, but I can't recall seeing pin count in a footprint filter except if there are loads of NC pins. I quick Google search doesn't show any other pin counts for PowerDI3333 package so it should be safe.

Ratfink:
To be honest, I didn't even think of it as a pin count because the package really only has 5 pins.  I suppose I could remove it though, if you think that makes sense to.

evanshultz:
I think it does. No harm I can see and it's consistent with our library.

PR #653, opened 2018-06-05
Fiducial

thomasonw:
This pull request is to add a symbol for Fiducials.   There are no component suppliers, it however allows the assignment of Fiducial footprints for PCB manufacturing aids.

image


thomasonw:
Travis asked for default footprint, added 1mm copper spot.

Travis is calling out no datasheet, there is none for this 'symbol' - is there something other then the field blank that would be appropriate?

Ratfink:
First off, I'm not sure this makes sense to have as a schematic symbol.  You can add fiducial footprints to the PCB, then lock the footprints to prevent them from being removed when updating the netlist.

I'll leave the decision of whether to include these or not to someone more experienced.  I can say that if it is to be included, there should not be a default footprint (ignore the warning), and you can silence the datasheet error by setting it to `~`.

thomasonw:
I kind of think of them along the same lines as other ‘Non-electrical’ additions to the schematic, like the mounting holes:

Misca1234:
Just of curiosity, why would you like to place the fiducials in the scheme already?
Is it so the designer don't "forget" them ?

poeschlr:
Mounting holes make sense as the screws might be good to appear in the BOM. That is not really the case for fiducials.

thomasonw:
That’s a good point, and having a Fiducial on the BOM could be a bit awkward being it will always be a DNP ‘component’.

evanshultz:
Agreed. No need for this symbol whether a board, panel, or component fiducial.

thomasonw:
Sounds like an overall thumbs down then!

PR #661, opened 2018-06-10
Assign better fitting qfn-64 footprints (purpose build ep sizes)

poeschlr:
As part of checking https://github.com/KiCad/kicad-footprints/pull/646 i noticed that a lot of symbols that had one of the two qfn-64-1ep footprints assigned would need a footprint with very different ep sizes.
The footprint pull request adding these new variants is: https://github.com/KiCad/kicad-footprints/pull/649

The only symbol where i am unsure as to what we should do is AT90CAN32-16MU as the tolerance range for the exposed pad given in the [datasheet page 414](http://ww1.microchip.com/downloads/en/DeviceDoc/doc7679.pdf) does not make any sense at all. (min size 3.25mm max size 7.5mm. The inch range given also agrees with this nonsensical range.) For now i assigned the same footprint as used by AT90USB646-MU (nominal ep size is 7.5mm for that one)

---

 MCU_Microchip_SAML.lib was still in the kicad 4 file format and line ordering. Meaning a lot of apparent changes come from the file conversion.

Travis found some errors in symbols that i did not touch. This will not be fixed here.


fauxpark:
Could we get #525 merged before this? Also I don't know why you touched the ATmega/AVR libs anyway; I put in a lot of work to make sure the footprints were correct, and added any missing ones too.

poeschlr:
Well i went the other way round. I found that the current footprints are awful to say the least. Checked where these two footprints are used and fixed only these two.

I can not see a footprint pull request referenced in your symbol pull request. Where exactly did you add footprints?

fauxpark:
#340 is where I did the work on the footprint assignments and filters (and the missing ones were added in kicad/kicad-footprints#434). There was a rather lengthy discussion on the EP sizing, and Evan discovered an Atmel appnote on the topic which helped immensely. I believe I only found one footprint that was not covered in that document, and it was agreed that if in doubt to just use the max size from the datasheet.

(I tried to link together all of my Atmel PRs as best I could but evidently the chain was broken somewhere along the line)

poeschlr:
How can you explain that a 5.4x5.4mm "lead" is connected to a 7.25x7.25mm pad on the pcb? This is just wrong.
The application note simply has some qfn footprint in it but it really disagrees with what the specific datasheets of the parts shows. The only option how your ep size can be correct is if the datasheets are simply wrong. (List the wrong ep size in their part drawings)

fauxpark:
Yep, I did see that note and was baffled by the reference to the paste guidelines as "formulas", as a result I more or less disregarded it. Now I see the mega165A-MU states nominal thermal pad size as 5.4mm, even though the matching footprint from the appnote (by pin count, package size, pitch, and lead length) is 7.25mm. But again I favoured the document as I was told those dimensions in the datasheet are for the *package*, not the *land pattern*.

poeschlr:
It is ok to use such an appnote if the ordernumber for your part is explicitly stated in that note. Or if the datasheet for your part references that appnote. Otherwise there is no knowing if that appnote was ever intended for the part in question.

---

What i get from your response is that indeed the footprints assigned here are a better more fitting choice for these symbols.

fauxpark:
> It is ok to use such an appnote if the ordernumber for your part is explicitly stated in that note. Or if the datasheet for your part references that appnote.

In that case, considering none of the QFN parts make reference to the appnote, nor vice versa, many more of them may be wrong. However it is the only document we could find by Atmel concerning land patterns...

The first result (for me) on Google for "atmel qfn land pattern" returns that very document. But underneath that is [this page](http://www.microchip.com//wwwAppNotes/AppNotes.aspx?appnote=en590909), which does seem to indicate that it only applies to the listed EEPROM chips -- that would explain the filename. A 64-pin QFN package for a serial EEPROM though? The appnote itself certainly doesn't make any particular mention of the specific devices it should be used for.

poeschlr:
One does not need a land pattern definition by the manufacturer. The dimensioned drawing of the package plus a bit of understating for industry standards (IPC) should be enough to create valid footprints.
In fact the footprints referenced here do just that. I wrote a script that uses the IPC definitions to generate footprints from package dimensions. (More details see the footprint pull request and the script referenced within that.)

fauxpark:
Pointing to the IPC standards is not so useful when they are locked behind a paywall. It's unfortunate that the Kicad libraries have to use them. (not saying they shouldn't!) So I feel quite limited in my ability to contribute without access to this stuff.

poeschlr:
Well we have generator scripts that follow the ipc recommendations. (And there is a russian server out there that has a copy of the ipc standard. We even found newer standards on some file sharing server so if you want you will find the information you need.)

evanshultz:
Yes, I can. Sorry that I was on vacation for the last week and couldn't get to this earlier. We found the best info we could on Atmel QFNs earlier using their documentation, but let me look at this tomorrow.

Regarding #525, I'm a bit leery of doing that until after the libs are tagged v5 since it's not a functional change but just moving symbols. Plus I've been busy with work and other PRs trying to stuff into v5. Rene, what would you like to do? One of these PRs are sure to break.

poeschlr:
@fauxpark i think i reduced the confusion for contributors by finding out that for the exposed pad size the nominal size of the "slug" should be used. (See https://github.com/KiCad/kicad-footprints/issues/678)

I included this here already.

poeschlr:
@evanshultz i assume you forgot about this.

evanshultz:
@poeschlr 
See above. I was wondering if you scripted this and you'd prefer to merge #525 first since that was done manually. Or if you would rather wait to move any MCUs to \obsolete until after v5. Your call.

poeschlr:
This was done manually.

evanshultz:
@poeschlr 
Here's my thought: moving symbols to an obsolete folder would have been nice, but I didn't review it in time. At this point, getting corrected footprints that match the 3D model name is best overall for KiCad. I will review and merge this first, then #525 and https://github.com/KiCad/kicad-symbols/pull/702 end up where they are.

Agree?

@fauxpark 
I apologize. Would you be willing to re-do your work? Otherwise, I don't want to turn off such great and willing contributions as yours, so I'll re-do your work above if it breaks when this PR is merged.

fauxpark:
I assume I just need to resolve the merge conflicts that will appear in my PR. AFAICT there is only a little overlap between these.

poeschlr:
The reason these footprints need to be fixed is not for having correct 3d models. It is such that the parts can be used in production. (For handsoldering it would not be that important. But if you use them for reflow soldering then the ep size must be correct. Or at least near to correct. Rignt now most of these symbols point to footprints with exposed pads that are way too large.)

evanshultz:
@fauxpark 
I believe so.

@poeschlr  
Yes. I get that. But matching the 3D model will, I think, result in less confusion for users as well.

evanshultz:
- No thermal vias on `DSPIC33FJ64GP306A`?
- Should `TUSB8041` use a manufacturer-specific footprint? Or does the generator and the datasheet end up with the same footprint?
- Those two above don't have `1EP` in the footprint filter.

All the Microchip changes look OK. Once I checked one part with a particular suffix, I just checked that all identical suffixes had the same footprint.

evanshultz:
You mentioned not wanting to fix other errors, but Travis looks to have found a broken footprint filter on `ATSAML21G16B-AUT`. Is that worth fixing?

evanshultz:
I merged all of your footprint PRs that seem to be relevant, and there are still a number of missing QFN64 footprints according to Travis.

poeschlr:
For some reason there was a "P" missing from all the footprint fields. Fixed now.

evanshultz:
OK. Nice!

I read above that you didn't want to touch anything else, but I would like you to confirm you won't do anything I listed at https://github.com/KiCad/kicad-symbols/pull/661#issuecomment-401138139 or https://github.com/KiCad/kicad-symbols/pull/661#issuecomment-401140937 before I merge.

poeschlr:
I don't have time to look into anything else.

PR #663, opened 2018-06-10
Added 1x18650 and 2x18650 cell battery holder

Misca1234:
Suggestion for symbols for battery holder

What are your thoughts about making symbols for battery holders?
I deliberately added the battery type as a suffix it in the scheme so it is visible what type of battery it is.

bild

bild

bild

bild

bild

bild

------------

### DIFF
--- a/Battery_Holder.dcm
+++ b/Battery_Holder.dcm
@@ -1,0 +1,15 @@
+EESchema-DOCLIB  Version 2.0
+#
+$CMP BatteryHolder_1x18650
+D Battery holder 1x18650 cells
+K Battery holder 1x18650 cells
+F ~
+$ENDCMP
+#
+$CMP BatteryHolder_2x18650
+D Battery holder 2x18650 cells
+K Battery holder 2x18650 cells
+F ~
+$ENDCMP
+#
+#End Doc Library
--- a/Battery_Holder.lib
+++ b/Battery_Holder.lib
@@ -1,0 +1,40 @@
+EESchema-LIBRARY Version 2.4
+#encoding utf-8
+#
+# BatteryHolder_1x18650
+#
+DEF BatteryHolder_1x18650 U 0 40 Y Y 1 F N
+F0 "U" 100 250 50 H V C CNN
+F1 "BatteryHolder_1x18650" 50 -250 50 H V L CNN
+F2 "" 0 0 50 H I C CNN
+F3 "" 0 0 50 H I C CNN
+$FPLIST
+ BatteryHolder*1x18650*
+$ENDFPLIST
+DRAW
+S -100 200 100 -200 0 1 10 f
+X + 1 0 300 100 D 50 50 1 1 w
+X - 2 0 -300 100 U 50 50 1 1 w
+ENDDRAW
+ENDDEF
+#
+# BatteryHolder_2x18650
+#
+DEF BatteryHolder_2x18650 U 0 20 Y Y 1 F N
+F0 "U" 100 250 50 H V C CNN
+F1 "BatteryHolder_2x18650" 50 -250 50 H V L CNN
+F2 "" 0 0 50 H I C CNN
+F3 "" 0 0 50 H I C CNN
+$FPLIST
+ BatteryHolder*2x18650*
+$ENDFPLIST
+DRAW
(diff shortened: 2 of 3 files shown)


Misca1234:
Note, some of the battery holder footprint will not work, sp I suggest we agree on a symbol, for 1x, 2x, 3x, ... battery holder and if they should have, perhaps invisible power flags, etc.

Then I remove this push and create pushes unique symbols for all the battery holders.

poeschlr:
For the symbol style i would go with something similar to our current battery symbol (can be encased inside a rectangle to show it is a battery holder.)

What do you mean with invisible power flags?

Misca1234:
You mean something like this then

bild

poeschlr:
Your latest screenshot is extremely strange. If the positive pin is power out then the negative one is as well.
There is also a pin number duplication.

Misca1234:
This thread should actually be among the "issue" threads, it is an attempt to coem to an agreement for a symbol for battery holders so it should actually not be merged.

The last screen shots was to illustrate evans suggestion of having battery symbols inside

This shows more better a final result

bild

poeschlr:
A symbol like this would violate ERC. Connecting multiple power output pins does that.

The only real option we have is to stack them. (and make the hidden one as passive)

Misca1234:
Then we stack them but we all agree that this could be the symbol for battery holder ?

Shackmeister:
Shouldn't the double one be 2 parallel battery's? This way they can be connected in series or parallel. I think most dual batteryholders are meant for both

Misca1234:
Yes, it will all depend on how the battery holder works,

What is your ideas about the number of cells?
Let say it is one cell, then there is one cell in the symbol
if it is 3 cells in the holder there will be 3 cells in the symbols
(of course a 9V compound batter holder will only have one cell

Shackmeister:
Perhaps yes, but looking at the suggested symbol it looks like they are connected in parallel

Misca1234:
It is burden on all of our shoulders to to live with those imprecise artist ;)

Note, when we all agree on the principle layout of a battery holder, I will close this thread (should have been among "issues" really) and then I will create the real push of the battery holder symbols.

But if we agree that a battery holder should have 
the rectangle shape, 
filled with yellow, 
Have cells inside to describe the layout up to , let say 3 cells, that shows if the cells are in parallel or serial

Now, lets say a battery (holder) have more than 3 cells, is it a good idea to show this as two cells, with a dotted line between them?
like in these images
https://github.com/KiCad/kicad-symbols/pull/663#issuecomment-397813378

poeschlr:
I would show the 3rd cell replaced with dotted lines for holders with more than 3 cells. Maybe also place a text with NsMp (N number of cells in series, M number of cells in parallel) Or at least use this as part of the naming convention.

Misca1234:
> with NsMp (N number of cells in series, M number of cells in parallel)

So for 
BatteryHolder_Bulgin_BX0036_1xC
It should have been named 
BatteryHolder_Bulgin_BX0036_1s0p_....
?

Do we want the battery size in the name as well, like
BatteryHolder_Keystone_2468_2xAAA.kicad_mod

Should be named instead 
BatteryHolder_Keystone_2468_2s0p_AAA.kicad_mod ?

evanshultz:
@poeschlr 
Ping!

myfreescalewebpage:
@KiCad/librarians what about this PR ? Haven't get answers at my previous comment just above this one, what do you think ? Is this symbol really needed ?!

myfreescalewebpage:
@KiCad/librarians ping :)

evanshultz:
I'm not a big battery user so I didn't chime in because I don't have any strong opinion.

Agreed on the ref des.

Here is a screenshot from IEEE 315-1975 which may or may not be helpful:
image

myfreescalewebpage:
@evanshultz thanks for this reference. THis clearly looks like the symbol we already have in the library for the batteries:

Capture

For me the content of this pull request is not useful and to be closed.

Joel

myfreescalewebpage:
@Misca1234 and you what about my proposition ? I think the existing battery symbols are enough, no ?

Misca1234:
Sure

PR #673, opened 2018-06-18
Added initial version of LM5088-1

johnfwhitmore:
This is a cursory addition of the LM5088-1 symbol. There is a LM5088-2 which changes the function of Pin 10 of the device.

I'm creating a pull request at this early stage as I'm hoping for some feedback on the position of the pins. I've placed the pins, more or less, in the way that the datasheet uses them in the "typical circuit" but that might not conform to the KiCad requirements.

I'm not sure about the addition of LM5088-2. As it changes the function of a pin I'm not sure it can be an alias. 

I should also point out that the datasheet as a 3.3x3.3mm gap in the solder mask whereas I've taken the available footprint in the library of 3x3mm. What difference does 0.3mm make ;)
 
Datasheet - http://www.ti.com/lit/ds/symlink/lm5088.pdf

screenshot from 2018-06-18 10-33-21
screenshot from 2018-06-18 10-33-38


johnfwhitmore:
I've completely changed this symbol to force it to look like the pin out used by Texas Instruments in their datasheet and in their quickstart spreadsheet for calculating the component values. That makes it easier for me to create my schematic circuit to match their documentation, but I'm sure that's not going to adhere to the KiCad guidelines. I've not had any comments on here so I'm not sure if this is just unacceptable as a pull request. If that's the case then I can simply move the symbol to my own private libs.

screenshot from 2018-06-22 22-58-17

johnfwhitmore:
Oops I'd not seen that comment. Obviously I never refreshed this page before I started commenting again. That's going to cause confusing cross talk for which I'm very sorry. :(

Misca1234:
Hi
Change the pin offset to 20
bild

dont include the components name in the description, instead do this
bild

You are missing the foot print filter, FP, change it to this
(in short, replace the part which indicates the pin number with an '*', replace single characters with a '?'

bild

Remove the data sheet link from the filed properties, this filed is not used (keep data sheet link in description)
bild


Pin 17 should be invisible
If you move the whole symbol up "one step 100", you get the symbol centred around origo

Misca1234:
You could add 'LM5088-Q1' as an alias to LM5088-1
If you do, add to the description the difference between them, that is 
For LM5088-1 'Spread Spectrum EMI Reduction'
For LM5088-2 'Hiccup Timer for Overload Protection, automative'

Texas is a little bit messy in the data sheet because they have a mix of the nomenclature for the component

LM5088 used as a common name for both
LM5088-1, LM5088MHX
LM5088-Q1, LM5088-2, LM5088QMHX

If you don't want to use LM5088MHX and LM5088QMHX then you should at least 
rename this one to, just 'LM5088' so it is not confused with 'LM5088-Q1'
http://www.ti.com/product/lm5088
http://www.ti.com/product/lm5088-q1



PS
Personally i think the most correct names for these two components are
LM5088MHX
LM5088QMHX

johnfwhitmore:
Hi Misca1234, thank you for all your comments and sorry for the delay in getting back to you. 

I should make a few commits. My original design had a tightly packed set of pins and then when I came to do a schematic circuit layout I found that very restrictive so Vin was moved to the top of the symbol and it got stretched so that I could fit the other components around the LM5088 more cleanly. 
I agree that I did get the name wrong and it should be the LM5088 and the alias should be as you've suggested. 
As they're sharing a symbol I'll change the name of the pin to RES/DITH which is the only difference between the two variants.

johnfwhitmore:
OK I now have Jenkins happy :) So now it's just down to getting the symbol to agree with best practice for a kicad symbol. I'll attach another screenshot

screenshot from 2018-06-29 16-49-33

johnfwhitmore:
screenshot from 2018-06-29 16-50-31
screenshot from 2018-06-29 16-50-34
screenshot from 2018-06-29 16-50-48

johnfwhitmore:
I hope that's more less correct. I'm sure improvements could be made to descriptions or keywords but let me know any changes required before a possible merge of the symbol.

Cheers

Misca1234:
Hi

No problem, we are all doing this voluntarily so no one can expect a dead line or demand a response within a certain time.

1)
Remove the component name from the description
Change the following 

> High Voltage (75V) Step-Down Switching Regulator, HTTSSOP-16
> LM5088

into
> High Voltage (75V) Step-Down Switching Regulator, HTTSSOP-16
> High Voltage (75V) Step-Down Switching Regulator

2)
I see that I was unclear with the name of the component name and it's nomenclature.
Either you do following 
1a)
Name the component to LM5088 and add an alias LM5088-Q1
or
1b)
Name the component to LM5088MHX and add an alias LM5088QMHX

I am sorry of you got the impression that you was supposed to add all Tex-mesh names 

I will do a complete review once more if you change the above

Misca1234:
Regarding the nomenclature, I think we have to many aliases now

If you want to stick with LM5088, then I suggest you do the following

bild

bild

bild

evanshultz:
- KiCad does not include temperature variants or automotive-grade parts in the official lib, so please remove all `Q` versions. We also don't need a generic 5088 if you have both 508801 and 5088-2 symbols. Lastly, the `MHX` is adding the `X` to signify different packaging (page 34 of datasheet) so we won't include that either. I believe the two variants should be unique symbols since that's the direction of the library (see https://github.com/KiCad/kicad-website/pull/272), so that would make the only two symbols needed `LM5088MH-1` and `LM5088MH-2`.
- Looking at your latest commit I think the documentation for LM5022 has been modified. I didn't check the symbol. Can you please review and revert any changes to parts that are now LM5088?
- While KLC doesn't specify this exactly, for ICs I like to start with the datasheet description, add any additional info, and then end with the package type. For this part, `Wide Input Range Non-Synchronous Buck Controller, 75V, HTSSOP-16` would be nice. Again, this is common in the library but not canon.
- Keywords should be actual keywords, mutually-exclusive with the description and not including the part number, like `step down buck high voltage` for this part.
- Personally think the pin layout of the symbol looks nice and adheres to the standard (inputs on left, outputs on right, power on top and bottom, group related pins together, etc.). Would it be possible, though, to squeeze down the symbol? The pins can still fan out to external circuitry as you have shown above, but just with a smaller symbol.
- Pin 10 should be Passive type. All analog pins that only go to passive parts can be Passive type also: pins 4, 5, probably pin 7 (it's not a logic output and it's almost never connected to anything else), 11 (it only goes to an inductor and is not a signal output nor the output of the circuit) 15 (only goes to a cap). Possibly even pin 3, but Input is OK too since it can accept an external signal.
- Pin 3 should not have an overbar. @Misca1234: It's not a logic pin nor active low in any way!
- Pin 9 should be Input type as it's an input to the ramp generator circuitry.
- Pages 29 and 32 of the datasheet seem to indicate that the exposed pad is connected to ground. The schematics on pages 30-31 show it this way as well. But I don't see anything saying that the pad must be connected to ground or is internally connected. Do you? Page 32 talks about PGND, which isn't a discrete pin, so I think that term is being used for the thermal pad. Therefore, I'm inclined to make pin 17 a separate pin in case there are other connections possible which might be preferred in some use cases. For example, using a net tie from the signal ground to the power ground so routing is forced to avoid high currents circulating through the ground connection of sensitive parts on the controller side. OK?

johnfwhitmore:
You have been busy with all your responses. Thanks for that but I have a few questions about your comments. Well I think I'll have a few questions but I'm working through your responses.
"Remove the component name from the description" I'm slightly confused about this as I thought I had removed the component name from the description but you're obviously referring to the keyword field. I will remove it from there but I'd assumed that why you type in the search box, when looking for a symbol, "LM5088" that's the keyword that's searched. Obviously my assumption is in error. Actually I should check.

On pin 3 not being negative logic. You'd previously asked me to add the '~' to the pin. I'll revert that change.
I've reduced the size of the symbol.
I'm happy with your comments about the thermal pad being a separate pin but I'm not sure what to call it. I'll go over the datasheet again and check. For now I'll push my commits to check Jenkins 

Your comment "No problem, we are all doing this voluntarily so no one can expect a dead line or demand a response within a certain time." I hope I was not being demanding! That certainly wasn't my intention at all, I do realise that this is all done on a volunteer basis.

Thanks for your comments and help

evanshultz:
Note that two librarians have made comments above. So if you want to direct a comment at only one of us, go ahead about mention the handle. Otherwise, I'll take all your comments as generalized.

In v5, the symbol name, description, and keywords are all searchable. So you don't need to duplicate the part number in the keywords since it's in the symbol name already.

No, it's all good for me! It's easy to forget things so I appreciate a ping if there's no motion in a while.

evanshultz:
More review comments (to separate them):
- Two separate symbols are needed because of pin 10.Use the symbol names I gave above. You do not need a generic `LM5088` symbol.
- Include the differences in the description (`Frequency Dithering` and `Hiccup Mode Restart`)
- Do not include pin count, thermal vias, EP size, and other elements in the footprint filter so the user can use the filter to include alternate footprints if they desire. Change it to `HTSSOP*1EP*4.4x5mm*P0.65mm*`
- You can name the thermal pad pin `EP` (page 3) or `PGND` (page 32).
- Please post a screenshot of the symbol. This is always helpful when you've made visual changes. Since the centering rule isn't important, your symbol doesn't need to have equal height above and below the x-axis, so something like this would be fine (moved top down 100mil, moved left side pins up 100mil, separated ground pins, renamed pin 17 and made it visible):
image

johnfwhitmore:
I didn't notice that there were two librarians in here. I'm just mindlessly reading comments and trying to implement changes ;) Sorry if I confused ye.

I think I've posted fresh screenshot on any visual change but I may well have missed one or two. If I did sorry I'll try and remember to do that in future.

So I've split the single symbol with 2 alias symbols into two separate symbols, but I'm sure I've got the dithering and hicup in the description not quite right. :( Let me know 

screenshot from 2018-07-05 12-23-21
screenshot from 2018-07-05 12-23-24

johnfwhitmore:
Oops just noticed that you'd moved the pins up on the left hand side to clear some space around the ground pins on the bottom. Let me know what you think about description and I'll change pins with that change.

Misca1234:
@evanshultz 
Are you happy with the latest changes ?

evanshultz:
Sorry! I missed this update for so long!

Description:
After the base description, just add `, Frequency Dithering` (or `, Hiccup Mode Restart`), then the voltage, and finally the package. In other words, you can replace ` with` with a comma.

Symbol:
Indeed it doesn't need to be quite so tall. It could also probably be more narrow if the left-side pins move up and GND/PGND move a bit left. But just see what you can do and what looks good.

Sorry again!

evanshultz:
The description fix looks good. If you can squeeze down the symbol, at least vertically, I will merge. Thanks!

johnfwhitmore:
Thank you for merging, I never got back to squeeze the symbol, been distracted. I'll leave it for now.

evanshultz:
Here are squeezed symbols if you would update your branch with this:
```
# LM5088-1
#
DEF LM5088-1 U 0 20 Y Y 1 F N
F0 "U" -350 650 50 H V C CNN
F1 "LM5088-1" 250 650 50 H V C CNN
F2 "Package_SO:HTSSOP-16-1EP_4.4x5mm_P0.65mm_EP3.4x5mm_Mask3x3mm_ThermalVias" 1800 -750 50 H I C CNN
F3 "" 2250 -750 50 H I C CNN
$FPLIST
 HTSSOP*1EP*4.4x5mm*P0.65mm*
$ENDFPLIST
DRAW
S -400 600 400 -700 0 1 10 f
X VIN 1 0 700 100 D 50 50 1 1 W
X DITH 10 -500 100 100 R 50 50 1 1 P
X CSG 11 500 -200 100 L 50 50 1 1 P
X CS 12 500 0 100 L 50 50 1 1 I
X SW 13 500 200 100 L 50 50 1 1 O
X HG 14 500 500 100 L 50 50 1 1 O
X BOOT 15 500 400 100 L 50 50 1 1 P
X VCC 16 -500 -100 100 R 50 50 1 1 w
X PGND 17 100 -800 100 U 50 50 1 1 P
X EN 2 -500 300 100 R 50 50 1 1 I
X SS 3 -500 -400 100 R 50 50 1 1 I
X RAMP 4 -500 -300 100 R 50 50 1 1 P
X RT/SYNC 5 -500 -500 100 R 50 50 1 1 P
X GND 6 0 -800 100 U 50 50 1 1 W
X COMP 7 500 -600 100 L 50 50 1 1 P
X FB 8 500 -400 100 L 50 50 1 1 I
X OUT 9 500 -300 100 L 50 50 1 1 I
ENDDRAW
ENDDEF
#
# LM5088-2
#
DEF LM5088-2 U 0 20 Y Y 1 F N
F0 "U" -350 650 50 H V C CNN
F1 "LM5088-2" 250 650 50 H V C CNN
F2 "Package_SO:HTSSOP-16-1EP_4.4x5mm_P0.65mm_EP3.4x5mm_Mask3x3mm_ThermalVias" 1800 -750 50 H I C CNN
F3 "" 2250 -750 50 H I C CNN
$FPLIST
 HTSSOP*1EP*4.4x5mm*P0.65mm*
$ENDFPLIST
DRAW
S -400 600 400 -700 0 1 10 f
X VIN 1 0 700 100 D 50 50 1 1 W
X RES 10 -500 100 100 R 50 50 1 1 P
X CSG 11 500 -200 100 L 50 50 1 1 P
X CS 12 500 0 100 L 50 50 1 1 I
X SW 13 500 200 100 L 50 50 1 1 O
X HG 14 500 500 100 L 50 50 1 1 O
X BOOT 15 500 400 100 L 50 50 1 1 P
X VCC 16 -500 -100 100 R 50 50 1 1 w
X PGND 17 100 -800 100 U 50 50 1 1 P
X EN 2 -500 300 100 R 50 50 1 1 I
X SS 3 -500 -400 100 R 50 50 1 1 I
X RAMP 4 -500 -300 100 R 50 50 1 1 P
X RT/SYNC 5 -500 -500 100 R 50 50 1 1 P
X GND 6 0 -800 100 U 50 50 1 1 W
X COMP 7 500 -600 100 L 50 50 1 1 P
X FB 8 500 -400 100 L 50 50 1 1 I
X OUT 9 500 -300 100 L 50 50 1 1 I
ENDDRAW
ENDDEF
```

image

johnfwhitmore:
If the symbol is going to be merged into the official library I'll just delete this branch and pull from upstream?

evanshultz:
It's not been merged yet. But if you'd update the symbols in your branch with the ones I pasted above, I will merge.

johnfwhitmore:
Oops, sorry I'll do that now I didn't realise the procedure

evanshultz:
Thanks!!

johnfwhitmore:
Thank you and sorry for the delay

PR #675, opened 2018-06-19
add DS90C402

HubertD:
add DS90C402 dual LVDS receiver
Datasheet: http://www.ti.com/lit/ds/symlink/ds90c402.pdf

ds90c402

------------


evanshultz:
- Could you put the power pins on the centerline?
- Is there any benefit to having the receivers in separate units instead of both in one symbol?
- Ref des goes in the top left (good) and part number in the top right.

Please post a new screenshot if you make major visible changes to a symbol? That helps reviewing.

One more suggestion: If you push a number of commits to the same lib, do them in one PR. This will avoid branch conflicts when the first PR get merged.

HubertD:
thanks @evanshultz !

Personally, I don't like split symbols, so I'd rather stick with one.
If required, I can do the split of course.

Made the other changes you requested:
ds90c402

evanshultz:
For some reason we (the librarians, or at least me) seem to get an email for each commit you push. This is atypical. How are you making commits?

Having a single symbol is OK. I was only getting your feedback and preference. :)

Thanks for contributing the KiCad's library! You have a bunch of open PRs and I hope to see more from you once these are all merged!

poeschlr:
I would assume you got an email because you where mentioned in the commit message.

evanshultz:
Yes, perhaps. Or maybe it's Gmail or GH. The formatting of the email for the commit after my comment was different than any other notification email I've ever gotten. You're probably right.

PR #678, opened 2018-06-21
Fix the ep size for texas mcu symbols

poeschlr:
This fixes the mistakes made in https://github.com/KiCad/kicad-symbols/pull/560 regarding ep size as mentioned in https://github.com/KiCad/kicad-footprints/issues/678.

The footprint side is handled in https://github.com/KiCad/kicad-footprints/pull/680.


evanshultz:
All footprints now present and no errors. Thanks!

But good God, this library is a mess. I'll try to do some quick clean up that hopefully can get merged soon.

PR #679, opened 2018-06-21
fix ep size for sensor humidity

poeschlr:
Fixes my mistake as described in https://github.com/KiCad/kicad-footprints/issues/678 for the symbol originally adapted in https://github.com/KiCad/kicad-symbols/pull/558


Misca1234:
Just to follow your reason, in 
https://github.com/KiCad/kicad-footprints/issues/678
you say " but i think in future we can go back to using nominal sizes for the EP."

But this push for Si7020-A20, you set the DFN-6-1EP_3x3mm_P1mm_EP1.65x2.55mm
But according to the data sheet
https://www.silabs.com/documents/public/data-sheets/Si7020-A20.pdf
The nominal size of the PAD (page 31) is 1.5x2.4
and the recommended Land pattern (page 43) is 1.60x2.5

You take DFN-6-1EP_3x3mm_P1mm_EP1.65x2.55mm because it is the closest existing match ?

PR #681, opened 2018-06-21
Add MCP401x/402x

RensBloom:
Hi,

Potentiometers:
- MCP401x http://ww1.microchip.com/downloads/en/DeviceDoc/21978c.pdf
- MCP402x http://ww1.microchip.com/downloads/en/DeviceDoc/21945e.pdf (similar, but with WiperLock technology. I have added those as aliases of the MCP401x)

kicadadd you are contributing


RensBloom:
I indeed pushed the wrong files. It should now have the correct files.
I had seen there are a lot of different drawings for potentiometers. I preferred not to use American style resistors in the drawing and it should be clear to what pins the potentiometer is connected to. The drawing of the TPL0401A is a good example of what I mean it totally should not be like (American resistor, poorly drawn connection to GND pin):
image
The DS1267 drawing is too small to make a clear distinction between an unconnected potentiometer pin and a potentiometer pin connected to GND. I therefore used the MCP45HV51 drawing as a base and made it slightly smaller. I agree to you that the drawings are still quite huge.

I think it might be good to decide for a preferred drawing style of potentiometers.

evanshultz:
A standard style would be a good idea. Using box vs wiggly resistors is your choice so this is fine. I have created an issue about standardizing the digital pot symbols at https://github.com/KiCad/kicad-symbols/issues/699. Thanks for mentioning it. While the library is growing and improving, there are certainly areas that still need real attention.

Review notes:
- Could you take a crack at shrinking these symbols? Keep the box style if you like, but maybe start with the resistor box being the same size as the discrete resistor symbol?
- Add `with WiperLock™ Technology` for relevant part descriptions. (You copied 401x descriptions so the 402x don't indicate they're non-volatile.)
- Add package(s) to the end of the description.
- Correct footprint filters as you've done elsewhere.
- Use 20mil pin name offset.
- Let's add the resistance (wildcarded) to capture the full MPN. We also wildcard temp rating of ICs and KiCad can't handle the slash in Microchip MPNs. So for example, change MCP4014 to `MCP4014-xxxxOT`.
- MCP4011 needs changed because the DFN footprint can't share a symbol as it has 9 pins (the thermal slug goes on pad 9 of the footprint). You will have one symbol and 3 ALIASes here if you group the SOIC and MSOP packages together, or split it into two symbols (one for SOIC and one for MSOP). See the example symbol name above. You are not required to add another symbol for the DFN package.
- Change MCP4012/3 footprint filter to `SOT?23*6*`. Fix MCP4014 similarly.

RensBloom:
All symbols now are smaller, names, descriptions and packages are updated.

image

evanshultz:
These symbols look really great! Thanks!
- Add package to the end of the description as requested above. `SOIC-8`, `MSOP-8`, `SOT-23-6`, etc.
- Do not include pin count in the footprint filter, so 5- or 6-pin packages can both be `SOT?23*`. My mistake above. Sorry about that.
- Do not include the `T` in the symbol name before the hyphen in 4012/13/14. This indicates T&R packaging.
- SOIC8 footprint filter should be `SOIC*3.9x4.9mm*P1.27mm*`. MSOP8 should be `MSOP*3x3mm*P0.65mm*`.
- If you do want to include the DFN package for 4011 you're welcome to do so.

RensBloom:
No new pictures, as everything looks the same (except for the removed T). I updated everything, maybe I'll add the DFN later on too.

evanshultz:
Looks good! Just one small thing left before merging: The MSOP footprint filter ends in `P0.65*` and it should be `P0.65mm*` (missing the `mm`).

RensBloom:
Done!

PR #682, opened 2018-06-21
Add OPA2325

RensBloom:
Hi,

Operational Amplifier:
- OPA2325 http://www.ti.com/lit/ds/symlink/opa2325.pdf

kicadadd

------------


evanshultz:

You've added a number of symbols which we can start to go through one-by-one. You can update all your PRs according if the comment below is a general one. For this PR:
- Add the package type at the end of the description. `SOIC-8` in this case.
- You will have 3 units; two logical and one for power. See other symbols in this lib. In fact, you could make this an ALIAS of several other symbols instead of a new symbol.
- Add a default footprint.
- Update the footprint filter to match other SOIC-8 opamps.

RensBloom:
I didn't read the KLC thorougly, but I see why you are using these conventions. I have updated the OPA2325 symbol.
I tried making the OPA2325 an alias, but I couldn't find any dual-opamp that has only an SOIC-8 footprint (most are available in multiple packages). I have found how to change the footprint filter of an alias, but I can't set a default footprint of an alias without changing the default footprint of the master.

evanshultz:
Sure. I will be separating out opamps by package (hopefully before v5's release) so I can touch that later. All other changes look good and I want to get this committed so I can work on this too. Thank you again for contributing to KiCad and I hope to see even more submissions from you!

PR #683, opened 2018-06-21
Add MCP3201

RensBloom:
Hi,

ADC:
- MCP3201 http://ww1.microchip.com/downloads/en/DeviceDoc/21290D.pdf

kicadadd

------------


evanshultz:
- Add package type to the end of the description.
- All pin should use the normal Line style.
- As this is an SPI slave device, CLK should be Input type.
- Could the symbol be a bit smaller vertically, especially when bullet 2 above is done? If that looks ugly, then no worries. Just asking.
- Fully qualify the footprint filters with package dimensions, lead pitch, etc.

poeschlr:
A tip: Set kicad to English for posting screenshots. (Kicad main window -> preferences -> language)

evanshultz:
- We have many old symbols that are not conforming the the library standards/guidelines.
- For this symbol shape 40mil is OK. Thanks for checking.
- Similar to http://kicad-pcb.org/libraries/klc/S4.7/, we use the normal line styles everywhere. While the various styles can be used if desired, the official library does not use line styles (there are probably some old symbols that do).
- Please fix this. CLK is Open Collector type.
- Pin lengths should be 100mil, so if you can make the symbol smaller and keep 100mil pins on grid that's great. If not, that's OK too.
- Looks nice. Thanks!

RensBloom:
I have changed the angle of the diagonal left edges. The pin name offset of 20 mil is now proper looking and the width of the symbol could be decreased. Pins are aligned to the 100 mil grid
image

If you want to have the height decreased (instead of the width), it is possible with a pin name offset of 30 mil  (or 40 mil as before). Only for the input pins, the pin length should be more than the default 10 mil, if we don't want to have a pin number colide with the symbol's edge:
image

However, with longer pins of 15 mil, I think it looks okay. Please tell me if you think this (image below) looks better than my commit (first image), because then I will push this version.
image

evanshultz:
- The description should have the packages separated by a slash. Sorry I didn't mention that clearly before. Please change to `DIP-8/SOIC-8/MSOP-8/TSSOP-8`.
- I'm good with the last version above. Please move the ref des and part number closer to the symbol.

Thanks!

evanshultz:
And I assume your "15mil" is actually "150mil" and it appears this only applies to the two IN* pins.

RensBloom:
Your assumption is correct, 150mil for only the IN- and IN+ pins. I have pushed the drawing and updated the description.

image

evanshultz:
The last commit included TLE* symbols. If you remove them and change the `SO-8` in the description to `SOIC-8` this looks good to me. Thanks!

RensBloom:
Done!

RensBloom:
I'm not managing to remove those commits on this branch. I'll open a new pull request.

PR #684, opened 2018-06-21
Add BPW85, TEPT4400

RensBloom:
Hi,

Phototransistors (optical sensors):
- BPW85 https://www.vishay.com/docs/81531/bpw85a.pdf
- TEPT4400 https://www.vishay.com/docs/81341/tept4400.pdf

kicadadd

------------


evanshultz:
General:
- You copied this symbol from `Q_Photo_NPN`, right?
- Make these symbols an ALIAS of each other since they're identical logically and in pinout.
- Is the long lead C or E? Or is there a flat side of the lens? I don't see it in the datasheet. We need some way to make sure our symbol pinout matches the footprint.

BPW85:
- Change description to `Silicon NPN Phototransistor, T-1`.
- Give something unique about each part in the description, for example `Ica = 0.8-2.5mA` for BPW85A.

TEPT4400:
- Change description to `Ambient Light Sensor, T-1` or `Ambient Light Sensor, NPN Epitaxial Planar Phototransistor, T-1`.

RensBloom:
- I copied it from the BPW40 that has a different footprint (5mm instead of 3mm). I think phototransistors should be in the Sensor_Optical group and not in the Device group.
- I now see that I could make it an alias of the SFH309.
- Good question. The TEPT4400 suggests that the short lead is C, similar to the SFH309. I know that other phototransistors exist (like the Sharp PT480) that should have the long lead as C (it would be better if it was standardized). From the components I can see that the long pin is the post (short pin the anvil). As far as I know, that means the short lead is C.
- The lenses do not have a flat side.

RensBloom:
Post and anvil are the correct terms, see for example [this image on Wikipedia](https://en.wikipedia.org/wiki/Light-emitting_diode#/media/File:LED,_5mm,_green_(en).svg). I have tested the leaded phototransistors (BPW85, TEPT4400, Sharp PT480 that has a different footprint) we have in our lab, and they all have the short lead C.

I'll make the BPW85 and TEPT4400 an alias of the SFH309.

evanshultz:
Looks good. Thanks!

PR #685, opened 2018-06-21
Add AMS1117

RensBloom:
Hi,

Linear Regulator:
- AMS1117 http://www.advanced-monolithic.com/pdf/ds1117.pdf

kicadadd

------------


evanshultz:
Since this part comes in 3 packages, you will need to differentiate between all versions. So you can make AMS1117 and AMS1117CD versions in a single symbol while leaving off the AMS1117CS, for example. Or you could make AMS1117 and AMS1117CD as unique symbols with a default footprint and only one footprint filter each. We do not enforce a specific method, but we do prefer unique symbols. Either way, you will need double the documentation entries with only one footprint for each.

evanshultz:
And the output pin should be Power Output type.

RensBloom:
> So you can make AMS1117 and AMS1117CD versions in a single symbol while leaving off the AMS1117CS, for example.
That is what I did. Do you mean that I should make a symbol too for the AMS1117CS?

evanshultz:
You just have `AMS1117`. Nowhere are you specifying the `CD` suffix. There should be one symbol with 13 ALIASes and 14 documentation entries if you're covering all variants of this part.

In fact, AMS1117 (no suffix) can be an ALIAS of the existing AP1117 symbol. And all of them could be an ALIAS of LM1117.

You are welcome to add the CS suffix if you like!

RensBloom:
The LM1117 is available in TO-220/TO-252/TO-263/SOT-223/(WSON not included in the library). AP1117 is SOT-223 only. For the AMS117 (TO223/TO-252/SOIC) I can add the AMS1117 (no suffix) as an alias of the AP1117 and create new symbols for AMS1117CS and AMS1117CD. Or, is there a method to associate different footprints with aliases?

evanshultz:
Yes, you have the right method. The existing symbols aren't always split up by package, as they should be, so this is a task for a later date.

Since the library isn't clear now, you can pile all the AMS1117 symbols together with many packages or split them apart into different symbols by package (which is preferred).

There is no way to vary the footprint of an ALIAS. It must be a new symbol.

RensBloom:
Updated and added the AMS1117 packages:
image
image
image

evanshultz:
Thanks! You got it!

- Description for SOIC package should be `SOIC-8` (missing the `IC`)
- Stack the VO pins of the SO8 symbol so it looks like the other symbols.
- Adjust `AMS1117CS-1.5` symbol so the pins are on grid.
- You are not including the adjustable versions (no `-1.5` suffix, for example). This is not required, but I'm just mentioning it in case you wanted to. If not, I will happily merge without them.

RensBloom:
I hope this is the correct way to do the pin stack. I added the adjustable versions too.

image
image
image
image
image
image

PR #686, opened 2018-06-21
Add TLE2425

RensBloom:
Hi,

Virtual Ground (reference voltage):
- TLE2425 http://www.ti.com/lit/ds/symlink/tle2425.pdf

kicadadd

------------


evanshultz:
- Remove ` for 5V/GND Analog Systems` from the description since this part can be used for many voltages.
- Add the package type to the end of the description.
- You can make the symbols identical in size. Stick the NC pins on the SO8 part around the body so the end (the little "x") is on the edge of the body so they're not easily connected.
- Please place the text around the symbol in the same location. The `xD` one looks good.
- Can you find the MPN for the various TO92 lead prep versions? Apparently they should both be available, but I don't see the clear MPN differentiation at http://www.ti.com/product/TLE2425/samplebuy.

RensBloom:
I think I was confused myself too. The TLE2425xLP is interchangeable with the TLE2426xLP, so I think I meant that I those two the same size. I made the TLE2425xD the same size as the TLE2426xD, but the TLE2426xD has an additional noise reduction pin. There is indeed no reason to have the TLE2425xD drawn larger than the xLP-version.

I have updated the drawing of the TLE2425xD, it is know the same size as its interchangeable components (TLE2425xLP/TLE2426xLP). I have updated the footprint filter of both the TLE2426xLP and TLE2425xLP as you suggested.

image

evanshultz:
Yes, this is a bit confusing :)
- Remove "2.5V" from the start of the descriptions.
- In fact, start all of them with `Precision virtual ground`, then add comma-delimited extra info if you like (`4 to 40V`, Noise Reduction Pin`, etc.), and end with the package type (either `SOIC-8 or `TO-92`). Remove any temp reference.
- Can the TO=92 symbols be ALIASes of each other? They have the same pinout and packages.
- The noise reduction pin should be Passive type.
- It seems there are 8-pin symbols for TLE2426, but not TLE2425 anymore. Did you intend to remove it?
- Please post a screenshot of all symbols together.

Thanks!

RensBloom:
Changed. Somehow the TLE2425xD was missing in my previous commit (don't know why). It should be back now.
image
image
image
image

evanshultz:
No commit was pushed. I think it went to the MCP3201 branch accidentally??

RensBloom:
It went indeed to the wrong branch, my fault. It should now be here :-)

evanshultz:
Great!

Please remove `2.5V output, ` from the descriptions. This is one application mentioned in the datasheet, but the part can be used in a wide range of voltages so that is misleading. Then I can merge!

RensBloom:
Sure I can remove. The TLE2425 always outputs 2.5V independent of the input voltage, similar to linear regulators. The datasheet gives 5V/GND-systems as an example application (the only case in which the output will work as a balanced rail splitter). When the input is 10V/GND, the output will still be 2.5V. Therefore, I think it would be important to state 2.5V in the description, similar to descriptions of linear regulators.

RensBloom:
Done! :-)

evanshultz:
Thank you!

PR #687, opened 2018-06-21
Add Symbols for DS90LV011A and SN65LVDS1DBV LVDS Drivers

HubertD:
Symbol borrowed from FIN1019M

ds90lv011a

Datasheets:
http://www.ti.com/lit/ds/symlink/ds90lv011a.pdf
http://www.ti.com/lit/ds/symlink/sn65lvds1.pdf

------------


evanshultz:
- Description as I mentioned in your other PR. Something like `LVDS Single High Speed Differential Driver, 400Mbps, SOT-23-5` for DS90LV011A.
- Move extraneous part-specific info like switching rate from the keywords and put in description.
- Your pin names don't match either part's datasheet. Even with a borrowed symbol let's do that.
- Are you willing to add the other SN65LVDSx symbols? Not required, just asking.

HubertD:
thanks again for your feedback, @evanshultz !
I made the requested changes with one exception - I would prefer to name the input pin "IN" instead of "TTL-IN" since that would really bloat the symbol?

I did not plan to make more LVDS driver symbols (simply don't need them myself), but I will gladly do some more if there is demand. I guess your point would be to factor out some Interface_LVDS library?

evanshultz:
- Keeping the pin name "IN" is OK.
- I was simply asking if you're willing to make more symbols to flesh out the SN65LVDxx series. Just a request. If not that's fine and this is already a nice addition.

Please put the part number at the top right above the symbol.

This branch is broken because of a PR in the same library. Please fix this so I can merge this PR. Instead of each PR being a single part, each PR being a single library works fine.

HubertD:
ds90lv011a
fixed.

evanshultz:
Thank you!

PR #688, opened 2018-06-21
Add Symbol for DS90LV027A dual LVDS driver

HubertD:
Datasheet: http://www.ti.com/lit/ds/symlink/ds90lv027a.pdf
Symbol drawing borrowed from FIN1019M
ds90lv027a

------------


evanshultz:
Please always post a screenshot, with a possible allowance if you're only adding an ALIAS. In this case you may have copied some geometry, but the symbol is clearly different so no pass.

- Put package type at the end of the description.
- Use the datasheet description with optional additional info after that for the description. Something like ` LVDS Dual High Speed Differential Driver, >600MBps, SOIC-8`.
- Remove package type from the keywords.
- Fully qualify the footprint filter with package size and lead pitch.
- For pins 5-8 please use TI's naming of `DO-1` instead of `DO1-`. Yes, I like yours better too.

HubertD:
I did upload a screenshot in the first place, but accidentally removed it when editing my initial comment - sorry for that. I uploaded a new one now.
Otherwise, I made the changes you requested - thanks for your feedback!

evanshultz:
Put the part number at the top right above the symbol. Then this is good to merge once the branch conflict is fixed.

HubertD:
ds90lv027a

- moved part number to top right
- updated label texts to "D1" and "D2" to better match the SN65LVDS047 symbol.
- waiting for new merge conflicts :)

evanshultz:
And why this once didn't develop a branch conflict is a mystery...

PR #689, opened 2018-06-21
Add Symbols for SN65LVDS047 quad LVDS drivers

HubertD:
Datasheet: http://www.ti.com/lit/ds/symlink/sn65lvds047.pdf
sn65lvds047

------------


evanshultz:
- Add `Differential Line` in the description to match the datasheet.
- Remove package type from the keywords.
- Please put ref des at top left.
- Name pin 8 `~EN` to get the overbar.
- Please use datasheet pin names.
- Fully qualify footprints filters.
- Your enable gate symbols looks a little smashed compared to a typical AND gate. Can you fatten it? Since the function is relatively understood I'd accept removing this gate in favor of simple pins and a smaller symbol if you prefer that style. Up to you.

As this symbol will look different, post a new screenshot after making updates. This makes reviewing much easier and faster.

HubertD:
thanks for your feedback @evanshultz !
made the changes you proposed... did also rework the AND gate part a little.
Still, I don't like the drawing very much (especially the VCC pin), but it's the best I can come up with right now.

sn65lvds047

evanshultz:
- Please move the part number back where it was and also keep the ref des where it is now.
- It looks like a little break between the line from ~EN and the circle. Can you close that gap?
- The AND gate looks fine to me. Thanks!

I will merge the symbol like this. But you could make the triangle have a smaller angle on the right side, so the top and bottom are flatter, and then maybe VCC can move to the middle. Similarly, with a more creative line routing from the output of the AND gate, GND could also move to the middle. I suppose you may want to shrink the AND gate and make the EN and ~EN lines step closer together to get the output of the AND gate to move directly vertical and meet the existing long vertical line. If you want to play around, see if you can find anything that looks better. If not, that's OK too.

HubertD:
sn65lvds047
a little better now, thanks for your input!

evanshultz:
Looks great to me too. Thanks!

evanshultz:
You will need to fix the branch conflicts so I can merge this. You have figured this out by now, but grouping a PR by library is easiest with regards to avoiding broken branches. So you can pile multiple changes in the same library into one PR.

HubertD:
didn't expect that git wouldn't manage "simple" additions to the lib files - I thought that it would be easier to review one symbol after the other...
thanks for your patience - will definitely group PRs by libraries in the future :-)

PR #701, opened 2018-06-26
Bq51050b

jneiva08:
Datasheet: http://www.ti.com/lit/ds/symlink/bq51050b.pdf

Symbol:
bq51

PR for footpint is [#701](https://github.com/KiCad/kicad-footprints/pull/701)

------------


jneiva08:
I don't know if have made best modification on some pins.

The symbol looks like this:
bq5105xb

DanSGiesbrecht:
Thanks for the quick changes! Unfortunately, my apartment doesn't have internet right now, so I can't properly check this. I'll check tomorrow on my lunch break, and if everything looks good, merge.

evanshultz:
Pin 10 is named TERM, not THERM.

And a few observations:
- Since ILIM is an analog pin and doesn't have voltage thresholds that could cause net contention, we usually use Passive type so that the ERC engine can't give a false negative. Same with RECT.
- Since RECT goes to a cap to ground and FOD through a resistor, perhaps put it under FOD?
- Isn't TS/CTRL related to the battery and therefore best on the right side?
- Since ~CHG is connected through passive parts to BAT, maybe put it on the right side below BAT?
- Since smaller symbols give more flexibility to part placement and more usable schematic canvas, and this one is wildly unbalanced, all the "passive" pins (11, 12, 14, 18, etc.) could go on the lower right where there is room. We've done this is a few similar situations.

Sorry if some of this is contradictory with the above. Placing pins on left and right sides is generally good for organization if there's no other reason, but we often break it to group pins by function or bus them together (for example, MISO and MOSI lines of a SPI bus are next to each other even though one is in and the other is out). And again, it's just some quick observations from me so when Dan is happy to merge then that means it's finished.

DanSGiesbrecht:
Thanks, @evanshultz ! I agree with new pin placement recommendations. I was unsure between following KLC and usability of the part. @jneiva08 sorry for the extra work! Thanks for accommodating the requested changes.

evanshultz:
Nice!

Would be best to put ~AD-EN on the top left since the FETs it drives connect to AD?

DanSGiesbrecht:
> Since smaller symbols give more flexibility to part placement and more usable schematic canvas, and this one is wildly unbalanced, all the "passive" pins (**11**, 12, 14, 18, etc.) could go on the lower right where there is room. We've done this is a few similar situations.

Actually, I could see value in having `EN2` on the right side, because that one isn't passive. Sorry for not noticing that sooner! @evanshultz , do you agree?

jneiva08:
Moved back ~AD-EN to near AD and the symbol grow a little to be symmetric.

bq

DanSGiesbrecht:
Thanks, @jneiva08 . I'll review this ASAP, and merge if everything looks good.

DanSGiesbrecht:
Looks great, @jneiva08 
Thank you for your contribution!

PR #707, opened 2018-06-28
Add symbol for TJA1059TK dual CAN-FD transceiver

HubertD:
Datasheet link: https://www.nxp.com/docs/en/data-sheet/TJA1059.pdf
tja1059tk

------------


bradfa:
This looks good to me.  The pin naming for TXD and RXD is confusing to me but the symbol matches the datasheet where TXD is an input and RXD is an output.

Misca1234:
Hi

You foot print (and capsule) have a thrmal pad beneth and according to the datasheet both GND pin should be connected external, therefore yo ushould stack all these pins intop of each other, you do it like this

1. Move pin 2 to the center of the package and rename it to 'GND'
2. Make pin 5 invisible and of type "passive", renamed it to 'GND' and place it exactly on top of pin 2
3. make a third pin 15 (this will be the pin to the thermal pad) name it GND, make it invisible and place it on top of pin 2.

myfreescalewebpage:

myfreescalewebpage:
Hi @HubertD, thanks for contributing,

- [ ] "seperate vio" in the description is not correctly spelled and not very useful (many device have it), package name at the end has not the "-". I propose you the following description based on the datasheet: `Dual High-Speed CAN Transceiver with Standby mode, HVSON-14`
- [ ] Keywords are not enough, I suggest `High-Speed CAN Transceiver` like many other in the same library
- [ ] Footprint filter is not correct, there should be no leading star, no EP size, and moreover I do not agree to use the DFN footprint. It should be HVSON-14 footprint. Seems it does not exists, maybe a scripted version can be made @poeschlr ?
- [ ] The EP is missing and should be stacked with the two GND pins, as indicated by @Misca1234 

Cheers,
Joel

HubertD:
I mostly agree, but I'm not sure about the package.
Re-checked it and the DFN-14-1EP_3x4.5mm_P0.65mm seems to be exactly what I would want to have on my board. 
A newly created "HVSON14" would be a exact copy?

myfreescalewebpage:
Thanks @HubertD for the fixes!
Well, the datasheet indicates HVSON, so we must use this here, and you will have to submit it to the footprint repo.
If you think it is the same as the DFN package you have selected, make a copy. I'm not very experienced with the standard footprint rules, @poeschlr is the best person to answer this. Rene, if you read this message, can you tell us if the HVSON-14 and DFN-14 are the same ?
Thanks,
Joel

HubertD:
Hi @myfreescalewebpage 

I did some research, and I think you can safely merge this PR.
In current Interface_CAN_LIN.lib there are multiple other symbols using DFN packages when the NXP datasheet says HVSON:

- TJA1021TK (HVSON8 / "Package_DFN_QFN:DFN-8-1EP_3x3mm_P0.65mm_EP1.55x2.4mm")
- TJA1029TK (HVSON8 / "Package_DFN_QFN:DFN-8-1EP_3x3mm_P0.65mm_EP1.55x2.4mm")
- TJA1043TK (HVSON14 / "Package_DFN_QFN:DFN-14-1EP_3x4.5mm_P0.65mm_EP1.65x4.25mm")
- TJA1049TK (HVSON8 / "Package_DFN_QFN:DFN-8-1EP_3x3mm_P0.65mm_EP1.55x2.4mm")
- TJA1051TK-3 (HVSON8 / "Package_DFN_QFN:DFN-8-1EP_3x3mm_P0.65mm_EP1.55x2.4mm")
- TJA1145TK (HVSON14 / "Package_DFN_QFN:DFN-14-1EP_3x4.5mm_P0.65mm_EP1.65x4.25mm")

Some, but not all of these were my own PRs :-)
If this really needs to be changed, I'd suggest to do this in a separate step?

Cheers,
Hubert

Misca1234:
If you want to read up on the difference 
https://en.wikipedia.org/wiki/Quad_Flat_No-leads_package
and
HVSON is Thermal Enhanced Very Thin Small Outline Package No Leads

So the difference is more of the terminal properties of the capsule itself (they are both SON's)

If you want, I can make a scripted HVSON capsule (it is very easy)?

HubertD:
Frankly, I don't care if we have separate HVSON footprints in the library (I still don't see a good reason to).

But if so, we should remain some consistency along these nxp devices, e.g. change all symbols mentioned in my last comment. 

Not sure about the impact when changing the default footprints of devices that are in the lib for a long time.
If that's not a concern, I will gladly provide a PR to change all TJA*-TK footprints once we have HVSON packages in the footprint lib.

myfreescalewebpage:
Well, if other devices have been accepted with another footprint, do you remember the reason. I think it's not good for the global process in any case because now we wonder if a HVSON package should be created (this is my position to avoid future question on the definition of the device) or if another one can be used safely (I'm not able to answer personally and there is no KLC equivalence list...)

HubertD:
Small addition: If we need a specific package, it should probably be a VSON-14, since e.g. for the TJA1043 in HVSON-14, Texas Instruments is selling a Drop-In replacement in the very same package, calling it VSON-14: http://www.ti.com/lit/ds/symlink/tcan1043-q1.pdf

HubertD:
And one more: Microchip is selling CAN transceivers, e.g. the ATA6565 with the very same package.
They name it "DFN-14" on their webpage as well as VDFN-14 in the datasheet.

Same is true for the MCP2517FD SPI CAN Controller, which, again, is in the kicad symbol library as "DFN-14-1EP_3x4.5mm_P0.65mm_EP1.65x4.25mm".

Shall we really duplicate the perfectly matching footprint (packages having the same looks, the same height, the same EP size) just because every manufacturer desires to invent it's own name for it?

myfreescalewebpage:
@HubertD I agree with you but I just follow the datasheet of the submitted symbol. Only @poeschlr or @evanshultz I think have the answer.

myfreescalewebpage:
@HubertD the footprint question has just been answered by @evanshultz and confirmed by @poeschlr on an other subject here: https://github.com/KiCad/kicad-footprints/issues/1450#issuecomment-468728773.

You should create the exact footprint.

HubertD:
so then :-)

@Misca1234 I would really appreciate your help. 

AFAIK we will need at least:
- HVSON-8 (e.g. https://www.nxp.com/docs/en/data-sheet/TJA1049.pdf)
- HVSON-14 (eg, https://www.nxp.com/docs/en/data-sheet/TJA1043.pdf, https://www.nxp.com/docs/en/data-sheet/TJA1059.pdf)
- VSON-14 (http://www.ti.com/lit/ds/symlink/tcan1043-q1.pdf)
- VDFN-14 (e.g. http://ww1.microchip.com/downloads/en/DeviceDoc/20005782B.pdf, http://ww1.microchip.com/downloads/en/DeviceDoc/MCP2517FD-External-CAN-FD-Controller-with-SPI-Interface-20005688B.pdf)

@myfreescalewebpage:
With these in the footprint library, I would create pull requests to update existing symbols?

myfreescalewebpage:
@HubertD I suggest you can create one PR for each footprint, then a PR for the symbols to be updated.

myfreescalewebpage:
@HubertD ping !

evanshultz:
@HubertD 
I believe these footprints can be scripted so you can do four at once. Reviewing is easy in that case.

myfreescalewebpage:
No news from the original author, indicate the PR is abandoned.

PR #709, opened 2018-06-29
Adding VSC8541XMV-0x

ki5libs:
Single Port Gigabit Ethernet Copper PHY with GMII/RGMII/MII/RMII Interfaces
[Datasheet](https://www.mouser.com/ds/2/523/Microsemi_VSC8541-01_Datasheet_10496_V40-1148034.pdf)
vsc8541xmv-0x
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/723)
------------


antoniovazquezblanco:
@ki5libs There is a branch conflict here. Could you please resolve it for @DanSGiesbrecht to be able to review the PR?

ki5libs:
@antoniovazquezblanco sure, it's done.
@DanSGiesbrecht please review it.

DanSGiesbrecht:
Thanks, @ki5libs ! I'll take a look through this ASAP (today or tomorrow).

ki5libs:
Hi @DanSGiesbrecht 
Thx for comments, I've made changes requested by you

evanshultz:
This is all just for information and clarity. Nothing above needs to change. Sorry if this is all known already.

---

We can't omit any pins and expect CvPcb to work with the footprint filters and when selecting footprints based on pin count (which is does by default and is usually the way to go). KLC allows two options:
1. Include all NC pins in the symbol so that the exact number of symbol pins matches the exact number of footprint pins. The footprint filter does not include the footprint pin count.
1. Omit any NC (unused) pins and include the pin count in the footprint filter.

The latter option is taken only in rare cases where the number of NC pins is excessive and typically more than functional pins. I recall this coming up just a couple times and it should only be done with caution since it makes the footprint filter more fragile. With only two pins (65, 66) needing to be handled on this symbol, option 1 should be taken as stated above.

---

The thermal pad pin (pin 69 for this symbol) must be named the same as the pin upon which it's stacked if it is stacked. They should only be stacked if the datasheet is explicit about stacking. Section 6.2 states that the pad should be grounded.

In cases where the pad is floating or the documentation is missing or not clear, the naming isn't fixed (pin numbers are matched by CvPcb so it's not important for the footprint). Here, `EP` would be reasonable since the datasheet calls it "Exposed Pad". However, `TP`, `PAD`, `Pad`, and other variations seem to ring a bell.

---

Pin placement on left (input) and right (output) is in lieu of having better information. Keeping all pins of an SPI bus together is better than splitting MISO and MOSI since both of those will go between the same two devices and so schematic routing is easier if they're together.

Here, grouping the signals going to the jack and the processor (as a PHY sits between those two things) is a great choice. There is plenty of flexibility since we don't want to cover every single use case in KLC but just provide a nice framework of guidelines than can be adapted to the myriad of symbols in the library.

---

Carry on! :)

DanSGiesbrecht:
@evanshultz thank you for all that information! I'm learning every day, it's awesome.

A few responses:

---
With regards to *why* to leave NC pins on the sheet for filtering robustness, I didn't know that, that's good to know! The KLC seems a bit ambiguous when I was trying to figure out which was preferred, but I had a gut feeling that leaving them on (as you suggest) was the right path. Maybe in the future, we could add this to the KLC.

---

So for the thermal pad: this part is unusual, in that there are **no** other GND connections. I suggested `GND` or `VSS` rather than `EP` so that the designer knows that this is the primary (only) GND connection. Thank you for your notes on stacking and otherwise naming it though, it's good information to have.

---

Agreed: arranging pins for magnetics (jack)/processor is the right call.

---

@ki5libs thank you for all your work. I'm going over your changes now. I have a suggested change for the stacking pins (sorry, I may have lead you slightly astray). I'll post again shortly.

DanSGiesbrecht:
@ki5libs ,

All your changes look really good! I have a few follow-ups:

- Thank you for changing the power pin names and stacking them. I think that some of them should actually be left un-stacked (but sharing the name is good :+1: ).
  - The analog pins `VDD25A` and `VDD1A` have a note in the datasheet (page 94) for each pin: `requiring additional PCB power supply filtering`, which I think justifies **not** stacking them. The designer may want to use a series filtering component going to each pin (resistor, ferrite), so they shouldn't be tied together through stacking (as the nets wouldn't be the same anymore).
analog-pwr-filter

  - The rest are acceptable to stack in my opinion (as you've already done, nice :+1: ), agreeing with the KLC.
  - This is how I mocked it up:
pin-stacks
    - Broke out each VDD25A pin,
    - Broke out each VDD1A pin,
    - Left all other pins stacked (as you have them)
- One other suggestion that I have is that the symbol can be narrowed slightly. I tried it out to see what it would look like: I brought in the left and right sides by 200mil on each side, for a total width reduction of 400mil:
full-part
  - I arranged the power pins (top) so that it's adjusted with more pins on the left side than the right, so that there's more room for the `VSC8541XMV-0x` label (see upper screenshot on power pins).
- With regards to pin sides, you're totally right - left vs. right doesn't really matter, because the part can be mirrored.
  - I am wondering though (and this comes from my lack of knowledge, so bear with me): should the `CLK_SQUELCH_IN`, `CLK_OUT` and `FASTLINK_FAIL` be on the jack/magnetics side (as in the block diagram), or are they on the processor side because the signals actually go to the processor?

DanSGiesbrecht:
Once we work through that last post of changes/suggestions, I think this can be merged!

DanSGiesbrecht:
@ki5libs looks great! Thank you for you contribution :)

Issue #635, opened 2018-05-29
Inconsistency in AudioJack_Switch symbols


Ratfink:
I believe all the switched pins in our audio jack symbols are supposed to disconnect when a plug is inserted.  That's the way that AudioJack3_Switch is drawn:
![screenshot from 2018-05-29 11-54-17](https://user-images.githubusercontent.com/320472/40670588-b15d2fd8-6337-11e8-9f4a-4ffcf74894b2.png)
But AudioJack3_Ground_Switch is drawn depicting switches that would close when a plug is inserted:
![screenshot from 2018-05-29 11-54-13](https://user-images.githubusercontent.com/320472/40670622-bfe99532-6337-11e8-86b0-a5ca9284d38b.png)
I believe this is a bug in AudioJack3_Ground_Switch, so we really should fix it.  It might also be nice to rearrange the pins to match AudioJack3_Switch's ordering.

evanshultz:
When the symbols were merged at https://github.com/KiCad/kicad-symbols/pull/310#issuecomment-379877419 I was soliciting feedback on these symbols because I think they can be improved as well, although I hadn't brought up this specific item, but that PR didn't need to get any more changes. I drew the upper symbol you show and retained the lower one since it could have been included in legacy schematics that would break connections, but I'm still OK changing it. With the nice assortment of audio jack symbols now in the lib I support a bit of polish and consistency for these symbols.

Ratfink:
Sure could! 🙂

evanshultz:
PR submitted at https://github.com/KiCad/kicad-symbols/pull/714.

PR #714, opened 2018-06-30
AudioJack* symbol consistency

evanshultz:
This is to fix https://github.com/KiCad/kicad-symbols/issues/635

The pinout and style is consistent for all *AudioJack* symbols. I made the updated symbols smaller where possible.

Old (from https://github.com/KiCad/kicad-symbols/pull/310#issuecomment-379877419):
image

New:
image

------------


Ratfink:
The one odd thing I see is that AudioJack3_Switch has less space between the S wire and the top of the outline box than the other symbols.  If you'd like to fix that, I'll let you.  Since it isn't a real problem though, I suppose this could be merged without a change too.

evanshultz:
Good catch. Thanks! Fixed now.
image

Ratfink:
Looks great, thanks for fixing these symbols!

PR #715, opened 2018-06-30
Typo in the name of the Power symbols / flags library

f6fvy:
Unless there is a legitimate reason, I suggest the name of the power symbols library should start with a capital P (despite the file name doesn't).

### DIFF
--- a/sym-lib-table
+++ b/sym-lib-table
@@ -176,7 +176,7 @@
   (lib (name Triac_Thyristor)(type Legacy)(uri ${KICAD_SYMBOL_DIR}/Triac_Thyristor.lib)(options "")(descr "TRIAC and thyristor symbols"))
   (lib (name Valve)(type Legacy)(uri ${KICAD_SYMBOL_DIR}/Valve.lib)(options "")(descr "Valve symbols"))
   (lib (name Video)(type Legacy)(uri ${KICAD_SYMBOL_DIR}/Video.lib)(options "")(descr "Video symbols"))
-  (lib (name power)(type Legacy)(uri ${KICAD_SYMBOL_DIR}/power.lib)(options "")(descr "Power symbols, special power flags"))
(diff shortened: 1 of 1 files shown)


f6fvy:
I see the library name and the file name must be in-sync. So, I guess the lower-cased p in the filename is on purpose, and the lib name can't be modified.

evanshultz:
@poeschlr 
Can you confirm this cannot be done?

Ratfink:
There's no technical reason that the name of the power symbol library has to be all lower-case, but renaming it would break the 5.0 freeze.

evanshultz:
@f6fvy 
I believe the CI error is because sym-lib-table lists `power.lib` but the actual file name now starts with an uppercase letter. If the file name and library nickname are in sync it should be OK. (Not sure why this is, and it may just be in the check scripts and not a real problem in KiCad, but I believe that's the issue.)

@Ratfink 
This shouldn't matter. Libraries can be renamed, it just means that changes may not be picked up. In this case, however, it's a rename and all relevant symbols are included either way (if the above is fixed). So any changes would be behind the scenes.

@poeschlr 
I finally found the discussion I was thinking about above but couldn't find earlier: https://github.com/KiCad/kicad-library/issues/1402#issuecomment-317230294. Since this repo is now only for v5 we are OK to rename if we like. And I like, since it's best for consistency.

The same applies for `pspice.lib`. However, as I mentioned at https://github.com/KiCad/kicad-symbols/issues/451#issuecomment-378701973, I think that one should be an actual rename to `spice.lib`.

Ratfink:
@evanshultz According to https://github.com/KiCad/kicad-symbols/issues/451#issuecomment-390485782, libraries can't be renamed during the freeze.

Ratfink:
Since this has to wait for 6.0, which will have a new symbol library format, this PR won't ever be something we can merge.  As such, I'm closing it.

PR #718, opened 2018-07-01
Add symbols for Microchip SAMD[09|10|11] devices

HubertD:
oh wait, there's more :-)

### SAMD09
http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-42414-SAM-D09_Datasheet.pdf

samd09c13a-ss
samd09d14a-m

### SAMD10
http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-42242-SAM-D10_Datasheet.pdf
PR for WLCSP-20 footprint: https://github.com/KiCad/kicad-footprints/pull/715

samd10d13a-ss
samd10d14a-u

### SAMD11
http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-42363-SAM-D11_Datasheet.pdf
(no symbols, just aliases from SAMD10 devices)

------------


HubertD:
anything blocking this from being merged?

MCU_Microchip_SAMD.lib
```diff
@@ -0,0 +1,146 @@
...
+#encoding utf-8
+#
+# SAMD09C13A-SS
+#
+DEF SAMD09C13A-SS U 0 20 Y Y 1 F N
+F0 "U" -400 650 50 H V L CNN
+F1 "SAMD09C13A-SS" 50 650 50 H V L CNN
+F2 "Package_SO:SOIC-14_3.9x8.7mm_P1.27mm" 0 -1050 50 H I C CNN
+F3 "" 0 -700 50 H I C CNN
+ALIAS SAMD10C13A-SS SAMD10C14A-SS SAMD11C14A-SS
```
zgyarmati: Just my 2 cents: as there are some significant differences between the SAMD{09|10|11} family's pin functions (for example SAMD11 has USB pins), i think it would be nice to have these indicated on the symbols as well, rather than simply have aliases to the SAMD09 series

MCU_Microchip_SAMD.dcm
```diff
@@ -0,0 +1,81 @@
...
+$ENDCMP
+#
+$CMP SAMD10D14A-U
+D ARM Cortex-M0+ MCU, 48MHz, 16KB Flash, 4KB RAM, 1.6-3.6V, 18 GPIO, WLCSP-20
+K ARM Cortex-M0+ SAMD10
+F http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-42242-SAM-D10_Datasheet.pdf
+$ENDCMP
+#
+$CMP SAMD11C14A-SS
+D ARM Cortex-M0+ MCU, 48MHz, 16KB Flash, 4KB RAM, 1.6-3.6V, 18 GPIO, SOIC-20
```
zgyarmati: Small hiccup: this a SOIC-14 package, not SOIC-20.

HubertD: thank you for pointing that out, fixed.

MCU_Microchip_SAMD.lib
HubertD: guess you are right here. will try to find some time to figure the differences out and use different symbols.

myfreescalewebpage:
Hi @HubertD thanks for this contribution,

A few comment during my review:

For all devices:
- [ ] Keywords a little bit redundant with description. Based on other SAM devices already in the lib, I propose you `32-bit ARM Cortex-M0+ MCU Microcontroller`

SAMD09D14A-M:
- [ ] We do not include EP size if the footprint filter. It should be `QFN*1EP*4x4mm*P0.5mm*`

SAMD11C14A-SS:
- [ ] Package indicated in the Description is not correct, should be SOIC-14

Also, can you resolve the branch conflict. SAMD family has been added since you proposed this PR, so the sym-lib-table file will no more to be updated.

Cheers,
Joel

bradfa:
Although I do like the way the reset, crystal, and SWD pins are broken out in these SAMD9/10/11 parts as it's handy and convenient to see, the already existing SAMD21 symbol in the library does not do this.  The existing SAMD21 symbol puts all of the A port pins on the left and B port pins on the right of the symbol like:

image

I don't necessarily think this is better or worse but I think we should have all SAMD parts with a similar convention.

What do you think?

myfreescalewebpage:
@bradfa do you mean all pin should be on the left ? No very convenient when they are small. The parts of the PR have no port B ....

myfreescalewebpage:
@bradfa your examples looks like the design of the symbol of the PR... STart top right corner then clockwise around the symbol... I do not understand your proposition.

bradfa:
> @bradfa your examples looks like the design of the symbol of the PR... STart top right corner then clockwise around the symbol... I do not understand your proposition.

I'm not really making a proposition, my goal was more to just point out that the PR here doesn't follow a similar convention to the existing SAMD part in the library.  The STM32 parts all seem to follow a similar convention and it seems nice to have a common convention.

I'm not trying to say this PR or the existing SAMD21 part in the library is wrong, just that they're different styles and maybe that's not the best.

myfreescalewebpage:
Ok, probably you refer to ATSAMD21G18A-AUT ? Probably it's better to follow STM examples I think and to correct the existing SAMD when it will be possible...

bradfa:
> Ok, probably you refer to ATSAMD21G18A-AUT ? Probably it's better to follow STM examples I think and to correct the existing SAMD when it will be possible...

@myfreescalewebpage yes, that's the part I'm referring to.  Sorry I didn't put the part number in my previous replies.

When would it be "possible" to fix the existing part that's already in the library?  Is this something that would have to wait for a KiCad v5 to v6 change or just has to wait for someone to actually take the time to make the change?

I do like the idea of copying what was done for STM32 as close as possible.  The STM32 parts all look clean and nice and easy to understand in the library.  SAMD can have up to 32 bits on a port while it seems most STM32 only ever have up to 16 bits on a port.

So would it be reasonable for SAMD parts which only have a port A to split the port into the lower 16 go on the right, upper 16 go on the left and then if there's more than just port A that A starts top right and the ports increase clock-wise like STM32?

poeschlr:
You can do small fixes to symbols right now. Especially if there is a bug. Examples of things that can be fixed right now:
- Symbol names if the current name could lead to a confusion of parts. (In that case further changes are possible as user schematics are already broken by the renaming.)
- Pins that have the wrong name, number or electrical type.
- Wrong or missing footprint assignments (This could mean that the name would also need to change. See point 1)

---


If the change is only to get them to a certain style (KLC) and it requires moving around pins then i would suggest to wait till shortly before version 6 as doing such things will break existing schematics with little to no benefit to the user of these symbols.
So what changes are therefore not allowed right now:
- moving the symbol around just to make travis shut up about the center requirement.
- moving pins around just to fit a particular look (exception: not connected pins.)

In short anything that would trigger in this check: https://github.com/KiCad/kicad-library-utils/pull/258

myfreescalewebpage:
Hi @HubertD are you still alive on this subject ? Thanks.

HubertD:
Sorry @myfreescalewebpage - not really.
Will gladly try to solve merge conflicts, join GND pins and fix descriptions.
But right now, I don't have the energy to discuss a global MCU pin layout convention (being obviously a fan of at least seperating JTAG / Reset / Oscillator pins from the rest), nor to do research regarding SAMD10/11 special pins.
Maybe I should remove the aliases to SAMD10 etc to allow these devices being implemented seperatly?

myfreescalewebpage:
@HubertD no don't worry about the more general discussion above. Just refer to my comment at https://github.com/KiCad/kicad-symbols/pull/718#issuecomment-451759828 and this is ok for me to merge your contribution !

Thanks for your time
Joel

HubertD:
@myfreescalewebpage did the proposed changes.
still offering to remove aliases to allow for better implementations in the future

myfreescalewebpage:
The implementation is great according to me.
You have added a new device (ATSAMD21G18A-AUT) I'm currently making the review.
Just realized the other SAMD of your PR are named SAMS instead of ATSAMD... Can you update it ?
Thanks
Joel

myfreescalewebpage:
Thanks @HubertD no more comment, merging also this PR !

Issue #510, opened 2018-04-20
Issues with Logic_LevelTranslator.lib


evanshultz:
From https://github.com/KiCad/kicad-symbols/pull/461

```
Checking symbol 'TXB0104RGY':
  Violating S3.6
    Pin name position offset
    Pin offset not preferred value
     - Pin offset (40) should be 20mils unless required by symbol geometry
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'Texas*S*PVQFN*N14*ThermalVias*' does not match the footprint 'Package_DFN_QFN:Texas_S-PVQFN-N14' set for this symbol.
     - could not match 'Texas_S-PVQFN-N14' against filter 'Texas*S*PVQFN*N14*ThermalVias*'
     - could not match 'Package_DFN_QFN:Texas_S-PVQFN-N14' against filter 'Texas*S*PVQFN*N14*ThermalVias*'
Changed symbol 'Logic_LevelTranslator.lib:TXB0102YZP'
Checking symbol 'TXB0102YZP':
  Violating S3.6
    Pin name position offset
    Pin offset not preferred value
     - Pin offset (40) should be 20mils unless required by symbol geometry
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'DSBGA?8*P0.5mm*' does not match the footprint 'Package_BGA:Texas_DSBGA-8_0.9x1.9mm_Layout2x4_P0.5mm' set for this symbol.
     - could not match 'Texas_DSBGA-8_0.9x1.9mm_Layout2x4_P0.5mm' against filter 'DSBGA?8*P0.5mm*'
     - could not match 'Package_BGA:Texas_DSBGA-8_0.9x1.9mm_Layout2x4_P0.5mm' against filter 'DSBGA?8*P0.5mm*'
Changed symbol 'Logic_LevelTranslator.lib:TXB0104PW'
Checking symbol 'TXB0104PW':
  Violating S3.6
    Pin name position offset
    Pin offset not preferred value
     - Pin offset (40) should be 20mils unless required by symbol geometry
Changed symbol 'Logic_LevelTranslator.lib:TXB0104ZXU'
Checking symbol 'TXB0104ZXU':
  Violating S3.6
    Pin name position offset
    Pin offset not preferred value
     - Pin offset (40) should be 20mils unless required by symbol geometry
  Violating S5.1
[...]

PR #720, opened 2018-07-01
Fix Issue #510: Improve logic level translator lib

HubertD:
fix travis warnings in Logic_LevelTranslator.lib and unify positions of reference, component name, footprint fields

transl01
transl02
transl03
transl04
transl05
transl06
transl07
transl08
transl11
transl12
transl13
transl14
txb0102dct
txb0102dcu

------------


evanshultz:
This is a lot of work, and the library now passes all the checks. Thanks for taking the initiative!
- I would prefer to see the footprint closer to the part body. It's no to avoid overlapping elements in the v5 symbol editor, but the footprint field isn't included in a "zoom all" command and it's easy to miss it if it falls out of the viewable window.
- Is there a significance of having filled versus unfilled triangles? If not, and they were just selected by individual users as they desired, this should be standardized. I don't see this type of schematic element in IEEE 315, but I vote for hollow. Does anybody else want to submit their opinion. (FYI, this doesn't have to be resolved to merge this PR.)

evanshultz:
Thanks!

PR #724, opened 2018-07-03
Add Symbol for ESP32-WROOM-32 RF module

HubertD:
Cannot believe this is not already part of the library :-)

Datasheet: https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32_datasheet_en.pdf

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/722

sym-esp32-wroom-32

------------


HubertD:
esp32-wroom-32

diegoherranz:
I've checked the pinout and it looks correct.
A minor comment: I would expect SENSOR_VP on top of SENSOR_VN, positive one first (and lower pin number too).

Thanks for submitting this.

diegoherranz:
Thanks for the update. It looks good to me.

diegoherranz:
While reviewing https://github.com/KiCad/kicad-footprints/pull/767, I've noticed that ESP32-WROOM-32D is the same as ESP32-WROOM-32 symbol-wise.
Could you make the ESP32-WROOM-32D an alias of ESP32-WROOM-32?

Thanks!

diegoherranz:
Hi, @HubertD 

Thanks for creating the alias.
The ESP32-WROOM-32U, however, despite having the same symbol, needs a different footprint so I believe that it requires an independent symbol so that it can be associated with RF_Module:ESP32-WROOM-32U.

So, can you either add another symbol for the 32U or simply remove it keeping the 32 and the 32D?
If it's the latter I can create a PR to add the 32U.

Many thanks!

HubertD:
i removed the ESP32-WROOM-32U alias again because it needs a different footprint (no keep-out zone necessary).
@diegoherranz  would be glad if you took the job to add a new symbol for ESP32-WROOM-32U since I don't have much time to spare in the next two weeks

diegoherranz:
Looks good to me. As soon as this gets merged, I'll submit a pull request with the ESP32-WROOM-32U.
Many thanks!

diegoherranz:
Is everyone happy with this one? When it gets merged I will submit the ESP32-WROOM-32U (different footprint). Thanks!

HubertD:
any news?

chertov:
Merge please! I'm waiting ESP32-WROOM in library
Thanks for the symbol!

DanSGiesbrecht:
I will revive this review ASAP. Thank you for your patience.

DanSGiesbrecht:
@HubertD looks great! Thank you again for contributing, and for your patience.

PR #725, opened 2018-07-04
Adding TPD6F003

antonlysak:
Adding TPD6F003 - Six-Channel EMI filter with ESD protection
[Datasheet](http://www.ti.com/lit/ds/symlink/tpd6f003.pdf)
tpd6f003
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/724)

------------


SchrodingersGat:
I don't think you should have the footprint filter for the WSON16 device - this seems to only apply to the TPD8F003 (8 channel).

So:

1. Remove the extra footprint filter
1. As there is only a single possible footprint, assign the default footprint field.

ki5libs:
Yes, you are right. I removed extra footprint and set default. There is an error now because [footprint PR](https://github.com/KiCad/kicad-footprints/pull/724) is not still merged...

poeschlr:
The footprint filter should not include the pin count (This is done by a special kicad internal filter)
Also use * instead of ?, this makes the filter more flexible if users want to use their own footprints but introduce "parameters" somewhere in between.

So it should be: `WSON*1EP*1.35x2.5mm*P0.4mm*`

ki5libs:
@poeschlr I have looked at the footprint one more time and found that it was renamed to `Texas_R-PWSON-N12_EP0.4x2mm`.  So I did it here and change filter to `WSON*EP0.4x2mm*`

poeschlr:
In that case the filter needs a * at the beginning as well.

ki5libs:
Sure, done.

PR #727, opened 2018-07-04
added WemosD1Mini after accidentally deleting it

Duckle29:
Added a symbol for the WeMos D1 Mini

------------
datasheet https://wiki.wemos.cc/products:d1:d1_mini

matching footprint PR https://github.com/KiCad/kicad-footprints/pull/726


Misca1234:
Hi

The v in 5v should be capitalized
I guess pin 16 should be output ?

antoniovazquezblanco:
@Misca1234 given that you started the review will you take this one? :)

Misca1234:
Thx, great, then this one will be merged when the foot print is merged

Misca1234:
@Duckle29 

The foot print was just merged, however, there is a conflict now for the symbol, 
I can merge when it have been resolved.

myfreescalewebpage:
@Duckle29 also I see that pin 8 (3V3) should be Power Output instead of Power Input.

Duckle29:
@myfreescalewebpage the 3.3V pin is the power input to the wemos, but the power output of the onboard LDO. You can power the board through the 3.3v pin, as I do in a few projects. Would output still be the preferred pin function for that? 

I agree that it should be U instead of A and have changed that.

Duckle29:
@Misca1234 I resolved the merge :)

myfreescalewebpage:
@Duckle29 I'm not sure of the use case you are doing. According to the schematic, the 3V3 pin is the output  of the internal regulator: https://wiki.wemos.cc/_media/products:d1:sch_d1_mini_v3.0.0.pdf

This is typically where the ERC check should record an error in Kicad to alert you it is possible you are making a mistake. Powering a regulator from the output with no input supply is a bad electrical idea!

In my opinion, 3V3 and 5V pin should also be crossed, 5V to the left of 3V3, as done on any regulator (input to the left, output to the right).

myfreescalewebpage:
@Duckle29 you should also update the footprint in this symbol to match the footprint PR which has been merged (see also travis error).

Duckle29:
@myfreescalewebpage I'm aware that the 3v3 leg is connected to the 3v3 of the internal LDO, but that 3v3 rail powers the entirety of the board, except for the USB-UART bridge, which would be powered from USB if one was plugged in.

I have marked it as an output, as I suppose "powered from 3.3v and plugged into USB" is something a PCB designer would have to consider though.

myfreescalewebpage:
@Duckle29 last error to be fixed (see also travis result):
- [ ]Footprint filter should be `WEMOS*D1*mini*`

Cheers,
Joel

Duckle29:
Done :)

myfreescalewebpage:
Thanks, just need to wait the footprint now.

Duckle29:
@myfreescalewebpage The footprint is already merged :)

myfreescalewebpage:
oups sorry you are right! So let(s go in this case :)
Thanks.

PR #729, opened 2018-07-04
Add PocketBeagle

twam:
Datasheet: https://github.com/beagleboard/pocketbeagle/wiki/System-Reference-Manual#70_Connectors_
Images:
screen shot 2018-07-04 at 20 32 02
screen shot 2018-07-04 at 20 32 10

------------


poeschlr:
Connector might not really be the correct repo for this. Might this better fit into the modules lib?

myfreescalewebpage:
Hi @twam 

Same comment already made by Rene, can you move this to "MCU_Module" library ?
Moreover, you have made 2 parts, but as single one is required, this is two headers to be connected to the same module. You can make a single unit with all pins, or two units in the same symbol.

Cheers,
Joel

twam:
If I make a single part, then it is not possible any more to use the usual connector footprints. Or do I miss something?

myfreescalewebpage:
No, this will require to create a specific footprint too, with all pins. This has the advantage to grant the distance between the two connectors, and also to create a silkscreen.

I don't know the position of @poeschlr but I don't think it is good to have the board split in two different symbols...

Misca1234:
Agree, this should be one symbol to match a dedicated foot print as Joel points out
Perhaps you should add the prefix P1_ and P2_ to the 3.3V and %V names so the designer knows what connector they are connected to the other pins seems to have unique names.

I guess the data sheet link should be 
https://github.com/beagleboard/pocketbeagle/blob/master/PocketBeagle_sch.pdf

@twam 
Do you want me to make the foot print or do you want to do it?

twam:
I can adjust the file and/or create the footprint. However it will take ~4 weeks due to other obligations.

Misca1234:
Foot print push is here 
https://github.com/KiCad/kicad-footprints/pull/1350

myfreescalewebpage:
Thanks @Misca1234 to help this PR :)

twam:
Finally managed to update the footprint. I hope I have tackled all wishes :)

Screenshot 2019-03-09 at 17 53 34

myfreescalewebpage:
@twam I was waiting for update following footprint proposition, I have not performed the review at the moment, but will do asap (today).

myfreescalewebpage:
@twam here are my comments:
- [ ] The correct name for this symbol should be `PocketBeagle`
- [ ] Do you have a better documentation link for this, a datasheet, not a schematic ?
- [ ] Spelling error in keywords : "beagleboard"
- [ ] Description is too long and not detailed, can you short this, add some kind of information about quantity of RAM/FLASH etc ? See for example how the other module symbol are described.
- [ ] "AIN.VREF+" is positive power supply and should be moved to the top of the symbol
- [ ] "AIN.VREF-" is negative power supply and should be moved to the bottom of the symbol
- [ ] "3.3V" is positive power supply and should be moved to the top of the symbol
- [ ] "VOUT" is positive power supply and should be moved to the top of the symbol
- [ ] Name of pin 46 is "GPIO52"
- [ ] AIN pins are Input
- [ ] RESET is Input
- [ ] PWR.BTN is Input
- [ ] BAT.VIN is Power Input and should be moved to the top of the symbol
- [ ] BAT.TEMP should be passive to avoid ERC errors
- [ ] I wonder about the USB1 pin types, ID and VBUS are probably Input, and DRVVBUS is probably Output isn't it ?

Better datasheet link will be interesting to fix the above comments and have a clear overview of the device.

Cheers,
Joel

twam:
I haven't found a (PDF) datasheet. I also like to use the https://github.com/beagleboard/pocketbeagle/wiki/System-Reference-Manual but declared the Schematic as a datasheet as this was suggested by @Misca1234

myfreescalewebpage:
I think the manual https://github.com/beagleboard/pocketbeagle/wiki/System-Reference-Manual is better in this case, it contain much more informations !

myfreescalewebpage:
My last comment:

- [ ] Description: please remove extra spaces between values and units (GHz and MB)

Else, great job!

Thanks,
Joel

myfreescalewebpage:
Footprint merged, just need the last fix in my previous comment to merge this symbol.

twam:
There you go. Thanks for the thoughtful review!

myfreescalewebpage:
@twam please do not force push, this breaks the commit history and make the review more difficult.
Else, this is OK for me, merging.

PR #733, opened 2018-07-05
Adding TPS51363

antonlysak:
Adding TPS51363
[Datasheet](http://www.ti.com/lit/ds/symlink/tps51363.pdf)
tps51363
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/730)

------------


jneiva08:
The VIN, SW and PGND should be stacked (passive and invisible),
The NC could be on edge of square,
The symbol can shrunk a little?

antonlysak:
Reworked

Misca1234:
Do you think EP, PGND and GND all be stacked ?

You should change the description 
from
High-voltage input, synchronous converter with integrated FET
to
High-voltage input, synchronous converter with integrated FET, QFN28

I can not match your foot print filter to a known foot print,
is it missing ?

ki5libs:
@Misca1234 
I've added default footprint, changed footprint filter and description.

According to KLC S4.3 stacked pins should have the same name. Pins `EP`, `PGND` and `GND` have different names and separate description in the datasheet (though they should be connected together on the schematic) so I think it should not be stacked. It's common practice isn't it?

Misca1234:
> High-voltage input, synchronous converter with integrated FET, QFN28
Sorry, it should be QFN-28

>  stacked pins should have the same name. Pins EP, PGND and GND have different names and separate > description in the datasheet (though they should be connected together on the schematic) 

In case of stacking, they got all the same name even if they 'are an EP'

It was a suggestion to stack them, but you think it is an bad idea?

ki5libs:
@Misca1234 
I've fixed the package type in description.

I've looked through different symbols in the library and found that `EP` is often made as a separate pin despite to mandatory connection to `GND`. As I already said datasheet separates these pins in the table **PIN FUNCTIONS** giving them different names. I'm not sure that they internally connected together (it doesn't show on block diagram). We could change pin names and stack them, but it would be a small falsification of datasheet information. Is there some clear rule or criteria when we can do this? I couldn't find it in the KLC.
func

I offer leave it as it is. I suppose that connection of these pins could be left up to schematic designer. Please, correct me if I make a mistake.

PR #742, opened 2018-07-08
Add ESP32-PICO-D4 system on chip

HubertD:
datasheet link: https://www.espressif.com/sites/default/files/documentation/esp32-pico-d4_datasheet_en.pdf
esp32-pico-d4

------------


herostrat:
Hey Hubert,

one question:
Is there a reason you did not use the standard pin names and in some cases used additional function names? 
Is it just for your projects usecase?
Because afaik the names of the pins should the pinname in 2.2 of the datasheet (but to be sure we should wait for a maintainer to confirm this).

Otherwise the grouping seems to make sense, even if the pins get named without the additional function name

HubertD:
Hi @herostrat,

afaik, I did use the standard pin names from the datasheet - please correct me if I made mistakes there.

But I did add additional names for pins that are used inside the System-On-Chip package (SPI flash), the JTAG pins and other ones that shouldn't be used without care, because they have special functions on boot. Not so sure about the SD2 and SD3 functions right now, maybe these labels should go away?

herostrat:
> afaik, I did use the standard pin names from the datasheet - please correct me if I made mistakes there.

I meant the usage of the additional names (from the function colomn) and not only the name (colomn 1 from table 2).

But I think the arguments are solid for including the special function names in the symbol.

HubertD:
any news here?

poeschlr:
What you call flash_sd1 and flash_sd3 are called sd0 and sd1, flash_sd2 is called cmd in the datasheet.
Where did you get your namings from?

All these pins are mentioned as "do not use for anything else then flash interface" inside the datasheet so it is definitely ok to add these names to the pins.

By the way there is a merge required as another rf module has been merged. (I can do the manual merge if you fell unconfident about that.)

HubertD:
> What you call flash_sd1 and flash_sd3 are called sd0 and sd1, flash_sd2 is called cmd in the datasheet.
> Where did you get your namings from?
Cannot imagine from anywhere else than the original datasheet. 
In currently available PDFs these names are not there any more, but googling for 'esp32 pico "flash_sd1"' does find some cached documents where these are still present.
Will fix.

HubertD:
esp32-pico-d4

diegoherranz:
The symbol for this part was added on https://github.com/KiCad/kicad-symbols/pull/2082. I didn't see this PR when I reviewed that one.
I think this can be closed now.

Thanks!

poeschlr:
@diegoherranz it was also partly my mistake. The label for "ready for review" was different in our repos (the symbol lib has had a space in it while the footprint lib did not) which meant i did not see this pull request when i filtered for the ones that i need to take a second look at.

diegoherranz:
Is it OK to close this one, then?
Thanks.

poeschlr:
@HubertD we are truly sorry that your contribution was missed and made redundant. As you can see above steps were taken to avoid the same problem in the future.

PR #746, opened 2018-07-09
Adding TPS51200DRC

ki5libs:
Adding TPS51200DRC
[Datasheet](http://www.ti.com/lit/ds/symlink/tps51200.pdf)
tps51200

------------


Misca1234:
My kryllic is not what is what but you should change the description 
from
Sink and Source DDR Termination Regulator
to
Sink and Source DDR Termination Regulator, VSON-10

Place pin VLDOIN, pin 2, on top of the symbol

May I also suggest you swap pin 9 and 5

please check if the foot print
Package_SON:VSON-10-1EP_3x3mm_P0.5mm_EP1.65x2.4mm
is the one you need for this, if so, specify as the dedicated one.

If you look at page 10, you see on the schematic indicate that pin 4 and 8 is connected internally, 
Also, on page 26, chapter 10.1 it is indicated that pin 4,8, and 11 should be connected together
this mean that they should be stacked.
That is, name Pin 4 to GND, make it invisible and set the pin type to passive and place it ontop of pin 8
That is, name Pin 11 to GND, make it invisible and set the pin type to passive and place it ontop of pin 8

ki5libs:
You are right about the footprint. I've checked and set it.
Everything else should be also OK now.
And sorry for inconvenience with cyrillic.

SchrodingersGat:
Thanks @ki5libs for the contribution :)

PR #748, opened 2018-07-09
NB-IoT SIM7020

jneiva08:
The symbol was created from symbol SIM800C and footprint is the same.

Datasheet: https://simcom.ee/documents/SIM7020/SIM7020%20Hardware%20Design_V1.01.pdf

Symbol:
nb-iot

------------


myfreescalewebpage:
Hi @jneiva08, thanks for contributing,

- [ ] I have not found a datasheet instead of the hardware design guide, have you got a link ?
- [ ] According to the hardware design doc, pin 25 should be called USB_DN instead of USB_DM
- [ ] I will say pin ANT is Bidirectional.

Cheers,
Joel

jneiva08:
It is the only datasheet can be found.
In the figure 2 of hardware design, you can found USB_DM and the table 3 you can find USB_DN. SIMCOM have made the same inconsistency on [SIM800C](https://github.com/KiCad/kicad-symbols/pull/634), in that link can find why ANT is passive and not Bidirectional.

myfreescalewebpage:
Hi @jneiva08 

> It is the only datasheet can be found.

Me too.

> In the figure 2 of hardware design, you can found USB_DM and the table 3 you can find USB_DN. SIMCOM have made the same inconsistency on SIM800C

Good point, that's good to keep library consistent so USB_DM accepted.

> in that link can find why ANT is passive and not Bidirectional.

Right. I do not know what is the Kicad strategy about that, if you look at several symbol in the lib, you can find Passive, Bidirectionnal, Input and Output antennas. According to me Passive is only for NC pins. @antoniovazquezblanco what is your advice ?

Cheers,
Joel

antoniovazquezblanco:
Fixed merge conflicts caused by other PRs merge.

> Right. I do not know what is the Kicad strategy about that, if you look at several symbol in the lib, you can find Passive, Bidirectionnal, Input and Output antennas. According to me Passive is only for NC pins. @antoniovazquezblanco what is your advice ?

We need help from one of the more experienced librarians here. @poeschlr, @evanshultz ?

Thanks!

jneiva08:
This answer of evanshultz from other gsm modem PR can help?

https://github.com/KiCad/kicad-symbols/pull/634#issuecomment-400774183

poeschlr:
NC (not connected) pins should not have type passive but type not connected.

I think antennas can be seen as input, output or even bidirectional. (depending on if your part is a receiver, transmitter or both.) The antenna it self however would be passive i think. Unless it is an active antenna.

myfreescalewebpage:
@jneiva08 I think the reason is not valid, we do not only use passive components on the ANT pins. Sometimes there are some amplifiers, RF switches, etc etc and of course there are active antennas.

I prefer the response of @poeschlr . Typically GPS ANT is input (can eb discussed because it also power the antenna itself throw the RF line), 433MHz transmitter ANT is output, GSM / LPWAN etc etc are Bidirectional. Some devices can transmit lots of power on output ANT.

@poeschlr  which good use of Passive pin ? If you have an example ?

Cheers,
Joel

jneiva08:
I'm guessing the other reason to be passive, is because you normally need to add a matching impedance circuit, when you are making ERC validation it do now give a error/warning.

poeschlr:
@myfreescalewebpage 
I would say passive is for anything that does not really fit any of the other options. (sometimes also used for workarounds)

@jneiva08 
Passive does not give you an error if you connect it to an input, output or bidirectional pin. So there is no danger in having false positives here.

myfreescalewebpage:
@poeschlr since there is no clear KLC rules about antennas, I suggest @jneiva08 can keep its proposition to set it as Passive. When there is no guideline, submitter is the responsible of the commit, so he should keep the last word. Is it okay for you ? In any case this can be easily modified later if KLC are created on this subject.

poeschlr:
We are working on a better way to give users guidance regarding pin types: https://github.com/KiCad/kicad-website/issues/342

myfreescalewebpage:
Any librarian, this PR is only blocked because of ANT pin type question. What should we do ?

myfreescalewebpage:
@poeschlr your position maybe ?

jneiva08:
thanks for all your time

PR #762, opened 2018-07-17
Corrected footprint for Atmega-48-MHN and Atiny-48-MHN

poeschlr:
It seems the original footprint was taken from a wrong document. It not
only had the wrong center pad size but also the wrong body size and pin
pitch.

Datasheets:
 - http://ww1.microchip.com/downloads/en/DeviceDoc/ATmega48A_88A_168A-Data-Sheet-40002007A.pdf page 479
- http://ww1.microchip.com/downloads/en/DeviceDoc/8008S.pdf page 16

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/752

Edit: i originally looked at the wrong page on the datasheet.


fauxpark:
Once again, #525 is still outstanding... ;)

poeschlr:
Replaced by https://github.com/KiCad/kicad-symbols/pull/769

PR #766, opened 2018-07-17
Add Inolux IN-PI554FCH and IN-PI556FCH RGB addressable LEDs

diegoherranz:
They are similar to WS2812B and WS2812 respectively and the
symbols have been created using those as starting point.

New footprint for IN-PI554FCH at https://github.com/KiCad/kicad-footprints/pull/756.
IN-PI556FCH uses existing WS2812 footprint.

IN-PI554FCH
in-pi554fch_symbol

IN-PI556FCH
in-pi556fch_symbol

------------


diegoherranz:
Travis fails on:

- `Pin offset outside allowed range. Pin offset (10) should not be below 20mils.` I didn't modify that from the original symbols. I've tried with 20 mils and it looks less tidy in my opinion (see screenshot).
bigger_offset
- `Specified footprint does not exist`: KiCad/kicad-footprints#756
- `Footprint filters should match all appropriate footprints. Minuses and underscores in footprint filter 'LED*IN-PI554FCH*PLCC*5.0x5.0mm*P3.2mm*' should be escaped with '?' or '*'. ` Even if the "minus" is part of the part name? e.g. IN-PI554FCH

diegoherranz:
Gentle reminder. Did anyone have a chance of looking at this?

antoniovazquezblanco:
It seems that nobody has started reviewing thing yet. Please be patient as librarians have a ton of pending work as you can see. Thank you :)

poeschlr:
In this case the minus in the footprint filter is allowed (travis will always complain about that. This is something that can not be checked automatically)

thanks for your contribution. all is well.

PR #747, opened 2018-07-09
Add pin names to opamps

evanshultz:
Continuation of https://github.com/KiCad/kicad-symbols/pull/627, adding pin names where missing. In general, the goal is to make pin names visible where possible using pin text size down to 20mil.

Image of a smattering of updated symbols:
image

More detailed breakdown of changes:
Amplifier_Audio:
- General: Change pin name offset to 10mil
- General: Name all unnamed pins
- OPA1622: change inverted pin type to Line type
- STK43x_Sanyo: cannot find a datasheet with pin names so some are left blank
- TDA7496: moved pin 7 to unit B to fit all pins with names

Amplifier_Buffer:
- EL2001: make pin names visible, add CN suffix to indicate DIP package
- LH0002H: fix pin 7 name, move pins 3 and 5 vertical, updated footprint filter

Amplifier_Current:
- INA28x: shift power pins to normal location, move REF* pins to the bottom, make pin names visible

Amplifier_Difference:
- Add AD628 (http://www.analog.com/media/en/technical-documentation/data-sheets/AD628.pdf)

Amplifier_Instrumentation:
- LTC1100*: make pin names visible, align symbols (other than additional pins on xSW symbol)
- INA128/129 moved from Amplifier_Operational

Amplifier_Operational:
- General: Fix up footprint filters throughout
- General: Change pin name offset to 10mil
- ADA4530-1: added package type to description, fixed footprint, reworked symbol a bit
- Add ADA4870 from https://github.com/KiCad/kicad-library/pull/1611 (http://www.analog.com/media/en/technical-documentation/data-sheets/ADA4870.pdf)
- Add MC33172/4 as ALIAS (http://www.onsemi.com/pub/Collateral/MC33171-D.PDF)
- Add OPA1602/4 as ALIAS (http://www.ti.com/lit/ds/symlink/opa1604.pdf)
- ADA4807-1: clarify pin names
- ADA4807-2ACP: clarify pin names, pin name offset to 10mil
- CA3080: name pin 5
- CA3130: name pins, pin name offset to 10mil
- INA128/129: move to Amplifier_Instrumentation
- LM13700: reworked symbol (http://www.ti.com/lit/ds/symlink/lm13700.pdf)
- OPA890: wildcard temp rating

Amplifier_Video:
- AD813: pin name offset, move disable pins and add pin name

Comparator:
- ADCMP35*: name input pin, set pin name offset to 10mil
- LM290*: set pin name offset to 10mil
[...]

PR #767, opened 2018-07-17
Add pin names to opamp symbols

evanshultz:
Continuing from https://github.com/KiCad/kicad-symbols/pull/747. See that PR for a detailed list of changes already implemented and a screenshot of a few symbols that have gotten pin names.

Changes:
General:
- Set pin name offset to 5mil on all triangle-shaped symbols

Amplifier_Audio:
- LM4950*: Add pin names on Unit C
- Cannot find datasheet for Sanyo parts with pin names
- TDA2030: Shrink V+ and V- pin name size
- TDA7266*/TDA7496: Shrink PG pin name size
- TDA7293: Name all pins

Amplifier_Current:
- Reducing pin name offset to 5mil (above) gives more clearance
- INA138: Use normal triangle symbol

Amplifier_Instrumentation:
- Use all long Ref pin length
- Fatten triangle outline to 10mil per KLC
- INA 327: Name pin 6
- AD8230: Use normal triangle symbol

Amplifier_Operational:

Comparator:
- LM2903: Changed pin 2 name from "_" to "-"
- MIC845H: Change pin 3 name from "IN" to "+"

I'll straighten out GH after this. Sorry it's still messy.

Comments from old PR:

> In general i think parts with +/- pins on the left could use a bit reduced pin name offset. Especially if the first slot on top/bottom is used. (The + and - collide with the text of that.)
There is also a slight inconsistency as not all op amps have the +/- visible.

> Amplifier Audio lib:
> 
> LM4950TA and LM4950TS is missing the pin names for the power pins (unit c)
> Sanyo parts have a few pins without names
> TDA2030 V+ and V- overlap (smaller fontsize would help)
> TDA7266M PG pin looks a bit odd with this large font next to the mute pin (they also nearly overlap)
> On other parts the PG pin is part of the last unit. They will suffer from the same issue if one overlays that unit with any other unit.
> TDA7266M Is pin 1 or pin 9 the stby pin (make the font smaller. The other pin has no name right now.)
> Amplifier buffer: everything ok
> 
> Amplifier Current:
> 
> INA193 and INA282: I would suggest to slightly reduce the font size for V+ and GND
> Amplifier Difference looks ok
> 
> Amplifier Instrumentation:
> 
> The ref pin is sometimes short and sometimes long. (even thought it is at the same "slot" in all symbols)
> Amplifier oerational:
> 
> ADA4807-2ACP pin 11 has no name
> OPA2333xxDRB pin 9 without a name
> OPA2691-14 Dis_A is longer than Dis_B (at the same position)
> Amplifier Video looks good
> 
> Comperator:
> 
> parts with a single IN pin might benefit from reducing the font size of the IN pin (overlap with power pins)

------------


poeschlr:
I scrolled through the symbols again. Looks good (except for the things you already mention above)

Thanks for this great contribution.

PR #768, opened 2018-07-17
Footprint for LPC111x had the wrong ep size

poeschlr:
should be 4.7mm or at most 4.85 (was 5.46mm)

footprint PR: https://github.com/KiCad/kicad-footprints/pull/757
datasheet: https://www.nxp.com/docs/en/data-sheet/LPC111X.pdf (page 108)


poeschlr:
The footprint PR for this has been merged. Meaning the symbol has no valid footprint as long as this is not merged.

poeschlr:
Guys? Without this merged we have invalid footprint assignments again as the footprint PR has been merged but the symbol side got forgotten.

PR #773, opened 2018-07-19
DDR3 SDRAM

vlunyov:
Added all modern Micron DDR3L SDRAM chips 1/2/4/8 Gb with all possible data bus width 4/8/16 bit.
One common (manufacturer independent) x16bit DDR3(L) SDRAM 1-8Gb was added too.
https://www.micron.com/products/dram/ddr3-sdram
Two footprints BGA-96 and BGA-78 were used. Pitch 0.8mm.

List of added components:
ddr3

Example drawing:
mt41k512m16

------------


vlunyov:
There are a lot of violations in the Memory_RAM.lib not in my components. Is it ok if I try to fix all of them? Some of the violations required to fix geometry of components. So, the libraty can not be compatible with old version after the fixes.

poeschlr:
You are certainly free to fix other symbols as well but it is definitely not required.

If you keep the geometry changes to a minimum it should be ok.

vlunyov:
It seems the PR is ready to review and merge. All checks have passed.

poeschlr:
Regarding the generic symbol. How did you determine its pin numbering scheme? Is there a standard for this? If not then this symbol might not be such a good idea.

vlunyov:
> You have filled out the footprint filter instead of filling out the footprint field.

I agree, it need to be fixed.

> The changes in this pull request more than make up for the problems created by making these 6 symbols break old designs. I therefore decide that this is still allowed into the lib. (Although i would have preferred these symbols not to be broken in such a way.) Meaning for the future please do not mix in changes to pins (for KLC compatibility) into more important changes.

Old components were changed because I have fixed violations not only my new components.

> Regarding the generic symbol. How did you determine its pin numbering scheme? Is there a standard for this? If not then this symbol might not be such a good idea.

The generic symbol is to represent all of SDRAM footprint pins. Some SDRAM manufacturers reduced some signals for example CKE1, CS1#, ODT1, ZQ1. Count of address bus depends RAM capacity. So, the generic symbol allow to route all possible wires from CPU/FPGA to DDR3 SDRAM. Then you can solder on the PCB any RAM chip manufacturer or capacity and it will be work properly.
image

> Right now there is a merge conflict. If you feel overwhelmed solving it than i can take care of it once all issues are resolved.

Yes, I have not yet enough experience with git and resolving conflicts.
But I can create a new pull request at current git position if it is necessary.

poeschlr:
A new pull request containing only your changes might indeed be the safest option.

PR #774, opened 2018-07-19
Improve pin types for Atmel MCUs

fauxpark:
#387 

I've checked most of the pin types are correct (as far as I understand, anyway). RESET and AREF are Input, I/O and XTAL1/2 are Bidi.

The ones I'm not sure of are:

```
ATmega169A-AU: LCDCAP
ATmega169A-MU: LCDCAP

ATmega169PV-8AU: LCDCAP
ATmega169PV-8MU: LCDCAP

ATmega16U4-AU: VBUS, UCAP
ATmega16U4-MU: VBUS, UCAP

ATmega3290V-8AU: LCDCAP

ATmega329V-8AU: LCDCAP
ATmega329V-8MU: LCDCAP

ATmega8U2-AU: UCAP

ATxmega64B1-AU: VLCD, BIAS1, BIAS2, CAPL, CAPH
ATxmega64B1-CU: VLCD, BIAS1, BIAS2, CAPL, CAPH

ATxmega64B3-AU: VLCD, BIAS1, BIAS2, CAPL, CAPH
ATxmega64B3-MH: VLCD, BIAS1, BIAS2, CAPL, CAPH

ATtiny43U-MU: LSW
ATtiny43U-SU: LSW

AT90USB82-16MU: UCAP
AT90USB162-16AU: UCAP

AT90USB646-AU: VBUS, UCAP
AT90USB646-MU: VBUS, UCAP
```
------------


evanshultz:
For all pin names, I checked the first datasheet I saw. Please confirm on subsequent parts if the pin function is identical.

LCDCAP: Passive since it just goes to an external capacitor according to page 11 of http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8284-8-bit-AVR-microcontroller-ATmega169A_PA_329A_PA_3290A_PA_649A_P_6490A_P_summary.pdf.

VBUS: Page 6 of http://ww1.microchip.com/downloads/en/devicedoc/atmel-7766-8-bit-avr-atmega16u4-32u4_datasheet.pdf suggests it should be Input type. Apparently the MCU cannot source current from VBUS?

UCAP: From pages 4, 6, and 263 of http://ww1.microchip.com/downloads/en/devicedoc/atmel-7766-8-bit-avr-atmega16u4-32u4_datasheet.pdf is seems Passive type as well. It looks like there a regulator there but apparently it cannot source current.

VLCD / BIAS1 / BIAS2: Power Output type from page 76 of http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8330-8-and-16-bit-AVR-Microcontroller-XMEGA-B-ATxmega64B1-ATxmega128B1_Datasheet.pdf.

CAPL / CAPH: Page 54 of http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8330-8-and-16-bit-AVR-Microcontroller-XMEGA-B-ATxmega64B1-ATxmega128B1_Datasheet.pdf shows these pins should be Passive since they go only to a capacitor in a flying-cap power supply.

LSW: Pages 35-36 of http://ww1.microchip.com/downloads/en/DeviceDoc/doc8048.pdf show this is the switch drain/collector for a boost converter. It connects to an inductor and diode anode, so Passive would be fine. You could also use Open Collector, but this pin has no relevance to ERC to I'd go Passive.

fauxpark:
Thanks, that's the rest of them done. VLCD/BIAS1/2 are moved to the right side, since they are (power) outputs now.

RE: VBUS pin - seems to be used for detecting when USB is connected vs. external power supply. Interestingly, some of the USB AVRs don't have this ability.

evanshultz:
Just a couple things here, so it's easier to find later even if the info is buried in the commits:
1. When ~RESET is one a GPIO pin, that pin is still Bidirectional, right?
1. Pages 9-10 of http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8284-8-bit-AVR-microcontroller-ATmega169A_PA_329A_PA_3290A_PA_649A_P_6490A_P_summary.pdf show that the GPIO are tri-stated during reset. Can the GPIO pins on any MCUs be Tristate during operation? If so, their pin type should probably be that instead of Bidirectional.
1. Are all AREF pins only inputs? None have an internal reference which can be driven out of that pin?
1. Do any support driving a clock source into the XTAL pin? The couple I looked at appear to have an inverter inside which must have an external crystal, but some ICs can also support a clock driven directly into one pin with the other pin floating (and no crystal). Can that happen on any of these parts? If so, the pin that can accept a clock input may be best set as Input type with the other (possibly?) as Passive.

Would you post screenshots of a few representative symbols? That will be much easier to compare with in the future. This is only important where pins moved, like on LCD-capable parts.

evanshultz:
BTW, thanks again for your hard work. I was a little nervous to pull the trigger on obsoleting symbols before v5 was tagged, but Rene was good to do that so we're back in business. :)

fauxpark:
I've only moved VLCD/BIAS1/BIAS2 to the bottom right on those symbols, but I think they might be better placed in the top right.
sym

 1. Yep, RESET is Input unless it shares a pin with GPIO.
 2. I assume that means Bidi + Hi-Z. I think most of the GPIOs are capable of being tristated. I'll have a closer look.
 3. ATmega16M1 has a Bidi AREF/ISRC pin - it is either an analog reference input or 100uA current source output. Some others share a pin with GPIO. It seems on some chips [AREF can also have a capacitor to ground](http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-42742-ATmega164P_Datasheet.pdf?page=311) connected for the internal voltage reference? I don't really know what that implies. Apparently you can measure Vref from the pin so maybe it should be Bidi too.
 4. I think some of them do support that, but they all also support a crystal too, so wouldn't both pins just be Bidi?

I'm afraid this is as much as I understand about pin types. So far I've just been making sure they're all consistent, as opposed to correct in terms of ERC.

evanshultz:
VLCD/BIAS1/BIAS2 can be at the bottom. Since they gets caps going to ground I think the bottom is good.

2. Yes, Tristate is Bidirectional + Hi-Z. Thanks for checking!
3. ATmega164P can pipe the internally-generated reference to an external pin, or use the voltage supplied to the AREF pin. In any case, a cap there should lower noise and give better resolution. ATmega16M1 is similar in that and internal current source (instead of an internal voltage reference) can be applied to the pin or else an external reference voltage. My inclination here is to go Passive for all AREF pins. They should only go to caps and resistors, and wouldn't go directly to any digital Output pin (even if the pin was PWM/PDN you'd need a passive network). If an internal voltage/current is put onto this pin and used by other circuit, it's probably an analog circuit (since it's an ADC reference) and so having a "digital" pin type doesn't matter. Are you OK with Passive then, or do you have some other suggestions?
4. The pins of the crystal are Passive type, so it doesn't matter what pin types the MCU uses (they could even be Passive too). If a XTAL pin accepts a clock input, it should be Input type to help catch connection errors. And if the other XTAL pin can supply a clock it should be Output type to go to digital input pins. Hmmm.... are there other connections to consider?

The ERC engine uses these GUI settings to help you understand when it will flag a connection:
image

I have been thinking about ERC (which is the reason to set pin types) and I believe the above is good. But, knowing what it means and having a productive discussion is great so feel free to take any time you need or ask questions so we can talk it out and agree. :)

fauxpark:
Looks like all of the ATmega, ATtiny and AVR symbols have tristate GPIOs - I'm not too sure about the 8051s though.

I also moved the ADC inputs to the left on the ATmega8L and shortened it a bit:
sym

evanshultz:
Leaving the 8051 pins as Bidirectional is fine, then. Thanks!

Is this ready to merge?

fauxpark:
It appears the 8051 GPIOs go high on reset, so no tristate.

I still need to investigate the XTAL and TOSC pins but these diffs are already pretty big... if you're good to merge then sure.

evanshultz:
Sure thing! We can do the other pins next.

PR #775, opened 2018-07-20
Add Asahi Kasei AK5720 24-bit stereo ADC

yankee14:
Add Asahi Kasei AK5720 24-bit stereo audio analog-to-digital converter.

[Datasheet](https://www.akm.com/akm/en/file/datasheet/AK5720VT.pdf)\
[Manufacturer development board](https://www.akm.com/akm/en/file/ev-board-manual/AK5720VT.pdf)
[Screenshot](https://i.imgur.com/kn6TACt.png)

I think my symbol matches the official footprint library. Please tell me if it does not!

[The footprint matches TSSOP-16_4.4x5mm_P0.65mm.kicad_mod.](https://raw.githubusercontent.com/KiCad/kicad-footprints/master/Package_SO.pretty/TSSOP-16_4.4x5mm_P0.65mm.kicad_mod)

------------


evanshultz:

Can you please create a GitHub account and use the same email account from which you pushed this commit? Then you can sign the CLA above. Without that, there's no sense in me reviewing this symbol, and I really like AKM parts so I'm happy to see that you've added it! :)

yankee14:
Thanks for the help!

Sorry for the trouble, I've never made a pull request or contributed before, so I'm not surprised I did it wrong! I've read the complete KLC but I don't think it has any information about this issue with the CLA. I think I remember signing the CLA a while back.

> Can you please create a GitHub account

I have to make a new GitHub account? It sounds like I just forgot to set the user.name and user.email fields for my local git repo before committing. Can I just amend my commit with the correct author information? Sorry, I don't understand what I need to do.

evanshultz:
- Datasheet page 22 says VCOM can't source current and only connects to a cap, so use Passive type.
- Apparently REGO is similar, so the same suggestion applies.
- Let's put analog pins on the left and digital on the right. So only LIN and RIN on the left side, all the rest on the right. Also swap VD and VA. Once that's done, shrink the symbol if you can.
- This is minor, but pin 13 only needs a tilde in front and not a trailing one.

Thanks!

evanshultz:
Also, VCOM should be on the left side along with REGO, I guess, according to the first page of the datasheet. Put those pins that only connect to ground through a passive part at the bottom of the symbol.

yankee14:
@evanshultz 

>This is minor, but pin 13 only needs a tilde in front and not a trailing one.

Fixed.

>swap VD and VA.

Fixed.

>Put those pins that only connect to ground through a passive part at the bottom of the symbol

Good call. I think that's an improvement on how the rest of AK's symbols are laid out in the library.

>Datasheet page 22 says VCOM can't source current and only connects to a cap, so use Passive type.
Apparently REGO is similar, so the same suggestion applies.

I do understand wanting to classify it as a passive pin, but from a design rules check perspective, should it not be classified as an output? Other AK ADC's VCOM pins are classified as an output.

>Let's put analog pins on the left and digital on the right...VCOM should be on the left side along with REGO.

I think that's okay, but again I was just taking other AK ADC's in the library as an example. It might wind up looking a bit lopsided with most of the pins concentrated on the right.

>shrink the symbol

Fixed.

scap2

evanshultz:
Thanks! This symbol is looking really nice.

I don't think other symbols are a good reference in many cases, such as this one. Since those pins don't source current Passive type is OK with the ERC engine and matches what's been done on other parts recently. Since they're not digital or power pins the ERC engine will of limited value.

After seeing the updated symbol and checking out the datasheet here are a couple things I will mention but you're free to do them or let me know you prefer the symbol as is:
- Space apart RIn and LIN so that horizontal caps can go next to the symbol.
- Space apart VCOM and REGO so that caps can go horizontally to the left of the symbol.
- Push CKS up a bit and leave the symbol slightly off-center vertically.
- I typically put the I2S clocks in order of frequency with data at the bottom: MCLK, BICK, LRCLK, SDTO.

One more thing: please put the ref des at the top left and part number at the top right.

yankee14:
>put the I2S clocks in order of frequency

Thanks, I was having trouble deciding on the order, and that sounds like a good policy! Fixed.

>Space apart RIN and LIN so that horizontal caps can go next to the symbol

Fixed.

>Space apart VCOM and REGO so that caps can go horizontally

Fixed.

>put the ref des at the top left and part number at the top right

Fixed.

>Push CKS up a bit and leave the symbol slightly off-center vertically.

I wanted to leave CKS further away to kind of convey that noise and parasitic capacitance should be kept low. It is also possible that many components will be connected to it, like a resistor divider network ([development board](https://www.akm.com/akm/en/file/ev-board-manual/AK5720VT.pdf#page=14&zoom=auto,-272,581)). I'd like to leave it like, but let me know and I'll change it!

I'm not sure what you mean by leaving it off-center.

evanshultz:
All looks fine. Thanks!

Regarding centering, the symbol is the same Y and -Y distance right now but even if it's off a bit that's OK. We often break that rule if the symbol is aesthetically better off-center.

PR #777, opened 2018-07-21
Lattice iCE40HX8K-BG121TR FPGA Symbol

awygle:
Add iCE40HX-BG121 symbol.

Documentation is found at http://www.latticesemi.com/Products/FPGAandCPLD/iCE40, datasheet is http://www.latticesemi.com/view_document?document_id=49312.

Images:
image
image
image
image
image

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/693

------------


awygle:
Hmm, I don't understand why the center was calculated as not being at the origin here. I measure all units as being as close as possible to centered without placing pins off the 0.1 grid. I could add blank space to center them exactly I suppose but I didn't think I was being asked to do that previously...

awygle:
I figured it out. The center check is based on pins and _only_ on pins - I was centering based on the yellow box (which makes a lot more sense to me, but oh well).

The only Travis failure now is the lack of the footprint, which is because https://github.com/KiCad/kicad-footprints/pull/693 hasn't been merged yet.

awygle:
Again, just a reminder this PR still exists - is someone available to review it? @evanshultz ?

awygle:
The footprint has been merged and the symbol passes all checks.

whitequark:
Friendly ping!

evanshultz:
Why am I getting blamed for so much lately!? :)

Centering isn't something we place high importance on and it is true that the algorithm is quite weak. Centering on the symbol outline is fine and I prefer that over just the pins for this symbol.

- Remove the trailing "TR" in the symbol name. This must be about packaging.
- Change the package type in the description to "BGA-121". The other one should be "QFN-48" so I must have missed that during review.
- We use 100mil pins if the pins have <=2 characters, so I think this symbol can use 150mil pin length.
- CDONE is pin K8 and CRESET_B is pin L9.
- I would be inclined to put the SPI pins with the other IOB ones since they can be normal IO. What's the reason to put them on the last unit?
- Why does the footprint name have "NSMD" in it? Since that's our default pad style it shouldn't necessary. Maybe just a legacy footprint or before we decided how to do pads?
- I haven't verified the footprint yet, but assuming that is correct the footprint filter should be `BGA*9.0x9.0mm*Layout11x11*P0.8mm*Ball0.4mm*`.

I'll check the footprint filter tomorrow. Sorry I'm out of time today.

evanshultz:
Yeah, that footprint looks good to me. So please update the footprint filter as mentioned above.

whitequark:
> * Remove the trailing "TR" in the symbol name. This must be about packaging.

Done.

> * Change the package type in the description to "BGA-121". The other one should be "QFN-48" so I must have missed that during review.

Done.

> * We use 100mil pins if the pins have <=2 characters, so I think this symbol can use 150mil pin length.

Changed.

> * CDONE is pin K8 and CRESET_B is pin L9.

Fixed.

> * I would be inclined to put the SPI pins with the other IOB ones since they can be normal IO. What's the reason to put them on the last unit?

The SPI pins are in their own bank and have a separate VCCIO. I guess they could be in another unit entirely.

> * Why does the footprint name have "NSMD" in it? Since that's our default pad style it shouldn't necessary. Maybe just a legacy footprint or before we decided how to do pads?

There's a lot of BGA footprints with "NSMD" in the name. Is this really relevant to this symbol PR? Seems like it'd be a separate bulk change.

> * I haven't verified the footprint yet, but assuming that is correct the footprint filter should be `BGA*9.0x9.0mm*Layout11x11*P0.8mm*Ball0.4mm*`.

Changed.

evanshultz:
I didn't catch that the SPI pins had their own supply.

The unit body sizes need to change to put the now-150mil pins on grid but I think that's it.

whitequark:
Done.

evanshultz:
Thanks once again! And apologies for taking so long to get to some PRs, like this one.

PR #778, opened 2018-07-22
INA181 Symbol

markus-k:
Add symbol for INA181 current sense amplifier. The symbol was copied from the INA282 and modified accordingly.

Datasheet for the INA181: http://www.ti.com/lit/ds/symlink/ina181.pdf

Image of the symbol:
image


evanshultz:

There's only one thing to change before I can merge: Pin numbers 3 and 4 are flipped.

markus-k:
Whoopsie, that could've ruined someones day, thanks for pointing that out. Fixed in latest commit.

By the way, the Amplifier_Current library seems to be CR-LF, but KiCad generates LF line endings. I didn't want to change the whole file so I converted the file for the commit to match what's there. Maybe the file can be converted in a separate commit.

evanshultz:
Thank you!

Yes, they should all be LF (Unix style). Thanks for pointing that out. There used to be a check and auto-convert during commit on that...

PR #781, opened 2018-07-22
Fix atmel MCU qfn-44 assignment

poeschlr:
All parts that list qfn-44 have the same dimensioned drawing of the
package. They should therefore reference the same footprint. (Two
referenced one with EP set to maximum "slug" dimension wheras the vast
majority used nominal "slug" dimensions.)


poeschlr:
The footprint PR for this has been merged. Meaning right now the symbols that would be changed here have no valid footprint!

poeschlr:
Guys? Without this merged we have invalid footprint assignments again as the footprint PR has been merged but the symbol side got forgotten.

PR #782, opened 2018-07-24
Change Atmel AVRs XTAL1/2 to Input/Output

fauxpark:
Continuing from #774.

I've gone with setting XTAL1 to Input as it often accepts an external clock pulse, and the datasheets all refer to it as the input to the inverter. XTAL2 does not seem to be able to output a clock signal, but it is again described as the inverter output.

There is not much info about the ATxmega256A3B(U)'s TOSC pins. I've made them passive.

xtal
xtal2

------------


fauxpark:
I see Travis is warning about AREF pins not being Input, is this correct?

poeschlr:
The pin type suggestions by travis are to be taken with a grain of salt. They are based on (personal) experience with the goal that the suggestions are correct for most parts. In this case they might just be wrong. (There is a reason it is a warning not an error)

If somebody has some time they could analyze the current lib and find out if the suggestions given are correct or if there should be some changes. (I only notice this check if it goes wrong, so my information is biased against it.)

fauxpark:
@poeschlr any chance of an update on this?

poeschlr:
Not sure which type of update you mean so i answer both ways to read this:

Update regarding travis: I doubt we will be able to get a better algorithm then we have now.

---

Update regarding review: I currently have very little time so if you want to wait for my review then it will take some time. Might be faster if somebody else takes a look.

fauxpark:
Sorry, I meant regarding the PR/review. No worries, I didn't mean to "push in line" or anything.

serkanxselcuk:
@fauxpark @evanshultz 
> We could really need additional maintainers.

I would like to provide support to review PRs as I can. I have also not always time but one more is better than one less.

evanshultz:
All the XTAL pins seem OK to me. Note on page 82 of http://ww1.microchip.com/downloads/en/DeviceDoc/doc8077.pdf that it mentions XTAL2 can be a GPIO when using an external clock to drive XTAL1. Can you see if that is common for other parts? If so, we may need to make some adjustments.

I didn't initially see anything that says TOSC couldn't be Passive type since they appear to only discuss going to a crystal. See page 188 of the doc linked above where TOSC1 also goes into a block called "External Clock" and several places in http://ww1.microchip.com/downloads/en/AppNotes/doc42042.pdf where it does appear to say TOSC1 can be a clock input. I stopped reading at that point since it seems like there's very likely an inverter between the TOSC pins and I'm fairly well convinced of that.

Now making TOSC1 Passive type won't cause any errors if a clock is driving the pin, and neither will Input. The only thing that could change if we make TOSC1 of Input type would be if that net only has pins of Input type and no Output or Bidirectional pins. It's probably a low chance to catch something that helps out a user, but what do you think? Do you see the same thing and is it worthwhile to change TOSC1's type?

evanshultz:
Here we go. See page 16 of http://ww1.microchip.com/downloads/en/AppNotes/AN2648-Selecting_Testing-32KHz-Crystal-Osc-for-AVR-MCUs-00002648B.pdf. There is an internal inverter between those pins so TOSC1 can be an input to accept an external clock while TOSC2 could be Output or Passive as far as I'm concerned.

fauxpark:
Where the datasheet states that XTAL pins are shared with GPIO I've left them as tristate (see eg. ATxmega64A1, they are usually `PR[0:1]`).

I will change TOSC1/2 to input/passive then.

evanshultz:
OK. Thanks!

PR #784, opened 2018-07-24
Added 74LVC2T45 2-bit level-shifter

Duckle29:
I added a symbol for the 74LVC2T45 level-shifter

[Datasheet](https://assets.nexperia.com/documents/data-sheet/74LVC_LVCH2T45.pdf)

------------

image
image
image
image


evanshultz:
- Change pin name offset to 20mil
- Put a hyphen between the package type and pin count in the description (so `VSSOP-8`)
- Could the description be simplified a bit using the datasheet phrase? Perhaps `Dual supply translating transceiver, 3-state, 2-bit`?
- The letter and number are flipped on pins 2-3 and 6-7.
- Footprint filters do not need pin count but do need pitch. For VSSOP use `VSSOP*2.4x2.1mm*P0.5mm`.
- The description should contain the difference between the PNs.

Duckle29:
Done :)

evanshultz:
Thank you again!

PR #785, opened 2018-07-24
Cd4504

Duckle29:
Added A CD4504 symbol. It seems kicad has re-arranged a ton of lines in the .lib file as well :/

------------

[Datasheet](http://www.ti.com/lit/ds/symlink/cd4504b.pdf)

image
image
image


evanshultz:
- You can't have a default footprint if the footprint filters are set to allow different packages. This also affects the description.
- In addition to catching the above, Travis also found that you duplicated pin 7 and left out pin 8.
- The datasheet linked above is not the one you included for this symbol.
- Since this part seems to have a specific purpose with regards to logic types, perhaps copy the datasheet description.
- Use the other pin names from the datasheet, like AIN and AOUT instead of Ai and Ao, GND should be VSS, and write out SELECT.
- Put refdes and top left and part number at top right.

Duckle29:
@evanshultz Done. Thanks :)

evanshultz:
The duplicate pin 7 mentioned above wasn't fixed.

Also, add `DIP-16/SOIC-16/TSSOP-16` to the end of the description to indicate all packages. The footprint filters do not get pin count and should therefore be:
- DIP*W7.62mm*
- SOIC*3.9x9.9mm*P1.27mm*
- TSSOP*4.4x5mm*P0.65mm*

Lastly, please include a new screenshot since you made visible changes. Thanks!

Duckle29:
Oops on the pins. All changes should be implemented:

Duckle29:
https://www.amazon.com/BOSCH-PKP18E-Glue-Melt-Sticks/dp/B00R7HUPBI/ref=sr_1_3?ie=UTF8&qid=1532951865&sr=8-3&keywords=bosch+glue+gun

Duckle29:
uuh oops

Duckle29:
Done :)

evanshultz:
Thank you!

PR #787, opened 2018-07-26
Adding Lattice XP2-5 in TQFP-144

stewi1014:
I'm using this FPGA in a project at the moment. The aliases are the same chip in the same package, just made to different speed specifications according to the data sheet. It might be more appropriate to remove them and name the symbol LFXP2-5E-XTN144 instead. Let me know.

Datasheet: http://www.latticesemi.com/view_document?document_id=24635
Pinout: http://www.latticesemi.com/view_document?document_id=28312
Lattice Page: http://www.latticesemi.com/Products/FPGAandCPLD/LatticeXP2
Have ran checklib.py and is passing.
Image:
lfxp2-5e-5tn144


poeschlr:
Sorry for the late feedback (we clearly don't have enough maintainers for the amount of contributions we get)

Is this symbol script generated or did you create it by hand?

Is there a reason for having some power pins on every unit while still having a central power unit?
Longer pins might make sense in this case. 100 mil is only enough for parts with less than 100 pins.

stewi1014:
I made it by hand primarily using the pinout spreadsheet for reference.

The power pins on each bank are there as they are supply pins for the banks, and are relevant to to their respective banks and the chosen IO standard for those banks. I think I was just trying to copy other FPGA symbols to be honest though. I am noticing that other symbols have the grounds on the "power" (E) unit though. I'd leave them by their banks, but I don't think there's much in it either way. Let me know your thoughts.

I think longer pins do make sense. I'll commit some longer pins.

stewi1014:
New image for completeness
lfxp2-5e-5tn144

poeschlr:
If the power pins are meant to supply that particular part of the chip then it is ok to have them with the other pins that go with it.

Well as it is not created by script i will now have the task to check if all pins are assigned correct numbers, names and pin types. Will take me a bit.

poeschlr:
Looks good, thanks

stewi1014:
Awesome, glad I could help.

PR #790, opened 2018-07-31
Ads868x ADC

SchrodingersGat:
http://www.ti.com/lit/ds/symlink/ads8681.pdf

Added symbol for the QFN variant of this part, with aliases for each three SPI data rate options

image


SchrodingersGat:
Closing, started from wrong branch.

PR #791, opened 2018-07-31
ADS868x adc

SchrodingersGat:
Added ADS868x ADC parts in QFN package

- ADS8681
- ADS8685
- ADS8689

http://www.ti.com/lit/ds/symlink/ads8681.pdf

image


SchrodingersGat:
Bump @evanshultz

evanshultz:
- Can you change the description to `16-Bit, High-Speed, Single-Supply, SAR ADC Data Acquisition System
with Programmable, Bipolar Input Ranges, 1MSPS, QFN-16`? (Change the samples/second, of course.)
- Pins 5 and 6 should have an underscore.
- Did you verify the footprint? At the very least, the footprint name doesn't match either (!) of the package names given in the datasheet on pages 71-73.
- The footprint filter is definitely wrong.

evanshultz:
That's all good, but I'm afraid a 20mil pin name offset won't make the pin names all fit inside the outline. Can you please check?

SchrodingersGat:
There was a very very slight overlap - pushed back out to 30mil

PR #793, opened 2018-08-02
add SST26VF016B serial memory flash

ChristianHirsch:
Add [SST26VF016B](http://ww1.microchip.com/downloads/en/DeviceDoc/20005262D.pdf) serial flash memory. I know that the symbol is slightly off-center (S3.1).

image


bradfa:
If you could make it centered I think that would be better.  There's a lot of KLC violations in the Memory_Flash.lib and adding more probably isn't best.  Some other SPI devices split the 6 pins up to have 3 on the left and 3 on the right, although I'm not sure that's a better solution either.

You could break the pins up like this having the clock enable and clock pins 1 pin's spacing away:
image

A lot of people that I know would call this a "QSPI Flash" for quad-SPI.  There doesn't seem to be many other SPI flash in the kicad-symbols but I think adding "QSPI" to your keywords would be worthwhile.

Since there's a ton of part numbers for this part based on package, temperature grade, and tape and reel/tube/tray, would it be worth having wild-card flags in the part name?  As per KLC S2.2: http://kicad-pcb.org/libraries/klc/S2.2/    I'm not sure as to the answer to this.

I don't think there's a valid SOIJ footprint with 5.28mm body.  There is an SOIJ with 5.3mm body which I believe should match the needs of this part.  Could you verify that your footprint filter list is correct please?

poeschlr:
@ChristianHirsch would you be willing to look into the comments above?

ChristianHirsch:
@poeschlr @bradfa finally I fixed your comments. regarding the wild-card flags: I'm not 100% sure how to change it. one example could be _SST26xF0xx_, but based on the [document](http://ww1.microchip.com/downloads/en/DeviceDoc/20005356B.pdf) from microchip I don't know if this is legit, because of the many different package information?

bradfa:
@ChristianHirsch I'm not sure about the wildcards and the partnumber, sorry, there are many many variations of that part numbering scheme and without a ton of review it's going to be hard to get it right.

Maybe it's best to just put in your original full part number now and then as people need/want the variations they can alias it or do the review to use wildcards at that time?  Basically kick the can down the road?

Sorry I don't know what's best here.

ChristianHirsch:
@poeschlr what do you think about using wild cards for the symbol?
@bradfa I did not insert any wild cards so far. In case you wanna keep it as is, it is ready to be merged..

poeschlr:
@evanshultz i think you have more expirience with wildcarding part numbers then me. Could you take a look at my comment and correct me if my interpretation is completely wrong?

We require wild-cards for certain part number parameters. To be more precise: If possible only electrical parameters should be included and part numbers can even be shortened.

I don't quite get the point made by @bradfa. Most datasheets even use a very similar wildcarding system as we require. (We do not even care which options there are then available. Only that there is more than one.)

If these non electrical options are specified at the end of the part number then we simply leave them out.
(If they are organized in a single block in the middle then we can also leave them out. Users will still be able to understand what is meant.)

---

For this part everything we would be interested in is on page 60 of the datasheet. We for sure do not care about tape and real and temperature. I am not even sure we would care about the operating frequency to be honest. But as there is only one we can include it.
In the end we would only need 3 part numbers to get all possible options of this part depending on footprint.
- WDFN: `SST26VF016B_MF`  (Or at most: `SST26VF016Bx-104x_MF` or even `SST26VF016Bx-xxxx_MF` if we do not want to include the operating frequency. But for the later i would choose the very first option i show.)
- SOIJ: `SST26VF016B_SM`
- SOIC: `SST26VF016B_SN`

myfreescalewebpage:
@ChristianHirsch I have just reviewed similar symbols you can check out https://github.com/KiCad/kicad-symbols/pull/1595#issuecomment-485269097 to look at the current design of them.

There are other issues to fix here, for example the part should be atomic.

Joel

myfreescalewebpage:
@ChristianHirsch are you still working on this symbol ?

ChristianHirsch:
@myfreescalewebpage well, that wasn't on my list anymore. But I checked the differences to the PR you mentioned. Is it still necessary to add this symbol? It almost look identical to MX25R3235F and SST25VF080B. Maybe it's just better make an alias?

myfreescalewebpage:
@ChristianHirsch you decide. We do not required anything from you.
Please tell me if you think you will spent time on this contribution or not.
Thanks
Joel

ChristianHirsch:
@myfreescalewebpage Well, I'd prefer to make an alias of another symbol. It's easier to maintain I guess.

myfreescalewebpage:
If this is apossible you can. Just remember that alias = same pin name, same pin numbering AND the same footprint.
You can continue on this PR to not create a new one.

PR #797, opened 2018-08-02
Add ST VNH5017A-E

NICOLASGON:
Hi, here is the symbol forthe VNH5018A-E from ST Microelectronic : https://www.st.com/resource/en/datasheet/vnh5019a-e.pdf

capture du 2018-08-02 23-44-54

The footprint is the Package_SO:ST_MultiPowerSO-30 so it's actually in the mainstream kicad footprint library.

The checklib tool told me that i'm violating S4.4 but i didn't find a solution for that.

Thank you for your help,
ngonzalez


jneiva08:
The Package_OS:ST_MultiPowerSO-30, may be at the bottom next to GNDA, where it is, is not readable.
The NC must be at the border of the square.
The length of connector could be half.

NICOLASGON:
- I moved the package label to bottom right.
- I moved NC pins to the maximum downwards.
I didn't understand for the length of the connector ?
Do you mean reduce by a half the length of all pin connectors ? 

capture du 2018-08-03 14-05-26

jneiva08:
The NC pins wanted say like this:
http://kicad-pcb.org/libraries/klc/S4.6/

The length wanted say this:
http://kicad-pcb.org/libraries/klc/S4.1/

NICOLASGON:
I pushed back the NC pins but for the length i respect the convention !
Length must be between 100 mil and 300 mil and here it's 200 mil.
To stay on the 100mil grid, i must choose 100, 200 or 300 mil and 100 is too short.

capture du 2018-08-03 22-12-39

NICOLASGON:
Hi,
Do you have news about this PR ?
Sincerely,

Misca1234:
Hi

100 is not to short, almost all symbols with pin number lower than 100 is using 100, see following as an example
https://github.com/KiCad/kicad-symbols/pull/925

It is ok if looks a little bit to narrow, I will have another look at it tonight

DanSGiesbrecht:
While having a pin longer than 100mil is allowed (up to 300mil), 100mil is the **preferred** value. I agree with @Misca1234 , I think with pin numbers that are less than 3 digits, (i.e. <100), 100 mil is appropriate.

evanshultz:
As I just noticed this is your first PR, thanks for contributing to KiCad! You've already made a lot of adjustments to make the symbol meet the library guidelines and been introduced to KLC. Thank you for climbing the learning curve to get this symbol into the library!

Here, I would put pins 5 and 9 on the left side. Grouping the pins together by function, and those pins will all go the the controller, are more important than forcing all outputs on the right side. Those pins are also tri-state and not just outputs.

Similarly, CS may be best on the left for the same reason.

CP is a little annoying but I suppose the top right, as it is, is a reasonable location.

Once that is all done, the symbol can be shrunk. NC pins will not appear in the final symbol so build the symbol as if they were not there and then fit them in wherever they can go, including burying them inside the symbol. Do not enlarge the symbol to fit them. You may be able to narrow the symbol 100mil on each side as well and shrinking vertically. The top and bottom pins do not need to straddle the vertical centerline of the symbol.

Lastly, the ref des and value can move down by 50mil to be just above the symbol.

poeschlr:
I took this on and finished it in https://github.com/KiCad/kicad-symbols/pull/1352

I hope nobody is mad at me for this.

PR #798, opened 2018-08-03
MCU_Microchip_ATtiny: add ATtiny212/412

aimylios:
This commit adds the Microchip ATtiny212 and ATtiny412 microcontrollers to the MCU_Microchip_ATtiny library. I tried to copy the general style of the other symbols from this library as much as possible.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/40001931A.pdf

Screenshot of the new symbol:
attiny412

The required footprint already exists: [SOIC-8_3.9x4.9mm_P1.27mm](https://github.com/KiCad/kicad-footprints/blob/master/Package_SO.pretty/SOIC-8_3.9x4.9mm_P1.27mm.kicad_mod)

------------


fauxpark:
I'm not part of the librarian team, but since I've been doing a lot of work on the Atmel libraries, thanks for helping out :) Here are my thoughts:

 - Could you swap the actual symbol to the ATtiny212? I've made it so that the parent symbol is always the "lowest-end" device in these libs.
 - For the sake of consistency could you also align the reference as left/bottom and the component name as left/top (but keep the positions - compare against the ATtiny4)?

I've deliberately kept the pin names short, with at most only one alternate function, but this symbol is simple enough that I think keeping the UPDI part is probably fine.

aimylios:
Thanks for your feedback! I made the ATtiny212 the main symbol and the ATtiny412 the alias. The reference is now aligned bottom-left and the name is top-left. I squashed my commits into one to avoid cluttering the history with these cosmetic changes.

I think we need a place to document library-specific design rules like these. Although the KLC is quite extensive, it can't (and shouldn't) cover every detail. This is OK as long as you have a single maintainer per library, but as soon as multiple people start to contribute, subtle inconsistencies will be introduced.

attiny212

poeschlr:
The placement of the fields does not really matter as they are auto placed in eeschema any ways. The only thing important in the lib is that they are placed such that the preview is readable.

aimylios:
The placement of the fields is indeed purely cosmetic. But there are other parameters that would merit to be documented as library-specific design rules.

Back to topic: Is there anything else I should change, or is the ATtiny212/412 symbol ready to be merged?

evanshultz:
Added in https://github.com/KiCad/kicad-symbols/pull/1139.

PR #799, opened 2018-08-06
Added TCAN33x series of CAN transceivers

SchrodingersGat:
Datasheet: http://www.ti.com/lit/ds/symlink/tcan337.pdf

* Four separate pinouts 
* Two available footprints
* Eight unique devices
   * TCAN330 / TCAN330G
   * TCAN332 / TCAN332G
   * TCAN334 / TCAN334G
   * TCAN337 / TCAN337G

image


SchrodingersGat:
Bump @evanshultz

evanshultz:
- Could you include `CAN FD` in the description?
- Put package types at the end of each description.
- TCAN334 says "low power mode" in the description but the datasheet calls it "standby mode".
- I'm not too familiar with CAN, but it seems the RXD pin should be tri-state (datasheet page 4).
- FAULT should be open collector type.

SchrodingersGat:
All seems quite reasonable. Thanks @evanshultz

evanshultz:
Thanks! Just one more thing: Add `-8` after `SOIC` in the description.

evanshultz:
Simply glorious...

PR #803, opened 2018-08-08
Added LTC6909 timer IC

SchrodingersGat:
www.analog.com/media/en/technical-documentation/data-sheets/6909fa.pdf

image


SchrodingersGat:
Bump @evanshultz

evanshultz:
- Pins 1 and 3 have names close together. Can you move left side pins down a bit?
- SET connects to V+A through a resistor while all other left side pins are V+D. Perhaps move SET to the top so that  can be "inside" the pins that connect to V+D.
- All OUT* pins are tri-state.

SchrodingersGat:
image

evanshultz:
Thanks!

PR #804, opened 2018-08-08
Added ADuM540x series of digital isolators

SchrodingersGat:
http://www.analog.com/media/en/technical-documentation/data-sheets/ADuM5401_5402_5403_5404.pdf

image


SchrodingersGat:
Bump @evanshultz  :)

evanshultz:
- Since we already have several digital isolators, could you use the full datasheet description to tell them apart? `Quad-Channel, 2.5 kV Isolators with Integrated DC-to-DC Converter`, note the I/O with `2/2 Input/Output` or `2 Input/2 Output`, and then add the package info?
- Are there any conditions/reasons why the two GND pins one each side wouldn't be connected? In other words, should they be stacked? I suppose from a layout perspective it may be inconvenient, and I don't see that you have to in the datasheet, which would make individual pins a good way to go.
- Capitalization is mixed. Subtext is lowercase for "GNDiso" but not "VDD1". Can you align it?

So pushy! There, that's all of your outstanding PRs. :)

SchrodingersGat:
image

each set of GND pins are connected internally so stacking them makes sense.

SchrodingersGat:
image

I think both pins must (or at least should) be connected according to the datasheet

evanshultz:
Hmmmm... it does only say recommended. Not required. And if they're internally connected it would seem that either one should be able to work. And whether they should be connected or not, keeping separate pins will work. Or perhaps we should ask ADI?

SchrodingersGat:
Fair enough, I'm not too fussed either way, the flexibility is good :)

evanshultz:
Alright. Thanks!

PR #806, opened 2018-08-08
Added MC33035 (alias NCV33035) and MC33039 (alias NCV33039) BLDC moto…

vlunyov:
Brushless DC Motor Controller (SOIC-24/DIP-24)
MC33035 (alias NCV33035) http://www.onsemi.com/pub/Collateral/MC33035-D.PDF
default

Closed Loop Brushless Motor Adapter (SOIC-8/DIP-8)
MC33039 (alias NCV33039) http://www.onsemi.com/pub/Collateral/MC33039-D.PDF
default

------------


poeschlr:
Sorry for the late review.

I would suggest to copy the symbols per package and assign a footprint to them.
I don't think power input is the best option for the Vc pin of MC33035. Maybe normal input is better here.

It should be possible to reduce the pin length to 100mil for these symbols (all pin numbers are at most 2 digits long)

vlunyov:
The Vcc and Vc pins are exactly power pins. Please see page 20 of the http://www.onsemi.com/pub/Collateral/MC33035-D.PDF
The pin Vc is power input for the Low (Bottom) Side of the output drivers.
Page 9:

> 18     VC     The high state (VOH) of the Bottom Drive Outputs is set by the voltage applied to this pin. The controller is operational over a minimum VC  range of 10 to 30 V.

Ok, I will reduce all pins length to 100mil. It will allow electronic circuits to be more compact.

poeschlr:
To me the wording of the Vc description suggested it to be a reference voltage input. I however suspect you have more knowledge of these parts. Meaning i let you make the decision on this one.

evanshultz:
If you come back to this, please remove the NCV parts as we don't include automotive-grade parts.

poeschlr:
Above i suggested one symbol per footprint and make them fully specified. Since that review we have gotten a lot stricter about that so now it is a requirement.

myfreescalewebpage:
@vlunyov are you still working on this subject ?

PR #807, opened 2018-08-09
Added IR2101, IR2102, IR2103, IR2105, FAN7888. Fixed FAN7842.

vlunyov:
Added:
IR2101 and IR2102 http://www.irf.com/product-info/datasheets/data/ir2101.pdf
IR2103 http://www.irf.com/product-info/datasheets/data/ir2103.pdf
IR2105 http://www.irf.com/product-info/datasheets/data/ir2105.pdf
FAN7888 http://www.onsemi.com/pub/Collateral/FAN7888-D.pdf

Fixed:
FAN7842 http://www.onsemi.com/pub/Collateral/FAN7842-D.pdf
The FAN7842 was redrawn according to standard representation of gate drivers and was fixed pin Electrical Type that caused check errors.

default

default

default

default

default

default

------------


evanshultz:
These are nice symbols. Thank you. Thanks for taking the initiative to fix up a symbol as well!

- Can IR2101 be an ALIAS of IR2016 or IRS21867S?
- Please use Infineon links since they've bought IR.
- IR2105 needs pin 5 as an invisible No Connect pin to match with the correct packages.
- Footprint filters do not use pin count, so remove the `?8` (or other pin count) from them.
- All symbols need pin name offset changed to 20mil.
- Add body dimensions to the footprint filter of FAN7888.
- Remove the extra height at the bottom of FAN7888.
- Since FAN7888 will have a more complicated wiring than a single half-bridge symbol, what about putting VDD at top and GND at bottom? That is the standard practice, but the single half-bridge ICs are "special" in this way to make them easier to wire into the rest of the circuitry. Not required, just asking.

Note that on some of the above items the library was out-of-date and not a good reference. Sorry.

vlunyov:
No, the IRS21867S can not be an alias of the IR2101 because IR2101 available in two packages DIP-8 and SOIC-8 but IRS21867S available only in SOIC-8.
Ok, I will fix datasheets URL to Infineon's site links.
FAN7888 has very different pins count at left (logic inputs) and right (output for three half bridges) sides of the footprint. VDD and GND are successfully took place at left side. So, putting VDD at top and GND at bottom will increse schematic size.
Why footprint filters can not use pin count? Where I can read about this rule?

Fixes:
default

default

vlunyov:
Pin name position offset is now 20 mils for all components in the Driver_FET library.

vlunyov:
NC pins were moved because there are an issue showed on images below.

Do not show hidden pins = ON:
default

Do not show hidden pins = OFF:
default

I have seen this decision at many KiCAD's libraries. I fixed this to improve neatness of the whole Driver_FET library for all users because I like Power Electronics.

vlunyov:
Yes, IR2101 and IR2106 has the same drawing and footprints and can be linked with an alias.
Pin count mask was removed from footprint filters.
I agree with power pins for the FAN7888:
default
NC pin of the IR2105 is now 100mil in length:
default

vlunyov:
Infineon's URLs to datasheets looks strange compared to IR links. For example:
https://www.infineon.com/dgdl/ir2101.pdf?fileId=5546d462533600a4015355c7a755166c
Compare it with IR link:
http://www.irf.com/product-info/datasheets/data/ir2101.pdf
The link https://www.infineon.com/dgdl/ir2101.pdf (without ?fileId=<big number>) does not work.
I suggest to use old IR links or use product page links instead a direct link to datasheet like this:
https://www.infineon.com/cms/en/product/power/gate-driver-ics/ir2101/
Which solution will be better in this case?

evanshultz:
Infineon, please.

evanshultz:
IR2105 appears to be discontinued, so I suspect that's why Infineon didn't include it on their site.

This is a great improvement. Thanks!

Note to self: Still a lot of pins in this lib that are not 100mil length...

PR #814, opened 2018-08-11
Add CH340G USB to UART Converter

GrantM11235:
Unofficial CH340G translated datasheet: www.datasheet5.com/pdf-local-2195953
Official CH340 English datasheet (no CH340G): http://wch-ic.com/downfile/79
Official CH340 Chinese  datasheet (including CH340G): http://www.wch.cn/downloads/file/79.html
Screenshot (updated for commit bc12df2)
screenshot
The layout is based on the unofficial datasheet with a few modifications to make it more consistent with other KiCad symbols:
- `VCC` and `GND` on top and bottom
- `UD+` and `UD-` to `D+` and `D-`
- `D+` and `D-` near center of left side
- ~~D- above D+~~
- `R232` configuration pin on left side

`V3` is meant to be connected to `VCC` when VCC is 3.3V, or simply a decoupling capacitor to ground when VCC is 5V. I set its electrical type to "passive", but I am not sure if that is right.

------------

- ~~If there are matching footprint PRs, provide link(s) as appropriate~~


GrantM11235:
Travis CI doesn't like that I didn't add a datasheet, but I couldn't find one that is official, includes the CH340G, and is in English. Should I add one? If so, which one? I am leaning towards the official Chinese one.

GrantM11235:
I put `D-` above `D+` because that is what I saw in some of the other `Interface_USB` symbols, but now I see there are other symbols with `D+` above `D-`, including all the USB connector symbols other than `USB3_*` and `USB_C_*`. Which way is preferred?

evanshultz:
First, thanks for adding this IC and for keeping an eye out when another PR was submitted. We really appreciate the community making sure desirable parts are in the library.

You've made good reference of other parts in the library. Some of them are old symbols, and the latest guidelines for the library are at http://kicad-pcb.org/libraries/klc/, if you hadn't found that already.

Ugh. There isn't a good choice of datasheet. We need an English one but I don't find anything better than the options you showed above. Perhaps we include both WCH links above. I have emailed WCH about this to see. @poeschlr ?

Since V3 cannot source current (that I can find), Passive type should be fine.

Review feedback:
- Put the ref des at top left and PN at top right.
- Use `USB serial converter, UART, SOIC-16` for the description to make it unique and extensible.
- Place pin 4 at the top.
- Pins 5 and 6 can be named `UD[+|-]` to match the datasheet pin names.
- Moving V3 to the top should allow pins 7 and 8 to space apart so a crystal symbol can fit there nicely.
- I think it's more salient to put pins 9-14 in order top to bottom. The KiCad library expressly does not make the logical symbol look like the physical package. But if you like it this way it's OK too.

GrantM11235:
I hope WCH gets back to you, it would be great to have an official updated English datasheet!

In the mean time, I made all of the other changes you suggested. This is what it looks like now:
ch340g

G-Pereira:
Why is V3 passive?

GrantM11235:
Here is my best guess about what V3 does:
USB signaling is 3.3V, so the USB part of the chip needs that voltage even when the UART part is using 5V. Therefore the chip needs an internal 3.3V regulator which requires a decoupling capacitor on the output. When VCC is 3.3V, the internal regulator will not be able to output 3.3V due to its dropout voltage, so it must be bypassed by tying V3 directly to VCC.

evanshultz:
I would agree about the function, but those guesses don't align into a single pin type. Nor does it matter since in neither condition is the pin tied to the pin of another IC or to be used as a current source. So, Passive works and shouldn't cause any false negatives in the ERC engine.

G-Pereira:
No No No!! USB signaling is 5V!
The thing here is that there are MCUs that you want to use this chip on, for example an STM32H7 which have 3.3V logic. If you want to program it you should use that V3 pin as @evanshultz metioned in the first comment (the second doesn't make sense at all, sorry nothing personal :p ) so as to be powered at 5V and receive the 5V signaling from USB BUT output to the MCU at 3.3V which I am looking for to program an STM32H7.
Hope that gets things more clear :)

G-Pereira:
In my opinion V3 should be an input since it acts like one.

GrantM11235:
USB VBUS is 5V, but the signaling for low-speed and full-speed is 3.3V https://en.wikipedia.org/wiki/USB#Electrical_specification

diggit:
V3 is probably regulator output pin for USB transceiver and core logic. Vcc defines logic levels on UART side. When you power this IC from 5V, internal LDO is used and V3 pin is power output of this LDO. If you want 3.3V levels on UART side, you'll just short V3 and VCC and power whole IC from 3.3V.
IMO **passive** type should be used here.

evanshultz:
@diggit 
Thanks for the brief description. I probably was less helpful rather than more by writing too much.

@GrantM11235 
If you're OK with everything, let's give it another day or two for WCH to respond and then I'll merge either way. Thanks for making all the updates so quickly!

GrantM11235:
Sounds good to me

G-Pereira:
Sorry for the wrong info

GrantM11235:
@evanshultz Have you gotten a response from WCH?

poeschlr:
What datasheets to distributors provide? (So digikey, farnell, ...)

evanshultz:
That's just it: none of them I can find carry the IC by itself. SparkFun hosts the unofficial English datasheet. Other than that, it's the options above from the vendor, neither of which are great.

Personally, I'd go with the unofficial one so at least it's English and covers this specific IC (the "G" suffix).

evanshultz:
@poeschlr 
Can you pick the option you prefer for the datasheet? You can see more details above.

The options I see are:
1. Official Chinese for CH340G.
2. Official English for other members of the family but not the G version created here.
2. Unofficial English just for the G version.

I prefer the third one.

evanshultz:
@poeschlr ?

poeschlr:
Ok lets go with option 3 here.

evanshultz:
Great.

@GrantM11235 
Please update the datasheet and I'll merge.

GrantM11235:
Done

evanshultz:
Thanks for sticking it out through this saga!

PR #824, opened 2018-08-15
Add LMV727x low power comparators

RockyTV:
Datasheet: http://www.ti.com/lit/ds/symlink/lmv7272.pdf


evanshultz:
- The singles should list only one package in their description
- The singles can be two ALIASes of LMV331 (I'm going to separate them by package once this PR gets merged so this is fine for now; see https://github.com/KiCad/kicad-symbols/issues/164)
- Assign a default footprint to 7272; it's annoyingly not quite our existing `Texas_DSBGA-8_1.43x1.41mm_Layout3x3_P0.5mm`
- 7272 should have 3 units: each logical gate on the first two and the last unit for the power pins (see LM2903 for an example)

RockyTV:
> The singles should list only one package in their description

Even though the datasheet specifies that both packages are available?

I'll get right onto it.

evanshultz:
Oh. Yes. You are correct on that point. Sorry. I was reading the table top-to-bottom and thinking the part number defined the package, but you're right. The output stage is different and both packages are available. Thanks for checking and verifying.

antoniovazquezblanco:
@RockyTV Could you please address the applicable comments above and please fix the merge conflicts?

Thank you!

RockyTV:
I am terribly sorry for leaving this, real life kept me busy. I've fixed the issues and hopefully it should be good now!

evanshultz:
Thanks for making changes! We understand real life is all-consuming sometimes.

Note that we prefer incremental commits so we can see changes from the last push, please.

A bit more work is left:
- Can you list the output circuit type for 7272 in the description as you have for the others?
- You added LMV331 as an ALIAS while leaving that symbol. What you should do is totally remove your 7271 and 7275 symbols and all them as an ALIAS to the existing LMV331 symbol.
- Are you willing to make the missing footprint for 7272?

evanshultz:
Also note the errors caught by the CI checks above.

RockyTV:
>     * You added LMV331 as an ALIAS while leaving that symbol. What you should do is totally remove your 7271 and 7275 symbols and all them as an ALIAS to the existing LMV331 symbol.

So, just remove the symbols then, and keep the `ALIAS` field?

>     * Are you willing to make the missing footprint for 7272?

I can try. I know that there's this image, taken from the datasheet, that tells me all I need to create the footprint. However, what's the distance between the pads and the outline? And what's the diameter used for pads - is it 0.335mm, 0.305mm, 0.275mm or 0.250mm?

evanshultz:
You will add your symbols as an ALIAS to the existing LMV331 symbol. And the two single-gate symbols you added get removed. I think that's what you're saying but hopefully I'm explaining it well.

The dimensions given are min and max, so you the mean value as nominal to construct the footprint. I mentioned a fairly similar footprint above you could use as a reference and we also have a nice Python BGA footprint generator at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Package_BGA if you don't want to do it manually. But either was is fine. Please use http://www.ti.com/lit/ml/mxbg270/mxbg270.pdf as the documentation for the footprint since it's part-agnostic.

RockyTV:
> You will add your symbols as an ALIAS to the existing LMV331 symbol. And the two single-gate symbols you added get removed. I think that's what you're saying but hopefully I'm explaining it well.

Sorry, I didn't understand it. What I did was edit the symbol properties, for both LMV7271 and 7275 and added an alias to LMV331. Is it correct? If not, what are the steps I should take to do it correctly?

evanshultz:
Here is what you should do:
1. Remove 7271 symbol.
1. Remove 7275 symbol.
1. Open LMV331 and add two ALIASes, one for each of the symbols removed above.

Does that help?

For the footprint, if making it manually is easier that's perfectly acceptable. I don't know of any usage instructions but in general you will execute the Python script and feed it an argument of the YAML file to which you've added the new entry for this package. And again, if you prefer to make the footprint by hand that's OK. If you're unsure about making footprints let's get the rest of this PR updated so only the footprint is missing and then we can focus on just the footprint. :)

RockyTV:
Would you like a separate PR for the footprint?

evanshultz:
Yep, that's it.

Please fix the pin position of LMV7272's V- pin.

So you are willing and able to make the footprint? The footprint will have to be a separate PR since footprints are in a different repo. Once the footprint is merged, it can be assigned to the 7272 symbol and this PR can also be merged.

evanshultz:
@RockyTV 
Could you please fix the pin position?

I can make the footprint for you if that helps. I'm delighted that you chose to contribute and I don't want the footprint to be a barrier. Let me know.

RockyTV:
There you go!

> I can make the footprint for you if that helps. I'm delighted that you chose to contribute and I don't want the footprint to be a barrier. Let me know.

I want to give it a try, by using the script. Should I just copy the remaining arguments from already existing footprints in the list, and just replace pin radius, body size and pin distance?

evanshultz:
The pins are good now. Thanks!

There is no footprint assigned to this symbol, which means it won't be usable right now. Would you like to add the footprint as discussed above?

RockyTV:
Sorry for the delay again! But it should be good now. I suppose the footprint PR must be merged before merging this one?

evanshultz:
Thanks!

Yes, please submit the footprint you created and, once that's merged, the symbol can also be merged. I see that you submitted a PR to update the generator script, but you also need to take the footprint itself and make a PR at https://github.com/KiCad/kicad-footprints to submit it.

evanshultz:
Please assign the footprint just merged as the default footprint to the relevant symbol and make the footprint filter for that symbol a bit more robust. Thanks!

RockyTV:
What do you mean with "footprint filter"?

poeschlr:
The footprint filter does not really assign a footprint. It just limits what is shown in cvpcb (if the filter is selected.)
To assign a default footprint you need to fill out the footprint field. For more details you might want to take a look at [this tutorial on the user forum (Section "Setting the footprint field")](https://forum.kicad.info/t/tutorial-how-to-make-a-symbol/13336/3?u=rene_poschl)

RockyTV:
Is the filter good?

poeschlr:
we typically do not include the pin count in the filter as this is done by a separate filter in cvpcb

`Texas*DSBGA*1.5195x1.5195*Layout3x3*P0.5mm*` might be a better choice.

evanshultz:
This looks good now. Thank you!

PR #828, opened 2018-08-17
Gate drivers

SchrodingersGat:
This PR cleans up symbols for many of the dual-channel gate drivers, and adds some new ones.

**IX442x Drivers ** - http://www.ixysic.com/home/pdfs.nsf/www/IX4426-27-28.pdf/$file/IX4426-27-28.pdf

_DFN-8 Package_
image

_SOIC Package_

image

**UCC2752x Drivers** - http://www.ti.com/lit/ds/symlink/ucc27528.pdf

_DFN-8 Package_
image

**FAN322x Drivers** - http://www.onsemi.com/pub/Collateral/FAN3224TU_F085-D.PDF

_DFN-8 Package_

image

**2EDN752x Drivers** - https://www.infineon.com/dgdl/Infineon-2EDN752x-2EDN852x-DS--DS-v02_03-EN.pdf?fileId=5546d462525dbac40152abcc7dbb1727

image

Footprint PR - https://github.com/KiCad/kicad-footprints/pull/831


SchrodingersGat:
The required footprints are available here - https://github.com/KiCad/kicad-footprints/pull/831

evanshultz:
@SchrodingersGat 
Can you fix merge conflicts and update the IXYS DFN8 package name to the one Rene submitted?

evanshultz:
Thanks!

- Do we use the Inverted pin style? I haven't seen it elsewhere except on decrepit symbols. Maybe an overbar on the pin name or even nothing for the most ALIASage? @poeschlr ?
- Many pins were changed from 200mil to 150mil, but should be 100mil.
- Can EL27* and MIC44* (and others?) be ALIASes? I only checked pin number and they seem the same.
- Please put the body size in the footprint filters.
- IX44*MTR should have the PAD connected to GND (datasheet page 13 note 2).
- UCC package is WSON, not DFN.
- TI parts don't have underscores in the pin names.
- The footprint you chose for the TI parts has an overly big EP, which could cause floating/shifting. Can you use a footprint with an EP the same size as the package's EP?
- Please note in the description that UCC2727528 has CMOS inputs.
- FAN* parts also don't have a matching EP size.
- Pad of Infineon pads must be connected to ground (datasheet page 8).
- Infineon package EP also doesn't match the footprint EP size.

SchrodingersGat:
Given that this PR has ballooned into a multitude of new symbols, I might start again, with a simpler subset, and then go from there. Your suggestions above are noted.

PR #831, opened 2018-08-18
Better descriptions and keywords for Device lib

fauxpark:
The generic device library has a few spelling/capitalization errors and inconsistencies. Many of the transistor, DIAC/TRIAC, SCR, dual diode symbols etc. didn't seem descriptive enough to me, so I added the pin ordering like so:

```
D_Schottky_x2_Serial_AKC
Dual Schottky diode, anode/cathode/center

Q_NMOS_SDGD
N-MOSFET transistor, source/drain/gate, drain connected to mounting plane

LED_RCBG
RGB LED, red/cathode/blue/green

Laserdiode_P_TYPE
Laser diode with photodiode, center on pin 2, PD cathode on pin 3
```

I have not modified the symbols themselves, only the DCM descriptions and keywords.

I have a question about the keywords though: is putting "transistor" in the keywords redundant when the description also contains the word, or is there some other use for the keywords field?

------------

### DIFF
--- a/Device.dcm
+++ b/Device.dcm
@@ -73,13 +73,13 @@ F ~
 $ENDCMP
 #
 $CMP CP1
-D Polarized capacitor, alternative symbol
+D Polarized capacitor, curved cathode
 K cap capacitor
 F ~
 $ENDCMP
 #
 $CMP CP1_Small
-D Polarized capacitor, small symbol, alternative symbol
+D Polarized capacitor, curved cathode, small symbol
 K cap capacitor
 F ~
 $ENDCMP
@@ -193,13 +193,13 @@ F ~
 $ENDCMP
 #
 $CMP DIAC_ALT
-D Diode for alternating current, alternative symbol
+D Diode for alternating current, filled shape
 K AC diode DIAC
 F ~
 $ENDCMP
 #
 $CMP D_ALT
-D Diode, alternative symbol
+D Diode, filled shape
 K diode
 F ~
 $ENDCMP
@@ -241,7 +241,7 @@ F ~
 $ENDCMP
 #
 $CMP D_Capacitance_ALT
-D Variable capacitance diode, alternative symbol
+D Variable capacitance diode, filled shape
 K capacitance diode varicap varactor
 F ~
 $ENDCMP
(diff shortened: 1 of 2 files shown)


fauxpark:
Some more notes:

 - Amperemeter_AC/DC should probably be renamed to Ammeter since that's what they're most commonly called (at least, that's what the Wikipedia page is called!).
 - There are a bunch of symbols with names containing all-caps words that aren't acronyms, eg. alternative symbols often have the suffix _ALT where they should probably be _Alt, dual diodes contain DUAL instead of Dual, the memristor and spark gap symbols are all caps.
 - Similarly, CP1 should be renamed CP_Alt to match the naming convention.

poeschlr:
I would at least include ampere meter (two words) in the description for every non native English speakers out there who will not get the idea that people could have the idea to shorten these words. (It would also allow searching for ampere)
I guess "Ampermeter"  comes from the German way of writing it. (Not sure how other languages deal with compound words.)

---

We should not do a massive renaming shortly after a release. (This should either have been finished before v5 or it will have to wait till version 6)
Right now this PR is ok from that point of view as you only update the descriptions.

fauxpark:
Fair enough on both points. I notice that searching for "meter" in the Choose Symbol dialog brings up the ammeter symbols regardless of whether the term is in the keywords, as the name and description both contain the string "meter" anyway. I would think you could get away with just having "ampere" in the keywords (if it is renamed to Ammeter). But I'm still not certain how and where the keywords are used.

evanshultz:
Thanks! This is a nice improvement.

I have a few pieces of feedback:
1. Net ties still have a capital letter in the keywords
2. Generic opamps probably shouldn't have any packages mentioned
3. If the symbol is an "alternate/alternative", the description should mention what is unique about it since the words "alternate" and "alternative" don't mean anything in isolation (or when looking at the symbol browser that has no way to compare/view multiple symbols at once

fauxpark:
I've done 2 and 3, but not sure what you mean by 1. As far as I can tell the keywords field is case-insensitive so I've just made them all lowercase apart from things like NPN, SCR etc. for consistency.

fauxpark:
@evanshultz just bumping this :)

evanshultz:
Thanks for pinging me!

By #1 above I meant that I am daft. All is good.

One more thought: I've recognized the curved cap symbol as US and the straight one as EU. Google "capacitor symbol eu us". I wonder if that's a better way to differentiate their descriptions? We do something similar for resistors.

fauxpark:
Yeah, that sounds like a better idea. Fixed.

evanshultz:
Thanks!

PR #833, opened 2018-08-19
Fix x10 and x11 voltage divider network rendering

fauxpark:
Currently these two symbols look like this:
sym
One side of the rectangles is in the wrong place (+/- 500mil).

They appear to be generated with [this script](https://github.com/KiCad/kicad-library-utils/blob/master/schlib/autogen/R_NETWORKS/make_R_NET.py), but the bit rot is starting to set in I think.

------------


jkriege2:
Thanks,
JAN

PR #838, opened 2018-08-20
Added symbol for cree CLS6B-FKW, QLS6A-FKW and QLS6B-FKW

jneiva08:
Added symbol for CREE QLS6A-FKW, CLS6B-FKW and QLS6B-FKW led's

Datasheet:

CLS6B-FKW: http://www.cree.com/led-components/media/documents/CLS6B-FKW.pdf
QLS6A-FKW: http://www.cree.com/led-components/media/documents/1381-QLS6AFKW.pdf
QLS6B-FKW: http://www.cree.com/led-components/media/documents/1397-QLS6BFKW.pdf

symbol for CLS6B-FKW and QLS6B-FKW:
led

symbol for QLS6A-FKW:
ledz

PR for footprint is [#850](https://github.com/KiCad/kicad-footprints/pull/850)

Hop i didn't miss anything. The datasheets is a litle confusing because of model with zenner.

------------


evanshultz:
- For QLS6A-FKW, can you please put 10mil filled circles where the lines touch to show the connection?
- That being said, aren't the PNs swapped? It looks to me like the zener parts are using the non-zener symbol and vice versa. Please check in case I've gotten confused too!
- Zener symbols should be unfilled. This isn't consistent everywhere in the library yet, but this is the IEC symbol which we specify. You can see the generic symbols in Device.lib.
- Please add `PLCC-6` to the end of the description. This library is especially poor in this respect but we should at least follow the rules with new symbols.
- Two symbols have an Avago datasheet.
- What's the reason to put the LEDs in this order? I would have normally thought R-G-B top-to-bottom, and that allow follows the pin numbering, but it's fine this way. I'm just wondering why it's G-R-B.

jneiva08:
Have added footprint name on description and updated datasheets.

Now the confusion part:
CLS6B-FKW
 - Datasheet: not mention Zener
 - Symbol (pag.10): have zener on Blue and Green
QLS6B-FKW
 - Datasheet: not mention Zener
 - Symbol (pag.10): have zener on Blue and Green
QLS6A-FKW
 - Datasheet: says zener protection
 - Symbol (pag.10): don't have zener

Have made the symbols based on datasheet description.
I can change the order to RGB instead of GRB, but will need to increase the size.

@evanshultz please advise how should preoceed.

jneiva08:
@evanshultz , Sorry for taking a long long time to fix.

Appears CREE have fixed the datasheets, I have made all the modifications.
All symbols are in RGB instead GRB
The QLS6A-FKW grow a little to put the zenner.

CLS6B-FKW/QLS6B-FKW
cls6b

QLS6A-FKW
qls6a

Hope this time can be merged :)

jneiva08:
@evanshultz , sorry to take a long time to note it have S3.3 KLC issue in one of symbols.

evanshultz:
- I checked against the links in the first post and now the zeners match for me as well. Great.
- Please unfill the diodes.
- Is there a reason why the left-side pins aren't 1, 3, 5 from the top to bottom? It's not wrong the way it is, but it would seem like the natural way to order the pins. (I may not have been clear above so this is me trying to ask in a different way.)

evanshultz:
Ah! I think you were saying this above about LED order but I didn't understand it before. Now I get it. Thanks for the clarification and this is good with me. I apologize for my slowness to grasp your point.

Two more things to align with the normal IEC LED symbol:
- Put a line through the LEDs (this actually saves one line)
- Place a 10mil diameter dot at each wire junction

Then it's ready to merge!

jneiva08:
Now the symbols looks like:

cls6b

qls6a

evanshultz:
Slight simplification of the symbols may be possible (like turning multiple polylines into a single polyline), but it's not worth it.

Upon a final review I think the footprint filter can be updated. Please change to `LED*Cree*PLCC*4.7x1.5mm*` to do two things: use asterisk whenever possible and remove the pin count. Then I think it's ready. Thanks!

jneiva08:
Have updated the footprint filter

evanshultz:
Great! Thank you and sorry I missed that earlier.

jneiva08:
Thanks for your time

PR #840, opened 2018-08-22
Added INA240

SchrodingersGat:
* INA240AxPW - TSSOP-8 package
* Aliases for each gain value
* http://www.ti.com/lit/ds/symlink/ina240.pdf

image


SchrodingersGat:
image

Pin 1 (marked as NC) should be connected to GND as per datasheet. The symbol has been updated to reflect this.

SchrodingersGat:
image

SchrodingersGat:
For some reason the unrelated AD8213 part was triggering a KLC error so I fixed that also.

evanshultz:
The description as some doubled info. Perhaps matching the datasheet with `High- and Low-Side, Bidirectional, Zero-Drift, Current-Sense Amplifier With Enhanced PWM Rejection, 20V/V, TSSOP-8`? Then I think it looks ready to merge.

It appears https://github.com/KiCad/kicad-symbols/pull/840/commits/1d5fb4ab47380118b1695a4949bc317bac2b5da1 added a second pin 8 to AD8213.

SchrodingersGat:
Good catch @evanshultz - I have fixed this now.

SchrodingersGat:
Thanks @evanshultz  -  I have a number of other outstanding symbol PRs that should be pretty good to merge. Do you have any time to take a look?

PR #841, opened 2018-08-22
Add MCP1804

evanshultz:
http://ww1.microchip.com/downloads/en/DeviceDoc/20002200D.pdf

image

image
image
image

------------


evanshultz:
Thanks for looking.

See http://kicad-pcb.org/libraries/klc/S2.2/. Temp and packaging are explicitly listed as things to wildcard since they don't affect the function (with the rated temp range, anyway) of the part. KLC is always up for debate, but I've followed the instructions as they stand now.

diegoherranz:
I'd understand that if there were different packaging (tube, tray, etc.) or temperature (industrial, commercial, etc.) variations to chose from. But in this case it's just Tape and Reel and Industrial temperature range. No other options are possible so I would say that that KLC doesn't apply here.

diegoherranz:
If you still think, @evanshultz, that those wildcards should remain, please let me know and I'll finish the review of these symbols.

evanshultz:
Yes, I do think they should. https://github.com/KiCad/kicad-library/pull/1474 is an (old) PR where this was discussed and requested by two librarians who are not me.

diegoherranz:
Part number combinations look correct to me.

MCP1804x-yy02xMB variant (SOT-89-3) reuses the symbol from MC78L05_SOT89 - Pin names do not match (IN vs VIN) but I guess it's OK not to have a new symbol (it's perfectly clear which pin is which).

MCP1804x-yy02xDB variant (SOT-223) despite being the same pinout as MCP1804x-yy02xMB, uses a new symbol because it's a different footprint - OK, however, the pin names could be fixed on the this one to match the datasheet although, again, it's pretty clear which is which.

MCP1804x-1802xMT (SOT-89-5) and OT (SOT-23-5) variants have new symbols too. Both pinouts and pin types look OK.

I've also checked the footprints used against the datasheet recommendations:
- Package_TO_SOT_SMD:SOT-89-5 is nearly identical to the footprint on the datasheet. Only difference I've seen is the trapezoidal shapes being slightly different but close enough - OK
- Package_TO_SOT_SMD:SOT-223-3_TabPin2 is slightly different but close enough.
- Package_TO_SOT_SMD:SOT-23-5 is notably different but I think that it should still work OK.
- Package_TO_SOT_SMD:SOT-89-3 is notably different but I think that it should still work OK.

evanshultz:
Thanks! I appreciate your time to take a look.

SOT89: That was my thought. We do this on opamps by always using `V+` and `V-` for power pins so that the symbols are consistent and the text fits within the same defined space. And the same thing happens for other linear regs where ALIASes are used.

SOT223: For the same reason, I stuck with our "standard" pin numbers for generic regulators. It could change, but keeping symbol size and style seemed more important since I also felt function was obvious.

Footprints: We're been moving to IPC-compliant footprints instead of manufacturer-specific ones if the package type is generic. All of these fall into that category.

@Ratfink @antoniovazquezblanco (or any other librarian)
If you're OK with the above, and don't find anything else, is this ready to merge?

evanshultz:
@Misca1234 @antoniovazquezblanco @DanSGiesbrecht 
Is anyone able to review the above and merge this?

DanSGiesbrecht:
@evanshultz my lunch break's in an hour. I can take a look at this.

DanSGiesbrecht:
@evanshultz just a few questions:
- I noticed a few deviations from the KLC in the Travis check. I assume that this was done to match existing symbols. Is there a desire to eventually make linear reg's conform to the following?
  - 20mil name offset from pins.
  - Centered bodies on origin.
- Is it a guarantee that the tab on SOT-223 is connected to pin 2 (GND)?
---
I only see one potential improvement:

I think that it would improve clarity by changing `SOT-89` to `SOT-89-3` in the description for the 7 symbols ending in `-MB`, to avoid confusion with the 5-pin parts.

evanshultz:
Thanks for looking!

- Yes. For these symbols with a desired look we used 10mil pin name offset. The centering algorithm is weak and often ignored. There are several cases like this where it's been brought up to do library-by-library rules in the check scripts but nothing came of it.
- Yes, as shown on page 16 of the datasheet:
image

Yep, that would add clarity. Updated those descriptions.

DanSGiesbrecht:
Excellent!

Issue #429, opened 2018-03-29
Add isolation lines to ACDC converter symbols 


antoniovazquezblanco:
As someone pointed out in https://github.com/KiCad/kicad-symbols/pull/415 it is desirable to clarify that some converters are isolated by adding lines to the symbol.

antoniovazquezblanco:
Related issue https://github.com/KiCad/kicad-symbols/pull/427

Ratfink:
Could someone provide a reference document that talks about these lines in schematic symbols?  I'm not familiar with them, but that could just be my I'm-not-an-EE showing.

poeschlr:
I always thought the diagonal double line was standardized but i can't find a reference to it. (So i might be wrong)

The only things i could find are these two images as reference for what we could use. I have a slight preference for the one using the diagonal line. But the other one might make fitting pin names easier.

![original source](http://indexxit.com/wp-content/uploads/2017/10/marvelous-converter-recommended-using-circuit-and-precautions-isolation-diagram-design-in-smps-circuits-for-current-voltage-sensing-purpose-monitor-breaker-optocoupler-ecg-wikipedia-970x719.jpg)
<img src="https://user-images.githubusercontent.com/18327350/38135233-89b419c0-3406-11e8-8294-f49567c88374.jpg" width=300>


![762084-dc-ups-din-rail-300x-00585384_0](http://www.lutze.com/fileadmin/tx_luetzeasimfetch/762084-dc-ups-din-rail-300x-00585384_0.jpg)

evanshultz:
I'm not familiar with any standard here. It's an important thing to communicate, but it's done differently by various folks who provide isolated products.

I skimmed through a few UL and IEC standards and didn't see any graphics they provide. I assume that would be considered a standard we could leverage if they had anything.

IMO, dashed vertical lines are nicer for building symbols as mentioned above. They also communicate something different than a solid line (usually a wire) in a black & white image (like a printout). It's unfortunate KiCad added line styles in Eeschema for v5 but not in the symbol editor. :(

More examples of integrated power supplies:
https://www.cui.com/product/resource/pbk-1.pdf
![image](https://user-images.githubusercontent.com/1936989/38145541-a636c7bc-33fe-11e8-8318-48f27aef37f5.png)

http://cdn.vicorpower.com/documents/datasheets/vibrick/VI_BRICK_175V_48V_330W_PFM.pdf
[...]

PR #845, opened 2018-08-23
3W and 5W AC/DC converters from Hi-Link.

Sthing:
Footprints in https://github.com/KiCad/kicad-footprints/pull/860

## 3W modules
HLK-PM03, HLK-PM01, HLK-PM12

http://www.hlktech.net/product_detail.php?ProId=60

image

## 5W modules
HLK-5M03, HLK-5M05, HLK-5M12

http://www.hlktech.net/product_detail.php?ProId=54

image

------------


myfreescalewebpage:
Hi @Sthing, thanks for contributing,

- [ ] I have not found datasheets in PDF format, not sure they are available on the manufacturer website, but it's better if you have a link ?
- [ ] HLK-PM12 datasheet is in PDF format but it is better if you redirect to the manufacturer website because link can be break at any time by this site
- [ ] You should cross -Vin and +Vin to get positive pin on the top right and negative pin at the bottom right
- [ ] Outputs pins are named "+VO", "-VO", not "+Vin" and "-Vin". To be consistent, I will call inputs "AC1" and "AC2" according to the manufacturer names "AC" and "AC" (we can't put the same name for this)
- [ ] Can you make the symbol horizontally symmetrical regarding to the "isolation dashes"  in the box ?
- [ ] You are violating KLC S3.6 and S5.2 that are easy to solve, please check at the travis output.
- [ ] For consistency with the other devices in the library, remove current in the description, for example look at "HS-40003" : "3.3V, 3W, AC-DC module power supply, Hahn"
- [ ] For consistency with the other devices in the library, remove current in the keywords, for example look at "HS-40003" : "3.3V 3W AC-DC module power supply"

Cheers,
Joel

evanshultz:
- Change pin name offset to 20mil.
- Pins 1 and 2 may be correctly named just `AC` as it may not matter which is connected to Line and which to Neutral. It would depend on the circuit design inside the module.
- At https://github.com/KiCad/kicad-symbols/issues/429 we decided to use a single dashed line to separate Primary and Secondary pins in this type of symbol.
- If some of the PNs are `PMxx` and some are `5Mxx` the footprint naming should indicate that, so please consider that in the footprint name and, once the footprint is merged, update this PR to match.

myfreescalewebpage:
@Sthing up ?

Encrypt:
Hello everyone!

I've just noticed that these symbols were added to the library in #1147 
I guess you can close this issue :)

myfreescalewebpage:
Hello @Encrypt thanks for pointing this to our attention! I was wondering also if original author as stop working on this subject. I label with Abandoned and Duplicated tags and close the PR now.
Also closing the footprint PR https://github.com/KiCad/kicad-footprints/pull/860 for the same reason.
Joel

PR #847, opened 2018-08-24
Added LM675

LJJS:
I added the LM675 from Texas Instruments to the KiCad Symbols due the fact it have different Pin numbers than the most opamps.
The associated footprints match the official footprint library from KiCad.

Datasheet:
http://www.ti.com/lit/ds/symlink/lm675.pdf


evanshultz:
- Add `-5` to the end of the description
- There are two Package Drawings listed on page 12 of the datasheet but only one drawing; http://www.ti.com/lit/ds/symlink/lm2941c.pdf appears to show they can share a footprint (not a 3D model) as the lead lengths are different but the stagger is the same; the current paradigm would result in a default footprint named `TO-220-5_P3.4x4mm_StaggerOdd_Lead3.71mm_Vertical` (or similar), but see https://github.com/KiCad/kicad-footprints/issues/764 since these should be changing (I'll try to get to this in the next couple days)
- The footprint filter should then be updated to match the new footprint

Also, please always put a screenshot of the symbol in the body of the PR for easy checking. Thanks!

LJJS:
The only difference between the two Package Drawings listed on page 12 is if they have a metal or a plastic flange.

LJJS:
Just for completeness, a screenshot of the symbol
lm675

LJJS:
Due an issue with Github Desktop I had to make an new PR https://github.com/KiCad/kicad-symbols/pull/940

This PR can be closed.

PR #849, opened 2018-08-25
Fix multiple Flash Memory symbols

DanSGiesbrecht:
**Full Disclaimer:** This is my first PR. Advice on how to improve is welcome.

- Is this PR too short?
- Should my branch be named differently (currently it is named to reflect a personal project)?

---
I have updated 8 flash memory symbols with the correct backgrounds:

- 29F010-TSOP-SP
- 29W040
- AM29PDL128G
- M29W004
- M29W008
- SST39SF010
- SST39SF020
- SST39SF040

E.g.
original

fixed


poeschlr:
We might want to wait with merging this till after 5.0.1 is released as it does change pin locations of existing symbols.

DanSGiesbrecht:
@poeschlr thank you for the reply. Will there be a stable master branch, and a 5.0.1 candidate? This workflow is very new to me. If it's valuable, I could cherry-pick some of these commits into a new PR (e.g. fix background, fix incorrect pin duplicate found on one of the parts).

poeschlr:
I doubt we have the manpower to work on two different branches. It might just be that some fixes are simply not allowed till we start the v6 development. It all depends a bit if we can get the devs to include the lib in 5.0.1 with the changes we already made. (I have not yet checked how much we changed since the v5 release)

poeschlr:
If you could revert the position changes to this and simply document in an issue that some of the symbols have off grid pins then we can merge your fixes.
Sorry for creating additional work but i feel it is worth it. (Teaching travis to report fixes differently to new additions might be a requirement if we want to go that route.)

DanSGiesbrecht:
It's no problem! It's a good learning process. I'd much rather do extra work to do it right. Would the best way to revert be to close this PR, and simply remove unnecessary commits from a new PR?

poeschlr:
Depends. If you remember which symbols you moved by how much then it might also be possible to manually move them back. This mostly depends on if it is easier to move them back or start fresh and do all other changes there.

DanSGiesbrecht:
If pin-alignment compatibility is the only issue and all other changes are acceptable (e.g. symbol field changes, datasheet links), I'll look into manually moving them back and submit another commit. Thank you again for the detailed feedback.

poeschlr:
The only things that can create problems with existing designs are changes to pin positions and changing the symbol name. Everything else is ok to change.

DanSGiesbrecht:
Rather than trying to messily revert commits involved in this PR, I have created a new PR #862 , focusing on important changes/fixes only. I've chosen to ignore other Travis-CI errors for now.

PR #857, opened 2018-08-29
Add FT24C04A - EEPROM Memory IC 4Kb (512 x 8) I²C 1MHz 550ns SOT-23-5

nhatkhai:
Add FT24C04A - EEPROM Memory IC 4Kb (512 x 8) I²C 1MHz 550ns SOT-23-5

* [datasheet](https://websites.godaddy.com/blob/d0f1813f-418c-4780-8b0e-f456f226885c/downloads/1bi7berke_536618.pdf)

* image


nhatkhai:
Why KiCad don't allow to have filed for manufacturer, part number, longer description? and referred supplier info for component that semi-atomic?

poeschlr:
Supplier is not universal, manufacturer part number is already in the symbol name.

Not sure what your point about the description is. We have no real restrictions for what you can put into the description field.

By the way atomic does not mean what you think it means. A fully specified symbol is a symbol with all the information added to it. Only if you also have the footprint specialized for this one part number can the pair be called atomic.

nhatkhai:
To me, symbol name is not equivalent with part number - it usually a family of a same ICs with same functions (except like performance, tolerant, package/footprint type, ... which are not need to be show in schematic, but very useful for selecting part, and generate complete BOM directly from schematic). So without a referenced manufacture name, full manufacture part number, event referenced supplier:
* In the long run it will be hard to find where is the symbol come from. 
* I will have to manually add all the field every time I drop one symbol into schematic. 
* Importantly scripting for searching parts, without these reference info, it would be hard to guarantee find you a correct parts (references supplier in this case help).
* Provide, multiple points checking for minimize errors from design to generate BOM.
* ICs manufacture now a day seem to shift, move, and obsolete parts constantly. With any addition info would help to re-track, find equivalent part that meet original intend of the design much more simple.
* Datasheet link are move around all the time, with these addition info will help you found it.

Giving that, I'm thinking you may need to add naming convention for these optional fields so people can share symbol/footprint more effectively. I think these are a least minimum info for tracking the correctness of symbol without present of datasheet.

Thank for clarify atomic definition. I still don't understand what is the issue by calling this part as semi-atomic.

poeschlr:
The official kicad library has its rules. We can not make it right for everyone. All of our rules have been discussed in the past. A large part of the decision is how easily maintained the lib is with a proposed rule change. All of your suggestions would add things that need to be maintained regularly. I am not really prepared to include this at this point in time.

nhatkhai:
I understand, quality is not free. But in the open source world, I though, maintaining for optional fields aren't be needed if you willing to let people going wild with these fields (it will eventually corrected). Regardless, I respect all open source maintainers decision. 

I will update this pull request next time around. I guest the only thing holding this request back is the extra fields I have in the symbol base on the travis-ci report. It that correct?

Misca1234:
I suggest the following 

Rename this into 
FT24C04A-UL  so it is declared that it is the SOT variant, so the other variants can be added laater

Remove filed F4, F5, F6 and F7

Add the aliases 
FT24C08A-UL and FT24C16A-UL
with changed description to reflect the larger memory size

You should change your symbol into something like this
if you feel it is to "crowded", you can expand it in x-led

bild

evanshultz:
- Change the datasheet to https://img1.wsimg.com/blobby/go/d0f1813f-418c-4780-8b0e-f456f226885c/downloads/1bs92tefg_846599.pdf (it looks horrible but it's the offical one from https://fremontmicrousa.com/products-memories).
- Comma separate the description.
- Pin 5 is apparently NC, not WP, so make this No Connect pin and bury it inside the symbol body.
- Using the new datasheet above, the symbol name should be `FT24C04A-xLx` (or just `FT24C04A-xL`).
- Make the footprint filter `SOT?23*`
- Capitalize the letters in pin 4 (as is done above).

nhatkhai:
@evanshultz - Thank to point out the datasheet of other same function part. This part seem to be fishy on the production (look like no longer manufacture, but digikey indicate Active!).
@Misca1234 - I known that Vertical space are not as large as horizontal space in the typical paper, your suggestion prevent me to maximize the usage of a page of schematic. 

I also realize that, your library already have many same EEROM symbols for this type of part, so this pull request may not add any value. I will drop this pull request. (Once can just pull out a symbol from  AT24CS32 as substitute of this - but my opinion is that, your symbol are eating too much space for optimize a schematic page)

PR #858, opened 2018-08-29
Added PIC18F2420

Misca1234:
Adding microchips MCU
PIC18F2420 both variants

(this a re push of 
in smaller pieces)

bild

bild

------------


Misca1234:
>     RE3/MCLR is input only and not bidirectional

Not if the pin is configured as an input/output RE3, this pin is a little bit special because it is a combined reset pin

>    Most PICs on the libraries have longer, more descriptive pin names (e.g. RB0/INT0/FLT0/AN12). 
> Was it intentional not to use long names here? What is the preferred way in the libraries?

There is a ongoing discussion about it and it leans towards the shorter names

> These ones split Port B (4 pins left and 4 pins right). 
> Most PICs on the libraries do not split ports even if the symbol ends up being less "symmetrical". 
> I think that is better for cases like connecting a bus to that port.

I was thinking of that too but decision was to make it as small as possible, but I change it

> Also, I would add the PIC18F2520 (both variants) as aliases. Thanks again!

It will come, this push is actually a re push of an already made push that had ALL variants included from the datasheet.
The reason for this push is to redo the original push in smaller pieces so each push is a smaller effort to review.

Misca1234:
The aliases have been added and the pin layout have changed into this

bild

diegoherranz:
Thanks for the changes.

I think that the pin layout looks better now. And having the aliases in this same PR is good too since it's not much effort.

Regarding RE3, when used as GPIO, it is input only (this is common on many PICs). So either if used as MCLR or GPIO, it's always an input.

Another minor thing. Descriptions use lower case "b" for bytes and that should be upper case "B".

Thanks!

Misca1234:
jeee, you are right it is input only, never noticed that becosue I always used the pin as  MCRR pin, fixing that

Misca1234:
Changed RE3 to input 

bild

bild

Misca1234:
Regarding the 'b' issue,
After these 4 pushes are merged I intend to overhaul this library and change the 'b' into 'B' for all and change the pin offset from 40 to 20 and the FP filter issues that is in this library

diegoherranz:
I'm back at this.

>Regarding the 'b' issue,
After these 4 pushes are merged I intend to overhaul this library and change the 'b' into 'B' for all and change the pin offset from 40 to 20 and the FP filter issues that is in this library

OK, it's OK if you fix them all in one go after this

Having another check now, I think that the DIP package (and filter) is incorrect, it's too wide. Or am I missing something? The QFN package looks correct.

Thanks!

myfreescalewebpage:
+ footprint filter should not include EP size for the QFN variant.
Joel

diegoherranz:
> Footprint filter should not include EP size for the QFN variant.

Thanks, Joel. I had missed that one.

Misca1234:
Removed the EP size from the FP filter for the QFN symbol

myfreescalewebpage:
Just looked with more attention to this pull request. Two comments:

- The EP pin number 29 is missing on the symbol (name EP, Passive, on the bottom of the symbol, NOT stacked with GND). Also just notice that the EP is not present in the footprint filter, should be exactly `QFN*1EP*6x6mm*P0.65mm*`
- The two package versions can share the same symbol design, i.e. with RA port on the left and RB and RC ports on the right for example.

Joel

diegoherranz:
Thanks, @myfreescalewebpage, for the extra review comments. Could you, @Misca1234, implement those changes?

Thanks!

Misca1234:
Fixed FP filter
Aligned the pin layout
Added EP pin for QFN

bild

bild

bild

diegoherranz:
Thanks for the changes.

I'd say that given no recommendation is done by Microchip regarding the exposed pad, it should be called `EP` and be passive as @myfreescalewebpage said.

Everything else looks fine, I think.

Thanks.

evanshultz:
See the comments at the end of https://github.com/KiCad/kicad-symbols/pull/875. There were description and keyword updates that I merged before getting implemented; since this is also the PIC18 library would you mind doing that here?

Misca1234:
EP pin fixed

@evanshultz 
You mean I should change an existing component in this push ?

diegoherranz:
The exposed pad is all good now. Thanks, @Misca1234.

I think @evanshultz meant updating these descriptions/keywords in a similar way (I think).
As far as I can tell that would be using capital "B" for Bytes (please keep lowercase "k" for kilo) and keywords: "Flash-Based 8bit CMOS Microcontroller".
Would that make sense?

Thanks!

Misca1234:
@diegoherranz 
@evanshultz 

the changes in the description is done according to diegoherranz instructions

diegoherranz:
Looks good to me. Just a very minor, super-picky comment.There's a space missing here before 1536B
If you fix that one, I'll merge it.
Thanks!

Misca1234:
@diegoherranz 
Done

diegoherranz:
Thanks for your changes (even for very minor things). Merging.

Issue #418, opened 2018-03-25
Spice


arildj78:
I\m trying to get started with KiCad and Spice, but it I can't seem to find any parts with spice-models loaded. Right-clicking a part->opening properties->Edit Spice Model seems to be empty on all parts that I'm checking... Are there special libraries for spice or am I looking in the wrong place?

diggit:
I guess it is because of model licensing being incompatible with KiCad repositories.

evanshultz:
@arildj78 
There is a library named `pspice` but it's marked as legacy. Did you check that library?

Right now, I honestly don't know. Let me take a look at this soon (hopefully with 24 hours). I haven't tried simulation yet in KiCad and at least one of the librarians should be well-versed in it. I am very familiar with circuit simulation from other tools so let me see if I can help.

evanshultz:
I figured I should start with Wayne's video at https://fosdem.org/2018/schedule/event/cad_kicad_v5/, since it publicly showed this off recently. Start at around 9:30. He references a video by Tom and Orson, which I believe is at https://www.youtube.com/watch?v=A2_-hdRcf4U.

I do not, at this point, believe any parts in the library have really been intended for circuit simulation. I could be wrong.

SPICE directives placed on the schematic page work OK. I can transform a resistor or other 2-terminal symbol into a seemingly functional V or I source. And simple parts seem to work fine using the R, L, and C primitives. But I currently haven't figured out how to get a diode or other more complex primitive to work. I can load in an external library file and it is parsed correctly, but the model isn't being attached. Only after that would I tackle subcircuits.

Are you also able to work with the normal SPICE primitives like R, L, C and sources? Those seem easiest to get going.

I'll try to figure out more tomorrow. There seem to be a number of analogs to LTSpice in the way things work in general, but I need to figure out the details. I feel pretty confident creating a nicer library of SPICE parts will be a longer-term objective that won't be ready for KiCad v5.

arildj78:
I've started to figure this out by now but it wasn't intuitive. I was hoping to see the SPICE directives when browsing the SPICE MODEL and did not realize that changing between the different tabs also changed the behavior. Often this kind of behavior is controlled by radiobuttons and not by tab selection.

arildj78:
I just realized that part of the problem was me choosing operating point as the simulation mode. I'm now able to run a successful transient analysis on a simple resistor puzzle, but when I'm changing to operating point i get this:

```
Circuit: KiCad schematic
[...]

PR #859, opened 2018-08-29
Add documentation information to pspice symbols

poeschlr:
Informs users that these are intended to be used for simulation only.

Will hopefully better protect users from this problem: https://github.com/KiCad/kicad-symbols/issues/418#issuecomment-417090171


evanshultz:
- The '0' ground symbol is missing the word "only" that you added to other symbols.
- I wonder if you should put "for simulation only" in brackets or some other special notation so it stands out more?

poeschlr:
The 0 symbol would be the only one that can realistically be used in normal designs (It is after all a normal power symbol) That is why i left out the "only"

I will think about the wording for the others tomorrow.

evanshultz:
@poeschlr 
Were you able to think about this? I'm working on the symbol update for this lib and I want to get this merged first to avoid conflicts.

poeschlr:
I did not yet come up with anything better. You are free to merge as is. (We can clarify it further later on if we can find something even better)

PR #862, opened 2018-08-30
Fix backgrounds, line thicknesses, incorrect pin number

DanSGiesbrecht:
This is a simplified PR, from items listed in #849 

Items addressed:
- All backgrounds are now filled.
- All pin name offsets are now 20mils.
- All box line thicknesses are now 10mils.
- M29W004 pin name A9 was incorrectly labelled as pin 9. It has been corrected to pin 7.

Travis-CI fails due to many other errors in the Memory-Flash library, many of which can not be resolved without changing symbol size and pin position (which would break previous design compatibility).


poeschlr:
I looked at the [datasheet of M29W004](https://pdf1.alldatasheet.com/datasheet-pdf/view/22931/STMICROELECTRONICS/M29W004/+Q2843UORlHDyRHOIpa/1XXyxeobHOoHpx+/datasheet.pdf). A9 was correctly assigned pin number 7.

poeschlr:
Scrap my comment. I confused the diff output. I thought it was at 7 and you changed it to 9.
(I really need a coffee it seems)

Thanks for your contribution.

PR #869, opened 2018-08-31
M.2 key m connector

blazra:
Hi,

adding symbol for M.2 socket with the mechanical key in the M position.

It looks like the official M.2 specification is not publicly available.
However, I did find some bootleg copy here: https://wenku.baidu.com/view/e367f8e10912a216147929e7?pn=150
Pinout for the M key socket is on page 150.

Matching footprint PR: https://github.com/KiCad/kicad-footprints/pull/879

m 2_keym_symbol


myfreescalewebpage:
Hi @blazra , thanks for contributing,

- [ ] For consistency with mini pcie connector, will say you should rename the part to "Bus_M.2_KeyM"
- [ ] Pin name position offset should be 20mils
- [ ] Footprint filter `*M.2*` will be more generic
- [ ] Remove default footprint
- [ ] Pin length should be 100mils (symbol<100pins)
- [ ] You should stack GND pins
- [ ] I will say that due to the distance between the 3.3V pins, they can be stacked in three groups: pins 2 and 4 together, pins 12-18 together and pins 70-74 together
- [ ] The connector represents the M.2 card, so I will say GND and +3V3 must be Power Input, PETxx are Inputs, PERxx are Outputs, REFCLKx are Inputs (consistent also with other signals such as SUSCLK etc)
- [ ] Pins 56 and 58 are NC in your doc, but I have seen here https://www.simms.co.uk/wp-content/uploads/2017/09/100-xBMDS-VDCTMB_MUSE-D_-TECH-SHEET.pdf that they are MFG Clock and Data, I suggest to put them visible and called MFG_CLK and MFG_DAT (without "Reserved"). A line to separate the SMBUS pins is welcome
- [ ] Module key pins 59-66 should all be invisible, and your text "(Missing pins)" is to be removed, it is not useful
- [ ] Text "For SATA signaling" to be modified to " For SATA signaling (PEDET low)" and to be moved down with SATA pins.

I haven't found a better spec so the review is performed with google, but if you can provide something better, it's welcomed !

Cheers,
Joel

blazra:
Hi @myfreescalewebpage ,

thanks for the review. I agree with most of your comments and I will change the symbol according to them, I just don't have much time right now (I should have some time again in 3 weeks).

However, there are a few things I don't understand:
> For consistency with mini pcie connector (#1196), will say you should rename the part to "Bus_M.2_KeyM"

What's the point of the "bus" prefix? It doesn't make any sense to me.

> I will say that due to the distance between the 3.3V pins, they can be stacked in three groups: pins 2 and 4 together, pins 12-18 together and pins 70-74 together

Why not stack all the 3.3V pins together?

> I have doubt about the position of the pins on the symbol, I wonder if grouping the signaling pins all together is not better, and not sure in this case the drawing inside is so much useful.

I agree, it will probably be better to just keep the diff pairs grouped and then have the rest of the pins below and ditch the drawing inside, it's not that useful.

myfreescalewebpage:
Hi @blazra 

Thanks for your response, no worries on the corrections, that was just to notify you to check if you were alive because I'm also personally interested by the subject. I have draft the following document i want to share in this PR also, can be useful for next M.2 connectors:

[M.2.docx](https://github.com/KiCad/kicad-symbols/files/2808961/M.2.docx)

Regarding your questions:

- "Bus_" because that'"s PCIe, miniPCIe, ISA etc family, that's for consistency with the existing parts in the lib, all the existing are not consistent but we decided previously to apply this logic for this kind of symbols (see my mini-pcie PR)
- 3.3V pins can have several capacitors and to indicate this we do not use pin stacking.

Joel

myfreescalewebpage:
@blazra any news here ? Thanks, Joel

myfreescalewebpage:
@blazra ping :)

myfreescalewebpage:

PR #875, opened 2018-09-01
Added PIC18F4420-xP and PIC18F4520-xP

Misca1234:
Adding microchips MCU
PIC18F4420 and PIC184520 40 pin DIP variant

(this a re push of 
in smaller pieces)

bild

------------


Misca1234:
Changed RE3 to input 

bild

evanshultz:
Since pin 1 can be a GPIO, I think it should stay Bidirectional type.

@fauxpark 
You can probably spot anything that is non-conforming now real quick. Would you mind taking a look?

MCU_Microchip_PIC18.dcm
```diff
@@ -252,6 +252,12 @@ K Flash-Based 8-Bit Microcontroller
 F http://ww1.microchip.com/downloads/en/DeviceDoc/39564c.pdf
 $ENDCMP
 #
+$CMP PIC18F4420-xP
+D 16K Flash, 768b SRAM, 256b EEPROM, ADC, DIP-40
```
fauxpark: The SRAM and EEPROM sizes are in bytes, so they should be uppercase here.

MCU_Microchip_PIC18.dcm
```diff
@@ -252,6 +252,12 @@ K Flash-Based 8-Bit Microcontroller
...
+K 16K Flash, 768b SRAM, 256b EEPROM, ADC
```
fauxpark: The keywords field should be a little more informative than just the description again. Seems like the usual thing for this lib is "Flash-Based 8-Bit Microcontroller"

MCU_Microchip_PIC18.dcm
```diff
@@ -354,6 +360,12 @@ K Flash-Based 8-Bit Microcontroller
 F http://ww1.microchip.com/downloads/en/DeviceDoc/39564c.pdf
 $ENDCMP
 #
+$CMP PIC18F4520-xP
+D 32K Flash, 1536b SRAM, 256b EEPROM, ADC, DIP-40
+K 32K Flash, 1536b SRAM, 256b EEPROM, ADC
```
fauxpark: And again for these two lines.

fauxpark:
Hah, I know even less about PICs than AVRs, but other than the description and keywords this looks good to me.

The datasheet has this to say about RE3 (10.5, p117):

> The fourth pin of PORTE (MCLR/VPP/RE3) is an input only pin. Its operation is controlled by the MCLRE Configuration bit. When selected as a port pin (MCLRE = 0), it functions as a digital input only pin; as such, it does not have TRIS or LAT bits associated with its operation.

So I think it is correct to have it as Input.

evanshultz:
@fauxpark 
Got it. Thanks! I figured your work would help to be a template for all MCUs.

@Misca1234 
If you can update the description, and the other PRs you have for the other packages of this MCU, this will be ready to merge and I'll move on to the other PRs.

Misca1234:
Keyword updated

fauxpark:
@Misca1234 could you also do the same for the 4520?

evanshultz:
Dah. Too fast. I didn't realize what the above was asking for until it was too late. Since there are other open PRs for this family I'll request it there.

PR #880, opened 2018-09-02
Bus_ISA_8bit fix

nickysn:
This pull request contains only a fix for the name of the two ~MEM signals in BUS_ISA_8bit. Previously, there were two signals named ~MEM and it wasn't clear which one is for memory read and which one for memory write. I renamed them ~SMEMW (System Memory Write) and ~SMEMR (System Memory Read), which also makes them consistent with Bus_ISA_16bit.

For some reason, loading and saving the Connector library in KiCad 5.0.0 resulted in various changes to the file. I don't know how to avoid that. I committed this change separately.


nickysn:
IMHO, this should probably be labelled "Bug" and not "Addition", because it fixes a bug in an existing symbol, and does not add any new symbols.

antoniovazquezblanco:
Sorry, my bad! Thanks :)

evanshultz:
While I'm not particularly familiar with ISA signaling, https://en.wikipedia.org/wiki/Industry_Standard_Architecture shows that pins 11 and 12 are common for both 8- and 16-bit busses so they should be named the same. Thanks!

PR #881, opened 2018-09-02
IS61C5128AL/AS IS64C5128AL/AS high speed CMOS SRAM 512KBx8

nickysn:
This adds the following series of high speed CMOS SRAM chips:

IS61C5128AL/AS
IS64C5128AL/AS

Datasheet is here:
http://www.issi.com/WW/pdf/61-64C5128AL.pdf

There are 7 versions of the chip, as described in the datasheet. 4 of them are actually different symbols (due to different pinout), 3 of them are aliases. Unfortunately, I haven't set the footprint and footprint filters, because I couldn't find the appropriate footprints in the KiCad library :( Maybe they are compatible with existing footprints with an existing name, but I don't have a lot of experience with these kind of SMD packages. I have added the names of the packages (as described in the datasheet) to the descriptions. These are:
SOJ-36
TSOP II-44
TSOP II-32
SOP-32
STSOP I-32


nickysn:
I messed up when creating this branch, because I didn't create this from master, but from a different branch that isn't yet merged. Closing this pull request. Will open a new one, containing only the relevant changes rebased to a new branch.

PR #885, opened 2018-09-03
Adding MAX3218

ki5libs:
Adding MAX3218 RS-232 Transceiver
[Datasheet](https://datasheets.maximintegrated.com/en/ds/MAX3218.pdf)
max3218

------------


SchrodingersGat:
Hi @ki5libs looks good, only a small suggestion for the description:

```Dual RS-232 Transceiver, 1.8V to 4.25V, AutoShutdown, DIP/SSOP```

evanshultz:
Here are a few more:
- Put the pin count at the end of the package types in the description above (`DIP-20/SSOP-20`).
- Pin name offset will be 20mil.
- The words "LOGIC" and "R232" are unnecessary.
- The symbol is overly large. It can shrink both vertically and horizontally without affecting usability.
- Move V+ and V- closer to LX (move INVALID down).

ki5libs:
@SchrodingersGat @evanshultz look now, please. 
I previously tried to save symbol compatibility with MAX232, but if you think that it's too large let's do it smaller.
max3218

evanshultz:
Unfortunately this library is full of old symbols that aren't a good reference. This symbol, once merged, should be the reference for MAX3218!

Only a few thoughts left from me:
1. Spread apart the C pins so a cap can fit there.
2. It would be nice if there wasn't so much dead space at the bottom because of the GND pin collisions. I wonder if pins 2-4 could go down there and make that space usable? Up to you if you want to investigate.
3. V+ and V- are a bit tight since each need a cap. V+ shouldn't go closer to LX because of the diode. However, I'm fine with this as-is if you are.

ki5libs:
@evanshultz I'm completely agree. Please look at the result.
max3218

SchrodingersGat:
@ki5libs this looks great, we should match the other symbols to this one :) Thanks for the contribution

PR #887, opened 2018-09-03
Adding CAT24C128 (alias for 24LC16)

ki5libs:
Adding CAT24C128 (alias for 24LC16)
[Datasheet](https://www.onsemi.com/pub/Collateral/CAT24C128-D.PDF)
cat24c128

------------


evanshultz:
I'll allow it since it's an ALIAS and the symbol already exists in this format, but I note the following issues:
1. The SOIC footprint filter doesn't have the pin pitch specified.
2. The DFN package should be a separate symbol because there is no pin 9.

Please let me know if you're willing make any of those updates. If not, I'll merge and they can be done later.

ki5libs:
@evanshultz I think that all of the issues of this symbol are the same for CAT24C256, which is already an alias of the 24LC16 in the library. And there are some more aliases for that symbol which have similar issues. For example 24LC16 do not have DFN package at all though it's present in the footprint  filter because of aliases. I don't know the exactly right way to resolve all of issues, but anyway I think that resolving of that should not be part of this small PR.
Please merge it, and tell me if you know how to fix these issues. I could do it later.

Why footprint filter is not unique for each alias like description or documentation?

antoniovazquezblanco:
Closing and reopening for CLA.

antoniovazquezblanco:
@evanshultz ping

evanshultz:
Missed this. Done now.

To answer the question above: There are two different ways to handle a part that comes in multiple packages. We can either put no default footprint and many FP filters (done here) or make many unique symbols, one for each footprint. The library is gravitating towards the latter for new symbols and when updates are made.

PR #888, opened 2018-09-03
IS61C5128AL/AS IS64C5128AL/AS high speed CMOS SRAM 512KBx8 (take 2)

nickysn:
This adds the following series of high speed CMOS SRAM chips:

IS61C5128AL/AS
IS64C5128AL/AS

Datasheet is here:
http://www.issi.com/WW/pdf/61-64C5128AL.pdf

There are 7 versions of the chip, as described in the datasheet. 4 of them are actually different symbols (due to different pinout), 3 of them are aliases. Unfortunately, I haven't set the footprint and footprint filters, because I couldn't find the appropriate footprints in the KiCad library :( Maybe they are compatible with existing footprints with an existing name, but I don't have a lot of experience with these kind of SMD packages. I have added the names of the packages (as described in the datasheet) to the descriptions. These are:
SOJ-36
TSOP II-44
TSOP II-32
SOP-32
STSOP I-32


Misca1234:
Hi

Thanks for the contribution,

Until the foot print exists the symbol wont be merged.
I will have a look at them this evening, some of them can easily be generated with scripts, especially the SOP and TSOP variants, the SOJ should also be easily done with scripts (more difficult to create 3D model.

I will link to the foot print push in this thread when I pushed them.

That is, unless you want to create the foot print yourself?

nickysn:
I've never created footprints using the python scripts and I don't have much experience with python, so it will probably be quicker if you created it. I might look into creating the 3D models instead. As far as I understand, these can be added later - only the footprints are needed for merging these components. Is that correct?

nickysn:
Btw, if you want to create the footprints, I think you can find these drawings useful:
http://topline.tv/TSOP.html
http://topline.tv/SOJ.html

Misca1234:
>  I might look into creating the 3D models instead. 

There are python script as well to create the 3D model ;)

Misca1234:
SSOP is named 
SSOP-32_11.305x20.495mm_P1.27mm
push here
https://github.com/KiCad/kicad-footprints/pull/944


TSOP is named
TSOP-I-32_11.8x8mm_P0.5mm
TSOP-II-44_10.16x18.41mm_P0.8mm
push here
https://github.com/KiCad/kicad-footprints/pull/943


There already exist a foot print with the name 
TSOP-II-32_21.0x10.2mm_P1.27mm
that should fit one of your needed foot print (page 16)

Wait until these are merged before you add the foot print names to your symbols, 
they need to be approved

nickysn:
Thanks for making these footprints! I'll wait until they get merged and then I'll apply them to my symbols. Btw, are you planning to do the SOJ-36 footprint as well?

Misca1234:
The SOIJ-36_10.16x23.49mm_P1.27mm push is here

https://github.com/KiCad/kicad-footprints/pull/950

Misca1234:
@nickysn 
All the foot prints are merged, if you update the foot print repository you should find them
Update your push to match them so I can review

nickysn:
Ok, added all the footprints. Btw, why did you name it SOIJ, instead of SOJ? I saw the other SOIJ footprint that we have in the library doesn't have J leads, at least not in its 3D model?

Misca1234:
> why did you name it SOIJ, instead of SOJ?

Probably happened per reflex, do you want it to change it (before we merge this push) ?

nickysn:
Yeah, I think we should change it. Microchip also have a SOIJ package that doesn't have J leads, so leaving it as it is can lead to confusion:
http://ww1.microchip.com/downloads/en/DeviceDoc/8L_SOIJ_Medium_5_28_SM_C04-056C.pdf

Misca1234:
Renamed push is here 
https://github.com/KiCad/kicad-footprints/pull/984

Misca1234:
@nickysn 
All your foot prints are merged, let me check this tonight for a merge

nickysn:
Ok, footprint now updated, please review! :)

nickysn:
Ok, fixed.

Misca1234:
Nice, thanks

PR #890, opened 2018-09-04
Add AD7746 capacitance-to-digital converter

fauxpark:
I don't know much about this part, I'm just adding it on behalf of someone else who wants to use it.

Datasheet: http://www.analog.com/media/en/technical-documentation/data-sheets/AD7745_7746.pdf
sym

------------


SchrodingersGat:
Thanks for the contribution @fauxpark

evanshultz:
Hmmm.... I would have suggested a few things be changed:
- Since the purpose is digitizing signals, I would put all inputs on the left and SCL/SDA/RDY on the right. See figure 43 of the datasheet.
- EXC* pins are connected to the caps on the left so those pins should be on the bottom left.
- CIN* pins should be Passive type since they only go to capacitors.

PR #891, opened 2018-09-04
Shape fixes for dual FET drivers

SchrodingersGat:
First PR in an attempt to address the issues in https://github.com/KiCad/kicad-symbols/pull/828

For the existing similar dual FET drivers:

- Reduced pin lengths to 100mil
- Changed pin style from rounded invert output to overline

image

image

Once these changes are merged I can continue to add the new variants.


evanshultz:
OK. I'll go with this.

There are a lot of bad/incomplete footprint filters and also it would be nice if some of them could be ALIASes.

PR #894, opened 2018-09-04
Footprint filter fixes for Driver_FET

SchrodingersGat:
Some small changes as a follow-on from https://github.com/KiCad/kicad-symbols/pull/891


evanshultz:
Could you add the body size to the SOIC footprint filters?

SchrodingersGat:
@evanshultz thusly?

evanshultz:
Indeed!

PR #896, opened 2018-09-05
RAC01-xxSGB

Cylindric:
I am adding a RECOM AC/DC converter symbol, footprint and model.

[Official RECOM datasheet](https://www.recom-power.com/pdf/Powerline-AC-DC/RAC01-GB.pdf)

rac01-xxsgb-symbol

[Symbol PR](https://github.com/KiCad/kicad-symbols/pull/896)
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/916)
[Model PR](https://github.com/KiCad/kicad-packages3D/pull/391)

I don't have Travis but `checklib.py` reports no problems.

------------


Cylindric:
The only failure in the Travis checks is the missing footprint, I assume that's okay as I have submitted that PR already? Or should I submit PRs in order of model -> footprint -> symbol?

Misca1234:
Do you think it would be a good idea to add the 3.3V and 24V variants as aliases as well ?
Do you think it would be too detailed to add the output current?

Cylindric:
I wasn't sure really, this was my first part so I'm still testing the waters.
If every alias is added for every part, that could end up with a LOT of very similar aliases making the lists huge, is that something we want? Happy to add them though.

Regarding the output current, I guess that could go in the description easily enough, as it is consistent within the series.

Misca1234:
> If every alias is added for every part, that could end up with a LOT of very similar aliases 
> making the lists huge, is that something we want?

That is ok, we normally ask for the aliases if people have not added them

Cylindric:
I'll add them. I just realised I'm working from an older Rev 2017 datasheet that doesn't have those, which is why I didn't add them. I'll check against the 2018 revision and update.

Misca1234:
Great thx

PR #898, opened 2018-09-06
change CRLF to LF in a bunch of files

ppelleti:
So, I'm finding it impossible to work with the `kicad-symbols` repository, because git keeps insisting a bunch of files are modified when they're not.  This seems to be because the files have CRLF line endings.

Is there some reason I'm the only one having this problem?  Is there something I should be doing differently?

But I believe the KLC specifies files should have LF endings, in which case they should be changed anyway.

------------


diegoherranz:
+1
I've been suffering this. It's a pain to work with as is.

Thanks!

evanshultz:
Yes, they should. See http://kicad-pcb.org/libraries/klc/G1.7/.

I think this was me. I'm a Windows user and I touched some or all of those files recently. I'm not aware of a way to make KiCad using a specific line ending when writing files so I convert the files manually after saving in KiCad (or rather auto-manually using Notepad++) - which I seemingly forgot to do here.

I recall having our repos catch this but apparently that only worked on footprint files. https://github.com/KiCad/kicad-library-utils/tree/master/schlib/rules doesn't have a rule for G1.7 but for footprints it's at https://github.com/KiCad/kicad-library-utils/blob/master/pcb/rules/G1_7.py. I created an issue for this at https://github.com/KiCad/kicad-library-utils/issues/259.

It looks to me like only line endings were changed. Please confirm that if you would and I'll merge. Sorry and thank you.

ppelleti:
> It looks to me like only line endings were changed. Please confirm that if you would and I'll merge. 

Yes, I did not change anything.  This is just what git insisted on changing automatically, which should be the line endings.

> I'm not aware of a way to make KiCad using a specific line ending when writing files so I convert the files manually after saving in KiCad (or rather auto-manually using Notepad++)

Git ought to convert the CRLF to LF when you check files in.  (After all, that's the problem I'm having: that git wants to automatically convert CRLF to LF.)  Maybe you need to [adjust your git configuration](https://help.github.com/articles/dealing-with-line-endings/)?

Although I thought that the `text=auto` in the [.gitattributes](https://github.com/KiCad/kicad-symbols/blob/master/.gitattributes) file should already be configuring the line endings to be handled automatically?

evanshultz:
Right, I thought that should do it as well. Not sure what's going on.

@SchrodingersGat
I think you set this up. Do you know for sure this was working and if or why things may have broken?

evanshultz:
Ummm... I feel a bit embarrassed to say but I use the GitHub Desktop app unless I need to pop to the command line. I may have done something dumb on the opamp/comparator commit. I don't recall but I'll keep an eye on it.

poeschlr:
The github desktop app might go through the github api instead of using git. 
This would explain why git settings do not help in that case.

PR #900, opened 2018-09-06
L293: change pin 11 to be named "3Y" instead of "3A".

ppelleti:
L293/L293D had two pins named "3A", which doesn't seem right.  I renamed the output pin to "3Y", to match the other outputs.

l293

A separate question is whether this is the right naming convention to use, since the names differ from the names used in [the datasheet](http://www.st.com/content/ccc/resource/technical/document/datasheet/04/ac/22/f9/20/5d/43/a1/CD00000059.pdf/files/CD00000059.pdf/jcr:content/translations/en.CD00000059.pdf).  But at least this small change makes it internally consistent.

------------


evanshultz:
Definitely not the right nomenclature for the ST datasheet. But see the TI one at http://www.ti.com/lit/ds/symlink/l293.pdf. And notice that the datasheet in our library is the TI one. So the pin names are OK with me.

Also:
- Pins 1 and 9 need the numbers and letters flipped to match the TI datasheet.
- GND pins should be stacked.
- Ref des should be at top left and PN at top right.
- Something should differentiate the "D" suffix's description; in this case it's output current.
- The description should include the package name at the end.

Do you want to make any other fixes now? If not, I can merge anyway.

ppelleti:
> Do you want to make any other fixes now? If not, I can merge anyway.

No, not really.  Fixing pin 11 was what was really important to me.

evanshultz:
OK. Thanks for the fix!

PR #901, opened 2018-09-07
Added LM4132B to Reference Voltage

Ragnoraq:
http://www.ti.com/lit/ds/symlink/lm4132.pdf
lm4132b-sot235

------------


evanshultz:

First, you need to sign the CLA above before this can be merged.

Next, review comments:
- Package has 5 pins, not 6. Fix the default footprint.
- Pin 5 is named `VREF`.
- I'm thinking this symbol might be better to draw like our linear regs, with VIN and EN on the left. I know it violates KLC, but they are guidelines. I see you made several adjustments to the symbol shape. What do you prefer? What is best for KiCad users drawing a schematic?
- Could you re-arrange the description for better grouping? Like `Precision Low Dropout Voltage Reference, 1.8V, ±0.1%, SOT-23-5`.

Last, your handle is awesome! :)

evanshultz:
Oh. One more thing. Scoot the NC pin inward so the connection point lies along the edge of the body and it is more difficult to accidentally connect.

Ragnoraq:
I've made the changes you suggested.  I agree with moving the VIN to the left with VREF on the right. This is how I prefer my regulator and reference voltage symbols to look.

I did change the shape a lot, but that had to do with complying with KLC and getting the spacing between the pin names to my liking.  I have no issue with the shape of the symbol.

One thought that occurred to me after making the part though.  The LM4132 comes in 5 grades, A-E, which call-out its accuracy.  So an LM4132A has an accuracy of 0.05% while and LM4132E has 0.5%. I know library size is an issue so would it be beneficial to make 1 generic symbol with a name of LM4132x instead of potentially 5 different symbols with the different grades.  Description could state that the accuracy is between 0.05% - 0.5%.  Everything else is exactly the same between the parts, the only difference is the grade.

evanshultz:
@Ragnoraq 
You didn't push a commit so I can't see any changes.

It's fine to change the shape of the symbol, I was just noting that I see you tried a few variations so I was interested in your feedback.

We are moving more towards an "atomic" library where each symbol is defined uniquely. In some cases that does inflate the library for no real benefit, but right now there's no hard and fast rule here. So leaving it as it is with only the "B" grade is OK, changing to `LM4132x` for each voltage with a wildcard is OK, or adding all grades is OK too. I think we should at least include each voltage and only wildcard the grade. Considering what you need, your time, and what other KiCad users might like, what would you prefer to do here?

Ragnoraq:
Changed the grade to a wildcard but kept the distinct voltages, LM4132xMF-[1.8].
Moved VIN from the TOP to the LEFT.
Renamed Pin 5 to VREF.
Moved Pin 1 so the connection point is on the edge of the body.
Changed the Footprint to SOT-23-5.

2018-09-10-220517_1280x800_scrot

evanshultz:
Thanks! The symbols is looking good.

Two more things:
- Footprint filter should be `SOT?23*` to avoid selecting SOT-223 footprints. Sorry I forgot that above.
- Other SOT-23-5 regulators I see have a bit of a differently style. Could you match it? See this one (the NC pin is placed improperly here so please ignore that):
image

And before I can merge you need to sign the CLA. See above.

evanshultz:
Looks good. Thanks for contributing to KiCad!

If you can sign the CLA above I will merge.

antoniovazquezblanco:
Please @Ragnoraq sign the CLA so that we can merge :)

evanshultz:
@Ragnoraq 
This PR is ready to merge but the CLA was not signed. Now branch conflicts have developed. Are you willing to at least sign the CLA, and ideally fix the branch conflicts too, so this can be merged?

evanshultz:
@Ragnoraq 
I fixed the branch conflict so if you can sign the CLA with the second account then we're sure all is well.

@chschlue 
What do you think about this? The CLA was signed by the account that created this PR, but apparently there was an email bungle. Can this be merged if we don't hear back from @Ragnoraq ?

chschlue:
IANAL, but I guess it's safe to assume that Tom Kellogg is not an unrelated party that had his copyrighted work stolen and published here without consent, so yes.

evanshultz:
@chschlue 
OK. I'll merge before moving to GH if there's no response before then.

evanshultz:
I waited a few days with no response. Based on the above comment, I am merging...

PR #909, opened 2018-09-10
Add TLV755P*DBV LDO aliases

awygle:
This adds the TLV755P LDO series SOT23-5/DBV parts from TI as a set of aliases to the TLV71209_SOT23-5 symbol.

Datasheet is http://www.ti.com/lit/ds/symlink/tlv755p.pdf

------------


awygle:
I don't understand why the Travis script is calculating the center the way it is - based on my other experiences I would expect this to say the component is centered. If someone can offer advice on how to convince Travis I'll make those changes, otherwise I should probably revert the movement (but keep the pin offset change).

evanshultz:
In general, the centering algorithm is rather weak (I seem to recall only pin locations are checked) and we often allow violations of this rule. Here is the existing symbol:
image

It's fine as it was. You can revert any movement changes.

And while you're at it: For completeness, can you add `Low IQ, Small Size,` to the description as in the datasheet title?

awygle:
Done!

evanshultz:
Thanks once again! Please keep the PRs coming!

PR #734, opened 2018-07-05
Adding IS25WP256D

ki5libs:
Adding IS25WP256D
[Datasheet](http://www.ti.com/lit/ds/symlink/tlv62080.pdf)
is25wp256d

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #914, opened 2018-09-11
Adding IS25WP256D-xM

ki5libs:
Adding IS25WP256D-xM
[Datasheet](http://www.ti.com/lit/ds/symlink/tlv62080.pdf)
is25wp256d-xm_new

------------


ki5libs:
@poeschlr thanks for your help. I found the only way to close [PR#734](https://github.com/KiCad/kicad-symbols/pull/734/) and open this PR, because I had to manually edit Memory_Flash.lib from the actual master branch to get the right result. 
Maybe I did something wrong, but I couldn't fix Travis errors though I merged actual state of master branch to my old IS25WP256D branch. So I decide to make it from scratch...

Misca1234:
You could point out the the foot print SOIC-16W_7.5x10.3mm_P1.27mm explicitly, you are aware of that ?

I suggest you remove the "inverse balls" on the pins like pine 1, in the text add a circum to the start like this this '~HOLD~/IO3', then you get a line above the text instead.
pin 3 will be '~RESET~/NC'
Pin 7 will be '~CE'

I suggest you shortened the pin length to 150 instead of the 200 you have now

You should group pin 15, 8 and 16 together, preferable left side.

Pin 3 is not RESET/NC according to page 9 it should be '~RESET~/IO3 and perhaps be bidirectional

Another idea to keep the size of the symbol down is to group the serial interface on the right side and "all the other" on the left side

ki5libs:
@Misca1234 
because of only SOIC package has 16 pins I thought that renaming this symbol to `IS25WP256D-xM` is a good idea. I've also set default footprint for it.

I've made name changing of inverse pins and removed circles, as you asked.

>I suggest you shortened the pin length to 150 instead of the 200 you have now

Sorry, but changing the length to 150 I would have to shrink the rectangle (to allocate pins in a 100mil grid according to KLC). But in this case I couldn't allocate invisible NC-pins correctly. I couldn't fix it out so decide to leave it as it is.

>You should group pin 15, 8 and 16 together, preferable left side.
>Another idea to keep the size of the symbol down is to group the serial interface on the right side and "all the other" on the left side

Done. I've updated the picture in the initial comment.

>Pin 3 is not RESET/NC according to page 9 it should be '~RESET~/IO3' and perhaps be bidirectional

I'm not sure. According to the table on the page 8 pin 3 of SOIC package could be only NC (for IS25WP256D-JM) or RESET (for IS25WP256D-RM), but not HOLD (it's only on pin 1 as shown on the page 7 first picture).

myfreescalewebpage:

Starting a new review now, and few comments I have:

- [ ] Description and keywords can't be in capital letters, please look at other similar devices for some examples. The description can be enhanced too, with the density and operating voltage of the memory at least.
- [ ] Footprint filter should be `SOIC*7.5x10.3mm*P1.27mm*`
- [ ] Pins length should be 100mil for all pins (symbol<100pins)
- [ ] Name of pin 3 is `~RESET/NC`, name of pin 7 is `~CE`, name of pin 8 is `SO/IO1`, name of pin 16 is `SI/IO0`
- [ ] You should groupe pin 7 with pins 5/15/16. The size of the device can be reduced. You can look at device AT25SF081-SSHD-X for an example.

Cheers,
Joel

ki5libs:
Hi @myfreescalewebpage and sorry for the delay.

I've done changes according to your comments. Screenshot is updated. I don't know how to make it smaller actually.

Thanks for review

myfreescalewebpage:
Hi @ki5libs 
Thanks for the modifications, great job! No more comments to do, merging.
Joel

PR #918, opened 2018-09-11
Security: Create new library and add ATECC608A crypto co-processor

antoniovazquezblanco:
http://ww1.microchip.com/downloads/en/DeviceDoc/40001977A.pdf

imagen

imagen

imagen

------------


antoniovazquezblanco:
@evanshultz or @Misca1234 can I ask for a review as a favour? I am about to use this in a product and would love to have someone look at this before that. Just in case

evanshultz:
- Remove the company name from the keywords.
- Missing package type in description.
- Actually, perhaps we should begin the precedent to always break out symbols by package? This would allow, by default, using the correct footprint for the DFN since we have several that differ only by EP size.
- Does the datasheet show how to connect the DFN thermal tab? Right now that pin is missing and unconnectable, more reason to make unique symbols.
- Since the interface (I2C vs single-wire) appears to be an orderable option that's another reason to make unique PNs split by interface since it would be unfortunate to install the wrong one.
- UDFN is missing the `1EP` in the footprint filter.
- UDFN body size is 3x2mm, not 2x3mm.

antoniovazquezblanco:
I could not find information about pinout of the SWI interface.
EP seems to be a NC as it is nowhere mentioned in the datasheet.
UDFN package does not exist.

evanshultz:
I just found there is a newer datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/ATECC608A-CryptoAuthentication-Device-Summary-Data-Sheet-DS40001977B.pdf. The

What do you mean the UDFN package doesn't exist? It's orderable from Microchip's website and you created the symbol for it. What am I misunderstanding?

- I'd like to have the EP has a pin so it can be connected. This might make routing easier. Or a future datasheet revision will include the info.
- UDFN EP size is 1.30x1.50mm. You have it reversed.
- Can you put "I2C" before the package and add the pin count to the description?

antoniovazquezblanco:
Implemented the suggested changes. What I've meant is that the required UDFN footprint has not been yet submitted to KiCad library. I was not planning on creating it. Maybe there is a fast way to do it from existing scripts.

Thanks for the review @evanshultz

antoniovazquezblanco:
The new components:
screenshot from 2018-10-25 15-58-09
screenshot from 2018-10-25 15-58-21

evanshultz:
The UDFN footprints does exist, but without the "U" (just "DFN"). Won't that be OK?

I would also prefer the EP pin on the bottom since it's likely to be grounded and matches what we usually do.

antoniovazquezblanco:
> The UDFN footprints does exist, but without the "U" (just "DFN"). Won't that be OK?

I don't really know if that is the case :sweat_smile: . I will check the dimensions in a minute and will let you know.

antoniovazquezblanco:
DFN footprint seems compatible although Microchip suggest to add a couple of thermal vias to the pad.

evanshultz:
Oh yeah. Good point. But considering even with the maximum 6V applied to the IC and the maximum current draw of 14mA it will only dissipate 84mW, it will be OK. This is probably the default footprint for this package.

Are you OK if I merge this?

antoniovazquezblanco:
Yes. Everything looks fine. Thank you very much @evanshultz !

evanshultz:
My pleasure, mi amigo.

antoniovazquezblanco:
hahahaha 😄

PR #922, opened 2018-09-12
Added SOIC-8 version of INA240AxD

SchrodingersGat:
The TSSOP-8 version is already in the libs. However the SOIC-8 has a different pinout:

image

Datasheet: http://www.ti.com/lit/ds/symlink/ina240.pdf


evanshultz:
Just change the `V+` pin to Power Input type and I'll merge. Thanks!

SchrodingersGat:
I have made this change to the new symbol and the old one too. Thanks @evanshultz

evanshultz:
Thanks!

PR #930, opened 2018-09-15
Add Texas Instruments TPL5010 - Nano Power Timer w/ Watchdog

yankee14:
This is an ultra low power timer from TI. It draws 35 nA while running. Timing interval set with a resistor, from 100 ms to 2 h. Can be triggered manually. It is designed to be used with a microcontroller e.g. to trigger an I/O ISR. It has a watchdog function for resetting the microcontroller if the microcontroller does not respond to the timer within 20 ms of the end of the current timing interval.

Datasheet: http://www.ti.com/lit/ds/symlink/tpl5010.pdf

Screenshot:
tpl5010

Footprint: SOT-23-6*

NOTE: The manufacturer's datasheet differs from the stock KiCad footprint. By the datasheet, the pad centerline distance between the two columns of pads should be 2.7 mm (no tolerance given), but it is only 2.2 mm in the stock KiCad footprint.

------------


Misca1234:
Thx for your contribution

> The manufacturer's datasheet differs from the stock KiCad footprint. 

Sometimes the manufactures don't agree among themselves how a standard foot print look like, but if you think the stock foot print does not match good enough you should push a manufacture specific foot print, this is not uncommon. (in this case it is 0.5 mm difference).


Please change the following 
bild

Add the SOT23 to the description , from
Timer, Nano Power, Watchdog, 35 nA, 100 ms to 7200 s, VDD 1.8 V to 5.5 V, Iout max 1 mA.
into
Timer, Nano Power, Watchdog, 35 nA, 100 ms to 7200 s, VDD 1.8 V to 5.5 V, Iout max 1 mA, **SOT23-6**

Pin 1 and 2 should be of the type "Power input"

Pin 6 is named  "RSTn" I guess you want to indicate that it is active low, instead of adding the 'n' at the end, you add '~' at the beginning instead like this "~DRV"
Also, it should be of the type open collector.

bild

Misca1234:

bild

bild

bild

bild

bild

yankee14:
>Pin 6 is named "RSTn" I guess you want to indicate that it is active low, instead of adding the 'n' at the end, you add '~' at the beginning instead like this "~DRV"

"RSTn" is the actual name of the pin from the datasheet. Would you like me to change it to "~RSTn" or "~RST"?

>Also, it should be of the type open collector.

Good catch, thanks! Fixed.

Thank you. Fixed.

>Pin 1 and 2 should be of the type "Power input"

They already are.

Fixed!

Finally, on the subject of the footprint dimensions:

I checked again the footprint in the KiCad libraries and compared it to the [dimensions of the *leads* given in the datasheet](http://www.ti.com/lit/ds/symlink/tpl5010.pdf#page=23&zoom=auto,-20,662) (not the recommended *land pattern* given in the datasheet). The maximum outside dimension of the leads is given as 3.05 mm, and the KiCad land pattern outside dimension is approximately 3.27 mm. This leaves about 110 μm (4.3 mil) of wiggle room on either side. Considering board fab house tolerance, pick-and-place machine tolerance, etc, do you think this is enough room? The pad is technically big enough at the outside. I feel it is okay.

The KiCad handsoldering version of this footprint measures about 4.27 mm at the outside, so that leaves about 500 μm of wiggle room on either side. Plenty.

The inside dimensions of the pads is okay, on both the regular SOT-23-6, as well as the handsoldering version.

Misca1234:
@yankee14 
> "RSTn" is the actual name of the pin from the datasheet. Would you like me to change it to "~RSTn" or "~RST"?

The golden rule is to follow the datasheet naming exactly, no exceptions, well, almost, in this case 
we make an exception, because the line "above" is much more clearer.
Having a line "above" and a 'n' at the end, some people would argue is a negation on a negation so 
this is the exception from the golden rule, that is
-Use line above and drop the 'n' at the end



@poeschlr 
Here is a question for you
"I checked again the footprint in the KiCad libraries and compared it to the dimensions of the leads given in the datasheet (not the recommended land pattern given in the datasheet). The maximum outside dimension of the leads is given as 3.05 mm, and the KiCad land pattern outside dimension is approximately 3.27 mm. This leaves about 110 μm (4.3 mil) of wiggle room on either side. Considering board fab house tolerance, pick-and-place machine tolerance, etc, **do you think this is enough room?** "

yankee14:
@Misca1234 
>Use line above and drop the 'n' at the end

Fixed!

I think I'm up to date on all the changes, just waiting on response about footprint. Thanks guys!

myfreescalewebpage:

- [ ] Done pin on the right side as proposed by the datasheet is also very nice

Else, your symbol is okay.

Cheers,
Joel

myfreescalewebpage:
@yankee14 please look at the git diff, there is a merge issue to fix.

yankee14:
@myfreescalewebpage yeah, I'm trying to understand where I went wrong :/

My plan was:

 1. git checkout master
 2. git fetch upstream master
 3. git merge --ff-only upstream/master
 4. git push origin master
 5. git checkout TI_TPL5010
 6. git merge --ff-only master

But there was a merge conflict and I don't think I resolved it properly. I'm going to reset back to the beginning and try again.

yankee14:
Closed? I never closed it.

poeschlr:
It says you closed it. Is it possible you deleted your branch? Or is it possible you pushed the current master state onto it? (You did force push to this branch. This can lead to trouble in some cases)

yankee14:
@poeschlr I'm really sorry, I guess I'm rusty with git and made a mistake. I had to force push this branch to reset it to the state it was in before I made a mistake. I think I fixed it, and just pushed my changes. But GitHub is telling me this PR is closed, the changes have been merged, and it is safe to delete the branch. I never deleted the branch, and I never closed the thread, and I certainly don't think you merged it.

yankee14:
Updated symbol screencap:

scap

myfreescalewebpage:
Thanks @yankee14 

Just seen a remaining fix to do:

- [ ] Pin 6 should be named `~RST` to get the over bar on it and type should be Open Collector.

Thanks,
Joel

myfreescalewebpage:
Thanks, merging.

yankee14:
@myfreescalewebpage 

>     * [ ]  Pin 6 should be named `~RST` to get the over bar on it and type should be Open Collector.

TPL5010

Fixed!

PR #931, opened 2018-09-15
Add Texas Instruments TPL5110 - Nano Power Timer, Active-Low

yankee14:
This is an ultra low power timer from TI. It draws 35 nA while running.
Timing interval set with a resistor, from 100 ms to 2 h, 1% accuracy
typical. Can be triggered manually. It is designed to be used as a
high-side P-channel MOSFET driver for gating the power to a
microcontroller. It has an optional one-shot mode for enabling the
MOSFET only once.

Datasheet: http://www.ti.com/lit/ds/symlink/tpl5110.pdf

Screenshot:
tpl5110

Footprint: SOT-23-6*

NOTE: The manufacturer's datasheet differs from the stock KiCad footprint. By the datasheet, the pad centerline distance between the two columns of pads should be 2.7 mm (no tolerance given), but it is only 2.2 mm in the stock KiCad footprint.

------------


Misca1234:
Thx for your contribution

> The manufacturer's datasheet differs from the stock KiCad footprint. 

Sometimes the manufactures don't agree among themselves how a standard foot print look like, but if you think the stock foot print does not match good enough you should push a manufacture specific foot print, this is not uncommon. (in this case it is 0.5 mm difference).


Please change the following 
bild

Add the SOT23 to the description , from
Timer, Nano Power, Active-High, 35 nA, 100 ms to 7200 s, VDD 1.8 V to 5.5 V, Iout max 1 mA.
into
Timer, Nano Power, Active-High, 35 nA, 100 ms to 7200 s, VDD 1.8 V to 5.5 V, Iout max 1 mA, **SOT23-6**

Pin 5 is named  "DRVn" I guess you want to indicate that it is active low, instead of adding the 'n' at the end, you add '~' at the beginning instead like this "~DRV"

Misca1234:

bild

bild

bild

bild

bild

yankee14:
@Misca1234 ,

>Pin 5 is named "DRVn" I guess you want to indicate that it is active low, instead of adding the 'n' at the end, you add '~' at the beginning instead like this "~DRV"

Pin 5 is named "DRV", and it is active-high, so I think it remains like it is. The TPL5111 is the one that is active-low, so I will make that change to that part.

Thank you. Fixed.

Fixed!

On the subject of the footprint dimensions, see [discussion](https://github.com/KiCad/kicad-symbols/pull/930#issuecomment-421822342) in #930.

myfreescalewebpage:

Else, your symbol is okay.

Cheers,
Joel

yankee14:
Symbol screenshot:

scap

myfreescalewebpage:
Thanks, symbol is correct for me now.
However, you have a file "fp-info-cache" in the git diff which is not expected... Please fix.
Cheers,
Joel

yankee14:
@myfreescalewebpage wow, I did not realize I added that, sorry! It is removed.

myfreescalewebpage:
No worries, that's why we made reviews :)
Merging your contribution, thanks.

PR #932, opened 2018-09-15
Add Texas Instruments TPL5111 - Nano Power Timer, Active-High

yankee14:
This is an ultra low power timer from TI. It draws 35 nA while running.
Timing interval set with a resistor, from 100 ms to 2 h, 1% accuracy
typical. Can be triggered manually. It is designed to be used as a low
side driver or active-high enabler, e.g. for the enable pin on a
voltage regulator. It has an optional one-shot mode driving high only
once.

Datasheet: http://www.ti.com/lit/ds/symlink/tpl5111.pdf

Screenshot:
tpl5111

Footprint: SOT-23-6*

NOTE: The manufacturer's datasheet differs from the stock KiCad
footprint. By the datasheet, the pad centerline distance between the two
columns of pads should be 2.7 mm (no tolerance given), but it is only
2.2 mm in the stock KiCad footprint.

------------


Misca1234:

bild

bild

bild

bild

bild

yankee14:
@Misca1234 ,

>Pin 5 is named "DRVn" I guess you want to indicate that it is active low, instead of adding the 'n' at the end, you add '~' at the beginning instead like this "~DRV"

Fixed!

Thank you. Fixed.

Fixed!

On the subject of the footprint dimensions, see [discussion](https://github.com/KiCad/kicad-symbols/pull/930#issuecomment-421822342) in #930.
tpl5111

myfreescalewebpage:

myfreescalewebpage:

Else, your symbol is okay.

Cheers,
Joel

yankee14:
@myfreescalewebpage I made the changes you suggested. Before approving, please look over the [discussion about footprint dimensions](https://github.com/KiCad/kicad-symbols/pull/930#issuecomment-421822342) on another TI part I submitted in the same family as this one in #930.

Thanks for your review.

yankee14:
Latest symbol screencap:

scap

myfreescalewebpage:
@yankee14 thanks for the fixes.

> Before approving, please look over the discussion about footprint dimensions on another TI part I submitted in the same family as this one in #930.

That's a footprint question, symbol and footprint are not linked, it is possible to modify footprint without modification of symbol. So the current footprint assignment is correct for me here.

> EDIT: Are you sure the footprint filter should be 'SOT*23*'? There should be no reason to use SOT-23-5, for example.

The design of symbol is made with the following guidelines: 1. all pin must be added to the symbol, NC pin are not visible. 2. because all pins are on the symbol, footprint filter do not need the number of pins, it is not possible to assign a 6 pins symbol to a 5 pins footprint.

> Also, I had left the footprint field blank because of KLC 5.1. SOT-23-6 and SOT-23-6_Handsoldering are possible choices.

No, it seems you have complete the footprint field, but it's good. HandSoldering is an option permitted with the footprint filter but of course the preferred, default footprint for this device is SOT-23-6.

No more comment on your symbol, merging.

Joel

yankee14:
@myfreescalewebpage thank you! I will now fix up the other two timers so you can merge them also.

PR #933, opened 2018-09-15
Adding 88SE9125C0-NAA PCIe SATA controller

ki5libs:
Adding 88SE9125C0-NAA PCIe SATA controller
[Datasheet](https://www.marvell.com/documents/bqcwxsoiqfjkcjdjhkvc/)
88_new2

[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/940)

------------


myfreescalewebpage:
Hi @ki5libs, thanks for contributing,

- [ ] Why using bold text for "PCIe SATA" ? It's very large. KLC S3.2 suggest to use 50mil text only (http://kicad-pcb.org/libraries/klc/S3.2/)
- [ ] You have forgotten the footprint field (do not confuse with footprint filter, it's not the same, that's why travis is not happy with you)
- [ ] You should remove leading star of the footprint filter
- [ ] The datasheet says "The main ground is the exposed die-pad (ePad) on the bottom side of the package." so I would suggest that the EP pin should always be connected to VSS. Thus, it should be renamed VSS and stacked with pin 28.
- [ ] I do not agree to stack the VDD and VDDIO pins. Example page 26 clearly shows several capacitors, one for each pin, and KLC indicates in this case that stacking is not expected. However, I don't realize how much it will increase the width of the symbol. Maybe we can ask more experienced librarians what should be done @poeschlr @evanshultz maybe ?

And finally a question, but I'm not an expert of PCI express bus, I wonder about the position of the pin on the symbol. Typically PERST and WAKE pins are on the PCI side, probably it's better to move them on the left. I would say PCI lines on the left, SATA lines on the right. Can you provide a sample schematic screenshot with this symbol to get a better overview ?

Cheers,
Joel

ki5libs:
@myfreescalewebpage, thx for review.

- `PCIe SATA` text size changed
- default footprint added
- footprint filter fixed
- EP renamed to VSS and stacked with it
- I'm not sure that you are right about unstacking `VDD` and `VDDIO` pins. Firstly number of capacitors is not equivalent to number of pins. Please look at page 29 of datasheet. Chapter 4.4.1 tells about placing capacitors on each side of IC, not for each pin. I think this is a common practice and power pins are usually stacked in such situations.

Finally I've exchanged SPI pins (moved to the right side) and `PERST_N`, `WAKE_N`, `TP` (moved to the left side). 

Sorry, but I don't have KiCAD schematic example yet. I think schematic example on page 26 of datasheet could give the general overview.

myfreescalewebpage:
Hi @ki5libs thanks for the modifications,

- [ ] Still not agree with your bold text for "PCIe SATA". On a schematic, what is the reason this specific device will need a bold text ? It is more important than the other parts on the schematic ? My answer is no.
- [ ] Just realize the EP was missing in the footprint filter, should be `QFN*1EP*9x9mm*P0.4mm*`

>  I think this is a common practice and power pins are usually stacked in such situations.

Another device in the library as example ?

Joel

ki5libs:
Hi @myfreescalewebpage sorry for my inattention:
- `PCIe SATA` is normal now;
- footprint filter is changed to `QFN*1EP*9x9mm*P0.4mm*`.

Maybe you are right and it's not actually a **common practice**. I was confused founding a very few devices with stacked power pins. Examples could be found however in library `MCU_Microchip_PIC32` (PIC32MX110F016D-IPT, PIC32MX210F016D-IPT, PIC32MX575F256H and aliases).

I still think that stacking is OK in `88SE9125C0-NAA` case, but it would be great to see some comment from more experienced librarians.

myfreescalewebpage:
> I still think that stacking is OK in 88SE9125C0-NAA case, but it would be great to see some comment from more experienced librarians.

Yes, me too :)

@poeschlr or @evanshultz your opinion if you get time to read this ? Thanks !

myfreescalewebpage:
@evanshultz ping! Question about stacking of positive power pins just above. Thanks

cpresser:
Rule S4.3 has been changed in the past:
https://github.com/KiCad/kicad-website/blame/master/content/klc/S4.3.adoc#L13

Exactly following the KLC here does not seem to work here. The datasheet recommends decoupling-caps on 3 sides.

Right now I see 3 options:
- Stack all
- no stacking
- split into 3 stacks (one for each side - according to datasheet)

My personal option is to stack power pins. always. even if each of them needs decoupling.
A PCB designer needs to invest significant time into proper decoupling anyway (exactly reading the datasheet), so the few hints that the symbol can provide are not really important.

myfreescalewebpage:
> My personal option is to stack power pins. always.

We do not in KiCad library.

> split into 3 stacks (one for each side - according to datasheet)

Can be a good compromise, yes. @ki5libs you can go this way I think.

Joel

myfreescalewebpage:

ki5libs:
Hi @myfreescalewebpage 
Sorry, I've overlooked this PR. I've done stacks splitting. `VDDIO` pins are separated now, because all of them are located on different sides of the IC. And `VDD` stack is splitted into four, because pins are located on each side. Some pins movement allowed to save symbol in the same size. Here is the screenshot:
88_new3

Hope, it's OK now.

myfreescalewebpage:
@ki5libs this is a good compromise, thanks. Merging.

PR #934, opened 2018-09-15
Add BD8314NUV

asukiaaa:
screenshot from 2018-09-16 01-10-18

Datasheet: http://rohmfs.rohm.com/en/products/databook/datasheet/ic/power/switching_regulator/bd8314nuv-e.pdf

Related PRs:
https://github.com/KiCad/kicad-footprints/pull/941
https://github.com/KiCad/kicad-packages3D/pull/400
https://github.com/KiCad/kicad-packages3D-source/pull/160


jneiva08:
Hi,

The GND and PGND probably should be in the bottom of the symbol.

asukiaaa:
Thanks for the reviewing.
I updated so.

screenshot from 2018-09-16 11-36-30

Misca1234:
Hi

Thx for your contribution

Please change the following 

Shorten the pin length to 100 

bild

According to page 12 it says 
"The heat sink on the rear should be a GND plane of low impedance in same potential with the PGND plane."
This mean you should add a pin, number 11 (representing the heatsink pad) and name it PGND.
(if you look at the foot print you can see that there is a "pin 11" as well

You should stack pin 5 on top of 4
you do it as follow,
make pin 5 invisible
Make pin 5 of type passive
Place pint 5 on the same position as Pin 4

You should stack pin 7 and 11 on top of 6
you do it as follow,
make pin 7, 11 invisible
Make pin 7, 11 of type passive
Place pint 7, 11 on the same position as Pin 6

I assume it is the foot print 
Package_SON:VSON-10-1EP_3x3mm_P0.5mm_EP1.65x2.4mm
that fits this one
There fore the foot print filter should be changed from 
VSON*10*3x3mm*
into
VSON*1EP*3x3mm*P0.5mm*EP1.65x2.4mm*


Because this component only have one type of foot print your explicitly select it via the field properties.
The following illustration shows it for the SOT-23 (I reused the screen shoots form another thread but the principle is the same)
You should select it for the VSON foot print instead


bild

bild

bild

bild

bild

Misca1234:
Please change the description from
3.0V to 12V, Integrated 2.5A MOSFET 1ch Boost Converter
Into
3.0V to 12V, Integrated 2.5A MOSFET 1ch Boost Converter, **VSON-10**

asukiaaa:
Thank you for a lot of feedback.
I update the symbol except for type setting because TravisCI tells me error.
screenshot from 2018-09-16 19-14-18
screenshot from 2018-09-16 19-11-24

Misca1234:
You should change 
pin 6 to the type "Power input"

I guess pin 4 should be the type "Power output"

asukiaaa:
I changed type of LX and PGND as `Power Input`.
I guess PGND is not power output because high voltage is created by inductor.

evanshultz:
- The EP size should not be in the footprint filter.
- The EP size should match the package (2.0x1.2mm). You may need a new footprint.
- PGND is Power Input type because that's how KiCad wants power source and sink nets to be. :)
- LX pin should be Passive type.
- VREG, INV, and SWOUT should also be Passive type since they only go to passive parts and not normally any logic gates (the datasheet doesn't indicate VREG can source any current).
- It will be easiest to use if power goes from left to right. I don't why the datasheet is backwards but perhaps because they stick with the package pin placement for the symbol which KiCad doesn't do. So: move INV and SWOUT to the right side below LX and VREG to the left side at the bottom. You may want to move GND to the left of PGND as well since there is more room there.

asukiaaa:
Thank you for the comments

> - The EP size should not be in the footprint filter.
> - PGND is Power Input type because that's how KiCad wants power source and sink nets to be. :)
> - It will be easiest to use if power goes from left to right. I don't why the datasheet is backwards but perhaps because they stick with the package pin placement for the symbol which KiCad doesn't do. So: move INV and SWOUT to the right side below LX and VREG to the left side at the bottom. You may want to move GND to the left of PGND as well since there is more room there.

I updated so.

screenshot from 2018-09-17 20-52-08


> - LX pin should be Passive type.
> - VREG, INV, and SWOUT should also be Passive type since they only go to passive parts and not normally any logic gates (the datasheet doesn't indicate VREG can source any current).

I'm not good at pin type but I can't agree that they are passive.
- LX pin is contolled by PRE driver.
- VRES pin is 5V output.
- INV pin controls PWM controller.
- SWOUT pin is connected to GND through 50 Ohm register when STB become under 1.0V.

Whow do you think about them?

> - The EP size should match the package (2.0x1.2mm). You may need a new footprint.

I assigned my own footprint (and 3D model).

https://github.com/KiCad/kicad-footprints/pull/946
https://github.com/KiCad/kicad-packages3D/pull/401

asukiaaa:
Thanks you for reviewing.

> Since the dimensions are XxY, and pin 1 is at the top left, EP size should be 1.2x2mm.

I updated footprint and 3d package.

> The pins were not moved as I mentioned. Please take a closer look. And pins should stay stacked.

Could you supply more information about what is lacking?
I stacked LX because they were not Power Input now.
I can't stack PGND because TravisCI told errors.

screenshot from 2018-09-18 23-42-47

> The pin types I listed above are the correct ones to use. This is based on what is best for KiCad symbols and not always strictly based only on the internals of the IC.

I'm not sure know what is correct so I trust you and change pin types as Passive.

jneiva08:
For stacking pins, one os Power input and visible and the others is passive and invisível.

evanshultz:
@asukiaaa 
You can find the guidelines for the KiCad library at http://kicad-pcb.org/libraries/klc/. Pin stacking is covered at http://kicad-pcb.org/libraries/klc/S4.3/.

- The package name is still wrong. See the dimensions I gave above.
- Thanks for the notes in pin stacking @jneiva08!
- The pin placement I described at https://github.com/KiCad/kicad-symbols/pull/934#issuecomment-421850022. STB and VREG on the left with VREG at bottom. LX, SWOUT, and INV on the right. LX at top and SWOUT at the bottom.

evanshultz:
Good catch. And again my mistake for pushing the scripted footprints too quickly.

SON packages can take so many forms. There's hardly a common theme you can recognize among all packages called SON.

I fixed this at https://github.com/KiCad/kicad-footprints/pull/962.

evanshultz:
The symbol could probably be a bit smaller (specifically more narrow), but this is OK. All green now that the footprint move was merged. Thanks!

PR #935, opened 2018-09-15
Interface_Ethernet Fix S3.6

jneiva08:
Have fixed in all symbols KLC S3.6

KSZ8081RNA and LAN8720A still have S3.1 warning. 

------------


jneiva08:
Travis says the footprint of ENC424J600-ML doen't exist.
Can be changed by QFN-44-1EP_8x8mm_P0.65mm_EP6.6x6.6mm.kicad_mod ?

evanshultz:
Yep, footprint update looks good to me too. Otherwise, the only change is correcting the pin name offset to 20mil.

This all looks good and I'm happy to merge. However, since I'm greedy and you've been so diligent and productive to add and improve the libraries, I'll selfishly list a few more quick improvements here in case you wouldn't mind piling on a few more things:
- `DP83848C` has 150mil pins that can be shorted (enlarge the symbol body to retain pin positions).
- `ENC424J600*` have 200mil pins.
- `KSZ9031RNXCA` has the value on top left, text inside the symbol that can be removed, 200mil pins, and make all pins Line style.
- `LAN8720A` have 200mil pins and some pins are not Line style.
- `LAN951*` have the ref des and value inside the symbol body.

If you just want me to merge as is, that's great too.

jneiva08:
I think is all done, need to add the screensots for all the symbols?
The ref and value of LAN951* looks a little strange :), perhaps the power pins could be more near of each other.
If not wrong the DP83848C, KSZ9031RNXCA will need in the future stacking the power supply pins.

evanshultz:
Thanks!

Screenshots are always nice but there are some more changes below so they wouldn't be final anyway. If you can read and correct the below I'll add screenshots when I merge so it doesn't require any more of your time.

Sadly I may not have been clear and strong about pin length changes: right now, because we are in the middle of a patch release (v5.0.x), pin placement should not change. So if pins get shorter, they must have their ends (connection points) at the same place and the symbol body must grow. Sorry if that wasn't obvious. We don't want to break schematics.

That affects `ENC424J600*` and `LAN8720A` as the pins are shorter but body size is the same. `DP83848C` had the body enlarged so it's fine.

Sorry again. :(

Can you make another set of changes to push the pins back to their original locations while keeping the shorter pins? I don't want to merge this as it is now since that will disconnect all pins on these symbols.

`KSZ9031RNXCA` can have the "10/100/1000" and "Ethernet PHY" deleted.

You are right there is still work to be done when the symbols can change more, like stacking pins and moving them around to look more attractive.

jneiva08:
Pins moved back to the original place.

evanshultz:
Thanks! Sorry again for the issues above.

I notice that `KSZ9031RNXCA` is 100mil too small in both directions. The original symbol was +/-1300mil wide and +/-1600mil tall. Can you please re-enlarge that symbol too? Everything else is good.

jneiva08:
Moved pins of KSZ9031RNXCA to original place, to not break schematics.

DP83848C:
dp83848c

ENC424J600-ML:
enc424j600ml

ENC424J600-PT
enc424j600pt

KSZ9031RNXCA:
ksz9031rnxca

LAN8720A:
lan8720a

LAN9512/3/4
lan951x

evanshultz:
Thanks for all these fixes!

antoniovazquezblanco:
Thanks!

Issue #509, opened 2018-04-20
Naming for USB Type-C connector symbols


Ratfink:
As mentioned in https://github.com/KiCad/kicad-symbols/pull/310#issuecomment-379883615, our nomenclature for USB Type-C symbols could use improvement.  In particular, there are several types of connector defined in [the standard (106 MB .zip)](http://www.usb.org/developers/docs/usb_32_030818.zip) which aren't covered by the current naming: USB 2.0 Type-C plugs (with only 12 pins) and USB Type-C Power-Only plugs (with only 9 pins), as well as full-featured USB Type-C plugs with 24 pins.

Plugs and receptacles name a couple pins differently (CC1 and CC2 in the receptacle connect to CC and V<sub>CONN</sub> in the plug), so I would suggest keeping the current naming intact and extending it to support more types of connector by adding the number of populated pins.  Ones I know we would need, with checks for the ones we already have under a different name:

* [X] `USB_C_Receptacle-24`
* [ ] `USB_C_Plug-24`
* [X] `USB_C_Plug-22`
* [ ] `USB_C_Plug-12`
* [ ] `USB_C_Plug-9`

Another thing to discuss: the plugs with very few pins could have smaller symbols, or they could be kept the same size as the larger ones to act as drop-in replacements.  Which approach would be preferred?

evanshultz:
Which document(s) and page(s) are you looking at? I see this sort of stuff in `\usb_32_030818.zip\usb_32_030818\USB Legacy Cable and Connector\USB 3.1 Legacy Cable and Connector Revision 1.0.pdf` for non-C and `F:\_Evan\kicad\usb_32_030818.zip\usb_32_030818\USB Type-C\USB Type-C Specification Release 1.3.pdf` for type C. Are you looking elsewhere?

Ratfink:
That looks like the right PDF for the Type-C connector spec.  The pin names are shown on pages 21-22.

* Page 45 shows a Type-C receptacle with 16 pins for USB 2.0 only.
* Page 204 indicates that there are Type-C plugs with 24 pins.
* Page 48 shows the 12-pin USB 2.0 only Type-C plug.
* Pages 31-32 describe the 9-pin power-only Type-C plug.

evanshultz:
Yeah, it does seem to me like a unique symbol for the plug and receptacle are useful. Thanks for digging in and finding this!

[...]

PR #937, opened 2018-09-16
Add USB 2.0-only Type-C receptacle and plug

diegoherranz:
Adding the USB 2.0-only Type-C receptacle, started from the existing USB_C_Receptacle.

Types of plug and receptacle listed in 3.1.1 of USB Type-C Specification Release 1.3
https://www.usb.org/sites/default/files/documents/usb_type-c.zip

Took advantage of this to fix broken documentation links on other type-c connectors as well as adding the official "full-featured" denomination to the existing receptacle.

Although I'm not a big fan of pin stacking, I used it on VBUS and GND to follow KLC. But not on D+/D- since they may be shorted together or routed independently to an USB switch.

The "full-featured" receptacle could use pin stacking too, but I haven't modified it to avoid breaking the library. I think that is what we're doing until KiCad 6. Let me know otherwise.

usb_c_2 0

Thanks!

------------


diegoherranz:
By the way, as an example, a real-world receptacle using the USB 2.0-only spec is GCT's USB4085:
https://gct.co/connector/usb4085
Although they call it "Type C Charging Connector" instead of "USB 2.0-only" which is the official USB name for it...

evanshultz:
Looks OK to me. A dislike having pin stacking just on a subset, though, and I think it's OK since stacking pin smartly on the existing connectors will only make them smaller but all connections can still be present (stacking them poorly could break them, though).

Note https://github.com/KiCad/kicad-symbols/issues/509.

@Ratfink 
Could you take a look since you were invested in this topic already?

diegoherranz:
> A dislike having pin stacking just on a subset, though, and I think it's OK since stacking pin smartly on the existing connectors will only make them smaller but all connections can still be present (stacking them poorly could break them, though).

It's true, I could stack the pins on the other ones as long as we make sure nothing is broken, maybe I give it a try after this PR is done?

Re #509, I knew I had read a thread on USB-C but I couldn't find it anywhere. I've read it again and if you'd prefer to use the number of pins in the name (e.g. USB_C_Receptacle-16 for this one) instead of the USB official names, that's OK with me. Just let me know.

Thanks.

diegoherranz:
I've added the USB 2.0-only plug based on the USB2.0-only receptacle:

usb_c_plug_2 0

Now we should have the 4 more common ones.

evanshultz:
Nice!

@diegoherranz @Ratfink 
Do you agree on a symbol naming convention? Pin names or USB official names? Official names seem like a good way to go, but I'm not terribly familiar so I don't have a strong feeling now.

Ratfink:
Thanks for adding the plug!

@evanshultz I don't feel that there's a disagreement on naming.

evanshultz:
Unclear English. Not that were was any disagreement, but I am checking for agreement. :) There is a question at https://github.com/KiCad/kicad-symbols/pull/937#issuecomment-421821511 about keeping the name here or changing it. I wanted to wrap that up before merging. So... keep it or change it?

@diegoherranz 
Are you willing to stack pins on the other USB symbols after this? You indicated you'd do it above but I would like to confirm before anything gets messy.

poeschlr:
Please do not touch existing symbols in the same PR as adding new ones.

diegoherranz:
Yes, the pin stacking on the existing symbols would happen on a different PR.
However, technically I've modified the existing symbols since I updated their keywords, datasheets etc. on the dcm file. If you want me to, I can rebase my PR to only add new symbols.
Either way, I'll be more careful next time not to mix new stuff and modifications to existing stuff.

evanshultz:
In this case I can handle it since it's just the descriptions of existing parts. Unless I'm told otherwise. Rene feels very strongly about this and he's the head librarian so in the future you know what's preferred. :)

So to wrap this up, @Ratfink and @diegoherranz like this symbol name?

diegoherranz:
Summarising on the name, we could go for:

1.  `USB_C_Receptacle_USB2.0` and `USB_C_Plug_USB2.0` as in the PR currently
2. `USB_C_Receptacle_USB2.0-only` and `USB_C_Plug_USB2.0-only` to be even closer to the official USB naming.
3. `USB_C_Receptacle-16` and `USB_C_Plug-12` as in #509.

I think I would go 1 or 2 (slight preference for 2) unless we plan on adding the plug-9 and plug-24 (which would be the least common ones?), in which case we can go 3 to use the numbering naming scheme for them all, and we'd need to change the name of the existing ones.

I think I'm happy either way (1/2 or 3), and happy to help making the changes required here or further PRs.

diegoherranz:
What do you think on the naming? I think everything else is looking OK now.
Thanks.

Ratfink:
Either 1 or 2 sounds fine.  1 is less verbose, but 2 is closer to what USB-IF calls them, as you point out.  I don't know if "only" will ever be an important distinction though, so I'd slightly favor 1.

diegoherranz:
Thanks @Ratfink.
In that case, everyone happy to merge?

evanshultz:
Alright. It's at 1 now so we'll go with that.

@diegoherranz 
Thanks! You can ping me when you stack pins and shrink the other USB symbols. Make sure you retain an existing VBUS (surely the lower one) and GND pin location so they're backwards compatible.

diegoherranz:
Thanks, @evanshultz. I've raised #1005.

PR #956, opened 2018-09-18
Delete old files

antoniovazquezblanco:
* TODO list should be kept under GH issues and it has been already finished.
* Replacement files are no longer going to be used as the moving task has already been finished.


antoniovazquezblanco:
Requesting review from @SchrodingersGat and @poeschlr at least given that they started and supervised most of the transition of the libs.

poeschlr:
I would keep the replacement files as they document part of what has been moved.

antoniovazquezblanco:
I personally think that that information is documented in the repo history.

I was thinking that those files clutter the repo and do not serve a purpose at the moment. Maybe you would prefer to move them to the utils repo?

SchrodingersGat:
I didn't realize these were still in here. I think they should either be moved to the utils repo (where the appropriate scripts are located) or removed entirely. They are not symbol data, which is the point of this repo.

poeschlr:
@antoniovazquezblanco the info is not really in the history as these are completely new repos.

antoniovazquezblanco:
@poeschlr I see what you mean. Anyway, do you believe those files should be moved to the utils repo with the corresponding moving scripts as @SchrodingersGat suggests?

antoniovazquezblanco:
File back up in:
https://github.com/KiCad/kicad-library-utils/pull/262

Should we move fordward with the cleanup? Do you think is a bad idea @poeschlr?

antoniovazquezblanco:
Ping @poeschlr  @SchrodingersGat 

Do we move fordward or should I drop this?

SchrodingersGat:
I'm happy to merge this and put a pin in this issue. I'll have a look at the other PR in utils repo

antoniovazquezblanco:
Thanks

PR #958, opened 2018-09-18
Add ATSAME53J18A-AN

cp-aquila:
Add ATSAME53J18A-AN

------------
- https://www.microchip.com/wwwproducts/en/ATSAME53J18A
- http://ww1.microchip.com/downloads/en/DeviceDoc/SAM-D5xE5x-Family-Datasheet-01507B.pdf
- Screenshot 
- No fooprint PR

* [x]  Provide a URL to a datasheet for the symbol(s) you are contributing
* [x]  An example screenshot image is very helpful
* [x]  Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
* [x]  If there are matching footprint PRs, provide link(s) as appropriate


jneiva08:
Please change the following

Set the "Pin name position offset" from 40 to 20,
Can shrunk a litle the symbol?

myfreescalewebpage:
Closing/opening the PR to refresh CLA.

myfreescalewebpage:
Hi @cp-aquila , thanks for contributing,

- [ ] The part is not available in WLCSP package and moreover this kind of footprint requires a new symbol due to the numbering of the pins. You have forgotten TQFP in the footprint filters.
- [ ] Page 47 of the datasheet: "VDDCORE pin is just an output for monitoring the internal voltage regulator. This is not an input for
an external supply" so VDDCORE should be Power Output

Cheers,
Joel

cp-aquila:
Hi Joel,

Thanks for the review. I missed those two points.

One more question: The file format of "MCU_Microchip_SAME.lib" is outdated. Saving to it rewrites a lot of lines, it seems there was a change in EESchema in the past.
Is there a clever way to fix this? Seperate PR to just update the whole file?

myfreescalewebpage:
Thanks @cp-aquila 

Just made a final global review and realize the part number is strange. According to the datasheet page 19, the part should be "ATSAME53J18A-A" for the TQFP version, and "ATSAME53J18A-M" for the QFN version. I suggest we have to make two symbols (that's what is often done with other MCU). It is not mandatory for you to submit the two, you can submit only one at the moment if you want, but the part name must be consistent with the package type.

Also just seen the footprint is not correct (you have set LQFP-100).

Cheers,
Joel

cp-aquila:
How about an alias?
There is only one issue, that is that the TQFP does not have 65 pins (no exposed pad).

myfreescalewebpage:
You can't set the footprint of a single alias :(

> There is only one issue, that is that the TQFP does not have 65 pins (no exposed pad).

Exact! I haven't realized it before, and the duplicated symbol for the TQFP version should not have the pin 65.

I suggest to rename your symbol to "ATSAME53J18A-M".

cp-aquila:
Hi,

The TQFP filter was removed.
I added some aliases for the different Flash sizes.
What would be the clever way to also include the other variants (E51, E53, E54)? They have the same pinout, only the possible peripherals muxes differ.

Once this PR is done, I will start a new one for the 100pin part. I ended up not using the 64pin device in my project :)

myfreescalewebpage:
Hi @cp-aquila 

We made several changes, so I'm currently reviewing the whole part and try to sum up (I realize I have not yet made a full review of your submission):

- [ ] TQFP filter has not been removed. Also need to add the QFN footprint value.
- [ ] Package name at the end of the 3 descriptions should be "QFN-64" instead of QFN64.
- [ ] Pins length should be 100mil for the part (symbol < 100 pins). Sorry for this, this will need moving the pins to get it align with KLC rule about the grid.
- [ ] VDDIO should not be stacked to allow several caps to be added.
- [ ] RESET is not Bidirectional, it is Input only

> What would be the clever way to also include the other variants (E51, E53, E54)? They have the same pinout, only the possible peripherals muxes differ.

You can do it simply adding some aliases, yes, thanks.

> Once this PR is done, I will start a new one for the 100pin part. I ended up not using the 64pin device in my project :)

Great!

Cheers,
Joel

cp-aquila:
> > What would be the clever way to also include the other variants (E51, E53, E54)? They have the same pinout, only the possible peripherals muxes differ.
> 
> You can do it simply adding some aliases, yes, thanks.

That would result in a huge matrix of parts (4*3 variants), just for one footprint.
With QFN-64, TQFP-64, that would make 24 symbols. Not counting the devices with a different pin count. Is this the way to go? 
Is using a 'x' or some other wildcard in the name allowed?

Regarding VDDIO stacking, I do not agree totally. It is up to the designer to add the correct number of caps into the schematic.
This how I usually do it:


I we would to this totally correct, the way to go is to split the different VDDIO domains as well. See Table 6-34 on page46.

cp-aquila:
Changes are pushed;
I also updated the screenshot in the first post.

myfreescalewebpage:
Thanks.

- [ ] Still missing the footprint, take care this is not the same field than footprint filter. Footprint is in the Field Properties dialog, icon below.
capture
Footprint should be `Package_DFN_QFN:QFN-64-1EP_9x9mm_P0.5mm_EP4.7x4.7mm`.

- [ ] Also realize the footprint filter miss the EP (sorry), it should be `QFN*1EP*9x9mm*P0.5mm*` to be more accurate.

Joel

myfreescalewebpage:
Thanks @cp-aquila for your job! No more comment, merging the PR.

Joel

cp-aquila:
> Thanks @cp-aquila for your job! No more comment, merging the PR.
> 
> Joel

Thanks for your patience :+1:

PR #960, opened 2018-09-18
Added DAC7750IR

cp-aquila:
Added DAC7750IR 

------------
DACx750 Single-Channel, 12- and 16-Bit Programmable Current Output Digital-to-Analog Converters for 4-mA to 20-mA Current Loop Applications

- http://www.ti.com/lit/ds/symlink/dac7750.pdf
- 
- DAC7850IR as an alias
- No realated footprint PR

* [x]  Provide a URL to a datasheet for the symbol(s) you are contributing
* [x]  An example screenshot image is very helpful
* [x]  Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
* [x]  If there are matching footprint PRs, provide link(s) as appropriate


Misca1234:
Hi

Change the pin length to 100 instead of now 150, so the symbol looks like others
Change the text to pin 2 from "ALARM" to "~ALARM", then you get a line above the text.
Pin 16 should be of the type input
You are missing Pin 21, NC
Pin 25 should be output

Perhaps it is a good idea to place pin 9 on the left side together with pin 7 and 8, even it is an output but they belong together.

I can not see that the component is isolated internally between the digital and analogue part, therefore you should stack all GND pins together (name them all GND)

You have specified the dedicated foot print to be 
Package_DFN_QFN:Texas_S-PVQFN-N40_EP4.15x4.15mm
but a more accurate foot print would be
Package_DFN_QFN:Texas_S-PVQFN-N40_EP4.6x4.6mm_ThermalVias



@KiCad/librarians 
How to handle the DVDD, pin 39, according to page 4 it can be both "power input" and "power output" depending on how to set pin 23

cp-aquila:
Hi,

Thanks for the detailed review :+1: 
- changed the pin-length to 100 (it was 200).
- alarm pin was renamed to ~ALARM
- Pin16 is now an input (I checked TLC5940NT which also has a current set resistor, there its an input as well)
- Pin25 is now an output
- changed footprint (the 4.6mm EP was just recently merged)
- stacked all the GND pins
- Move SDO next do DIN.
- no change on Pin21, as its already there.

evanshultz:
I prefer incremental commits to changes are more obvious. I recall this being requested by other librarians as well but I don't know if it's prefered by all or an official request.

@Misca1234 
That one is tough. What are the typical configurations? External DVDD where the pin would be Power Input, or internal DVDD where the pin would be Power Output. As DVDD can source current for other devices Power Output is certainly a reasonable choice.

So what if things get mixed up? If we go Power Output, we get an error if an external supply is used (ERC thinks there are two power sources). There's no remediation there.

If we go Power Input and the internal supply is used, with no external circuitry there is an error but the No Connect symbol can be added to the pin to fix that. If external circuitry is connected with the internal regulator running there is also an error (ERC thinks there is no power source) but in this case the user can add a PWR_FLAG symbol.

Passive type would also be a potential choice. In that case, there will be errors in any case and it doesn't have any implications to pin function. Still, my slight preference is for Power Input but I could be convinced otherwise since I see it's a gray area. As this exact situation doesn't come up often, though, it's certainly a plausible place to bend or break rules since one size can't fit all.

@cp-aquila 
- Symbol name should be `DAC8750xRHA` to wildcard temp rating and include the full part number section that defines the package. And then replace the 8 with a 7 for the other symbol.
- Smash the symbol down to only the size it needs to be for the visible pins and then sprinkle the NC pins wherever they can go within the symbol body. Don't enlarge symbols to fit NC pins.
- Put the value on the top right.
- Pin 16 should be Passive type. You may have found an old symbol, but in cases where the pin goes to ground or to a R/C, that's what we've been using.
- Pin 2 should be Open Collector type (KiCad doesn't have Open Drain).
- I'd keep DIN on top left and move SDO to bottom left or bottom right. This makes daisy-chaining easier to wire.
- Pins 16-8 should be on the right side since they're analog pins. 24 too, I think. Keep CAP1 and CAP2 at the bottom so they're easier to get to ground and you may want to shift the power pins to the left to make more room.
- The description should mention just the bitness for the specific IC and not both.
- But the pin count after the package type in the description.
- The footprint filter should be `Texas*PVQFN*`.

cp-aquila:
Hello,

I did most of the changes suggested by @evanshultz.
- Rearranged pins, changed pintypes
- Updated description and aliases
- Change footprint filter
- Change Name to "DAC8750xRHA". Sidenote here, only the 'I' temperature-version exists right now. Both according to the datasheet and distributors
- The CAPx pins are still at the top, in most applications they will be connected towards a positive supply rail. See the attached screenshot from a related appnote (Ti: SBAU205A)
- I kept pin 24 on the left side, since it is an input. In most applications it will be driven via capacitive coupling from some kind of modulator.  This kind of satisfies S4.2 because pin24 and pin26 are on the same height.
- The symbol box is 100mil higher than needed, to keep the origin aligned and the box symmertrial to it.
- For DVDD I choose PowerInput because it allows easy DRC fixes for both powering options. 

As pointed out, this time its not a squased force push, but a seperate commit.
I could not find any rule for 'where do analog pins go' in KLC. Did I miss that, or is it not there?

evanshultz:
Thanks! This looks nice!
- Please put a hyphen between VQFN and 40 at the end of the description.
- I see about CAP1/2. Thanks. Could they go at the top and ~ALARM below them?
- I understand pin 16 will go to ground through a R or float, while pin 18 could come from some external circuit. So I'd put 16 below 17/18 to avoid a potential need to cross wires.
- Could it be a bit less wide? It seems like there's room to thin it down 100mil on each side.

Funny that the screenshot has an error, I think. The names of pins 24 and 25 don't match the datasheet.

cp-aquila:
Hi,

I did incorporate all your hints.
The screenshot is actually from a different device. The DAC8760 also has a -10..+10V Analog out which is not present on the DAC8750. Since they are so similar, there seems to be only one evaluation board for both of them. I guess the different pinout is a result of a differend bond-wire-scheme.

evanshultz:
Got it. Thank you!

cp-aquila:
Thanks for the great reviews and maintaining effort!

PR #980, opened 2018-09-18
Fix pin names for two Adesto Flash parts

DanSGiesbrecht:
Both AT25SF081-SSHD-X and AT25SF081-XMHD-X had pin 2 mislabelled as 'SO/IO2'. It has been corrected: 'SO/IO1'.

Please see the [datasheet](https://www.adestotech.com/wp-content/uploads/DS-AT25SF081_045.pdf) and screenshot for more information.

image

------------


evanshultz:
Thanks once again!

PR #982, opened 2018-09-19
Close arrow polygon for pot arrows

poeschlr:
Small change to close the polygons of the potentiometer symbols. The problem was discovered during the GAL rebuild of EEschema as that renders open polygons differently right now. (The devs will need to change that as otherwise our logic symbols will no longer display correctly.)


evanshultz:
Just going for it...

PR #983, opened 2018-09-19
Add TCA9406DC

cp-aquila:
Add TCA9406DC

------------
Symbol is based on pca9306 which has a similar function.
TCA9406 adds one-shot accelerators and is faster.
DC-Suffix was added to the part name, so it maches the DCT and DCU packages from TI. The DSBGA-Version has a different pinout and would require another symbol.
The pin-names with and without underscores are present like that in the datasheet. So I choose not to change/unify them.

- http://www.ti.com/lit/ds/symlink/tca9406.pdf
- no footprint PR for this symbol


SchrodingersGat:
Thanks for the contribution @cp-aquila

cp-aquila:
> 
> 
> Thanks for the contribution @cp-aquila

It would have done that symbol anyway.
I have to thank you for reviewing and maintaining this repo! :100:

PR #986, opened 2018-09-20
Add LT3092

evanshultz:
http://www.analog.com/media/en/technical-documentation/data-sheets/3092fc.pdf
image
image
image

Copied design from existing symbols.

------------


evanshultz:
xST: The asterisk should be fine since there isn't a SOT-1223 or some other footprint that would get eaten by a greedy footprint filter. Questions marks are used only when needed like to catch SOT-23 but not SOT-223. So I believe this should be good.

I fixed the other footprint filter you noted and did some minor simplification of the keywords. KLC isn't too specific about the datasheet or keywords so I like to stick with the datasheet name for the description and I can't think of a simpler keywords that describe the part. If you do want more changes to merge let me know. Thanks for catching my mistakes and making good suggestions.

DanSGiesbrecht:
Looks good! Thanks for the quick reply.

PR #991, opened 2018-09-21
add MAX14763 analog switch

penoud:
add MAX14763 analog switch
------------

https://datasheets.maximintegrated.com/en/ds/MAX14759-MAX14763.pdf
image


Misca1234:
Hi

Place pin 3 and 6 together on right side
Place Pin 4 on left side and pin 7, pin 5 on right side, this match the symbol in the data sheet better
Pin 8 should be of the type input

You have chosen foot print
DFN-10-1EP_3x3mm_P0.5mm_EP1.55x2.48mm
This FP have a pitch of 0.5 mm, your component need a pitch of 0.65 so you can not use this foot print

Do you want to create a foot print your self with help of scripts, I can show you how to do it it's name will be TDFN-8-1EP_3x3mm_P0.65mm_EP1.58x2.35mm

evanshultz:
In the description the pin count comes after the package type, so use `TDFN-8. I believe there is a typo above.

Also, page 11 clearly states that VN and VP cannot be used to power other circuitry so they should not be of type Power Output. Make them Passive type so if someone tries to use them as a current source, and runs ERC, they will be notified that those pins are not a power source.

evanshultz:
Also don't include pin count and the EP size in footprint filters.

penoud:
Corrected version:
image

evanshultz:
Thanks!

I'd rather see the description match the datasheet with additional information, like Rds(on), later. Something like `Above- and Below-the-Rails Low On-Resistance Analog Switches, 2ohm max`. Pasting `2 ohm` in the middle as written above is bad English. Also, the description needs changed anyway because the pin count is 8, not 10.

Page 11 says connect the EP to VN. To do that, name pin 9 "VN" and make it Passive type and invisible. Then place it at the same location and in the same orientation as VN. Sorry for the bad advice above.

Lastly, the symbol above doesn't quite match the "standard" analog switch symbol in our lib. Could you match it a little better? The A1 and A2 pin styles are different but otherwise I think it's nice.
image

penoud:
Updated version of the symbol:
image

penoud:
I have several questions:
1) How to do symbol with two units but a different drawing inside each unit. I tried to do 2 units version but I wasn't able to have a different drawing inside each unit?
2) I placed the EP pin over the VN pin to reflect the fact they are connect in the datasheet. But seem's to violate KLC. How to represent VN is connected to EP pin?

evanshultz:
Good question. If you click on an element (rectangle, polyline, circle, etc.) you can select to place it on all units or not. This can help but I find I end up editing text files manually for complex symbols.

SEL should stay on the left side since it's an input and placing the EP pin close to VN makes sense, but otherwise it looks good to me. Thanks!

penoud:
Thanks, I understand now. I change SEL location to the right:

Misca1234:
uhh, I did not follow it so closely so I would suggest you merge it

evanshultz:
I reverted this PR since I realized pin 11 should be stacked on VN. My apologies. See my notes at https://github.com/KiCad/kicad-symbols/pull/991#issuecomment-423851359 for how to stack those pins on this symbol. Maybe submit a new PR with all 3 switches?

I haven't done this before using the web interface so hopefully it went fine.

PR #994, opened 2018-09-23
Sensor_Distance library added. GP2Y0A41SK0F symbol added.

serkanxselcuk:
* Library Sensor_Distance added.
* Simplified schematic of symbol GP2Y0A41SK0F added.
* Sensor_Distance is added into sym-lib-table

Datasheet: http://www.sharp-world.com/products/device/lineup/data/pdf/datasheet/gp2y0a41sk_e.pdf
Footprint is on PR: https://github.com/KiCad/kicad-footprints/pull/949

image
**Poll:** Symbol is only simplified logic of device, not real internal schematic. Would you prefer to show pin names instead of simplified scematic?

------------

- ~[ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)~


serkanxselcuk:
CI error is only package library, which is on PR process.

poeschlr:
Sorry it took so long. The symbol in the lib does not agree with your screenshot (pin numbers are exchanged)

The part also does not seem to have pins that are directly soldered to the pcb. So i am a bit confused as to why it has pins in the symbol. My guess would be that this is connected with a short cable via some connector. This connector could then however have any pin numbering that the user wishes. (Can not really be defined by the kicad library)

Your footprint has two pads defined. These are however not used for connecting to the sensor but only connect to the mounting pads. (I am not even sure why these are created as plated through holes.)

serkanxselcuk:
Hi @poeschlr , right, sensor doesn't have pins to connect directly to pcb. It would pass better into a new library, which will include electronic cabinet or machine automation components.

serkanxselcuk:
> Hi @poeschlr , right, sensor doesn't have pins to connect directly to pcb. It would pass better into a new library, which will include electrical cabinet or machine automation components.

Some examples on at #1846

serkanxselcuk:
Component is not be placed on PCB. Because of that I close this PR.

PR #996, opened 2018-09-23
add analog switch MAX14761

penoud:
add analog switch MAX14761
------------

https://datasheets.maximintegrated.com/en/ds/MAX14759-MAX14763.pdf
image


SchrodingersGat:
The EP pad must be exposed, as it should be connected to VN:

image

penoud:
Correct by changing EP pin to visible:
image

evanshultz:
- Correct the description as discussed before.
- Note that the package has 10 pins.
- Put EP on the bottom.
- This is quite cramped. Space apart the two switches to the pins aren't jammed together or use two units.
- Stack pins as noted at https://github.com/KiCad/kicad-symbols/pull/991#issuecomment-423851359.

And, of course, branch conflicts.

poeschlr:
If we go with a multi unit approach than i would guess we would have two units for the switches plus one for supporting stuff (power pins plus maybe VP and VN depending on what these two pins do.)

penoud:
Updated screenshoots:
Unit A:
image
Unit B:
image
Unit C:
image

evanshultz:
Push the ref des slightly to the right so it sits above the symbol better.

Looks fine to me otherwise. On second, though, perhaps EP should be on the size next to VN?

@poeschlr @SchrodingersGat 
Any other comments and/or requests?

penoud:
I correct the ref pos.
image

evanshultz:
Thanks for stacking pins! If you fix the branch conflicts and Travis is green I can merge.

penoud:
Updated unit c:
image

evanshultz:
Thanks again!

PR #997, opened 2018-09-23
New symbol for the Recom RAC04-GA and RAC04-GB parts

Cylindric:
I am adding a RECOM AC/DC converter symbol, footprint and model.

The GA and the GB are identical in all but datasheet, but I wasn't sure if it was preferred to call it RAC04-xxGx and have lots of aliases with different datasheets, or to just create two different symbols. The package, symbol and footprints are identical for both series, they just have a different EMC class.

[RAC04-GA Datasheet](https://www.recom-power.com/pdf/Powerline-AC-DC/RAC04-GA.pdf)
also suitable for the 
[RAC04-GB Datasheet](https://www.recom-power.com/pdf/Powerline-AC-DC/RAC04-GB.pdf)

rac04-symbol

[Symbol PR](https://github.com/KiCad/kicad-symbols/pull/997)
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/963)
[Package PR](https://github.com/KiCad/kicad-packages3D/pull/407)

------------


Cylindric:
Weird, my symbol check missed this pile of errors. Gimme a sec.

Cylindric:
That'll probably be from my recent RAC01 part :) I'll update the PR in a bit.

Misca1234:
Thx

PR #998, opened 2018-09-24
Add SMLVN6RGB

asukiaaa:
I add an symbol of full color led.

Datasheet: https://www.rohm.com/datasheet/SMLVN6RGB1U

screenshot from 2018-09-24 22-19-49


evanshultz:
- The description could be enhanced with something like `High Brightness Tri-Color LED, RGB, 3.5x2.8mm`.
- Set the pin name offset to 20mil.
- All pin types should be Passive for diodes.
- Instead of pin names, it might be more salient to name all pins `~` and show the 3 diode symbols with one letter for their color. Here are two examples from our current lib (filled symbols should be hollow, though):
image

Apologies for popping in, Dan.

DanSGiesbrecht:
@evanshultz beat me to it! Haha that's ok. The more input the merrier.

I was going to bring up - what's the benefit of having this symbol in the LED library, rather than Devices? It seems to me that the symbol would be better served as a generic, rather than atomic symbol.

evanshultz:
Device.lib is for simple parts which don't have a manufacturer attached. It functions as a reference from which symbols can be copied and adapted into other libraries when a manufacturer-specific component is created. These generic parts also could apply to many footprints or physical parts, whereas other libraries can have a specific footprint attached that matches a tangible component.

LED.lib is therefore correct for this part.

If there is a missing symbol permutation in Device then it can be added, but it should be done in a separate PR to avoid cluttering things up.

LED direction isn't controlled, as far as I know. The anode is normally the "input", I suppose, but that's not an argument. To me it makes sense to have diodes point to the right. But for whatever reason KiCad's generic diodes have pointed left with cathode as pin 1 since before I started helping out.

DanSGiesbrecht:
@evanshultz That clears up my confusion a lot. I thought that the generic LED symbols were to be used whenever possible. It sounds like the primary path for LED addition should always be atomic.

asukiaaa:
Thanks for the information.
I updated this symbol by copying LED_RGB.
screenshot from 2018-09-27 22-17-47

evanshultz:
Nice.
- Pins 1 to 3 are B, R, G in this ROHM part. Please update the text on this symbol.
- Remove the period from the end of the description.
- Footprint filter should be `LED*ROHM*SMLVN6*` or `LED*SMLVN6*`.

evanshultz:
@DanSGiesbrecht 
Please review. That's all I see so if you're good you can merge if you like.

asukiaaa:
Thanks for reviewing.
I update the symbol.

DanSGiesbrecht:
Looks good! Thank you for your contribution. :)

asukiaaa:
Thanks for checking and merging.

PR #1001, opened 2018-09-25
Add BAV70 series diodes

fauxpark:
A handful of common cathode dual/quad switching diodes. BAV70S is essentially two dual diodes in a single 6-pin package.

bav70

Datasheet: https://assets.nexperia.com/documents/data-sheet/BAV70_SER.pdf

Footprint PRs: KiCad/kicad-footprints#969, KiCad/kicad-footprints#970

------------


evanshultz:
How did our library not include these diodes?! Thanks for adding them!

A couple issues (which are more recent discoveries and so the existing libraries may not take them into account):
1. You should not duplicate pins (for example, only one cathode pin on BAV70). This helps ensure CvPcb does only pick compatible footprints.
2. Since series diodes share a pin, they cannot be separated and must be near each other on the schematic. I personally like multi-unit parts, but here I think internally-connected symbols should be on the same unit. Could you use a single unit for BAV70 and two units for BAV70S?

Also:
- Please use a similar description to some existing parts that matches the datasheet. `High-speed Switching Diodes, 100V 0.25A, Dual Common Cathode, SOT-363` or similar for BAV70S.
- BAV70 If is 215mA and BAV70W is 175mA (might as well fill it all the way out).
- Can BAS70 be an ALIAS of BAT54C? (That symbol needs the footprint filter fixed.)

fauxpark:
I also like the idea of having them be individual diodes that you can put anywhere, but I can see how that might cause some problems.

On the description: should I just use mA since we're being so precise?

The lack of a consistent format/structure in symbol descriptions irks me a little and I'd like to see more guidelines here, but I don't know if this is something the KLC really aims to address. The general description for the BAV70S says "quadruple common 
cathode/common cathode" where the others say "dual common cathode". How about: "Quad 100V 0.25A high-speed switching diodes, dual common cathode, SOT-363" for the BAV70S?

BAT54C is a dual Schottky diode, while these BAV70s are your garden variety signal diodes - the drawing is different.

evanshultz:
mA is fine with me.

Yep, descriptions are all over the place. I usually request, since this is what I often saw, three fairly distinct sections:
1. <datasheet description>
2. other information as desired (which must differentiate otherwise similar parts)
3. finally <package type>-<pin count>

In this case, having "quad" and "dual common cathode" in the description sounds quite wrong and that prompted me to ask for a change. "Common cathode" already implies at least two diodes so I thought sticking "dual" onto that phrase was accurate and descriptive. But it could be something else, though I feel the suggestion above is not good as I just noted.

Yep. You're right. My mistake. I went searching through the library to look for other symbols with the same packages and didn't catch that difference. Would you mind looking or verifying you did the same and these can't be an ALIAS of anything that exists already?

fauxpark:
Done the milliamps and BAV70S description. I'm not sure what is so wrong about having "quad" and "dual" in the same description as one refers to the number of diodes (4) and the other refers to the number of cathodes/units (2).

I can't find any other symbols using the same drawing and pinout to make them aliases of. There is the BAV99, but it's not common cathode (series) and it's split into two units.

fauxpark:
I interpret the description to provide, separately, the total number of diodes in the package ("Dual/Quad high speed ... diodes"), and the configuration they are in ("Common cathode/Dual common cathode"). Thus the BAV70S has four diodes with two cathodes, one for each pair of diodes.

evanshultz:
I can see that if each section is considered separately.

Well... what about if all parts of the quantity are put together? Would that make it clear? So avoid "dual" or "quad" at the beginning? That is done for BAV70S already, but not the others. Or should I merge like this? Or would you suggest something else?

fauxpark:
Considering the forward current ratings in the descriptions seem to be per diode (and also the reverse voltage but it's the same for all of them), I would think they *should* be considered separately. But if you think it's clear enough as is, I'm fine to merge.

evanshultz:
OK, I'll just merge then.

You're quite right about the lack of consistency in the descriptions of diodes and transistors. That is something that could, and should, be reviewed and structured.

evanshultz:
Bah. Wrong button. Sorry.

PR #1006, opened 2018-09-27
Added MCP14A030x

SchrodingersGat:
Three gate drivers from Microchip:

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/MCP14A0303_4_5-Data-Sheet-20006046A.pdf

image

image

image

Footprint PR - https://github.com/KiCad/kicad-footprints/pull/974

Added the DFN-8 package variant, there are three different versions


SchrodingersGat:
Only error is due to footprint not yet being merged.

Misca1234:
Hi

I guess the GND and EP should be stacked, page (page 13, chapter 3.6)

As you remembered Evan pushed a scripted version of the foot prints, but with a slightly different name, 
the foot print name is 
WDFN-8-1EP_3x2mm_P0.5mm_EP1.3x1.4mm.kicad_mod

evanshultz:
@SchrodingersGat 
Ping!

evanshultz:
@SchrodingersGat Ping!

evanshultz:
Usually we use the recommended pad size for the EP if given one instead of basing it off the package thermal pad, but I think this will be fine. Thanks!

PR #1007, opened 2018-09-27
Added FDMC8032L mosfet

SchrodingersGat:
Datasheet: https://www.fairchildsemi.com/datasheets/FD/FDMC8032L.pdf

image

image

image

image

Footprint PR - https://github.com/KiCad/kicad-footprints/pull/975


evanshultz:
There are now two S2 and two D1 pins.

SchrodingersGat:
Good catch @evanshultz I had the "edit all units" button pressed by mistake..

evanshultz:
Looks good now. Just waiting for the footprint to be merged.

evanshultz:
And it is!

PR #1008, opened 2018-09-27
Add HopeRF RFM69CW symbol based on RFM69HW

SeanDS:
This is the symbol for the HopeRF RFM69CW module. This is similar to the RFM69HW symbol already in the KiCad library. It is functionally similar to the RFM69HW, but it has 14 pins instead of 16, and a different pin mapping.

I created a pull request for the corresponding footprint [here](https://github.com/KiCad/kicad-footprints/pull/978). Please see it for the datasheet URL. I suggest that this symbol is not merged until the footprint is also merged, as I have referenced it within the symbol.

image


SeanDS:
Oops, it appears this pull request does not contain the part I intended... closing and will resubmit...

PR #1009, opened 2018-09-27
Add HopeRF RFM69CW symbol based on RFM69HW

SeanDS:
**NOTE:** the build only failed because the [companion footprint](https://github.com/KiCad/kicad-footprints/pull/978) has not yet been merged.

This is the symbol for the HopeRF RFM69CW module. This is similar to the RFM69HW symbol already in the KiCad library. It is functionally similar to the RFM69HW, but it has 14 pins instead of 16, and a different pin mapping.

I created a pull request for the corresponding footprint [here](https://github.com/KiCad/kicad-footprints/pull/978). Please see it for the datasheet URL. I suggest that this symbol is not merged until the footprint is also merged, as I have referenced it within the symbol.

image


myfreescalewebpage:

myfreescalewebpage:
Hi @SeanDS , thanks for contributing,

- [ ] "Transciever" not correctly spelled in the Description
- [ ] Pin offset should be 20mils (KLC S3.6)
- [ ] I would suggest to move GND pins to get them centered on the symbol
- [ ] Datasheet link is broken. Should be https://www.hoperf.com/data/upload/portal/20181127/5bfcb767eb0f2.pdf
- We usually put reference in top left corner and name in top right corner even if it is not mandatory

I know other HOPERF modules to not fit all those rules but they should be fixed in the future.

Cheers,
Joel

myfreescalewebpage:
@SeanDS ping! Are you still alive on this subject ?

SeanDS:
Hey @myfreescalewebpage, I don't have time to fix this now and I already printed my board with this footprint (and it fits fine). If you're happy to wait I might eventually get round to doing it, but it may be months.

myfreescalewebpage:
Hi @SeanDS no worries, that was just to notify you, you fix when you have time :)

myfreescalewebpage:
@SeanDS indicated in the footprint PR he is no more working on this subject for the moment.

PR #1013, opened 2018-09-28
Add QFN-32 packages for ATmega8/16/32U2

fauxpark:

8u2

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/doc7799.pdf

------------


evanshultz:
Thanks! A couple things:
- Please put the `1EP` in the footprint filter, to make it `QFN*1EP*5x5mm*P0.5mm*`.
- Can you shrink the symbol by 100mil on top and bottom?
- To make it less cluttered, perhaps move down the left side pins?

fauxpark:
Fixed the footprint filter, whoops. As for the shrinking, I've deliberately left that space on all of the AVR symbols as I think it looks less cluttered. Same goes for the pins on the left, all of the USB AVR symbols look like this. Probably there should be a little more space between the XTAL/USB data pins, but I think that should be a separate PR since that would also affect many more symbols.

evanshultz:
OK. I'm fine going for consistency over possible goodness. Whenever is the right time we can revisit more extensive pin placement to align all symbols. Thanks!

PR #1014, opened 2018-09-28
Add LMZM23600 Power Module

cp-aquila:
Add LMZM23600 0.5A DC/DC Converter
An alias for the 1A part is included as well.

It matches the footprint from this PR: https://github.com/KiCad/kicad-footprints/pull/953
Datasheet: http://www.ti.com/lit/ds/symlink/lmzm23600.pdf

------------


myfreescalewebpage:
Hi @cp-aquila, thanks for contributing,

- [ ] There are several versions of the device in the datasheet (adj and fixed versions), You should create some alias for each and set the description according to the alias
- [ ] The description should look like "0.5A, _XXX_ V Step-Down DC/DC Power Module in 3.8x3mm Package" with _XXX_ is the output voltage, not the input voltage range, for consistency with other devices in the library
- [ ] I would change the keywords to "Step-Down DC/DC Module" to be consistent with other devices in the library
- [ ] "MODE" should be called "MODE/SYNC"
- [ ] Please check the travis output and solve the KLC S5.1 error
- [ ] Still for consistency with regulators, please move the VIN pin to the left side in the top left corner, EN centered on the left side, MODE on the left side in the bottom left corner
- [ ] Description and datasheet of the LMZM23601 are not correct (should be with 1A and http://www.ti.com/lit/ds/symlink/lmzm23601.pdf)

Cheers,
Joel

evanshultz:
- Voltage and current do not need to be duplicated in the keywords (another way of saying the above).
- Add the package to the end of the description (`MicroSIP-10`).
- Place the value (PN) on the upper right.

myfreescalewebpage:
@cp-aquila thanks.

- [ ] For consistency with other devices, description should be `1A Adjustable Step-Down DC/DC Power Module in 3.8x3mm, MicroSIP-10` (format of footprint at the end).
- [ ] I do not agree with the name "LMZM2360x", it should be "LMZM23600", with aliases "LMZM23601", "LMZM23600V3", "LMZM23600V5", "LMZM23601V3" and , "LMZM23601V5". Your first commit was better on this point of view, I was just pointing you should add 4 aliases for fixed versions.
- [ ] Description can contain "adjustable" in place of the voltage for "LMZM23600" and "LMZM23601"
- [] Symbol not very sexy, maybe you can do something like:

capture

Joel

myfreescalewebpage:
> That was done because the partname in the editor always shows "LMZM23601", which is not correct for all versions.
Somewhere I fucked up and did not make this consistent over all aliases.
TBH, the realtion Partname <-> AliasName is not that clear to me.

In the symbol editor, yes. When you add a part in a schematic, the name is correct. The display is a limitation of the symbol editor only.

- [ ] Just seen current description is not correct for LMZM23600V5 alias (only this one), should be 0.5A instead of 1A.

Joel

cp-aquila:
Sorry, that was a stupid copy/paste error.
New push is in.

myfreescalewebpage:
@cp-aquila thanks for your quick fix.

@antoniovazquezblanco I think this PR is ready to be merged.

antoniovazquezblanco:
The FB pin is similar to the ADJ pin in other devices of the lib. Should we wait to have consensus on wether that pin should be input/passive?

myfreescalewebpage:
@antoniovazquezblanco other device (https://github.com/KiCad/kicad-symbols/pull/1169) decides also the FB/ADJ is an input. No more modifications to do on this one. PR ready to be merged according to me.

antoniovazquezblanco:
Thank you everybody!

PR #1017, opened 2018-09-28
Add LAN Ethernet Transformer

cpresser:
I could not find a symbol for a single ethernet lan transformer.
This PR adds one.

https://www.bourns.com/docs/Product-Datasheets/PT61017PEL.pdf

Footprint: https://github.com/KiCad/kicad-footprints/pull/986

------------


myfreescalewebpage:
Hi @cpresser , thanks for contributing,

- [ ] I not agree with the name of the symbol, there are many of Ethernet LAN Transformer, many are different. Should be atomic, not generic. Name "PT61017PEL" is okay for me. If other devices have the same symbol and pinout, it is still possible to add some aliases.
- [ ] Footprint field should be blank to allow aliases in the future.
- [ ] Generic or atomic, pin number must be shown and pin name must be inside the symbol. Pin name position offset set to 20mil. Pin length 100mil. See for example symbol "Wuerth_7499010121A". It is a Ethernet connector with drawing inside. Not bad to follow.
- [ ] Keywords a little bit redundant with the description. I propose `single port ethernet transformer` based on other existing symbols.
- [ ] Description is not enough, I propose `Ethernet LAN 10/100 Base-Tx Transformer with Center Taps`

Need to wait for the footprint actually.

Cheers,
Joel

cpresser:
Hi Joel,

I did update the symbol. And added an alias for the BEL transformer that has a footprint.

myfreescalewebpage:
Hi @cpresser thanks for the modifications.
However, I disagree with footprint and also the alias. You chosen to make the device atomic (which is good choice), in this case KLC recommends to associate the right footprint (it isn't). Also, the Bel device have not the same footprint (package dimensions are not the same, so it should be another device with the right footprint assigned too.
Joel

poeschlr:
For anybody else wondering:

bourns body size: 12.8+/-0.25mm x 6.9mm(ref)
bel body size: <12.32mm x < 7.24mm

In other words: bel is shorter but wider than bourns.

cpresser:
Hi,
tbh, I fail to see where the problem is.
Each one (symbol/alias) has the respective footprint set (with https://github.com/KiCad/kicad-footprints/pull/986 still open...).
Why would the BEL transformer need another symbol? It has the same pinout.

poeschlr:
One can not assign different footprints to aliases.

myfreescalewebpage:
@cpresser same pinout do not mean same footprint and as indicated by @poeschlr it is not possible to assign footprint to the alias. With other words, in the editor, "Footprint" field is common to the symbol and all the aliases.

cpresser:
> 
> 
> One can not assign different footprints to aliases.

Sorry, I was not aware of that.
Do different footprint-filters works for aliases?

cpresser:
> 
> 
> > One can not assign different footprints to aliases.
> 
> Sorry, I was not aware of that.
> Do different footprint-filters works for aliases?

After looking at KLC again, that would not help. It seems like atomic symbols need the footprint symbol to be set.

poeschlr:
This is a known issue with kicad as it is right now. The alias system is simply not even close to powerful enough. This is one of the motivations behind the new file format that will most likely come with version 6.

myfreescalewebpage:
> After looking at KLC again, that would not help. It seems like atomic symbols need the footprint symbol to be set.

Yes, that's why it's required to separate the symbols.

> This is a known issue with kicad as it is right now. The alias system is simply not even close to powerful enough. This is one of the motivations behind the new file format that will most likely come with version 6.

That's really welcome :)

myfreescalewebpage:
Footprint has been merged, merging the symbol now.

PR #1018, opened 2018-09-29
Added ADP3336ARMZ

tmineno:
## Added ADP3336ARMZ, a 500mA LDO

Datasheet: http://www.analog.com/media/en/technical-documentation/data-sheets/ADP3336.pdf
adp3336armz

Note: Diff shows many differences in Microchip LDOs but I did not change these parts. My KiCad may change orders of these parts in .lib file, I think.

------------


evanshultz:
Thanks for making changes!

The footprint filter should be just `MSOP*3x3mm*P0.65mm*` without the pin count, assuming this is indeed the correct footprint. Can you check out my questions above? https://www.alibaba.com/product-detail/Electronic-Component-ADP3336ARMZ-L22-MSOP8_60696987280.html and http://www.componentsindustry.com/details.html?product_id=5565 look like relatively normal MSOP pins, but I don't see a picture of the underside to confirm it's all epoxy.

Also, since we're changing the look of the symbol can you update the screenshot in the original post or put a new screenshot in a comment? That will help the review.

tmineno:
Thank you for the confirmation, @evanshultz .
I fixed my patch.

>     * It appears there is no exposed metal under this part, so a normal MSOP footprint is OK. Also, Figure 4 of the datasheet looks like it has wide pins while Figure 5 looks like normal pin widths. Can you confirm either or both of those items?

Yes, ADP3336ARMZ has ADI's RM-8 package (normal pin width of MSOP) as shown in Figure 5. 
Figure 4 just illustrates "Paddle-Under-Lead" architecture.

>     * Footprint filter doesn't contain a leading `*:`, or pin count, and use asterisks instead of question marks. We can finalize this once the above item is closed.

At this moment, I changed the filter to `MSOP*8*3x3mm*P0.65mm*`.

evanshultz:
@tmineno 
See my comment above. We were probably cross-posting.

tmineno:
@evanshultz 

Apologize. I missed your comment about footprint filter. I fixed that as you pointed out.

evanshultz:
Thanks!

PR #844, opened 2018-08-23
Meanwell IRM-30/45/60 AC/DC converters

Sthing:
Symbols for Meanwell IRM-30/45/60 AC/DC converters.

Footprints: https://github.com/KiCad/kicad-footprints/pull/858

Datasheets:
 - http://www.meanwell.com/Upload/PDF/IRM-30/IRM-30-SPEC.PDF
 - https://www.meanwell-web.com/content/files/pdfs/productPdfs/MW/IRM-45/IRM-45-spec.pdf
 - https://www.meanwell-web.com/content/files/pdfs/productPdfs/MW/IRM-60/IRM-60-spec.pdf

image

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #1020, opened 2018-09-30
Added AD6645 and AD6644 analog to digital converter.

pgrudzinski:
Datasheets:
- AD6644: http://www.analog.com/media/en/technical-documentation/data-sheets/AD6644.pdf
- AD6645: http://www.analog.com/media/en/technical-documentation/data-sheets/AD6645.pdf

Footprint: 
- merged with PR [#844](https://github.com/KiCad/kicad-footprints/pull/844)

Picture:
image

------------


evanshultz:
@DanSGiesbrecht 
Are you going to review this?

DanSGiesbrecht:
@evanshultz I intend to, but I don't have internet in my apartment, so this is limiting my ability to complete reviews pretty much until the weekend. If you have time and want to do it before I get a chance, go for it!

evanshultz:
- Add package type to the end of the description (TQFP-52).
- Remove the EP size from the footprint filter, change the first character to "T", and force a match to footprints with a single exposed pad. It should be `TQFP*1EP*10x10mm*P0.65mm*`.
- Put ref des at top left and value at top right.
- VREF should be Passive type.
- Move VREF towards the bottom so it's cap is easier to get to GND.
- Space apart C1 and C2 so a cap can fit there.
- Pins 32, 35, and 52 could go on the left side since they will likely interface with the other pins there.
- You have two pin 27s.

pgrudzinski:
>     * Add package type to the end of the description (TQFP-52).
Done
>     * Remove the EP size from the footprint filter, change the first character to "T", and force a match to footprints with a single exposed pad. It should be `TQFP*1EP*10x10mm*P0.65mm*`.
Done
>     * Put ref des at top left and value at top right.
Done
>     * VREF should be Passive type.
Done
>     * Move VREF towards the bottom so it's cap is easier to get to GND.
VREF is used to drive common mode input on differential amplifier like on the drawing, so it is better to leave it close to AIN pins 
image

>     * Space apart C1 and C2 so a cap can fit there.
Done
>     * Pins 32, 35, and 52 could go on the left side since they will likely interface with the other pins there.
These are digital pins and will in general go wherever data pins are going, especially DRY which is output clock.
>     * You have two pin 27s.
Done

evanshultz:
Thanks for making changes.

> VREF is used to drive common mode input on differential amplifier like on the drawing, so it is better to leave it close to AIN pins

Yes, but the ENCODE pins are in the way of getting to ground. What if pins 5 and 6 were at the top, then the AIN pins, and finally VREF, C1, and C2 at the bottom? That seems best for wiring up this symbol to me. Do you have another suggestion?

> These are digital pins and will in general go wherever data pins are going, especially DRY which is output clock.

I could argue against DMID, but fair enough.

evanshultz:
It's always convenient to include a screenshot when the symbol is visually changed. Here it is now:
image

There are a couple things I note, which aren't incredibly critical:
1. We typically put the "highest" symbol name as the ALIAS. So the symbol would be named AD6644 and the ALIAS as AD6645. Here it is backwards.
1. The symbol is horizontally unbalanced. While the centering algorithm is often not what we would visually choose, and it's ignored in favor of other preferences when there is convention, I don't see a reason that it needs to be off-center in this case. I would think ENCODE (and its complement) could be next to each other to save a bit of vertical room, if C1 is moved up 100mil a typical 400mil tall cap symbol can fit there nicely, and then we can end up with this:
image

What do you think? Would you make a few more tweaks?

pgrudzinski:
C1 and C2 don't need capacitor between them, only to ground. On the other hand ENCODE might need termination resistor or limiting diodes, so I would keep C1 and C2 together and in turn ENCODE apart.

evanshultz:
Oh yes. Thank you. Sure, that sounds good to me. Please make the change and post/update the screenshot.

pgrudzinski:
Done. Screenshot updated in the first post.

evanshultz:
Thank you!

PR #1021, opened 2018-09-30
Add AD9851 and AD9850 DDS synthesizers

pgrudzinski:
Datasheets:
- [AD9851](http://www.analog.com/media/en/technical-documentation/data-sheets/AD9851.pdf)
- [AD9850](http://www.analog.com/media/en/technical-documentation/data-sheets/AD9850.pdf)

Screenshot:
image

------------


evanshultz:
Our other ADI DDS ICs are in Interface.lib. Can you please take a look at them and match the style?

evanshultz:
And thanks for contributing to KiCad! I see this is your first PR and having support from the community is so important!

I'll take a closer look at the symbol later.

evanshultz:
- Please use the exact datasheet description followed by the package type, which in this case is `SSOP-28`.
- All pins should be 100mil in length.
- Remove the leading question mark from the footprint filter.
- While they may be interchangeable, these parts aren't exactly pin compatible so use two symbols.
- Put refdes at top left and value at top right.
- 9851: DACBP pin doesn't have a pin number.
- 9851: Although I like the symmetry of the VIN/VOUT pins, I would like to see pins 12 and 17 at the bottom of the right side so if anything is connected to them there is an easy path to ground.
- 9850: You can do more stacking of power and ground pins on this symbol if you like, but since it might make the symbols appear less interchangeable I'm on the fence about this.

pgrudzinski:
I am not sure what do you mean about parts not being interchangeable. From AD9851 datasheet:

> The AD9851 is available in a space-saving 28-lead SSOP, 
> surface-mount package that is pin-for-pin compatible with the 
> popular AD9850 125 MHz DDS. 

The only difference is naming comparator outputs QOUT vs. VOUT. Otherwise they differ only by maximum frequency.

I moved comparator inputs to the right, as they are  either used by connecting DAC outputs to them or not used at all. 

As for further stacking, I would prefer not to, as they are grounds with different names and defined functions, so it is better to keep them separate.

evanshultz:
Logically they are interchangeable, yes, but because of the pin name differences I'd like unique symbols. That way, if the user is only going to use one part or the other the symbol pin names match the datasheet.

The comparator is uncommitted and could be used for any purpose. The datasheet even mentions this. However, it's included because it does have a specific and common purpose for this IC. that being said, I'm OK placing all comparator pins on the right side as you have.

Sure. No more stacking is fine too.

One thing I would like to see: Can you spread apart the two output pins to fill up the space on the right side? That will allow more room for the resistors that go on the output for easier routing. Then please include a screenshot of both symbols for easy future reference. Thanks!

evanshultz:
Here are the screenshots I requested above:
image
image

- AD9850 pin names are wrong on pins 5, 6, 9, and 17.
- Change Pin 12 type to Passive.
- DACBL/DACBP seems like it should not be connected, but to what could it be connected? I don't see anything in the datasheets. If it could be a cap then make the type Passive. I can't think of a more suitable type since I suppose it could be connected in some possible application.

pgrudzinski:
> Here are the screenshots I requested above:
I have uploaded them and put them in the first post. Is it not the correct way to do that?
> 
>     * AD9850 pin names are wrong on pins 5, 6, 9, and 17.
Done.
>     * Change Pin 12 type to Passive.
Done.
>     * DACBL/DACBP seems like it should not be connected, but to what could it be connected? I don't see anything in the datasheets. If it could be a cap then make the type Passive. I can't think of a more suitable type since I suppose it could be connected in some possible application.
Done.
Datasheet is kind of vague about that, so I left it visible just in case.

evanshultz:
Thank you.

AD9850 now has pins 6/23 and 5/23 stacked. Above you said you didn't want to stack them to keep the symbols more interchangeable and I agreed. Do you want to unstack or should I merge?

pgrudzinski:
I was not aware that designation of those pins is different in AD9850,
hence my comment. You can merge as it is.

pon., 8.10.2018, 19:07 użytkownik evanshultz <notifications@github.com>
napisał:

evanshultz:
Just to confirm because I'm not sure what comment you're referring to, this means that if a user places the AD9850 symbol and then wants to switch to AD9851, they cannot use a different power supply for the PVCC/PGND pins. While I think this is fine in isolation, it makes the parts seem less compatible.

I would still think having a separate pair of DVDD/DGND pins on AD9850 gives the most flexibility.

It's not clear to me why ADI didn't use the same pin numbering or share a datasheet...

PR #1022, opened 2018-10-02
Add UFBGA-32 packages for ATtiny48/88

fauxpark:

t48

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/doc8008.pdf

The ATtiny48/88 pin positions are a bit unbalanced, but I heard I shouldn't be moving pins around at this stage of the dev cycle. For now I've used the existing layout.

I also renamed the -MMH package to -MMU. H indicates NiPdAu finish on the leads, U is matte tin (page 283) and apparently the default for Atmel MCUs. Not really something that needs to be specified in the symbol name - if it needs to be x'd out I'd like to do that in a seperate PR. I suspect there are more instances.

------------


evanshultz:
It looks to me like the H or U suffix neither one matter since our current philosophy is not to capture that in MPNs. The same with the R suffix as KLC explicitly says not to include packaging. So I think we can lop all of that off after "MM". I'll add this to the big Atmel issue.

fauxpark:
Fixed the underscore. (why doesn't Travis pick that up?)

evanshultz:
Fair point. I'm not sure why. Maybe someone has a valid reason to allow underscores at one time but I can't think of a reason at the moment.

Thanks once again for all your comprehensive and diligent work!

PR #1023, opened 2018-10-02
Add DFN-8 packages for ATtiny4/5/9/10

fauxpark:

t4

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8127-AVR-8-bit-Microcontroller-ATtiny4-ATtiny5-ATtiny9-ATtiny10_Datasheet.pdf

------------


SchrodingersGat:
Ironically for an `ATTiny` the symbol looks very large. Is this to stay in line with the existing symbols? I would think it should be a bit more compact there's much empty space.

fauxpark:
Yeah, I settled on a minimum size of 1000mil in #504, although apparently I just said width there. I guess I can do another round of shrinking once all these outstanding PRs are done.

poeschlr:
I would suggest to make new symbols as small as they can reasonably be made without reducing readability.

As soon as symbols are in the lib (and published via a release) we can no longer change things that move pins. So my suggestion is to either make this new symbol the size you want it to be or we merge as is and leave it till version 6 development is open.

evanshultz:
This one is fine to shrink since it's a new symbol. It can be noted as a reference of how we would like other symbols to look when the time comes to update existing symbols.

The footprint is a bit wrong. The pad size is 0.9x1.5mm nominal. Since the clearance between the thermal pad and outer pads is only 0.2mm nominal, we can't have a bigger EP and still meet IPC requirements for solderability (no bridging) on this part type. I don't see the correct footprint in our lib so it may have to be created. Would you be willing to do that?

Please remove the EP size from the footprint filter.

fauxpark:
Technically this is not a new symbol, as these MCUs are already in the library in SOT-23 package. For the sake of consistency (and being able to switch between the two) I don't really want to shrink just this DFN one.

I'll look at creating the footprint.

evanshultz:
Thanks. All that is missing is to make the right footprint and then link to that footprint here.

evanshultz:
Please update the symbol to use the new footprint and then I can merge this.

evanshultz:
Thanks once again for your fine work!

PR #1024, opened 2018-10-02
Add QFN-28 packages for ATmega48

fauxpark:

m48

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2545-8-bit-AVR-Microcontroller-ATmega48-88-168_Datasheet.pdf

------------


evanshultz:
The EP size is on this and another footprint filter and it should not be there.

The description doesn't differentiate what the V suffix means. This is likely a common issue amongst Atmel MCUs so I'll make it an issue.

Note that in this datasheet, on page 12, it appears the H suffix is for T&R reel packaging and so it should be wildcarded or simply left off since it's the last character. The R suffix is for an alternate lead finish which we also don't need to concern ourselves with according to the current philosophy.

fauxpark:
Sorry, I recalled Rene doing stuff with the footprint filters in the process of improving the QFN footprints and had it in my head that he added the EP size instead of removing it.

I suspect the V may have something to do with low voltage as they are lower speed and have a lower minimum voltage rating.

The datasheet actually contains an error in the annotation numbers: H is for the alternate lead finish and R is for tape & reel (they are correct on pretty much every other datasheet).

evanshultz:
Ha! Figured this is the one datasheet that is messed up. Anyway, it still holds that neither thing is something we're adding to MPNs in the official library.

PR #1029, opened 2018-10-03
Add symbol for XP Power JTE series dual output DC to DC converter.

stambaughw:
Add symbol for XP Power JTE series dual output DC to DC converters.

Datasheet: https://www.xppower.com/Portals/0/pdfs/SF_JTE06.pdf

Footprint pull request: https://github.com/KiCad/kicad-footprints/pull/992

------------

xp-power-jte-series-symbol


stambaughw:
Just so I understand the footprint filter issue, I should replace *XP_POWER_JTExxxxDxx* with *XP?POWER?JTExxxxDxx*.  This doesn't seem right to me since XP_POWER_JTExxxxDxx is the absolute footprint name.

poeschlr:
We typically do not fill out the datasheet field that is stored in the lib file but only the one in the dcm file. (As far as i know it will be copied automatically into this field on symbol placement if it is empty in the library.)

Only one of the power output pins can have "power output" as pin type. ERC would complain otherwise if they are connected to each other. So my advice would be to stack them (make one of them invisible and give it the pin type passive.)

As footprint filter i would propose either `*XP_POWER*JTExxxxDxx*` and ignore the error by travis or use `*XP?POWER*JTExxxxDxx*` (In other words use a "*" between the manufacturer name and the part number. Makes it a bit more flexible for users and also a bit more future proof.)

Everything else looks ok.

stambaughw:
I will remove the unnecessary datasheets and change the outputs to used
passive invisible stacked pins.  Do you make the pins zero length when
you stack them or should they be the same size as the visible power
output pin?

poeschlr:
The visible and invisible pins should be equal in everything except pin type and visibility flag. At least in kicad 4 one needed to setup the pins before stacking them as opening the pad properties dialog for any of them changed some things for all pins in the stack. Not sure if this is still the case in v5. (I have not yet designed any symbol with v5)

stambaughw:
Does it make sense to stack the power input pins as well?  They get tied
together at the board level so it might make for a cleaner schematic.

poeschlr:
As long as they do not need separate capacitors per pin then yes stacking them makes sense. Just make sure the invisible one is also of type passive as it will otherwise be converted into a global label.

stambaughw:
I just pushed the fix for the stacked pins and the footprint filter so
this symbol should be good to go.  Please let me know if I need to
change anything else.

poeschlr:
The only problem left is the datasheet field. You set it in the lib file while we require the use of the dcm file for that. I know that this rule could theoretically be more relaxed now as both fields should behave the same in v5. But as it makes the test script easier i still prefer this more flexible solution that allows setting different datasheets for different aliases. (Even if in this case this will mean that all aliases have the same datasheet assigned.)

In current nightly this would be done in the aliases tab instead of the general tab of the symbol properties dialog.
I seem to remember that in v5 the place where you set the datasheet in the dcm file is reachable via the symbol properties dialog (The button with the gear) whereas the .lib field is reachable via the symbol field dialog (the button with the letter T). But i am really uncertain how these two dialogs have been called.

stambaughw:
Sorry about that.  I misunderstood your original request.  Just so I
have this straight, no datasheet in the lib file and all aliases get a
datasheet link.

poeschlr:
Yes this is correct.

stambaughw:
I pushed the changes so this should be good to go.

poeschlr:
Thanks, looks all good now.

PR #1030, opened 2018-10-03
Adding the TI THGS5641A DAC

marcusmueller:
This adds the smallest (8 bit) of the CommsDAC family.
Both supported footprints (SOIC/TSSOP) available with this symbol.

Easily extended to further (more-bitty) DACs from the same family by converting the
appropriate (invisible) NC pins to D[8:15] pins (and making them
visible, and enlarging the symbol).

Component website: (Datasheet PDF as specified in symbol)
http://www.ti.com/product/THS5641A

Image: image from TI website

------------


marcusmueller:
I hope I'm right adding this to `Analog_ADC`; is it the appropriate category? Or does "Analog" mean "ADI"?

evanshultz:

- Please include screenshot of the symbol from the Symbol Editor, not the package 3D model from the datasheet. Like this:
image
- Yes, that is the correct library. No, Analog does not mean ADI. No, it doesn't make sense to me either.
- Please buff up the description and add the packages. Something like `8-Bit, 100 MSPS, CommsDAC, Parallel Current DAC, SOIC-28/TSSOP-28`.
- The keywords are space-separated, not comma-separated.
- Include the body sizes in the footprint filters; so `SOIC*7.5x17.9mm*P1.27mm*` and `TSSOP*4.4x9.7mm*P0.65mm*`.
- All pin should be 100mil long.
- Ref des goes at the top left of the symbol and value at the top right.
- I would expect to see D7 at the top since it's the MSB.
- Make BIASJ Passive type since it goes only to a resistor to ground.
- Put COMP1 at the top right and COMP2 at the bottom right.
- We only use the Line style so please fix pin 21.
- Correct the names of pins 21 and 22.
- SLEEP is always mentioned as being connected to analog power/ground, so move it to the right side.

evanshultz:
Also, once you've got all the pins moved around try to shrink the symbol. Surely it can be less tall then, but you can also make it less wide.

marcusmueller:
Wow, thanks for the intense work and the extensive feedback!

Not quite sure I follow on the SLEEP aspect, though: that pin is a digital "enable" input, and as input I've put it left.

If you're preferring to have it right, I'll still do that, just wanted to confirm.


The data pin ordering has practical background: with D0 on top, the user can place a label, e.g. "data0", and then use the "insert" key to duplicate-and-increment. I'd rather not change that; MCU pins are arranged the same.

On 4 October 2018 00:35:21 GMT+02:00, evanshultz <notifications@github.com> wrote:
>Also, once you've got all the pins moved around try to shrink the
>symbol. Surely it can be less tall then, but you can also make it less
>wide.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/KiCad/kicad-symbols/pull/1030#issuecomment-426826274

-- 
Diese Nachricht wurde von meinem Android-Gerät mit K-9 Mail gesendet.

evanshultz:
Since the analog pins are on the right in a DAC, and the datasheet explicitly says those are the pins to connect SLEEP to, yes, let's move it. Pin placement recommendations are the default if there's no information that would make us reconsider, but in this case we do have a reason to break the normal guideline for input pins on the left.

Data pin ordering is OK as it is. The MCU pins are arbitrary, but it does make labeling the nets easier which is quite helpful.

evanshultz:
@marcusmueller 
Are you planning to finish off this PR? Can you look at my comments above?

aris-kimi:
Hi @evanshultz 
i did something as shown:
THS5641A
How should i proceed? Create a PR and refer to this one as duplicate?

aris-kimi:
I am closing this one since symbol has been submited by newer PR #2830.

PR #1032, opened 2018-10-04
Fix the VDD pin type for the PIC16F83 and PIC16F84. (from Bidirection…

RTrioux:
…al to Power pin)

Fix the pin type for VDD in MCU_Microchip_PIC16.lib (Some VDD pins were stated as 'bidirectionals' whereas a VDD pin should be stated as a 'power pin')
Moreover according to the datasheet these pins are unidirectionals 

https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&ved=2ahUKEwigjL6sqO3dAhUEgHMKHc_0CN4QFjAAegQIBhAC&url=http%3A%2F%2Fww1.microchip.com%2Fdownloads%2Fen%2FDeviceDoc%2F30430c.pdf&usg=AOvVaw0Tabj1A2bWEHudO92Cr7xD


evanshultz:
Better to post the direct datasheet link of http://ww1.microchip.com/downloads/en/DeviceDoc/30430c.pdf.

Good catch! Thanks for noticing this issue and submitting a fix. We appreciate all the community support and I hope you will contribute more in the future!

PR #1034, opened 2018-10-04
Visible pin numbers on R networks

evanshultz:
As noticed at https://github.com/KiCad/kicad-footprints/issues/961, there are no visible pin numbers on resistor network symbols. This makes knowing which resistor in the package you've connected to difficult, which is annoying since they're interchangeable.

This PR turns on pin number visibility for the entire group of symbols.
image

At higher pin counts the pin number runs over the symbol body, which looks ugly, but I think having pin numbers with this caveat is better than no pin numbers. Unless there's a real reason to do more, I'd rather have this merged and then think about a better way later once we have more freedom to change symbols.

------------

### DIFF
--- a/Device.lib
+++ b/Device.lib
@@ -3217,8 +3217,8 @@ S 50 50 50 50 0 1 0 N
 S 50 150 50 250 0 1 0 N
 S 50 250 50 250 0 1 0 N
 S 110 330 -110 -300 0 1 10 f
-P 2 0 1 0 -50 -150 -50 -250 N
 P 2 0 1 8 -50 -150 -50 -250 N
+P 2 0 1 0 -50 -150 -50 -250 N
 P 2 0 1 0 -50 0 -100 0 N
 P 2 0 1 0 -50 0 50 0 N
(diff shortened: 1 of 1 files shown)


fauxpark:
These are script generated. I have an open PR at KiCad/kicad-library-utils#256 that makes the script much less of a headache to read and modify :)

evanshultz:
Thanks. I'd rather get this merged quickly and switch to the nicely-updated script versions later.

evanshultz:
Confirmed that the script-based symbols are different, so please merge this if acceptable and we can do more in the future.

myfreescalewebpage:
@evanshultz @fauxpark The pin numbers are on the symbol itself, probably it will be better if the pin are longer to get a good visibility (especially above pin 10).

evanshultz:
@myfreescalewebpage 
These symbols did not have visible pin numbers. We are not breaking schematics during the 5.x cycle, which includes now, so that means no moving pins at this time.

antoniovazquezblanco:
Looks fine to me as a temporal fix until version 6.

Thanks!

PR #1038, opened 2018-10-06
Add LMH6551 fully differential opamp

pgrudzinski:
Two parts, LMH6551MA with SOIC footprint and MM with VSSOP footprint

Datasheet: http://www.ti.com/lit/ds/symlink/lmh6551.pdf

Screenshot:
image

------------


pgrudzinski:
I am not very fond of the idea that output pins are not differentiated in any way. They are not interchangeable and that should be clearly marked. I can settle for + - labels on outputs as this seems to be established alternative way of labelling.

evanshultz:
I understand what you're saying and we have made a recent effort to label all pins on opamp-like symbols. The geometry of the triangle shape makes this a little hard because the pin name offset applies to all pin while some pins look better with varying offset distances. We preferred to use pin names rather than floating text and these pins without pin names are the ones ones I now see in a quick search.

Do you have a proposal on how to show the labels you mention above?

pgrudzinski:
I would propose this:
image

Alternatively circle can be put as part of the geometry and not as a pin feature.

evanshultz:
The crash between the lower pin number and pin name with the triangle is the issue. This is with a long pin. If the pin is shorter there is no crash with the pin number, but then the pin name is outside the triangle (and covered up for a pin named `-`) and we need a short line to join the pin and the triangle body. I don't see a good solution with our current tools. At least not what I consider good. I do like having both pin names and numbers but we're limited by what KiCad can do right now.

If having both name and number in this case is required the pin could be shorter with a number outside the body, the pin could be named `~` to suppress the pin name from being visible, and text of the pin name could placed inside the triangle inline with the pin.

I do feel rather strongly that it's most obvious to put the positive pin on top.

@poeschlr @Misca1234 @antoniovazquezblanco @DanSGiesbrecht 
Thoughts?

pgrudzinski:
For this reason I like the notation with negation dot. If using negated pin
symbol is not a possibility (why?) then the dot could be part of geometry.

pon., 8.10.2018, 19:14 użytkownik evanshultz <notifications@github.com>
napisał:

pgrudzinski:
One more thing that was bugging me about those LTC6362 and I just confirmed that it is true. They have non inverted output on the same side as non inverting input, which is very misleading, because correct connection of feedback looks like this: 
image
If you do not cross feedback paths like this, you end up with positive feedback. For me this is reason enough to label pins somehow, even if it is ugly or overlaps with something. I prefer my schematics ugly than wrong.

evanshultz:
Thanks. These are good points. Let's gather from feedback from the other librarians then we can decide what to do here and I'll update any other symbols that need it.

evanshultz:
@poeschlr @Misca1234 @antoniovazquezblanco @DanSGiesbrecht 
Any thoughts or preferences? There are two items I see surfaced above:
1. How to best show pin numbers and names in this scenario.
1. In which position should the output pins be in a differential opamp.

I'm inclined to do what is proposed at https://github.com/KiCad/kicad-symbols/pull/1038#issuecomment-427671491 and accept the slight overlap. At least then all pin numbers and names are shown and uniform, and there is (by default) contrast between those elements so they're visually distinct even when crashing.

Misca1234:
what about this idea

bild

evanshultz:
I believe previous conversations, including this one, desire both pin numbers and pin names for having total clarity on the schematic sheet even when printing to paper. This is the only please I'm aware of where this now isn't done for opamp/comparator symbols.

evanshultz:
@pgrudzinski 
You have a major typo in your commit, probably from cut n paste. Please open in KiCad or review it and you'll see. It's missing the two lines that close the symbol.

I'm happy with the style you showed at https://github.com/KiCad/kicad-symbols/pull/1038#issuecomment-427671491 and will merge unless somebody objects. (That also means fixing any other similar symbols, which I will do.)

pgrudzinski:
I have fixed the missing lines and added another commit changing labels to a normal text which is positioned to avoid overlapping with the outline. It looks better for me, but you can leave off this commit from merge if you prefer the earlier solution.

evanshultz:
Thanks for fixing the files. Please always open your commit in KiCad to catch issues before pushing it to GitHub. (I'm guilty of doing that sometimes so this is a general request.)

Here is the latest commit:
image

Here is the previous one using the pin names and not text:
image

I prefer the previous commit without text both visually and for file maintenance and self-documentation (as the pin itself is named). However, hopefully in the future KiCad will provide better options. Since this question of how to visually portray the pin name is opened again, any more comments?

evanshultz:
I'll merge the previous commit (the lower image above) unless there is a comment against it in the next 24 hours.

PR #1042, opened 2018-10-07
Added symbols for AD849x Thermocouple Amp

johnfwhitmore:

This is a Pull Request for the four members of the Analog Devices AD849x family of Thermocouple amplifiers.

Datasheet - http://www.analog.com/media/en/technical-documentation/data-sheets
/ad8494_8495_8496_8497.pdf

screenshot from 2018-10-07 20-38-17


evanshultz:
- Please use the datasheet description of `Precision Thermocouple Amplifiers with Cold Junction Compensation`, then you can add `J-Type Thermocouple, 0 to 50C` or some other helpful information, and finally include the package of `MSOP-10`. Some variation is allowed but the general format seems to work well throughout the library.
- What would you think about using a regular opamp triangle symbol? It can fit two pins on the bottom for REF and SENSE (or one on top and one on bottom).
- I think SENSE would be best as a Passive type.

The datasheet link got messed up above but it's not a big deal.

johnfwhitmore:
screenshot from 2018-10-10 11-59-34
screenshot from 2018-10-10 11-59-44

johnfwhitmore:
I'm not happy with the SENSE pin as it don't fit in the symbol. In the same way the '-IN' and '+IN' pins have had their names truncated to make them fit in. I might have to go to a size up.

evanshultz:
We allow smaller text sizes if required to fit in the symbol body. That would work here. Take some cues from similar parts in the lib, such as this one:
image

And using a rectangle is OK if you don't like this or it simply doesn't work; this part seems to map naturally to a triangle so I thought it would be worth investigating.

johnfwhitmore:
Thanks a million for that pointer. I'm happy with the Amp symbol just wasn't happy with text. Didn't think that text size could be changed, due to guidelines but should have checked other parts.

johnfwhitmore:
screenshot from 2018-10-14 21-07-24

evanshultz:
Thanks! It looks nice.

A couple more things:
- Can any pin text be bigger? For example, if REF and SENSE are flipped could they use 30mil text? Same for pins 1 and 8, if possible.
- Push the NC pin inside so the pin terminus is on the symbol outline. Then it will be unlikely to be accidentally connected.
- Can you slide the ref des and value both on the right side like most (all?) of the triangle symbols?

johnfwhitmore:
Sorry just getting back to this. I swapped the 'REF' and 'SENSE' pins so that I could make the text bigger. Don't know if I had them in the original order for a reason, I can always change them back again.
screenshot from 2018-10-17 10-38-06

evanshultz:
Looks good to me. The connection point of the NC pin isn't on the symbol body but it will still be rather ugly if someone connects it.

Thanks!

johnfwhitmore:
Oops sorry about NC & Thank you

PR #335, opened 2018-02-28
Atmel MCU DIP footprints/filters

fauxpark:
Fixed up footprint association and filters for some Atmel MCUs - just the DIP packages here, but the scope of the PR can be extended to include TQFP/QFN/etc. as I have a few of those done too.

I didn't fix any of the other issues but if they don't get fixed before the new release, they are still usable and will at least get proper footprint associations.

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

Issue #300, opened 2018-02-14
Atmel (ATMEGA) MCU symbols have the wrong footprint assigned and wrong footprint filters


poeschlr:
Especially most ATMEGA symbols have footprints with very old footprint naming assigned (no body size, pitch, ... in the name)
They sadly also do not use wildcard chars in the footprint filters but simply copied the footprint field. (So not even the footprint filter will help the user out here.)

Originally reported over at the forum in: https://forum.kicad.info/t/no-footprint-for-tqfp-32/9625

bobc:
ATMEGA also has problems with too long pin names. It looks like the footprint name is copied from source data. I guess they were generated by script, does anyone know?

poeschlr:
The long pin names are a different topic (and covered in a different issue.)

The reason i separate these two is that the pin name stuff is a KLC issue whereas this is a bug that prevents the symbols to be used as intended.

bobc:
Ok, that makes sense. If there was a script to generate the symbols, it seems to be lost in the mists of time, the symbols appear in github in 2008 and I can't find anything earlier.

I think it will have to be done from datasheets.

fauxpark:
I have started some work on this in #335 :)

fauxpark:
The merge of #340 has resolved this issue (I should have linked it so that it autoclosed).

poeschlr:
Thanks

PR #1043, opened 2018-10-07
Add missing QFN packages for ATtiny2313A/4313

fauxpark:
Added ATtiny2313A/4313-MU as aliases of ATtiny2313V-10MU (4x4mm QFN-20), and ATtiny2313A/4313-MMH (3x3mm QFN-20) as a separate symbol.

sym

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/doc8246.pdf

------------


evanshultz:
- The packages in the description could be named MLF-20 and VQFN-20 to differentiate them.
- The symbol `ATtiny2313V-10MU` has a footprint with an EP of 2.25mm sq, but the actual package is 2.6mm sq nominal. `ATtiny2313A-MMH` is using a 1.6mm sq EP footprint while the nominal package size is 1.55mm.

Yes, the latter EP/pad size difference is basically inconsequential but can we allow any exceptions when making footprints is so easy and it will be of benefit if the delta is larger? This makes me think an entire review of all Atmel footprints is likely to turn up at least some minor issues. I see #300, #335, #338, #340, etc. have all gone through this but there are probably stragglers.

fauxpark:
What is the correct EP size for these then? I'm not sure how to calculate it.

I see this in the [YAML definition](https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_DFN_QFN/size_definitions/qfn-2x.yaml) for the 3x3mm QFN:

```yaml
# For backwards compatibility use 1.6mm even thought nominal would be 1.55
EP_size_x: 1.6
EP_size_y: 1.6
```

The 4x4mm footprint with the 2.25mm EP is not scripted, but there is one with a 2.5mm EP. Much closer but still too small.

evanshultz:
I see.

Well, no reason not to add the actual size now that we've got a script and we're not concerned with backwards-compatibility on this new symbol. And for the 2.6mm one, no harm in making a correct footprint.

Would you like me to submit a PR with those two footprints using the script? Or perhaps you'd like to do this one?

fauxpark:
Ah, the 1.6mm comes from our old friend [Atmel Appnote 8826](http://ww1.microchip.com/downloads/en/AppNotes/Atmel-8826-SEEPROM-PCB-Mounting-Guidelines-Surface-Mount-Packages-ApplicationNote.pdf). Is this what the "backwards compatibility" is referring to?

I think you might have a better idea of the appropriate values to plug into the generator, so by all means go ahead, if you have the time. Otherwise I'll have a crack at it.

poeschlr:
@fauxpark wrote:
> What is the correct EP size for these then? I'm not sure how to calculate it.

I did not yet come around to add this to the KLC: https://github.com/KiCad/kicad-website/issues/332

By default the EP size should be equal to the nominal size of the metal pad (lead or slug) of the package.

If the manufacturer specifically states in the datasheet or appnote for this particular part that it should be sized differently then this size is to be used. (Your mentioned appnote however is not really fitting for the part in question.)

Even if there is a document sugesting a different size for the EP, care must still be taken. Especially if the suggested pad size is outside of the min/max size ranges for the package pad. (For larger suggestions it will most likely be that the pad will be a soldermask defined pad -> soldermask will have a cutout with the size inside the tolerance ranges.)

evanshultz:
Footprints submitted at https://github.com/KiCad/kicad-footprints/pull/1005.

evanshultz:
The footprints were merged so can you update the description and default footprint and footprint filter to match and then this is good to go. If you shrink the symbol that's fine with me, but for commonality with other packages I can understand leaving it alone.

fauxpark:
I think I should fix the description in a separate PR because that also affects most of the QFN symbols. Currently they all just use "QFN" and their footprints are the same - it may require renaming or duplicating footprints to match what the datasheets call them.

evanshultz:
Sure. The description can be done later. And any actual footprint renames later, if required.

PR #1046, opened 2018-10-08
Adding MICRF112

gcormier:
DataSheet : http://ww1.microchip.com/downloads/en/DeviceDoc/MICRF112.pdf
image

image

Footprint good. 3x3mm, 0.5mm pitch.

Only thing I was unsure of was the ENABLE pin. I know inputs are on the left but it makes the right side very lonely and increases the height overall.


DanSGiesbrecht:
Also, I can see why you have multiple footprint filters. I think there's some ambiguity in the KLC. Please see: https://github.com/KiCad/kicad-website/issues/343.

gcormier:
Thanks so much for the detailed feedback! I will look at these this afternoon.

What is the guideline/etiquette on multiple packages.. in this case there are two.. however, let's say an IC has 5 packages, are contributors expected to do a PR with 5 packages if they are using only one?

(This one has 2, not a big deal :) )

DanSGiesbrecht:
Thanks for taking a look!

With regards to etiquette - because this is a volunteer-based system, there's no requirement for people to contribute *all* variants. I think contributing what you're comfortable with is reasonable.

Say there were 5 packages: I think contributing one symbol is fine, and that the important thing is making the library compatible for others to contribute the remaining 4 packages.

DanSGiesbrecht:
Sorry, I don't quite understand what you mean. You don't know how to get *what* back to your branch?

gcormier:
Ah, disregard :)

gcormier:
Let me know how this round looks!

DanSGiesbrecht:
@gcormier will do! I'm at work right now, but I'll check it later tonight. Thanks for the quick fixes!

DanSGiesbrecht:
Your changes look really good!

I did notice a few things:
- Please change add the body size to the footprint filter: `MSOP*3x3mm*P0.5mm*`
- Please swap the position of `ASK` and `FSK` (see comment in previous review), unless there's a reason you'd like to keep it as-is (not a big deal).

I think then we can go ahead and merge this. 😄

DanSGiesbrecht:
Sorry, I wasn't clear. Please also remove the first `*` from the footprint filter, in front of MSOP: `MSOP*3x3mm*P0.5mm*`

DanSGiesbrecht:
@gcormier Everything looks great. Thank you for contributing to the libraries! 👍

PR #669, opened 2018-06-15
Makeover of Sensor_Magnetic and add AH1806

jneiva08:
Hi,

This PR will be the biggestes done by me up to now.
Tried put Sensor_Magnetic to respect KLC, the changes are:
* Pin name position offset to 20mil,
* Centered  symbols to be symmetrical,
* Moved the NC ot the edge,
* Updated the datasheet link was broken,
* Updated some footprint filters,
* Added A110x  P/N for SIP-3 package and updated description,
* Added AH1806.

Need you help to try understand why A1101ELHL, A1101EUA, A1301KLH and A1301KUA-T still getting error on S4.4, It looks how to should be.
On checklib-py -v gives this message:
"    Pin electrical type should match pin function
     Pin types should match pin function"

Need you option about should proceed the footprint for AH1806. For SOT23W need be other pin order and SOT553 need to have other measures.

Now the symbols looks:

a110x_sot

a110x_sip

a130x_sip

a130x_sot

ah1806_sot553

ah1806_sot_sip

as5045a

as5045b

as5048b

as5055a

mmc5883ma

sm35xlt

tlv493d

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #1048, opened 2018-10-08
Added AH1806 family

jneiva08:
Added Symbol for Diodes AH1806

Datasheet: https://www.diodes.com/assets/Datasheets/AH1806.pdf
Screenshot:
ah1806

ah1806_sto-553

PR for SIP-3 Ammo Pack: https://github.com/KiCad/kicad-footprints/pull/1001

This symbol was from closed PR: https://github.com/KiCad/kicad-symbols/pull/669

------------


jneiva08:
@evanshultz 
If I remove ex "-P-A", all the symbols will have the same name, how should naming to have SC59,SOT533 and SIP3 independent?

evanshultz:
The last two characters are a hyphen and a letter, so in the example above retain the "-P" but ditch the "-A".

jneiva08:
@evanshultz 
The footprint filter for AH1806-P, i think it should be "Sensor?Current:*SIP*", if only have "*SIP*" it will appear footprint from Racom. What is your opinion?

evanshultz:
Yes, I think you're right that we need to be more specific. I don't have any information if this SIP-3 package is specific or generic. So unless we get more information, we could have two footprints named `Diodes_SIP-3_4.1x1.5mm_P[1.27|2.65]mm`. How does that sound to you?

Some other things:
- Remove leading asterisks.
- Use a generic SOT-553 footprint for -Z.
- Add default footprint to -P and update footprint filter as discussed above (or add new info).

jneiva08:
Sorry for taking some time.

Most of SIP package i know is DC-DC for example Recom/Traco.
Allegro have  1.27 (already on master), but the 2.65 have discontinued.
Diodes have SIP-3 2.65 and 1.27 can find other brands have other size of pitch, can find on the following link: https://www.digikey.com/products/en?keywords=sip%203%20hall%20switch

evanshultz:
So then it should be good if we use the manufacturer name and footprint as above, right? Are you bringing up an issue I'm overlooking in your comment above?

Thanks for coming back to this!

jneiva08:
I have renamed the footprint plus Allegro SIP-4 to be the same of SIP-3 on PR [#1072](https://github.com/KiCad/kicad-footprints/pull/1072).

Make sense to update in this PR the footprint that use Allegro SIP-3/4 or create new one?

I'm missing anything you requested?

evanshultz:
Let's limit this PR to just the Diodes Inc part. Thanks for asking.

We shouldn't include the pin count in the footprint filter. And since there are two possible lead preps and they're both Diodes, Inc. parts, what about add these two footprint filters: `Diodes*SIP*P2.65mm*` and `Diodes*SIP*P2.65mm*`?

We use asterisks unless a question mark is necessary so that explains the above and there's also an question mark that can be converted in the SOT553 package.

jneiva08:
The AH1806-P have two possible packages Allegro_SIP-3_4.09x1.52_P1.27mm (Diodes Bulk Pack compatible) and Diodes_SIP-3_4.1x1.5mm_P2.65mm, i have changed footprint filter to:
- Allegro*SIP*P1.27mm*
- Diodes*SIP*P2.65mm*

Allegro did discontinued the Pitch 2.65mm or equivalent.

And AH1806-Z change back to be *SOT*553*

evanshultz:
I think we should only have Diodes SIP footprints here. While the Allegro may just happen to have the same body size (from a quick check they looked to same to me) I still want all footprints here to start with `Diodes`. Or remove any vendor prefix and just have `SIP-3_Px.xxmm`. But not mixed vendor prefixes as it's too weird and confusing with the potential to break stuff in the future.

The -Z version should not have a leading asterisk.

jneiva08:
sorry for take too long to reply.
The AH1806-Z, footprint filter needs to be "SOT*553*", right?

evanshultz:
No problem. :) Just remove the leading asterisk from the two SMD symbol footprint filters and this is good.

evanshultz:
Thanks! I'd like to get the footprints merged before this, so let's finish that off next and then come back to this. :)

jneiva08:
already added on https://github.com/KiCad/kicad-footprints/pull/1072, in the last 2 commits

evanshultz:
Now that the footprints have been merged, can you please add the body size to the footprint filter for the -P version. Then I think this is ready to merge as well.

jneiva08:
@evanshultz, Did you want to mean this?

ah1806-p

evanshultz:
`Diodes*SIP*4.1x1.5mm*P1.27mm*` and `Diodes*SIP*4.1x1.5mm*P2.65mm*` should be good. Thanks!

jneiva08:
Have change with the filter you recommended.

evanshultz:
Thanks!

Now seeing this again an hour later we could have left off the pitch so both footprints would be selected and other pitch options later would be included without touching this symbol, but all footprints in the current datasheet are included. Plus, this PR has been opened a long time and you've been patient and diligent to make changes so it's ready to merge. No sense in pushing things out any more.

jneiva08:
Thanks

PR #1005, opened 2018-09-26
Connector: pin stacking on USB_C_Receptacle and USB_C_Plug

diegoherranz:
Follow up of https://github.com/KiCad/kicad-symbols/pull/937

In order to keep the compatibility with the previous symbol,
the symbol hasn't been recentred.

plug
receptacle

Thanks!

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1049, opened 2018-10-08
remove trademarked USB superspeed logos from USB3 connectors

dlharmon:
I don't believe the USB 3.0 A/B and USB C connector symbols are usable by anyone in the USA who is not a USB licensee due to a trademark. https://trademarks.justia.com/775/82/ss-77582723.html

All USB trademarks:
https://trademarks.justia.com/owners/usb-implementers-forum-inc-2584569/

Example of modified symbol.
screenshot from 2018-10-08 17-35-15

The USB trident does not appear to be trademarked, so I left it. They were not as aggressive with such things in the past.


evanshultz:
@poeschlr @stambaughw

diegoherranz:
A few days after working on #937 and #1005, I started wondering whether we can use the USB logos or not.

In case we had to remove the SuperSpeed from the logo (don't know the legal/trademark stuff), I think we should at least make it look pretty and not just remove the "SS". E.g. having the complete USB logo (that screenshot above is missing the big dog at the bottom end which looks weird), recentring the logo, etc.

Thanks.

myfreescalewebpage:
@KiCad/librarians anybody able to help here ?

cpresser:
I did a quick search and read some pages.
The most important seems to be this: https://www.usb.org/logo-license

This is what I learned so far:
 * The USB logo must be reproduced exactly, which is not the case here since "SS" is written in the KiCAD font.
* There are no Test-Requirements for the SS-Trident-Logo in this case (N/A) in this document: https://www.usb.org/sites/default/files/trademark_license_agreement_licensed_mark_requirements_final_as_of_november_30_2018_locked_0.pdf
* There is a licence-agreement required for the use of any usb logos: https://www.usb.org/sites/default/files/trademark_license_agreement_fall_2018_clean_final_0.pdf
* I don't assume that this agreement is compatible with the librarys open-source-licence.
* Besides, one has to pay $3500 to enter that agreement

As a next step I wrote an email to the usb-if and asked about the requirements of using the logos in the context of EDA software. Lets see If I get a usefull answer.

cpresser:
I actually got an answer from the USB-IF Administration:
>The USB-IF logos may only be used with products that are certified and posted to the Integrators List under that company's name, and the company must execute a USB-IF Trademark License agreement. 
> Please note the USB-IF does not certify software.

My interpretation is that all USB logos that are managed by the USB-IF should be removed from KiCad.

poeschlr:
@cpresser does this include the normal usb symbol or just the superspeed one? (meaning is the "trident" ok as indicated by @dlharmon?)

cpresser:
> @cpresser does this include the normal usb symbol or just the superspeed one? (meaning is the "trident" ok as indicated by @dlharmon?)

That includes all USB Logos in this file: https://www.usb.org/sites/default/files/trademark_license_agreement_licensed_mark_requirements_final_as_of_november_30_2018_locked_0.pdf

The "trident" in its standalone-version is not included in that file.
My other research also suggests it is not trademarked. So the conclusion is that we can keep it.

For clarification, this is what I am calling the trident-logo:
image

poeschlr:
Perfect. So we can at least include something that will show it is a usb connector. @dlharmon if you could add the missing circle then this should be good to go.

PR #1053, opened 2018-10-09
Add ATmega48/88/168P

fauxpark:
Just like #1050, seems these are older/lower-end versions of the ATmega48/88/168A, so again I've swapped them out to be the parent symbols.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8025-8-bit-AVR-Microcontroller-ATmega48P-88P-168P_Datasheet.pdf

48pv

------------


evanshultz:
This looks fine and I can merge, but I have one question first: You made one parent symbol `ATmega48PV-10AU` that used to be `ATmega48A-AU`. Would the "A" not come before "P" and thus be the parent?

fauxpark:
I'm ordering the aliases from lowest to highest end - note that the 48PV is the parent, with the 48P as the first alias, *then* the 48A. This is because the 48PV runs at 10MHz vs the 48P's 20MHz, thus it is (to me anyway) lower end. As I mentioned above, they also seem to be much older. The 48A's datasheet claims its first revision was in 2016, and the 48P's datasheet says 2007.

Obviously which is the parent and which are the aliases isn't critical, but I like consistency :)

evanshultz:
Right. That's why I asked so I know know your methodology. Thanks again!

PR #1060, opened 2018-10-11
Fixed pinout of USB6B1

kajusK:
Hi,
I've noticed the symbol for USB6B1 does not match the [datasheet](https://www.st.com/content/ccc/resource/technical/document/datasheet/3e/ec/b2/54/b2/76/47/90/CD00001361.pdf/files/CD00001361.pdf/jcr:content/translations/en.CD00001361.pdf).

The fixed symbol looks like:
image

Also it breaks KLS in two points according to the checklib script:
Checking symbol 'USB6B1':
  Violating S4.2
    Ground and negative power pins should be placed at bottom of symbol
  Violating S4.4

As it's the ESD protection, all pins are passive, therefore these errors are false positives.

Kajus


evanshultz:
Good catch on the pinout error.

And you're right that the errors are OK for this symbol.

Now that I think symbols can change a bit more, I would rather stack pins and change to this:
image

DanSGiesbrecht:
@kajusK by stacking the pins ([KLC S4.3](http://kicad-pcb.org/libraries/klc/S4.3/)), you can represent this symbol with 4 pins instead of 8.

kajusK:
@DanSGiesbrecht Thanks for the explanation. But does it really make sense in this case? Imagine the board layout when pins 1 and 8, 2 and 7, 3 and 6, 4 and 5  should be connected together. There would be 4 traces running below the chip, possibly complicating the layout as no other traces can be routed below the chip on the same layer. In that case it would make more sense to use a chip in smaller footprint (e.g. sot 23-5).

The original schematic keeps the possibility to route the signals through the chip. What's your opinion?

DanSGiesbrecht:
@kajusK that's a good point! Thanks for pointing that out. I do see the validity in forcing the path through the device, rather than bypassing it underneath the part.

kajusK:
Any update on this? Is there anything that blocks this pull request from merging?

evanshultz:
Hmmm...

Here's the original symbol:
image

The changes above do fix the pinout, but the pin placement is dramatically different than before. Could you retain the pin placement while fixing the pinout? I think the symbol you've made is not as clear as having the I/O pins on each side and I'd like to keep the VBUS and GND pins on top and bottom. While it's valid to have the symbol's pin location match the package's pin locations, this library does not do this but instead intentionally separates logical from physical representations.

Thanks for taking this time to catch this error and fix it, and sorry for the delay. If you can make a few more adjustments I'll get this merged.

kajusK:
Hi @evanshultz,

the pin order had to change as two more pins were added (one for gnd and one for vbus). Also the original part makes it hard to do the wiring on schematic. Check this image. If you wire all the signals from usb through the device, it's much easier and cleaner to do it this way:

2018-11-13-212935_1920x1080_scrot

Also the power pins are not powering anything, these pins are passive, it doesn't really make sens to consider them to be different from the IO pins. This way, it just follows the convention - "inputs" on the left side, "outputs" on the right.

Changing the part should not brake any existing schematics as the pinout was not correct before anyway so no boards with this part could possibly work.

What do you think?

evanshultz:
I do understand about the wiring. However, wiring `I/O1` and `I/O2` gets messy if you want to connect both terminals together. I don't see a single symbol style that is without any caveats.

I'm also interested in maintaining consistency for similar symbols. And while this library is a mess, it's gotten a bit better now and the symbol I pasted in my first post is the common style used now. Matching the USB6B1 style, including shrinking it, would be nice.

When updating from an old symbol, the schematic will not break at all and the pinout will be fixed if pin placement is retained as much as possible. Here is a style quite similar to the existing one but with the fixed pinout that you noticed and helpfully brought up:
image
```
# USB6B1
#
DEF USB6B1 U 0 0 Y Y 1 F N
F0 "U" -250 350 50 H V C CNN
F1 "USB6B1" 300 350 50 H V C CNN
F2 "Package_SO:SOIC-8_3.9x4.9mm_P1.27mm" -975 -400 50 H I C CNN
F3 "" -950 -100 50 H I C CNN
$FPLIST
 SOIC*3.9x4.9mm*P1.27mm*
$ENDFPLIST
DRAW
C -200 0 10 0 1 0 F
C -100 0 10 0 1 0 F
C 0 -350 10 0 1 0 F
C 0 -250 10 0 1 0 F
C 0 250 10 0 1 0 F
C 0 350 10 0 1 0 F
C 100 0 10 0 1 0 F
C 200 0 10 0 1 0 F
S -300 -300 300 300 0 1 10 f
S -100 250 100 -250 0 1 0 N
P 2 0 1 0 -200 0 -100 0 N
P 2 0 1 0 -100 -350 100 -350 N
P 2 0 1 0 -100 350 100 350 N
P 2 0 1 0 -60 -110 -140 -110 N
P 2 0 1 0 -60 190 -140 190 N
P 2 0 1 0 0 350 0 -350 N
P 2 0 1 0 60 -110 140 -110 N
P 2 0 1 0 60 190 140 190 N
P 2 0 1 0 100 0 200 0 N
P 3 0 1 0 40 40 -40 40 -40 20 N
P 4 0 1 0 -300 100 -200 100 -200 -100 -300 -100 N
P 4 0 1 0 -140 -190 -60 -190 -100 -110 -140 -190 N
P 4 0 1 0 -140 110 -60 110 -100 190 -140 110 N
P 4 0 1 0 -40 -40 40 -40 0 40 -40 -40 N
P 4 0 1 0 140 -190 60 -190 100 -110 140 -190 N
P 4 0 1 0 140 110 60 110 100 190 140 110 N
P 4 0 1 0 300 100 200 100 200 -100 300 -100 N
X Vcc 1 -100 500 150 D 50 50 1 1 P
X I/O1 2 -500 100 200 R 50 50 1 1 P
X I/O2 3 500 100 200 L 50 50 1 1 P
X GND 4 -100 -500 150 U 50 50 1 1 P
X GND 5 100 -500 150 U 50 50 1 1 P
X I/O2 6 500 -100 200 L 50 50 1 1 P
X I/O1 7 -500 -100 200 R 50 50 1 1 P
X Vcc 8 100 500 150 D 50 50 1 1 P
ENDDRAW
ENDDEF
```

But most important, more than anything else above, is that the symbol is fixed and you are happy to have contributed to KiCad and would be interested in contributing more. So, if you don't want to make any further changes let me know and we'll book this PR to be merged in a couple weeks after 5.0.2 is released.

kajusK:
In the same way wiring becomes messy when you want to pass all signals through the device.

If you wouldn't mind, I would stick with this variant, it looks better to me (plus it's already used in my latest project :) ).

EvilMav:
In case it ends up a democratic choice, in the spirit of making the common schematic case clean and considering old netlists will be broken with numbering change anyway, I'd place my vote on kajusK's version...

evanshultz:
@kajusK 
Would you mind moving the power pins to the top and bottom? We are nearing 5.1's release so this is a good time to merge this PR.

kajusK:
@evanshultz How exactly should I move these? Should I stack them, or keep them separated? Should I keep the internal diodes connected to these pins as on the original symbol proposal?

evanshultz:
Thanks for coming back to this PR!

If I understand your comments above you would like separate pins (not stacked), so please do that for both VBUS and GND pins. Yes, please keep internal lines going to the pins so connectivity is clear.

evanshultz:
@kajusK 
Are you able to make the change above to get this merged quickly?

evanshultz:
@kajusK 
Did you have any more questions about the last request above? You have caught a real error in the library and I would like to get it fixed for 5.1's release. Are you able to help get this done soon?

kajusK:
@evanshultz 
Sorry for a long delay, I was quite busy lately... I've moved the power pins to up and bottom as requested, however I don't find this look very elegant. It's huge and will take a huge portion of the shematic.

2019-02-24-201756_1920x1080_scrot

With pins arranged as requested, it might be better to ignore the internal wiring and make it a bit smaller. Also stacking power pins makes sense when placed like this, though it  removes to option to wire power line through the device itself.
2019-02-24-202636_1920x1080_scrot

evanshultz:
@kajusK 
Thank you!

I am OK with either style. Hiding the pin names, shorting the pin lengths, and leaving the internal "schematic" is another option. Which of the three options do you prefer?

kajusK:
Well, I would go with my latest proposal (hidden pins and hidden internal schematic)

evanshultz:
OK. Let's go for it. :)

evanshultz:
Not stacking GND and VBUS pins is OK as well, but please put them on top and bottom.

kajusK:
Ok, pushed the latest change

evanshultz:
For reference, here is the symbol now:
image

Please change pin name offset to 20mil. Then I'll merge.

kajusK:
Done

evanshultz:
Thank you. I'm happy to see this fix will be included in the library.

PR #1061, opened 2018-10-11
Add OPA355NA high speed opamp in SOT-23-6

pgrudzinski:
Datasheet: [OPA355](http://www.ti.com/lit/ds/symlink/opa355.pdf)
Screenshot:
image

------------


evanshultz:
- Change pin name offset to 5mil.
- Please remove the commas from the description and adhere to the datasheet description.

pgrudzinski:
I will fix the offset, but how does your comment correspond to S6.3 of KLC?
I'm quite sure that I did exactly as it says. It does not say that
description must be 100% in accordance to datasheet, although it makes
sense to use the same terms as manufacturer. It does say it have to be
comma separated, so I took what is used in datasheet and separated terms
with commas. It looks exactly like illustration in KLC. I would leave the
description as is.

evanshultz:
Then please fix the pin name offset and I'll merge.

evanshultz:
Thanks again!

PR #1062, opened 2018-10-11
Add LTC1742/4/6/8 high speed ADCs

pgrudzinski:
Datasheets:
- [LTC1742](http://www.analog.com/media/en/technical-documentation/data-sheets/1742f.pdf)
- [LTC1744](http://www.analog.com/media/en/technical-documentation/data-sheets/1744f.pdf)
- [LTC1746](http://www.analog.com/media/en/technical-documentation/data-sheets/1746f.pdf)
- [LTC1748](http://www.analog.com/media/en/technical-documentation/data-sheets/1748fa.pdf)

Screenshot:
image

------------


evanshultz:
- Add package type to the descriptions.
- 1742 has a double space in the description.
- `Pitch` in the footprint filter should be just `P`
- I'm thinking pin 1 should be Input type?
- And perhaps pin 2 should be Output or Power Output? I don't see an output current listed but it obviously can source at least a bit of current with only 4ohm Zout. Those two are a bit tricky.
- Can REFLA/HB and REFLB/HA be spread apart a bit to fit a cap in there?
- Maybe pins 23/24 should go below pins 4/5/2/1 so that ~ENC has an easier path to ground if unused?
- Since I can see ~OE being tied to ground, what about putting at at the bottom on the right side?
- On the right side I might think about grouping D* pins then the other 4 in a separate group.

evanshultz:
Thank you!

PR #1064, opened 2018-10-13
Cypress touch sensor fix

evanshultz:
http://www.cypress.com/file/46236/download

image
image
image
image
image
image

Update of https://github.com/KiCad/kicad-symbols/pull/350.

------------


myfreescalewebpage:
Hi @evanshultz , thanks for contributing,

- [ ] Can you fix also S4.4 (see travis output) by moving VCC to the top of the symbols and changing the type to Power Output ?

@KiCad/librarians if my understanding is correct, when all pins have been moved, it is necessary to wait Kicad 6, right ?

Cheers,
Joel

evanshultz:
@myfreescalewebpage 
Welcome. Thanks for commenting. It's important to read the datasheet and understand what's going on during a review. The check scripts help to catch items that are easy to overlook, but they're basic and manual intervention coupled with a bit of thinking is often required. That is the case here.

Passive type and the location of the VCC pins is correct. See the linked PR if you want the whole story.

It's more of a judgement call about moving. Rene has asked to avoid large-scale changes within a version cycle, but for smaller changes there's potential wiggle room. I don't mean that applies here, just in general. And, if the change is to provide a fix for a broken symbol that's possible any time.

poeschlr:
Could you provide a short description of what the changes are you did here?
(just some bullet points would be enough.)
The changes to the file are too much to easily see the intended changes and especially the reasoning behind these changes. (The discussion in the linked pull request is also a bit convoluted and hard to follow so it is of no particular help.)

evanshultz:
Sure! It's been a little while, but here is what I recall and can find from looking at the diff:
- Pin name offset to 20mil.
- Assign default footprints and update footprint filters.
- Simplified pin names to only the basic function.
- Change SCL pins to Bidirectional type.
- Change VCC to Passive type due to the connection possibilities.
- Make all CS* pins Passive type since they can be touch inputs.
- Align pin placement for all symbols in this family: Data pins on left side, power pins on top/bottom, etc.

Now that I'm reviewing this again, and with the passage of time that allowed you to write the nice footprint generator, we can do better footprints. The QFN24 should have a 2.65mm EP and the QFN16 should be 1.7mm as can seen on page 35 of http://www.cypress.com/file/46236/download.

evanshultz:
I've submitted ideal footprints at https://github.com/KiCad/kicad-footprints/pull/1179.

myfreescalewebpage:
@evanshultz ok for your comments

evanshultz:
Footprints have been merged.

@myfreescalewebpage 
Please take up your crusade regarding pin types in a dedicated issue, or add onto an existing ones if there is a suitable open issue. This PR conforms with the way we've been doing pin types and I would not like to block this PR for an much "larger" issue.

myfreescalewebpage:
@evanshultz as said in another PR few days ago, I agree with the use of Passive to get ERC checking working without unwanted errors.

I have looked at this PR a long time ago when I was not in the team, so without the possibility to change tags etc etc. I'm currently checking another PR, but I can check this one just after :)

myfreescalewebpage:
Checked, no more comment. Merging.

PR #1071, opened 2018-10-16
Add AD7606 series AD Converters.

cybertale:
Add AD7606 series AD Converters.

------------

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing : http://www.analog.com/media/en/technical-documentation/data-sheets/ad7606_7606-6_7606-4.pdf
AD7606:
ad7606
AD7606-6:
ad7606-6
AD7606-4:
ad7606-4


evanshultz:

- Add the package (`LQFP-64`) to the end of the description.
- Please stack AGND and REFGND pins. (AVCC and REGCAP pins should not be stacked according to notes on page 25 and 15, respectively.)
- Pins 3/4/5 are configuration pins and could go next to 6/7/8/34 instead of leaving a gap.
- REFGND can be on the bottom, on the left side, of the symbol.
- You can make the symbols smaller (both height and width) once this is all done.
- Spread apart the analog input pins once the symbol is smaller. Put a gap between each pair of pins for clarity. I think that will make the best use of the area on the left side.
- Pin 7 should have a tilde after the words "PAR" to limit the overbar to only that function.
- DB14/15 are clear why they are Bidirectional type, but for DO0-13 I don't see when they are inputs (even though the datasheet says that on page 15). Do you?
- The DB pins are Bidirectional now, and can be high-Z depending on the state of ~CS and ~RD, but I don't see any reason to make them Tri-state. Either way, no ERC errors will result if they're connected to multiple AD7606 devices and only one is addressed at a time. The only difference is if the same pins are also connected to other pins which are Output type, as Tri-state will give a warning, but I think Bidirectional should be OK. Is that OK with you too?
- REFIN/OUT should perhaps be Passive type as well. What do you think?
- I have a slight preference to capitalize all letters in pins 24 and 25, if you agree.
- You have used an underscore instead of a space in many pin names, which is good, but not pins 9, 10, 34.

PurgatoryHydra:
Hi Evan,

Thanks for your reply!

> ...
> 
>     * DB14/15 are clear why they are Bidirectional type, but for DO0-13 I don't see when they are inputs (even though the datasheet says that on page 15). Do you?
> 
>     * The DB pins are Bidirectional now, and can be high-Z depending on the state of ~CS and ~RD, but I don't see any reason to make them Tri-state. Either way, no ERC errors will result if they're connected to multiple AD7606 devices and only one is addressed at a time. The only difference is if the same pins are also connected to other pins which are Output type, as Tri-state will give a warning, but I think Bidirectional should be OK. Is that OK with you too?
> 

So Bidirectional should be OK if I'm understanding this right.

>     * REFIN/OUT should perhaps be Passive type as well. What do you think?
I'm getting a little confused here. This pin can be configured as a reference output or reference input, so input is definitely not right. But passive type should not be compatible with reference output, either.
Actually I referred AD7616 for the type of this pin without having a throughout consideration. AD7616 is just like a bigger AD7606. 
I think this pin is quite like a programmed I/O pin described in S4.4, and bidirectional is more fit. 
What do you think?

> 
>     * I have a slight preference to capitalize all letters in pins 24 and 25, if you agree.
I checked some other symbols and see all the 'DOUT' of them are capitalized, I think you are right!
> 
>     ...
Other problems will be fixed soon!

cybertale:
Screenshots of symbols updated.

evanshultz:
Thanks for making updates!
- For DB0-13, I think they should only be Output type.
- Pin types are part of the ERC engine, and here this pin isn't a digital logic pin. Passive is the best fit with an analog pin like this to avoid any false positive ERC errors.
- Please fix pin 6's name as mentioned above. (Sorry that I originally typed pin 7, not 6.) The tilde is only needed if the entire pin name does not need an overbar; pin 7, for example, just needs a leading tilde.

cybertale:
Screenshots of symbols updated.

evanshultz:
Thanks for making changes! DB14/15 should be Bidirectional type, that's why I only mentioned DB0-13 above. If you can fix that, this looks good and I'll merge. Thanks again!

evanshultz:
没问题，谢谢你！

PR #1072, opened 2018-10-16
Add MICRF220AYQS

gcormier:
Open for feedback.

image

http://ww1.microchip.com/downloads/en/DeviceDoc/micrf220.pdf

------------


gcormier:
Need suggestion on footprint filter. It's a single package, unsure why using wildcards is recommended...

gcormier:
I'd like to close this - can I get some help on the footprint filter?

antoniovazquezblanco:
Please use the following FPFilter:
`QSOP*3.9x4.9mm*P0.635mm*`

We are a little bit busy so full review of the symbol may take a little bit more time. I am sorry for the wait.

gcormier:
No problem. Thanks for the help, I see what you mean by wildcards now.

Should be all set whenever you get the time. Let me know if there is any other changes you'd like to see.

antoniovazquezblanco:
Thank you very much!

myfreescalewebpage:
Closing/opening the PR to refresh the CLA.

gcormier:
I agree it is tall. I find the KLC specifying inputs on the left and outputs on the right is not enough for us to make decisions independently.

Following your logic, I would support moving SEL0/SEL1 to the right side as well, as these would likely be controlled by an MCU. SQ could be as well, but it's likely it has a pull-down resistor to reduce the noise when the MCU is sleeping, so I think it is fine on the left.

gcormier:
image

gcormier:
Let me know your thoughts before I do the usual alignment dance.

myfreescalewebpage:
Hello @gcormier thanks for your work

> I agree it is tall. I find the KLC specifying inputs on the left and outputs on the right is not enough for us to make decisions independently.

That's a guidance but it is not strict and for such kind of symbol it's better to think how they are wired in a schematic.

> Following your logic, I would support moving SEL0/SEL1 to the right side as well, as these would likely be controlled by an MCU. SQ could be as well, but it's likely it has a pull-down resistor to reduce the noise when the MCU is sleeping, so I think it is fine on the left.

SEL0/1 can be modified while the device is powered ON ? In this case OK.
Agreed with SQ, good to keep on the left.

> Let me know your thoughts before I do the usual alignment dance.

The above proposition is fine I think, just need to solve issues regarding to the gris and stacking of pins.

Joel

gcormier:
Perfect! I will tidy it up and submit a final PR :)

gcormier:
Hopefully this fixes it up!

myfreescalewebpage:
Thanks @gcormier, no more comment to do following the last modifications, merging the PR.

Joel

PR #1073, opened 2018-10-17
Added symbol for PIC16F1704

johnfwhitmore:
The alias menu has changed, possibly since I upgraded Kicad. I'm not
recognising the interface and how aliases are entered. They're not
showing up in the parts list as yet.

I'm going to create a pull request and see if CI gives me any clues as to why I've got this
lack of aliases and a datasheet in the symbol.

Signed-off-by: John Whitmore <johnfwhitmore@gmail.com>

Datasheet - http://ww1.microchip.com/downloads/en/DeviceDoc/40001715D.pdf

screenshot from 2018-10-17 14-59-56
screenshot from 2018-10-17 15-00-26

PR #1075, opened 2018-10-17
Add PCF8523 RTC

johnfwhitmore:
This is an initial commit of symbol for the PCF8523 RTC,
Upload for CI check, but struggling with datasheet entry for base symbol. Earlier today for another
component it's flagged as missing but if you add it to the general tab it gets flagged for being included.

screenshot from 2018-10-17 20-02-06


johnfwhitmore:
I seem to have lost the knack of this. I thought that PCF8523T and PCF8523TK could be added together as they share the same symbol, but I'm struggling with footprints. 

Should my base class be "PCF8523" and then have two alias components?

myfreescalewebpage:

myfreescalewebpage:
Hi @johnfwhitmore 

For this PR, see also my comment at https://github.com/KiCad/kicad-symbols/pull/1073. Devices need to be duplicated. See also for example RTC already in the library: DS1302 and all the variants.

Joel

johnfwhitmore:
I responded to #1073, but same applies to this one. Getting started but I'll have to try and find that work. It's been a while and out of site out of mind, or vice versa. Thanks for the pointer to DS1302, I'll have to look at it as I'm a bit confused it sounds like the process has changed somewhat.

johnfwhitmore:
Have to add a new footprint for this as the SO-8 in the libs is different?

https://github.com/KiCad/kicad-footprints/pull/1309

myfreescalewebpage:
Hi @johnfwhitmore I have seen SO-8 package has been merged for this device, can you address the above comment, i.e. create two symbols, so we can review and merge. Thanks, Joel

myfreescalewebpage:
@johnfwhitmore ping :)

johnfwhitmore:
Sorry been distracted. Working it now.

johnfwhitmore:
Bare with me it's been a while and I usually build kicad from source. Then
I usually hop between distros so I'm building on this machine for the first
time. This takes a while. Sorry.

myfreescalewebpage:
@johnfwhitmore no worries my previous message was just to get news from you :) Thanks for the pong :)

evanshultz:
We have been driving towards unique symbols if footprints are different, so things indeed may have changed a bit. At since since you and I worked through some PRs. However, between T and TK there would always have needed to be unique symbols because there is a thermal pad on the TK package, so that part should be the same for years.

myfreescalewebpage:
I do not clearly understand you previous message. T and TK versions are different because they have not the same package. Moreover the TK version has one more pin (the exposed pad); So the symbol are different.
Joel

johnfwhitmore:
Sorry deleting this to start afresh.

myfreescalewebpage:
Ok.

johnfwhitmore:
This time #1707

So far so good, but pin placement might be an issue. Trade off between
clearance & size

myfreescalewebpage:
Thanks!

johnfwhitmore:
No problem, hope it's close. Going to sort other footprints.

PR #1076, opened 2018-10-17
More audio jacks

evanshultz:
For the footprints I've been promising (threatening?) at https://github.com/KiCad/kicad-footprints/issues/191, there are a few symbols needed. They are here.

image

I'm not totally happy with them and have the following questions/thoughts:
1. These audio jacks could be grouped together. Perhaps all prefixed with `Audio_`? Then you would have `Audio_Jack2*`, `Audio_XLR*`, `Audio_speakON*`, etc.
1. I also think indicating poles that are switched versus the XLR with a discrete SPDT switch is not clear. Perhaps putting `SPDT` into the name of the XLR jacks would be better than trying to use the word "switch" in two contexts?
1. Many of the audio jacks have lines on odd grids (5mil for example). With small changes I could make them all on at least a 25mil grid without changing sizes substantially.
1. I kept pin placement consistent on XLR combo jacks. Should I instead make the smaller like the corresponding AudioJack symbols?
1. "AudioJack" doesn't communicate much to me. Farnell uses the term "jack" which may be the origination of this term in our library. DigiKey uses "barrel" and Mouser uses "phone". The small DC power jacks use "barrel" so perhaps "phone" is more unique and still understood in our library rather than such an arbitrary term as "jack"? Call them `Audio_Phone3*`?

Shall I make any other changes now? These jacks were made/overhauled not too long before v5 was released so the affected users shouldn't increase significantly if we make changes now and then have audio jack symbols that won't change for the foreseeable future.

I also moved around some of the refdes and value where they previously collided with the symbol.

I do not know why KiCad (release v5) decided to jiggle around a bunch of lines. Sigh.

------------


antoniovazquezblanco:
Closing and reopening to trigger travis.

evanshultz:
Resolved a slew of CI issues and now all green. Also updated some footprints filters to match the footprints I'll be submitting soon (which I believe fit the KLC rules for this footprint type). Off-grid symbols were only shifted by 50mil so no new image since that won't change it on a schematic.

The only possible sticking point I can see is that some G (ground) pins are not a 50mil increment in length. To do this the entire AudioJack symbol will have to be reworked. I can do that if required.

antoniovazquezblanco:
I hope I have time to have a look at this during this week

antoniovazquezblanco:
I would rework the symbols to meet KLC if possible. Being strict, pin lenght should be a warning but the 50mil increment rule should trigger an error.

imagen

Should we ask why the 50mil rule is in place and try to make it into a soft constraint?

evanshultz:
If you're still willing to review and merge for 5.0.2, I'll rework them. :) The existing symbols weren't on a 50mil or 25mil grid which caused the silliness. I'll retain pin positions and keep the symbol size and style similar so it shouldn't break existing schematics.

evanshultz:
OK. I reworked all the AudioJack* symbols to place elements on a 25mil grid while keeping relatively the same sizes (compare with the screenshot above as the symbols auto-updated):
image

All AudioJack* symbols (with XLR combo or standalone) were horizontally compressed to be as small as possible for the pin count (matching the regular AudioJack* symbol for that pin count).

I also removed the "S", and/or "R", and/or "T" text from the XLR_AudioJack* symbols.

Anything else?

antoniovazquezblanco:
I had a look at the symbols.
Naming is a bit confusing sometimes due to the high amount of configurations available for this kind of connectors but I can see what you intended to do and I think I cannot improve it with any suggestion.
Many footprints missing for all of this configurations but obviously it is not your job.
Pin numbering seems to be consistent.

Everything looks fine for me. I would like to merge this ASAP so that reviewing of a couple of footprints like https://github.com/KiCad/kicad-footprints/pull/930, https://github.com/KiCad/kicad-footprints/pull/1014 and https://github.com/KiCad/kicad-footprints/pull/1073 becomes easier but if anyone wants to have a look at this before I merge please write a comment in the next 24h please :).

evanshultz:
Thanks for looking! I agree about naming but I couldn't come up with anything better. The only thing that might help is if we start to divide up Connector.lib and Device.lib to be more specific and then perhaps the context of the library would allow the names to be streamlined? But that can't happen now.

Next week I hope to finally submit a huge slew of Neutrik footprints which drove creation of these symbols.

antoniovazquezblanco:
Thank you!

evanshultz:
@poeschlr @jkriege2 @Misca1234 @antoniovazquezblanco @DanSGiesbrecht 
There are a number of questions in my first post; any thoughts about them?

poeschlr:
Generic symbols have the drawback that we need loads of semi generic variations to be able to have a symbol for most usecases. There is simply a limit what can be done with them. Everything else really requires fully specified symbols that are specific to one part with a footprint already connected. 

If you need longer to choose the correct symbol from the sea of generic options than to draw a new one then you definetly have gone too far with trying to have a generic option for everything. Similarly if we need longer to decide how to get a generic naming scheme that works for everything than we would take to make a few specific symbols then we might as well choose the specific route. (If finding a naming scheme takes long then there is a high chance that we already missed something.)

evanshultz:
That's a good point to not get lost in the weeds. These symbols are all used by the footprints I'll be submitting shortly, but perhaps they should have been specific. Point taken.

The above discusses item 2 and item 5 is not, IMO, worth changing though I wanted to include it. That still leaves item 1. Should we try to group together related symbols within the generic libraries or break apart the massive libraries (`Connector_Audio`, `Device_Transistor`, etc.)?

poeschlr:
For the v5 cycle i would not break apart libs. This is something that might be better done shortly before v6 is nearing feature freeze. (I would say about half a year before the planned release date. Similarly to how we did it for v5.)
This means the next year or two might be a period where we keep the lib organization stable. This gives us maintainers the option to work on other things and it keeps users happy as they get a kind of stable library.

poeschlr:
I did not want to imply that this pull request went too far. My previous answer was more a general answer to the topic of generic symbol. I did not really look at the specifics of this pull request.

If this PR only contained the symbols shown in the screenshot then i am fully ok with the decision to add them. I would also support the additional addition of a separate library for fully specified audio connectors. (This could even happen during the v5 release phase. Adding additional libs is not a problem. Moving stuff around however might be one.)

PR #1077, opened 2018-10-19
Add symbol for CSD16301Q2

jneiva08:
Datasheet: http://www.ti.com/lit/ds/symlink/csd16301q2.pdf
screenshot:
n_mos

------------


evanshultz:
- Perhaps describe it as 5A instead of 20A since that's the package limit.
- Shall we describe the package as `SON-6` since that's what the datasheet shows on page 12? Is there a better name? Usually it's `<package type>_<pin count>` but this one is tricky.
- For most flexibility replace the question mark with an asterisk in the footprint filter. We only use question marks if there's a reason (like SOT-23 vs SOT-223 confusion), and here there isn't.

evanshultz:
Thanks!

jneiva08:
Thanks for the merge :)

PR #1080, opened 2018-10-21
Added INA260

Misca1234:
Re push of 
https://github.com/KiCad/kicad-symbols/pull/470

INA260
http://www.ti.com/lit/ds/symlink/ina260.pdf

It is in the Sensor.lib
GND is not stacked, see page 29

bild

bild

bild

------------


evanshultz:
- IN- should not be Output type. Passive would be OK with me. Perhaps IN+ and VBUS should be too since they all go to a mux.
- I'd be inclined to put the I2C pins and ALERT on the right side since they all interface to a processor and it also keeps the D pins on the "output" side.
- Similarly, I think it would be best if IN- is on the left side below IN+ and VBUS.
- Why can't pins 6 and 11 be stacked? Page 29 only shows separate vias, but that could be getting to a plane quickly for lower inductance. Page 4 indicates those two pins are exactly the same.
- Pin 7 defaults to active low but can be active high. So remove the overbar?

Travis is complaining because the visible pin in the stack is of a different type.

Misca1234:
Some thing like this then

bild

evanshultz:
That looks good to me. Can you shrink the symbol a bit in height and then I'll mege?

Misca1234:
Here we go

bild

evanshultz:
Great. What about moving VBUS back up next to IN+? It seems like they're intended to be connected to the same net.

Misca1234:
Yes, but it looked more symmetric in that way, but it is moved now

evanshultz:
Thanks!

PR #564, opened 2018-04-30
added LP2951 voltage regulator series

jkriege2:
- datasheet: http://www.ti.com/lit/ds/symlink/lp2951.pdf
- missing PWSON-N8 footprint: https://github.com/KiCad/kicad-footprints/pull/541
-remaining KLC-errors due to missing new (!) footprint, see PR above

2018-04-30 18_51_05-symbol library editor - regulator_linear d__kicad_kicad-symbols_regulator_linea
2018-04-30 18_50_41-symbol library editor - regulator_linear d__kicad_kicad-symbols_regulator_linea

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1081, opened 2018-10-21
Added LP2951 8 pin variants

Misca1234:
This is a partly overtake of push
https://github.com/KiCad/kicad-symbols/pull/564

LP2950 is already an alias to
LM79L05_TO92

LP2951 
http://www.ti.com/lit/ds/symlink/lp2951-n.pdf

Pin GROUND was renamed into GND
The DIP package was added despite it is not mentioned in the TAPE AND REEL INFORMATION section

LP2951-30N    3.0V
SOIC
VSSOP
WSON

aliases
LP2951-33N    3.3V
LP2951-50N    5.0V

LP2951-50NP    5.0V
DIP

bild

bild

bild

bild

bild

bild

------------


Misca1234:
Does anyone knows why the FP "WSON*4x4mm*P0.8mm*" , does not find the 
Package_SON:WSON-8_4x4mm_P0.8mm.kicad_mod ?

evanshultz:
- Separate packages with a forward slash not commas. So `VSSOP-8/WSON-8/SOIC-8`.
- Since you're sharing packages, why aren't DIP, VSSOP, and SOIC all sharing the same symbol?
- WSON needs a unique symbol since it has a thermal pad. Page 33 says the EP pin should not be stacked.
- And since the WSON needs a unique symbol, I don't see how the others can be combined into one symbol.
- The symbol names would seem to be formatted as `LP2951<package type>-<voltage including a decimal>`, which doesn't match these symbol names of, for example, `LP2951-33N`.

It looks like a properly-formatted footprint filter. I don't know why it wouldn't find that footprint.

Please include footprint filters in grave accent symbols (used for code in GH comments) so asterisks aren't lost. This will make it much more clear.

Misca1234:
> Since you're sharing packages, why aren't DIP, VSSOP, and SOIC all sharing the same symbol?

DIP only comes in 5.0V variant

> WSON needs a unique symbol since it has a thermal pad. Page 33 says the EP pin should not be stacked.

Jee, true, that is what happens when you hurry to be finished before Total Recall is aired 
I have renamed them, but the DIP one remains as you pointed out in the earlier thread.
So I used the name from the lp2950 datasheet but use the datasheet for lp2951 for the DIP-8

I splitted them also into separate symbols

bild

bild

bild

bild

bild

bild

evanshultz:
- Sorry to just notice this now, but http://www.ti.com/product/LP2951/samplebuy shows the PNs in a different format, with the voltage before the package. Which is right?
- For the DIP package, the TI ordering info in the datasheet and online don't show it. But https://www.digikey.com/products/en/integrated-circuits-ics/pmic-voltage-regulators-linear/699?k=lp2951&k=&pkeyword=lp2951&pv1291=177&pv1291=862&sf=1&FV=ffe002bb&quantity=&ColumnSort=0&page=1&pageSize=25 shows TI, ON Semi, and Microchip all make this part and it comes in DIP with various voltages. Perhaps we should make this a more universal symbol to capture all vendors? In that case, as we've done in a few other places, use the package type (`_DIP`, `_SOIC`, etc.) as the suffix.
- Please use EP instead of EP1 for the WSON package's pin 9 name.
- I think it would be easier to connect if FEEDBACK was on the right side, when the user is not using the internal voltage divider. And that also means VTAP should be on the right side too since if the internal resistors are used those pins are connected together.

Misca1234:
VSSOP is not on 
 http://www.ti.com/product/LP2951/samplebuy

So what do you want
SOIC
LP2951-30DX
LP2951-33DX
LP2951-50DX

SON
LP2951-30DRGX
LP2951-33DRGX
LP2951-50DRGX

VSSOP?

DIP
LP2951-30_DIP
LP2951-33_DIP
LP2951-50_DIP

The mystery is even deeper
http://www.ti.com/lit/ds/symlink/lp2951-n.pdf
contain the vssop variant but 
http://www.ti.com/lit/ds/symlink/lp2951.pdf
does not

However 
http://www.ti.com/lit/ds/symlink/lp2951-n.pdf
is of later date

BUT 
Texas "official data sheet"
http://www.ti.com/product/LP2951/technicaldocuments
points to
http://www.ti.com/lit/ds/symlink/lp2951.pdf

evanshultz:
We know the websites are updated more frequently than the PDF docs. So the webpage is the best information.

See the DigiKey link I sent. More than just TI make this regulator and so you can follow the advice I gave above on symbol naming. This is what we do on, for example, TL072 opamps. I think this used to be in KLC but I can't find it now.

Misca1234:
So you want

LP2951_SOIC
LP2951_WSON
LP2951_VSSOP
LP2951_DIP

evanshultz:
Yes. That is what we have done elsewhere when the same "base" PN is interchangeable from various vendors but the orderable PNs do not match. And that is what we have here.

Misca1234:
What about this ?

bild

evanshultz:
Yes, I think so.

ON Semi Micro8 is the same package as TI VSSOP, but I think that's fine to combine them into a more common single SSOP symbol.

Also note that TI has a SON and WSON package of different sizes. Both packages come in all 3 output voltage versions. See http://www.ti.com/lit/ds/symlink/lp2951.pdf and http://www.ti.com/lit/ds/symlink/lp2950-n.pdf.

evanshultz:
Why add the vendor name? LP2951_WSON and LP2951_SON should work with all parts now in the market.

Misca1234:
This should be it then 

bild

bild

antoniovazquezblanco:
@evanshultz ping! :)

evanshultz:
New scripted footprint merged at https://github.com/KiCad/kicad-footprints/pull/1175. Would you mind assigning that footprint to the mating symbol? Then I'll check this out.

evanshultz:
That footprint PR is referencing the LP2951 datasheet. So yes, that is what I'm saying.

Misca1234:
Foot print changed, look like this

bild

evanshultz:
- WSON symbol: The footprint filter should be `Texas*S*PWSON*N*`, or something similar. This removes the pincount, the underscore, and the EP size.
- `LP2950-5.0_TO92` has the wrong description in Vout and package.
- The SON symbol isn't present, as discussed at https://github.com/KiCad/kicad-symbols/pull/1081#issuecomment-433110851 above. It's not required, but are you leaving it off intentionally?
- Shouldn't ~ERROR be on the right side?

evanshultz:
@Misca1234 
Can you check out my comments above and resolve branch conflicts so this can be merged? We still have time to make 5.1.

Misca1234:
Fixed the FP and the description, moved Error here

bild

evanshultz:
- LP2950-5.0_TO92 did get the "TO" in the package in the description.
- Moving the ERROR and VTAP pins should happen on all LP2951 symbols, not just WSON.

Misca1234:
Now so
The pin layout for the LP2951 could perhaps be a little changed but it can be done later

evanshultz:
`LP2950-5.0_TO92` still needs the description fixed a bit. Otherwise it's ready to merge. Thanks!

Misca1234:
It is now
Micropower Voltage Regulators with Shutdown, 100mA, Fixed Output 5.0V, LDO, /=-92, TO-92

How would you like it to look like ?

evanshultz:
Just change `/=-92, TO-92` to `TO-92`.

Misca1234:
Done

evanshultz:
Thank you!

PR #1082, opened 2018-10-22
Add LT1019 precision voltage reference

ferdymercury:
https://www.analog.com/media/en/technical-documentation/data-sheets/1019fd.pdf

------------


ferdymercury:
image

ferdymercury:
@antoniovazquezblanco I corrected this symbol, it should be good now.
image

evanshultz:
- Add the package type at the end of the description (`DIP-8` in this case).
- Since you've picked the package, you need the suffix on the symbol name to indicate this and differentiate between other LT1019 symbols in other packages. So rename to `LT1019xN8`.
- I prefer a description of the datasheet name, then extra info, then package. KLC isn't too specific on this so there is some variability. Here, it would be `Precision Reference, Ultra Low Drift, DIP-8`
- TRIM should be Passive type and I suppose TEMP should be too.
- Don't include the package text in the symbol style since it doesn't show up. Snuggle the ref des at top left and value at top right (in general, they may move depending on the specific symbol).
- Your symbols are rectangles with pins ordered counter-clockwise like the package, and that is not what the KiCad library does. Take a look at the symbols in the library you're submitting to and match that symbol style, asking questions if you have them or the symbols look decrepit (there are some outdated libs). You will need to apply this to all of your PRs. You could take inspiration from this, perhaps adding TRIM on the right side and TEMP on the left):
image

evanshultz:
Please use commit messages that are specific. Titling them "Update Reference_Voltage.lib" won't help us figure out the purpose of a commit in the future if each individual commit is merged into the Git history.

In general, I will ask for changes. If I didn't mention something then the item shouldn't change or I just missed it. This affects two items here:
- You changed the pin names to just IN and OUT, which do not match the pinout drawings in the datasheet. The screenshot I showed was for reference, but saved typing and was more clear. Your part may be different, but page 1 shows these shorter pin names so it's fine.
- Retain the DNC pins. These are required for the footprint pin count filter in CvPcb to work.

That addresses your second bullet point. Here is the answer to the first one: See at https://github.com/KiCad/kicad-symbols/pull/1082#issuecomment-433513114 how the ref des and value are way above the symbol body? Ignore that the footprint text is visible in the symbol editor (it's not visible in the schematic) and place the ref des and value close to the symbol body. The screenshot you just posted is nice in this respect.

Did that answer all your questions?

And now one more thing: The DNC pin should have their terminus along the edge of the symbol body so it's not easy to connect them. Sorry that I forgot to mention where to place them above.

ferdymercury:
Thanks for the fast reply.

New version:
image

ferdymercury:
This would be a more compact version with the shortened pin names. Which one do you prefer?

image

evanshultz:
I like the smaller one. If you like it too go ahead and push a commit for it. Thanks!

ferdymercury:
Done, thanks!

evanshultz:
Thanks!

You could move the left side in 100mil. The TEMP pin text still won't bump into GND. Is that OK?

ferdymercury:
I would say it is too tight

image

evanshultz:
I'm OK with it but will let it go. The pin name offset could go down to 10mil to give a bit more space.

If you can push the ref des and value down by 50mil so it's closer to the symbol I will merge.

ferdymercury:
image
Ok!

evanshultz:
Thank you! We'll work on your other PRs next and I'm hopefully you will be able to keep contributing more in the future!

The last commit had a generic message but the previous few were detailed. Thanks for that.

ferdymercury:
Yes, I tend to forget about the message because it gets autofilled. Is it possible to deactivate the autofill?

PR #1083, opened 2018-10-22
Include AD8001ARZ amplifier

ferdymercury:
https://www.analog.com/media/en/technical-documentation/data-sheets/ad8001.pdf

------------


ferdymercury:
image

ferdymercury:
@evanshultz I found out that this symbol can be just an alias of OPA333xxD, so no need to do more work on it.

evanshultz:
Almost.

You the `xD` means SOIC package, and that's why this symbol had a default package which you've removed. That makes is harder to use this symbol.

So, you can add the AR suffix as an ALIAS to `OPA333xxD` (and restore the default package and footprints filter) but you will need make a new symbol for the AN suffix or find an existing symbol that already specifies a DIP package and makes logically (`MCP601-xP` might work).

The symbols are not split by package, but I'm working on that now so I understand comparators and opamps are a bit messy now. Sorry about that.

ferdymercury:
Changed as suggested, thanks.

evanshultz:
Thank you!

Did you verify that MCP601-xP is a valid symbol to use as an ALIAS? It looked OK to me but I'd like you to check in case I missed something and since I you must know the part you want to add. Thanks!

ferdymercury:
Yes, I verified it. They are identical.

image

image

evanshultz:
OK. Thank you!

PR #1084, opened 2018-10-22
Add AD8561ARZ comparator

ferdymercury:
https://www.analog.com/media/en/technical-documentation/data-sheets/ad8561.pdf

------------


ferdymercury:
image

evanshultz:
I see you've added screenshots. Thank you.

Rather than repeat myself many times, please see my comments at your other PR and do them for all of your PRs, such as using background fill for symbols. Also, you need to do some work on the footprint section of the symbol. And for opamps and comparators use a triangle symbol as is standard in our library.

ferdymercury:
I know, thanks. I will work on this next month.

ferdymercury:
I noticed that this symbol is already available as TL3116. So no more work needed, just adding as alias.

evanshultz:
OK. This basically works (I'll fix the packages later as I mentioned).

But there are two small things:
1. Remove the comma from the keywords.
1. Add the packages to the end of the description (`DIP-8/SOIC-8/TSSOP-8`).

ferdymercury:
Done, thanks.

evanshultz:
Looks good. Thank you!

PR #1088, opened 2018-10-22
Add SN74AHCT123ANE4

ferdymercury:
http://www.ti.com/product/SN74AHCT123A
Dual Retriggerable Monostable Multivibrator

Use with Package_DIP:DIP-16_W7.62mm_LongPads

------------


evanshultz:
@ferdymercury 
You've submitted a lot of parts recently and we haven't had a lot of time to review yet. Thanks for contributing! I'll try to get to these soon.

There are several requests in our GH PR template. One of them is to include a screenshot of the symbol. Would you please add that to all of your open PRs?

ferdymercury:
No hurries. I just finished two PCBs and decided it was a good time to suggest all my new designs to KiCAD, as they might be helpful for someone else. I am learning how to contribute, today I just corrected three footprints so that they match the Travis rules. I will do as you suggested, thanks for the patience.

ferdymercury:
image

evanshultz:
- You didn't submit the documentation for this symbol (description, keywords, and URL in the DCM file).
- Fill out the package you identified above as a footprint, but with normal pads.
- The footprint filter will then be `DIP*W7.62mm*`. We remove pin count and the trailing asterisk will easily allow big pads and other variations of the base footprint to be selectable in CvPcb.
- This is probably a good symbol to use two units.
- The symbol name can be just `SN74AHCT123A`. We don't include automotive PNs, high-temp PNs, etc. That means all remaining PNs have the same pinout.
- Power pins on top and bottom.
- Fill symbol with background color.
- The Q pins should be Output type.
- The REXT and REXT/CEXT pins should be Passive type.
- Use a leading tilde (`~`) instead of an asterisk to get an overbar.

You will find the guidelines to the library at http://kicad-pcb.org/libraries/klc/.

ferdymercury:
What do you think of this new version?

image

ferdymercury:
One problem is that the  FK PACKAGE has different pin counts (20 instead of 16)

jkriege2:
Shouldn't there really be three symbols? One for each of the separate monoflops and one for the power pins ... basically look at the existing 74xx symbols and model your symbol to match these.

Also we usually use a logical pin-ordering, not one that's oriented on the actual package.

JAN

ferdymercury:
image

ferdymercury:
Final version here. Let me know if I need to do more changes.

image

jkriege2:
Hi!

I had a closer look at the existing libraries: in 74xx There already exists a 74LS123 with three separate units and conforming to KLC. Therefore you don't need a new symbol for the AHCT-variant. Simply delete your symbol and add an ALIAS to 74LS123, which is calles 74AHCT123 and add your datasheet and description there. I think the naming should be the same as with the other 74xx in our lib, i.e. drop the SN.

Best,
JAN

ferdymercury:
Hi, good point, I have done do as suggested. Is it ok now?

Why are there several 74xx symbols in the Logic_LevelTranslator.lib library? Shouldnt be they moved to 74xx.lib ?

ferdymercury:
Thanks for the tips and your patience.

I have put now the NXP datasheet.

evanshultz:
You're welcome. Thank you!

PR #1089, opened 2018-10-22
FT245BM: fix footprint filter

jeanthom:
An attempt at fixing issue #971 : the footprint filter for the FT245BM does not match the specified footprint for the component.


evanshultz:
Thanks again!

Issue #972, opened 2018-09-18
MCU_Intel: S5.1 and S5.2 errors


antoniovazquezblanco:
KLC Rules:
http://kicad-pcb.org/libraries/klc/S5.1/
http://kicad-pcb.org/libraries/klc/S5.2/

Splitting issue #28 I have found the following violations:

```
Checking symbol '8086_Max_Mode':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'PDIP*W15.24mm*' does not match the footprint 'Package_DIP:DIP-40_W15.24mm' set for this symbol.
     - could not match 'DIP-40_W15.24mm' against filter 'PDIP*W15.24mm*'
     - could not match 'Package_DIP:DIP-40_W15.24mm' against filter 'PDIP*W15.24mm*'
    Symbol has a footprint defined in the footprint field, but several (2) footprint filters set. If the symbol is for a single default footprint, remove the surplus filters. If the symbol is meant for multiple different footprints, empty the footprint field.

Checking symbol '8086_Min_Mode':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'PDIP*W15.24mm*' does not match the footprint 'Package_DIP:DIP-40_W15.24mm' set for this symbol.
     - could not match 'DIP-40_W15.24mm' against filter 'PDIP*W15.24mm*'
     - could not match 'Package_DIP:DIP-40_W15.24mm' against filter 'PDIP*W15.24mm*'
    Symbol has a footprint defined in the footprint field, but several (2) footprint filters set. If the symbol is for a single default footprint, remove the surplus filters. If the symbol is meant for multiple different footprints, empty the footprint field.
checking ruleS5.2
checking ruleS5.1
Checking symbol '8087':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'PDIP*W15.24mm*' does not match the footprint 'Package_DIP:DIP-40_W15.24mm' set for this symbol.
     - could not match 'DIP-40_W15.24mm' against filter 'PDIP*W15.24mm*'
     - could not match 'Package_DIP:DIP-40_W15.24mm' against filter 'PDIP*W15.24mm*'
[...]

PR #1090, opened 2018-10-22
808x: Remove secondary footprint filters

jeanthom:
An attempt at fixing #972. The 8086/8087/8088 have two footprint filters :

 * `DIP*W15.24mm*`
 * `PDIP*W15.24mm*`

The second footprint filter is, if I understand well, useless because the footprint for a PDIP or DIP package is the same.


evanshultz:
Yep, that should do it. The other symbols could use some work with footprints but at least this error is now gone. Thanks!

Issue #973, opened 2018-09-18
MCU_Microchip_PIC18: S5.1 and S5.2 errors


antoniovazquezblanco:
KLC Rules:
http://kicad-pcb.org/libraries/klc/S5.1/
http://kicad-pcb.org/libraries/klc/S5.2/

Splitting issue #28 I have found the following violations:

```
Checking symbol 'PIC18F13K50-EP':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'PDIP*W7.62mm*' does not match the footprint 'Package_DIP:DIP-20_W7.62mm' set for this symbol.
     - could not match 'DIP-20_W7.62mm' against filter 'PDIP*W7.62mm*'
     - could not match 'Package_DIP:DIP-20_W7.62mm' against filter 'PDIP*W7.62mm*'
    Symbol has a footprint defined in the footprint field, but several (2) footprint filters set. If the symbol is for a single default footprint, remove the surplus filters. If the symbol is meant for multiple different footprints, empty the footprint field.

Checking symbol 'PIC18F2331-ISP':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'PDIP*W7.62mm*' does not match the footprint 'Package_DIP:DIP-28_W7.62mm' set for this symbol.
     - could not match 'DIP-28_W7.62mm' against filter 'PDIP*W7.62mm*'
     - could not match 'Package_DIP:DIP-28_W7.62mm' against filter 'PDIP*W7.62mm*'
    Symbol has a footprint defined in the footprint field, but several (2) footprint filters set. If the symbol is for a single default footprint, remove the surplus filters. If the symbol is meant for multiple different footprints, empty the footprint field.

Checking symbol 'PIC18F2450-ISP':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'SPDIP*W7.62mm*' does not match the footprint 'Package_DIP:DIP-28_W7.62mm' set for this symbol.
     - could not match 'DIP-28_W7.62mm' against filter 'SPDIP*W7.62mm*'
     - could not match 'Package_DIP:DIP-28_W7.62mm' against filter 'SPDIP*W7.62mm*'
    Footprint filter 'PDIP*W7.62mm*' does not match the footprint 'Package_DIP:DIP-28_W7.62mm' set for this symbol.
[...]

PR #1091, opened 2018-10-22
PIC18F: removing secondary footprints filters

jeanthom:
Hi again !

This PR is quite similar to #1090 (redundant footprint filters for special DIP package denominations), except for the PIC18F96J60-IPT where one of the footprint filters wasn't both matching the associated footprint for the component, but also erroneous (the complete part number indicates **only** one package).


evanshultz:
Thanks for fixing several issues today!

Almost. This one is weird because it's just wrong. The "PT" suffix is a 12x12mm TQFP with 0.4mm pitch. The footprint and filter are OK for that package.

But the "PF" package is a 14x14mm TQFP with 0.5mm pitch. So would you please copy `PIC18F96J60-IPT`  and then remove the three "PF" part number from one of them? For the second symbol leave only the "PF" part numbers and change the footprint and footprint filter to the bigger TQFP. This footprint already exists in the library.

Then this should be ready to merge.

evanshultz:
The referenced issue this will resolve is https://github.com/KiCad/kicad-symbols/issues/973.

jeanthom:
That commit should do the trick. I also updated 18F96J65 and 18F97J60 since they have the same issue as they are both aliases of 18F96J60.

antoniovazquezblanco:
Closing and reopening to re-trigger travis

evanshultz:
@jeanthom 
Why are there now so many added symbols? I was expecting just one new symbol with 2 ALIASes (and the existing symbol to have only two ALIAes as well). I believe the pinout is the same for all parts in the family so each PN doesn't need a unique symbol. What am I missing?

jeanthom:
IIRC I split it into multiple parts because of problems with part filters. I'll try to work on that issue next week.

evanshultz:
@jeanthom 
We only have PT and PF packages, and all the pinouts are the same, so only two new symbols are needed. I still don't see why my suggestion above doesn't work. Please re-review and use exact symbol names for clarity when you respond.

PR #1093, opened 2018-10-24
Add DRQFN packages for some ATmega MCUs

fauxpark:
Here are the Dual Row QFN symbols for the ATmega164A/PA, 324A/PA, 169A/PA, 169P, and xmega128B3.

164a

Datasheets:
http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8272-8-bit-AVR-microcontroller-ATmega164A_PA-324A_PA-644A_PA-1284_P_datasheet.pdf
http://ww1.microchip.com/downloads/en/DeviceDoc/atmel-8284-8-bit-avr-microcontroller-atmega169a_pa_329a_pa_3290a_pa_649a_p_6490a_p_datasheet.pdf
http://ww1.microchip.com/downloads/en/DeviceDoc/doc8018.pdf
http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8074-8-and-16-bit-AVR-Microcontroller-XMEGA-B-ATxmega64B3-ATxmega128B3_Datasheet.pdf

Footprint PR was https://github.com/KiCad/kicad-footprints/pull/990

------------


evanshultz:
Thanks for doing so much work to expand and clean up many parts of the libraries.

Man, there are a few issues in these datasheets that mess with my head. Tables with missing info and other sorts of stuff. I suppose it's all good, though.

- Please put the value on the top right.
- Where is the DRQFN EP connected mentioned? Page 3 of http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8272-8-bit-AVR-microcontroller-ATmega164A_PA-324A_PA-644A_PA-1284_P_datasheet.pdf mentions it for other packages but I don't see anything for DRQFN. (One may assume it's the same die in a different package and the substrate is also bonded to the thermal pad so the connection is the same, but you never know...)
- I assume you're trying to match the pin locations with the other packages for the same parts? I notice that `ATmega164A-MU` has pin PA0 at (600,-1700) but `ATmega164A-MCU` is (700,-1700). If you are trying to match pin locations, could you please take a closer look at these new symbols? If you're not trying to match pin locations, they can shrink a bit.
- Why aren't all the 169s (A, PA, P, PV) ALIASes? I can't find a difference that necessitates unique symbols but I'm probably missing something obvious.
- 128B3: Can the CAP* pins be spread apart a bit to fit a capacitor symbol between them?
- 128B3: The figure on page 7 indicates PDI_DATA should be either Output or Bidirectional type.
- 128B3: Can VLCD, BIAS1, and BIAS2 source current?

fauxpark:
Figure 28-1 in the 128B3 datasheet shows VLCD/BIAS1/2 to be part of the LCD power supply, so I suppose they must be able to source current, as far as I understand what that means.

evanshultz:
Most of these are fine. Thanks for explaining.

Why a 500mil body width? Would matching pin locations be better? A 450mil width would place the pins in the same place on a 100mil grid. I'm not opposed to having bigger symbols, but my first thought was that the symbol would be the same irrespective or package (excluding perhaps a thermal pad pin) so I wanted to know why you did what you did.

I'm fine changing the PDI_DATA pin type here, but others like a separate PR so let's do that. I don't know if it should be Output or Bidi but I'm sure you do.

I wanted to see a demo board for ATxmega128B3 using the LCD to know the pin connection, but I'm not seeing one. Surely there are several and I'm just missing them.

evanshultz:
Thanks for the info on the LCD pins. Hmmm, I reasoned through a similar pin at https://github.com/KiCad/kicad-symbols/pull/960#issuecomment-429988388. We have also used Passive type for many VREF pins on ADCs and DACs which are power pins that can be internally- or externally-sourced. Leaving them as Power Output is probably fine in the majority of circuits and won't cause too many false . What would you like to do?

I added the other items above to https://github.com/KiCad/kicad-symbols/issues/387, which can be done when pins can move.

I've asked Microchip about the pad connection of the DRQFN package so we can get a definitive answer. I think that and the pin type above are the only outstanding items for this PR, right?

fauxpark:
I'm happy to change them all to Passive in another PR if you think it's more fitting/less error-prone.

I would think the EP connection is device-specific, not package-specific. In any case, for all of these AVR devices I take the silence on the matter in the datasheets to mean that you *can* (not *must*) connect it to ground. Perhaps that might require it to be unstacked from the other GND pins to leave it up to the user.

fauxpark:
@evanshultz any response from Microchip yet?

evanshultz:
Yes, several responses but no answer. :(

evanshultz:
OK. Got an answer. The thermal pad is internally connected to ground but it is not required to connect it externally. It is only recommended to connect the thermal pad to a ground plane and that is more for mechanical stability than for thermal reasons. However, this appears to be the same guidance for the VQFN package where pins are stacked. Since they're joined internally, I'm OK with stacking as that seems to be standard practice for these MCUs and following Microchip's implementation recommendations.

fauxpark:
As I suspected. Thanks for that.

Anything else left to do here?

evanshultz:
Nope. Just wanted to make sure it was all good with you before merging. Thanks again!

PR #1095, opened 2018-10-26
Add DIN rail adapter

ferdymercury:
image

Matching footprint PR: https://github.com/KiCad/kicad-footprints/pull/1041

------------


ferdymercury:
Datasheet is https://www.phoenixcontact.com/online/portal/us?uri=pxc-oc-itemdetail:pid=1201578&library=usen&tab=1

myfreescalewebpage:
Hi @ferdymercury , thanks for contributing,

Librarians, is it really a "device" that is to be added in the lib ? If  going this was we will add many screws, nuts, etc etc and I'm not sure Kicad aimed to replace Solidworks !!!!

Cheers,
Joel

ferdymercury:
Well, I can remove the screws. But this has the same character than a "Mounting Hole". It is just an alias for three mounting holes. The point is that its footprint will tell you where not to put components based on the courtyard etc.

poeschlr:
I would allow this in the lib.

The question is if all din rail connectors come with 3 mounting holes. Maybe a better symbol would be to show a simplified part outline as viewed from the side instead of showing one from the top. (The side view might be a lot more generic)

myfreescalewebpage:
@poeschlr @ferdymercury I do not agree, it's not a mounting hole ! A mounting hole is important to add a hole in the PCB. In this PR we are taking about a piece of plastic on a DIN rail ! Do you really fix your PCB on it directly with no other enclosure ?! This kind of part is made to be fixed on the back of an enclosure to be mounted on a DIN rail....

I agree with @poeschlr that if the part need to be added to the library, a symbol representing the form of the clip is better.

> The question is if all din rail connectors come with 3 mounting holes.

No, some have two holes only, others are just some clips.

Cheers
Joel

ferdymercury:
I really fix my PCB directly to this adapter. Did it last month.

ferdymercury:
See for example https://eckstein-shop.de/Adafruit-DIN-Rail-RJ-45-To-Terminal-Block-Adapter-Vertical-Jack?curr=EUR&gclid=CjwKCAiA9efgBRAYEiwAUT-jtFRlM3fOtaPwg4uSzSYQx-1dPlB9dj0kyxGrYXxToUJik9T1SRVthhoCqnQQAvD_BwE

myfreescalewebpage:
Okay... never seen this in the industry but why not.

Can you rework the design of the symbol, for example something looking like the profile of the part ?

ferdymercury:
Reworked.
image

myfreescalewebpage:
@ferdymercury thanks.

Additional comment:
- [ ] You should swap reference and name, see MountingHole part for example.

Cheers,
Joel

ferdymercury:
Thanks Joel. Fixed.

myfreescalewebpage:
@ferdymercury thanks, no other comment to do.

If other librarians are okay, this is ready to be merged when the footprint will be accepted (https://github.com/KiCad/kicad-footprints/pull/1041).

Joel

myfreescalewebpage:
@ferdymercury I think this is ready to be merged. Thanks for the job !

Cheers,
Joel

PR #1057, opened 2018-10-10
Add ATmega324PB

fauxpark:
This one could *almost* have been an alias of the 324, except that it has a port E.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/40001908A.pdf

324pb

Also, I guess I forgot to double check the reordering of the 48PB after doing it manually...

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1099, opened 2018-10-28
Added LPC122x MCUs.

m5k8:
Added NXP LPC122x MCUs in LQFP48 and LQFP64 variants.

https://www.nxp.com/docs/en/data-sheet/LPC122X.pdf

screenshot_20181028_124323
screenshot_20181028_124342

Unfortunatelly KiCad symbol Library Editor reordered some other parts pins on save, so diff is not so clean.


m5k8:
OK, so there are some geometry errors to correct, I'll do that.
But I'm signing CLA and it still shows as not signed...

poeschlr:
You used a different mail address for creating your commits than you use in your github account

poeschlr:
In addition, please do not list alternative functions in the pin names. This results in unusably large symbols. So please only include the identifying name (PIDO_xx) plus maybe one function for pins that are almost always used with this function (reset, programming interface, ...)

m5k8:
> You used a different mail address for creating your commits than you use in your github account

I added this different address to github, but not as main. Should work, according to the instructions. Will fight this matter.

> In addition, please do not list alternative functions in the pin names. This results in unusably large symbols. So please only include the identifying name (PIDO_xx) plus maybe one function for pins that are almost always used with this function (reset, programming interface, ...)

For me and further viewers, full names act as documentation and allow for functional review without consulting the reference manual all the time. Just a glance and I know whether rx/tx pair is properly routed and not flipped, properly connected at all, etc, etc. Without this, schematic is much more cryptic and black box'ish.
Hence smaller size for pin names, just to make it not too wide. If it's unacceptable for public KiCad library, I'm afraid these symbols stay within my private library.

m5k8:
I've done some corrections to pass KLC checks. I'm not truncating pin names, full pin names are required for my expected documentation quality, clarity and usability. If it's not up to your expectations, I'm OK with keeping them private, no hard feelings.

evanshultz:
@m5k8 
First, thanks for contributing to KiCad. We really appreciate your submission and initiative to grow the official KiCad library.

As @poeschlr is the head librarian, that means we won't be able to accept these particular symbols. However, thanks for digging into KLC and learning your way around. I hope it will be helpful with your next submission in the near future! :)

Glad you got the CLA figured out.

For reference, here is what this symbol looks like on an A4 page:
image

m5k8:
OK, so I'm keeping the symbol with verbose names for my own use. No problem, and I even dare to think it's your loss :) I can't see the problem with symbol width, I prefer schematic to be as self-documenting as possible instead of as dense as possible.
To make my effort useful for others, I've just trimmed all names and commited the change, please review.

screenshot_20181029_195713
screenshot_20181029_195730

evanshultz:
Thanks! I tend to agree with you considering the current limitations in KiCad with regards to pin names. We have hopes that pin naming will improve in the future but that won't be for a while.

I'll take a look at these symbols in the next day unless somebody else gets to it first.

poeschlr:
I am aware that there will never be a consensus on this particular topic. For me the most important topic is to keep the library rules consistent and stable for the time being. (We had a period of rapid changes near the end of v4. A similar period will come again near the release of v6. So in a few years time.)

---

Having all this information in a small MCU might not be a problem. But try it with some of the larger high density STM MCUs. The symbols for these take up a full A4 page.
The problem is where do draw the line. For the sake of consistency we simply decided not to include any alternative function information in any symbol (With the exceptions listed above)

I also doubt that reading a schematic is easier if all functions are listed for all pins. You still have no clue which function is selected for this particular pin. For that you will need to have project specific symbols based on the official symbols. In these project symbols you would rename the pins such that the used function is used as the pin name. In addition i would even suggest to set the electrical type to something more descriptive for the selected function.

---

The hope is that the future file format will support alternative functions for pins. Ideally both an alternative for the pin name and the electrical type are supported. (would make ERC much more useful for such parts.)
However i fear this might not even make it into v6. My guess is that v6 will bring the new file format with feature parity to previous versions of kicad. I expect new features only in v7. (So at least 4 years unless kicad gets a massive boost in development resources)

m5k8:
> I also doubt that reading a schematic is easier if all functions are listed for all pins. You still have no clue which function is selected for this particular pin.

I think it's obvious more often than not. All comm interfaces, when wire nets are named correcly: RX/TX/MOSI/MISO/SCK/SCL/SDA/USB/ETH ADC inputs, buses, special signals - RESET/ISP/SWCLK/SWDIO, etc. But for the sake of library consistency I agree with you, let's have it one way in official library, hence my last commit with base names only. Is it OK?

evanshultz:
Regarding the pin naming, I appreciate you making shorter pin names. Hopefully there will be a better solution for all perspectives in the future, but this is where we are right now.

Review feedback:
- Please ref des on top left and value on top right.
- It appears this library currently has only specific parts, so could you either pick one part from each package or add them all (using the ALIAS feature)?
- The description can follow the others in the library, but some of them are missing the package at the end. So use `32-bit ARM Cortex-M0 microcontroller; 128 kB flash, 8 kB SRAM, LQFP-48` for `LPC1227FBD48-301`.
- Remove the leading asterisk on the footprint filter.
- The "L" is missing before "QFP" in the 48-pin footprint filter.
- Since we only need 2 characters for the pin numbers, all pins should be 100mil in length.
- Please change the VDD3V3 pin name to just VDD.
- I think I'd rather see the crystal, clock, ADC reference, and other non-GPIO pins on the top left. Then PIO0 can fill out the right side and the 64-pin can have other GPIO on the left side below the fixed-function pins. That matches all but one symbol in this lib (which is in the lower right as you have it). I'll merge either way since I suppose you would prefer to touch this as little as possible.
- Pin 40 of the 64-pin symbol is missing the GPIO function.
- For the GPIO with a secondary function shown on the left side, would it be better to arrange them `PIO<port>_<number>/<function>` so that the GPIO info is closer to the pin? You can see where we've done this for new symbols in another library at https://github.com/KiCad/kicad-symbols/pull/1057.

Note to self: `LPC11U37FBD48-401_` needs the underscore removed.

m5k8:
It's quite a drill for a newcomer! I get that, you want as strict conformity as possible and that's understandable. Let's see whether I'm fit for it :)

>     * It appears this library currently has only specific parts, so could you either pick one part from each package or add them all (using the ALIAS feature)?

I see, you like all variants listed explicitly. I'll conform, regardless of my opinion.
Should I leave the wildcard as main type, and add all specific MPNs as aliases, or should I use selected MPN as canonical one (but which one is the superior one?), and all others as aliases?

>     * I think I'd rather see the crystal, clock, ADC reference, and other non-GPIO pins on the top left.

It's more convenient to have them on the bottom. When you draw the schematic, there is resonator and 2 capacitors hanging down from these pins. When they are at the top, these parts often protrude into other pins area, unless we leave enough space between XTAL pins and pins that are further down. When XTAL pins are at the bottom, we have as much space as we want.
Are you sure you want it reordered? I can do it, but please acknowledge.

evanshultz:
Just a single variant is fine. No need to make them all if you don't want to. In some cases I see some wildcarding, but in this library each symbol is specific.

We usually start with the "lowest" (letter or number) one and make others ALIASes. "Increasing" will be any increment in the part number read left to right.

I see on the crystal pins. That's fine then. Thanks for pointing that out.

m5k8:
I applied all (I think...) suggested corrections.
screenshot_20181107_182822
screenshot_20181107_182814

evanshultz:
Great! Thank you very much!

I appreciate you taking the time to make these updates for the entire KiCad community.

PR #1100, opened 2018-10-29
Added TC1017.

m5k8:
Added Microchip TC1017 LDO Linear Regulator in 3 variants: SOT23, SC70 and SC70 reversed pinout.

http://ww1.microchip.com/downloads/en/DeviceDoc/21813F.pdf

screenshot_20181029_192756
screenshot_20181029_192914
screenshot_20181029_193121


evanshultz:
- Please see our existing linear reg symbols and take design design ideas from them. This includes ref des and value placement, pin length (use 100mil), and putting GND at the bottom of the symbol.
- You need to add NC pins to the symbol and place their terminus at the edge of the symbol body.
- You do not need a trailing tilde on the SHDN pins.
- The footprint filter on the SC70 parts is a bit simpler if it's `SOT*353`.
- Add the package type (`SC-70` or `SOT-23-5`) to the end of the descriptions.
- Since you haven't added symbols for each individual MPN, let's simplify the symbol names to `TC1017-x[C|L]TTR`.

m5k8:
OK, bear with me, I'm new to this, thanks for your patience. Please review the changes.

screenshot_20181029_204436
screenshot_20181029_204445
screenshot_20181029_204454

evanshultz:
Thanks! It's a lot to figure out. The guidelines for the library are at http://kicad-pcb.org/libraries/klc/ and there are plenty of other items which aren't captured there because it's too specific or KLC hasn't been updated.

What is the reason for the symbol naming? Some symbols in the regulator library have the package at the end but that doesn't apply here. Was there an error in my proposed names above?

Other than symbol names, it looks great.

evanshultz:
Oh, and please remove the "reverse" from the description. Having the "R" in the symbol name should suffice.

m5k8:
>     * Since you haven't added symbols for each individual MPN, let's simplify the symbol names to `TC1017-x[C|L]TTR`.

Enlisting every possible MPN is counterproductive, from my point of view. It's maybe helpful when there are only few possibilities, but too often they go into tens, or more. The listing of library contents becomes unreadable. Maybe it's reassuring for freshmen to find exact MPN (hey! It's my part!), but professionals don't need this and are able to supply complete and proper MPN, based on the part type with wildcard characters.

> What is the reason for the symbol naming?

The wildcard template conforms to your own KLC rule S2.2:

`Where an MPN has options for such non-functional variations, these portions of the MPN should be replaced with a wildcard character (x).`

So I'm doing exactly this: TC1017-xCTx_SOT23
Here, first 'x' is to be substituted with voltage and temperature range (I proposed more x's as a replacement for every option, but that maybe was excessive), and second 'x' is packaging option. Your suggestion to put constant TR here is based on the doc, but that's as of today - spec says TR is tape&reel. Maybe tomorrow manufacturer will add another packaging option, and suddenly we have assumed too much. Packaging is non-functional variation, hence 'x'. CT/VT letters in the middle are functional variation - they select package type, and even pinout, so these are different parts.
I feel like applying S2.2 literally. I attached SOT23/SC70 at the end as a hint pointing to the package. For one, letters in MPN are not obvious what they mean (looking at library listing, when one has to select proper device), and second - I've seen this method is used in the library.

Maybe I've misunderstood your proposal, so how these parts should be named to match KiCad style?

Thanks again for constructive discussion! Few more cases and I'll be out of kindergarten :)

m5k8:
Thanks for your input, much appreciated.
So, how many separate symbols should we have here? I can see two ways:

First one, as my proposition:
TC1017-xCT (SOT23)
TC1017R-xLT (SC70)
TC1017-xLT (SC70)
this way, we have unequivocal selection of three distinct variants and every symbol has distinct footprint assigned for PCB.

Or, do I understand correctly, that you propose to make one universal part number? As first two parts share the same pinout, maybe it's possible to group them, but third variant has different pinout, so we can't unify it. It leaves us with:
TC1017-xCT/TC1017R-xLT (SOT23/SC70, but identical pinout)
TC1017-xLT (SC70) (different pinout)
and then make the first name something like TC1017[R]-[C|L]T - but this way we can't assign unique PCB footprint for this symbol, and it even suggests TC1017R-xCT possibility, which is invalid.

How should we proceed?

evanshultz:
I should have used different syntax to indicate optional and required fields. I think we're on the same page. :) The symbols will be:
- TC1017-xCT
- TC1017R-xLT
- TC1017-xLT

m5k8:
OK, another - let's hope the last one :) - commit with shorter MPNs.

m5k8:
> Oh, and please remove the "reverse" from the description. Having the "R" in the symbol name should suffice.

I don't think R belongs in package name, there's no such thing as SC-70-5R. Being "reverse" here is about pin assignment, not package type. It's still plain SC-70-5.

But, if that's the convention, I'm OK with it. SC-70-5R ? Please acknowledge.

evanshultz:
I didn't mention the package name (you're talking about the package in the description). I asked you to remove the word "reverse" from the description because the _symbol name_ includes the "R" which is part of the MPN and thus the symbol name. Just remove the one word "reverse" from the description.

evanshultz:
Looks great. Thank you!

PR #1101, opened 2018-10-29
ICE40HX8K-BG121: split out GNDPLL0/1

marcan:
These aren't supposed to be connected together, they should only connect
to a decoupling network. Pair them up with VCCPLL0/1 for neater
schematics.


evanshultz:
@marcan 
Thanks for looking at the library and taking the initiative to reach out.

Can you do a few things to help get this going?
1. Provide a screenshot of your changes. Also add one of the before symbol if that's helpful.
1. Provide the documentation that drives this change.
1. Sign the CLA, because without it your PRs won't be merged.

@awygle 
You committed the original symbol and are more familiar with this IC than me. Can you please take a look?

marcan:
The [documentation](https://www.latticesemi.com/-/media/LatticeSemi/Documents/ApplicationNotes/IK/iCE40sysCLOCKPLLDesignandUsageGuide.ashx?document_id=47778) says:
> The sysCLOCK PLL has the DC ground connection made on the FPGA, so the external PLL ground connection (GNDPLL) must NOT be connected to the board’s ground. Figure 12 also includes sample values for the components that make up the PLL power supply filter.
This is slightly ambiguous because the doc talks about a single GNDPLL/VCCPLL pair but this FPGA has two, but I think the intent is that they should be independent supplies (they are on opposite sides of the chip), and the official [breakout board](https://www.latticesemi.com/-/media/LatticeSemi/Documents/UserManuals/EI/EB85.ashx?document_id=50373) definitely has them with separate decoupling networks and no cross-connection of GNDPLL.

The new symbol looks like this:
image
The old one just had the VCCPLL pins at the top, and a single overlapped GNDPLL pin at the bottom, with nothing on the right side. I moved them to the right side and spaced them out so that you can fit capacitors directly between the pins, like this:

image

marcan:
Fixed the grid issue.

awygle:
@marcan is right, this was an oversight on my part.

I'm absolutely astonished that this didn't fail Travis for having VCC pins not on top. I like marcan's design a lot better, but I think it violates KLC.

I also think GNDPLL should be an input not an output.

marcan:
I wasn't really sure of the precise semantics of "input" vs "output" for power. My thinking was that VCCPLL gets fed externally, but GNDPLL is tied internally.

marcan:
Changed the GNDPLL pins to power input.

evanshultz:
Thanks!

Just one more thing: Will you please restore the ref des and value positions? They can auto-place when put on the schematic, but if they're not their placement is only good for the power unit (unit E). They're on top of the symbol for all other units.

If you can do this with an incremental push instead of overwriting each Git commit it will be a bit easier to check.

Thanks!

marcan:
Done.

evanshultz:
Excellent. Thanks so much for catching and fixing this issue!

PR #1102, opened 2018-10-30
Remove duplicate ATtiny1634 definition

fauxpark:
The ATtiny1634-MN and -MU are, for the purposes of the libraries, completely identical. The N/U suffix indicates the lead finish on the package itself. U (matte tin) is the most common here so I have chosen that one as the one to keep. Obviously they should be removed from all of the symbol names, but of course that kind of mass renaming must wait until later in the development cycle.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8303-8-bit-AVR-Microcontroller-tinyAVR-ATtiny1634_Datasheet.pdf


evanshultz:
-MN is extended temp range which should not be in the library according to http://kicad-pcb.org/libraries/klc/S2.2/, anyway.

A question: Why only this part? We both know not to make big changes now, so why not fix all instances of this issue later? This symbol has been in the lib since https://github.com/KiCad/kicad-library/blob/master/library/atmel.lib, so it could be used in many designs. I'm sure you have a good reason, so I don't mean to sound offensive, I'm just wondering why this part and why now. Thanks!

fauxpark:
> Why only this part? We both know not to make big changes now, so why not fix all instances of this issue later?

Because this is the only instance of this issue. I don't really see what the big deal is if it's used in many designs as the -MU symbol is identical except for a single letter in the name. I doubt it is that painful to resolve.

In my list of missing ATmega/ATtiny devices I also have an entry for renaming ATtiny13A-MMH to -MMU to match the rest of the library, but there is probably not much point when the U is just going to get removed later anyway.

evanshultz:
@poeschlr 
I'm fine to remove this one symbol as a compatible one is present, but does this need to wait until after 5.0.2 is released or is it OK to do now? If I don't hear from you in a few days I'll merge.

evanshultz:
I'll merge this in about 24 hours if there are no replies.

PR #1105, opened 2018-10-30
Add Analog Devices ADuM764x digital isolators

henryhallam:
1 kV RMS 6-channel kV digital isolators based on "iCoupler" in-package
magnetic coupling

ADuM7640A / ADuM7640C - all channels same direction
ADuM7641A / ADuM7641C - 5 channels A to B, 1 channel B to A
ADuM7642A / ADuM7642C - 4 channels A to B, 2 channels B to A
ADuM7643A / ADuM7643C - 3 channels A to B, 3 channels B to A

"A" grades are 1 Mbps / 25 ns max pulse width distortion
"C" grades are 25 Mbps / 6 ns max pulse width distortion

[Manufacturer datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/ADuM7640_7641_7642_7643.pdf)

Existing library symbol ADuM7643 was split into separate ADuM7643A and
ADuM7643C aliases since the speed grades are an important functional
distinction (ADuM7643 had description fields matching the "C" grade). - **Is this acceptable or not due to breaking backward compatibility? If not, what is the recommended resolution?**
The other devices in the family were not present in the library.

adum7643c

------------

- [X] If there are matching footprint PRs, provide link(s) as appropriate - **N/A**


myfreescalewebpage:
Hi @henryhallam, thanks for contributing,

- [ ] "6Ch" in the keywords should be "6-channels"
- [ ] "QSOP20" at the end of the description should be `QSOP-20`for consistency with other devices
- [ ] "VDD2B" should be Power Input on all symbols, not Input only.
- [ ] GND1 pins should be stacked
- [ ] GND2 pins should be stacked
- [ ] Inputs and outputs are called "VI" and "VO", not "IN" and "OUT", you should name the pins according to the datasheet.
- [ ] You have removed ADuM7643 and it's a problem. Moreover, it seems that the difference between A and C versions is only the speed, right ? I'm not sure it's good to create two devices in this case.

Cheers,
Joel

myfreescalewebpage:
Hi @henryhallam thanks for the modifications

- [ ] ADuM7641C, QSOP-20 and 6Ch not fixed
- [ ] Reference should be in the top left corner. Example:

capture

> The way I see it, the differences between the A and C versions are important functionally - significantly different current consumption, propagation speed, skew etc. These differences are reflected in the description and keywords fields. That said, they are pin-compatible, so I suppose it comes down to how one defines "functional". KLC S2.2 is a bit ambiguous on this. The differences between these speed grades seem more significant to me than the KLC S2.2 examples of operating temperature range and packaging.

I understand your position. Need a more experienced reviewer for this point. @poeschlr @evanshultz @DanSGiesbrecht for e xample if you can have a look to the name of the device according to my comment above just written again below:

> You have removed ADuM7643 and it's a problem. Moreover, it seems that the difference between A and C versions is only the speed, right ? I'm not sure it's good to create two devices in this case.

Cheers,
Joel

PS : no force push, just push to your branch to make an additional commit, thanks :)

myfreescalewebpage:
Hi @henryhallam are you still alive on this subject ? Thanks.

myfreescalewebpage:

evanshultz:
While I understand that the part functionality is different, the symbol is not affected. I believe that is the intent of KLC but I agree it's not worded clearly.

However, I would accept either way so that contributing to the library is easier when the guidelines aren't clear. Additionally, the library is migrating towards more specificity over time.

myfreescalewebpage:
New pull request for this: https://github.com/KiCad/kicad-symbols/pull/2880. Closing this PR.

PR #1108, opened 2018-10-31
Add STC15W201S series

fvollmer:
I added the STC15W201S series MCUs (www.stcmicro.com/datasheet/STC15F2K60S2-en.pdf).
The datasheet mentions the SOP16 footprint, but the dimensions seem to match the kicad SOIC16 footprint.

stc

------------


myfreescalewebpage:
Hi @fvollmer , thanks for contributing,

- [ ] Description should ends with the package name, `, SOP-16`
- [ ] Comma is not required in the keywords, to be removed
- [ ] We usually put reference in the top left corner and name in the top right corner. Actually the name override the VDD pin which is not very convenient.
- [ ] You can use capital letters for GND and VCC
- [ ] You can move port P1 100mil up, so it is in aligned with port p3.
- [ ] You can cross P5.4 and P5.5 so it's in the same order than the other GPIO which is probably better to not make some mistakes later.
- [ ] I not agree with the footprint, even if it looks like the SOIC-16, it should be SOP package, and it has a base metal indicated page 156 of the datasheet. Is it an expose pad ?

Cheers,
Joel

fvollmer:
Thank you for your review! I already implemented most of the requested changes. Except everything regarding the footprint. I just would like to make sure that we have the correct one :)

Regarding the footprint: 
- Why do you think that it is a SOP16 package? I think the description page 156 is pretty much a perfect match with the JEDEC SOIC specification [here](https://www.jedec.org/standards-documents/docs/ms-012f). I couldn't find a SOP specification that matched the footprint from the datasheet. 
- I don't think that this part has an exposed pad. I found a nice image [here](http://www.inkocean.in/stc-microcontroller-stc15w201s-35i-sop16).

ht9170d

myfreescalewebpage:
Thanks for the modifications.

> Why do you think that it is a SOP16 package?

Because your part names finishes with "SOP16". Also SOP package is described page 156 of the datasheet. I agree they looks like the SOIC-16. Same dimensions too.

Need to ask a more experienced librarian about the decision to make. @poeschlr @evanshultz maybe ? We wonder about the difference between SOP-16 and SOIC-16 packages. Seems the SOP-16 version of this part looks like exactly a SOIC-16 IC...

Joel

fvollmer:
> Because your part names finishes with "SOP16". Also SOP package is described page 156 of the datasheet. 

I see your point. This would require a new SOP footprint, even through the SOIC footprint seems to fit. Let's wait for the opinions of the other librarians.

soic

myfreescalewebpage:
> Let's wait for the opinions of the other librarians.

Yes :)

fvollmer:
Any opinions? (ping @poeschlr @evanshultz)

To summarize problem:
 * According to the [datasheet ](http://www.stcmicro.com/datasheet/STC15F2K60S2-en.pdf) (page 61 and 62) this part uses a **SOP16** package 
 * But the package dimensions at page 156 seem to match the kicad **SOIC**-16 package (and also the [JEDEC **SOIC** specification](https://www.jedec.org/standards-documents/docs/ms-012f)). This means the kicad **SOP**-16 wouldn't be a good fit.

The question is: how should we handle this package mismatch? I could see two ways:
 * Use the SOIC-16 package
 * Create a special "STC-SOP-16" package

myfreescalewebpage:
@fvollmer the footprint question has just been answered by @evanshultz and confirmed by @poeschlr on an other subject here: https://github.com/KiCad/kicad-footprints/issues/1450#issuecomment-468728773.

You should create the exact footprint.

myfreescalewebpage:
@fvollmer any news about the footprint ? If you have created a new pull request in the footprint repository, can you indicate a link here ? Thanks, Joel

fvollmer:
I created the footprint

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1488
Footprint Generator PR: https://github.com/pointhi/kicad-footprint-generator/pull/314

fvollmer:
Travis fails, because the new footprint doesn't exist (yet). I guess this is an expected error. I should go away as soon as the footprint is merged.

myfreescalewebpage:
Yes this is an expected error when the footprint is not merged yet, no worries for that. Will wait the footprint integration before terminating this pull request.

fvollmer:
@myfreescalewebpage FYI: I think this PR should now be ready. The Footprint was merged and travis seems to be happy.

evanshultz:
I didn't catch this earlier. Sorry. The "QFN" issue linked was because the footprint was the same but the 3D model was not; using unique footprints allowed an accurate 3D model to be assigned.

If the footprint and package (3D model) used by this "SOP" part is identical to our SOIC footprint I've got no problem using an existing footprint. We do this in other cases if the manufacturer choose a name that doesn't fit our convention.

fvollmer:
@evanshultz So you want to revert https://github.com/KiCad/kicad-footprints/commit/113ceb4a66eb2a3b5912e63d56a668fce392c48f?
Actually there is a small difference for `lead_len` and `lead_width`. Another advantage of this new footprint is, that there is no confusion regarding SOP16 vs SOIC16.

evanshultz:
I didn't check anything before, just used the word "if" as a general rule.

The dimensions roughly fall within the JEDEC spec MS-012. A bit different, but it should be fine. However, height is different so I say do whichever way you like. But I personally lean toward commonalization where possible if there's no reason not to.

myfreescalewebpage:
@fvollmer one required modification has not been made on this symbol:
- [ ] Description should ends with the package name, `, SOP-16`
Joel

fvollmer:
@myfreescalewebpage I forgot about the description. Thanks for checking!
Should I go with `SOP16` (datasheet) `SOP-16`, or `STC_SOP-16`?

myfreescalewebpage:
The format adopted in the Kicad library for this is SOP-16.
Thanks,
Joel

fvollmer:
Done

myfreescalewebpage:
@fvollmer thanks for the last fix, no more comment, merging.
Joel

fvollmer:
Thanks for the review!

PR #1112, opened 2018-11-01
add ADG901/902 and ADG918/919 RF switches

lorem-ipsum:
[datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/ADG901_902.pdf)
902 is the reflective variant, 901 is absorbative. Added the package type as the LFCSP package has an EP. For the LFCSP foorprint there is already a [PR](https://github.com/KiCad/kicad-footprints/pull/1061). The second variant of the ADG901 is identically to the one displayed except the missing EPAD pin.

adg901bcpz
adg902brmz


lorem-ipsum:
Also added ADG918 and ADG919 which are very similar, but are SPDT instead of SPST. Both have the same package variants (with and without EPAD). Posting the variant with EPAD, again the others are identically, but missing that pin. 
[datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/ADG918_919.pdf)

adg918bcpz
adg919bcpz

myfreescalewebpage:

myfreescalewebpage:

Great job, a few comments I have during my review:

For all devices:
- [ ] "Analog Devices" not useful in the keywords (we never put the manufacturer), to be removed

ADG901BCPZ / ADG918BCPZ / ADG919BCPZ:
- [ ] We do not put EP size in the footprint filter, should be `LFCSP*1EP*3x3mm*P0.5mm*`
- [ ] The footprint is not correct, should be `Package_CSP:LFCSP-8-1EP_3x3mm_P0.5mm_EP1.45x1.74mm` (https://github.com/KiCad/kicad-footprints/pull/1061). See also travis error.
- [ ] The datasheets say "THE EXPOSED PAD MUST BE TIED TO THE SUBSTRATE, GND." so pin 9 is to be renamed GND and stacked with other GND pins

Cheers,
Joel

myfreescalewebpage:
Hi @lorem-ipsum 

Thanks, perfect for me, no more comment to do, merging.

Joel

PR #1117, opened 2018-11-03
Added D_TVS_Uni.

m5k8:
Unidirectional TVS diode.

screenshot_20181103_123652
screenshot_20181103_123726


evanshultz:
The KiCad library uses the IEC symbol convention, and you can find IEEE 315-1975 online if you search.

The diode falls into section 8.5.6.1 of the aforementioned document, I believe, and KiCad has chosen a single "tip" on the cathode mark (shown as Style 1).

Removing the second "tip" makes these look just like the existing D_Zener symbols. I was surprised this diode isn't in the library already, but perhaps using the zener symbol was the expected user action before (that's what I've already done). I suppose for ease of finding the symbols they could be duplicated, though, but I'd like another librarian to weigh in.

One thought rather than merge as-is: there already exists a D_TVS symbol but it's a bidirectional TVS. Since both uni- and bi-directional are common, if we add both symbols I'd expect the symbol name to state this. So I'd propose renaming the existing one to D_TVS_Bidirectional and spelling out "Unidirectional" on this one for clarity.

myfreescalewebpage:
Maybe I'm wrong, but following IEEE 315-1975, not sure our current zener diode symbol is correct ? Look like more to the TVS proposition of @m5k8 ....

Anyway, I have already seen the two symbol for the zener diode, as shown in examples below:
https://fr.wikipedia.org/wiki/Fichier:Zener_diode_symbol.svg
https://www.astuces-pratiques.fr/imagesarticles/24/1_alimentation-resistive-sans-transfo.png
https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcSq4RH2DYw9AirMEFBKJruBCR7aM-ztyRnNF8DLZJqIe-aSrFXL

But I agree that the TVS is usually represented as shown in this PR.

Joel

evanshultz:
Honestly, I think it was Jan who started doing the hollow diodes with a line through it and I don't know what reference document he used. All diodes I see in IEEE 315-1975 are filled so I don't know any of them apply.

It appears BS 3939 and IEC/EN 60617 hold the same symbols and may be what Jan used. Anyway, I only see a single symbol for a unidirectional breakdown diode and it only has a single tip.

I'm used to seeing the TVA symbol above but we need to fit the convention of all worldwide KiCad users and also look for consistency within the library. Taking that into account, the symbol shown above should change. This cathode mark is used for bidirectional breakdown diodes in the BS/IEC/EN standards I mentioned above.

Honestly, since all unidirectional breakdown diodes use a common symbol I would be happy with just a single one in our library. Also, there isn't anything special about the packages of a normal silicon diode, a zener, a TVS, etc. that would drive unique symbols because of the footprints. But... the zener symbol is already more specific than just "unidirectional breakdown" so I'm not against merging this to be more helpful and searchable for the users. Though my comments about the descriptions above still apply.

myfreescalewebpage:
@m5k8 are you still working on this ? Can you address comment of @evanshultz at https://github.com/KiCad/kicad-symbols/pull/1117#issuecomment-435895780 please ?

PR #1119, opened 2018-11-03
Fix wrong type of pin 3 of 7408.

bullestock:
Pin 3 was incorrectly configured as input - this patch correctly sets it as output.

(Only Unit A was affected, the other three units were correct).

See e.g. https://www.futurlec.com/74/IC7408.shtml


evanshultz:
Thanks for catching this mistake and posting a fix. I see this is your first contribution and hopefully we will see more from you in the future!

On a side note, my the "logic" libraries are in awful shape...

PR #1121, opened 2018-11-04
STM32F103 Blue Pill development board

jeanthom:
bluepill

No official documentation, but there is a bit of documentation here : https://wiki.stm32duino.com/index.php?title=Blue_Pill

------------


jeanthom:
I did find some manufacturer info here : http://www.vcc-gnd.com/cp/stm32hxb/stm32f1xxhexinban/

diegoherranz:
Thanks for submitting this symbol.

What's the idea behind this symbol? Is it creating a) add-on boards for the blue pill or b) using the blue pill as a component on a bigger board? Just out of curiosity because I guess that the symbol is the same either way.

Have you got any plans of creating footprint, if b), or a template, if a), for it?

On a very quick look, I've got a couple of comments:
- Some power pins could be stacked. See http://kicad-pcb.org/libraries/klc/S4.3/
- I would expect the NRST pin on the top left

Thanks!

jeanthom:
Hi !

It is more about using the Blue Pill as a component on some schematics, but I'll try to create a footprint sometime soon.

I'm not sure about S4.3 : in order to stack the power pins, should I set the visible one to **Power input** and the hidden ones to **Passive** ?

antoniovazquezblanco:
> I'm not sure about S4.3 : in order to stack the power pins, should I set the visible one to Power input and the hidden ones to Passive ?

Yes

jeanthom:
bluepill

myfreescalewebpage:

- [ ] Have you got a btter documentation link ? At least something in English...
- [ ] Pin offset should be 20mil (see travis output)
- [ ] Pins PC13 to PC15 are number 22-23-24, not 42-43-44
- [ ] Name of pin 37 should be `~RESET`
- [ ] Type of pin 18 (5V) should be specified Power Input or Power Output (I don't know..)
- [ ] You should define a footprint (called "BluePill" in "Module" library and push it to kicad-footprints repo), and add footprint filter (`BluePill*`)

Cheers,
Joel

myfreescalewebpage:
@jeanthom ping! Are you still alive on this subject ?

diegoherranz:
>  Have you got a btter documentation link ? At least something in English...

What if we use https://wiki.stm32duino.com/index.php?title=Blue_Pill as the datasheet link?
I think it's a good collection of information about it (schematics, pinouts, explanations...), maybe the best single like that can be provided?

jeanthom:
Hi @myfreescalewebpage, thanks for your comments.

Regarding pin 18, it can act both as a power input and a power output, depending on the way your are powering the development board (through the USB port or using pin 18). That's why it was `Unspecified`.

About the documentation, @diegoherranz's link is probably your best source of information if you don't speak chinese, but it is not official.

I'll try to make a commit with the requested changes soon.

diegoherranz:
> Regarding pin 18, it can act both as a power input and a power output, depending on the way your are powering the development board (through the USB port or using pin 18). That's why it was Unspecified.
Maybe `Passive`?

myfreescalewebpage:
Thanks you both !
OK for the link, I do not found better too.
Pin 18 should be Power Output, previous comment updated. It is the output of one regulator, don't think it's really good to power the board from this pin....
Joel

diegoherranz:
> Pin 18 should be Power Output, previous comment updated. It is the output of one regulator, don't think it's really good to power the board from this pin....

I think I may be getting confused with the pin numbering, but the 5V pin is not not the output of a power regulator. It is the input of the 5V->3V3 reg and also connected to USB_VBUS (see the [schematic](https://wiki.stm32duino.com/images/c/c1/Vcc-gnd.com-STM32F103C8-schematic.pdf)). It can be used for powering this board without any problem, as long as you don't provide power through USB at the same time :)

The thing is that pin 18 on that schematic is both VCC3V3 and PB14 (2 headers), but not 5V as in the symbol submitted. 

How has the numbering being assigned when comparing to the schematic linked?

Thanks!

jeanthom:
I did my own numbering as the Blue Pill official schematic doesn't have unicity in its pin numbering.

800px-stm32_blue_pill_top

myfreescalewebpage:
Ouppps....
I was previously thinking to pin 18 of the symbol proposed in the PR (5V), should be Power Input and pin 18 of the header P3 on the schematic of the module, 3V3, Power Output.

Now about the correct numbering to use, I suggest you can keep your assignment, but in any case you have to propose the footprint for this module (see also my review in a previous comment), the orientation of the module should be clearly identified on the footprint and of course it should be consistent with the symbol.

ikarakaya:
This page has better ezplanations about board module;
https://stm32-base.org/boards/STM32F103C8T6-Blue-Pill

myfreescalewebpage:
@jeanthom any news of the footprint ? Thanks

myfreescalewebpage:

PR #1125, opened 2018-11-07
Added the 18 Pin SOIC version of the MCP2515

johnfwhitmore:

Currently the kicad-symbol lib contains the 20 Pin version of the MCP2515 component. I've copied that symbol and renumbered the pins, (Removing the two NC pins) to add the 18 pin version of the device.

------------

Datasheet = http://ww1.microchip.com/downloads/en/DeviceDoc/21801e.pdf

screenshot from 2018-11-07 13-28-34


evanshultz:
Thanks for your continuing contributions.

- Temp should not be included in the PNs, so please just make just one symbol with a `xSO` suffix and remove the part about temp from the description. Sadly, this rule was not followed for other symbols so you saw a poor reference. Sorry about that.
- OSC1 should use the Line style without the little chevron. Again, sorry for a poor reference.

johnfwhitmore:
:( Sorry that my continuing contributions still aren't better first time out. But as you say I was simply doing a copy and paste from a questionable source ;)
screenshot from 2018-11-09 15-56-27
screenshot from 2018-11-09 15-56-30

evanshultz:
No need for a sad face! You and I both realize this library contained cruft.

Thanks for making these changes and fixing up the other bad symbol too!

- One more thing needs to be done, and it was my mistake to forget this with my first comment: change the footprint filter to `SOIC*7.5x11.6mm*P1.27mm*`.
- And, if you don't mind, push the NC pins of the TSSOP-20 in so the pin connection point is on the symbol body edge. This will make accidentally connecting to the pins much harder.

johnfwhitmore:
I'm sure you might not think it's a good idea ;) but I explained how I contribute to the libraries on IRC the other day. People seem to think it's very difficult, which has not been my experience. There's always bits you miss but that puts the pressure on ye maintainers. Was half thinking of doing a youtube video on the subject, but Chris Gemmel should be doing that one.

johnfwhitmore:
screenshot from 2018-11-09 16-19-52

evanshultz:
Looks great. Thanks for all your work here and elsewhere to help KiCad!

johnfwhitmore:
Thanks a million for merging, very happy to help out as much as I can

PR #1129, opened 2018-11-09
Added TPS54233 regulator

kajusK:
Added TPS54233 regulator, rule S4.2 fails as the pins are arranged for easy schematic wiring.
2018-11-09-125158_1920x1080_scrot

Datasheet: http://www.ti.com/lit/ds/symlink/tps54233.pdf


SchrodingersGat:
Fantastic, thanks

PR #1131, opened 2018-11-10
Add Fiducial as schematic symbol

ferdymercury:
Analogously to the Mounting Hole, that is also a symbol in the schematics and not only a footprint.

image

------------

### DIFF
--- a/Mechanical.dcm
+++ b/Mechanical.dcm
@@ -1,7 +1,7 @@
 EESchema-DOCLIB  Version 2.0
 #
 $CMP Fiducial
-D Fiducial Marker without connection
+D Fiducial Marker
 K fiducial marker
 F ~
 $ENDCMP
--- a/Mechanical.dcm
+++ b/Mechanical.dcm
@@ -1,8 +1,15 @@
 EESchema-DOCLIB  Version 2.0
 #
+$CMP Fiducial
+D Fiducial Marker without connection
+K fiducial marker
+F ~
+$ENDCMP
+#
 $CMP Heatsink
 D Heatsink
 K thermal heat temperature
+F ~
 $ENDCMP
 #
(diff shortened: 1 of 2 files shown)


DanSGiesbrecht:
@evanshultz @antoniovazquezblanco @Misca1234 what do you think of this suggestion? I like the idea of having a dedicated symbol. Currently, I have been using an unconnected testpoint with the fiducial selected as the footprint in my designs.

antoniovazquezblanco:
I have been using fiducial footprints without sym up until now but I would like to include them.

evanshultz:
I personally prefer to do it on the PCB and avoid a symbol, but I wouldn't object to its inclusion either.

- What does IEEE 315-1975 (or other standard reference) say about symbol style and ref des?
- Is `Mechanical.lib` the right library for this?
- Pin name offset (even if it isn't relevant) should be 20mil.
- AOI can't work if the fiducial has exposed (and maybe even unexposed) copper connected to it, so I'm not sure why a fiducial symbol would ever have a net connection. Unless there is a reason, remove "without connection" from the description.

thomasonw:
Just noting prior pull request:  #653 Myself, I still like having Fids on the schematic, akin to mounting holes, package, etc.

ferdymercury:
Thanks for your comments. I did some commits.
- I have matched now the IEEE convention designator
- Pin offset fixed
- I think mechanical.lib is fine, because it helps for alignment
- fixed the description.

diegoherranz:
+1 for having fiducials as a schematic symbol.
I also think that mechanical.lib is the right place.

Thanks

SchrodingersGat:
Thanks @ferdymercury this looks like a useful addition to the libraries. Absent any better location I think `Mechanical.lib` is fine

PR #1134, opened 2018-11-10
Add RAYEX ELECTRONICS L90 Series 30A Relay

Vlad-23:
General purpose SPDT/SPST 20A/30A (40A) PCB Relay from RAYEX ELEC. L90 Series (also compatible with TE Connectivity T90, Omron G8P and many other). PR for the footprint will follow.

Matching footprints: KiCad/kicad-footprints#1084

Datasheet: https://a3.sofastcdn.com/attachment/7jioKBjnRiiSrjrjknRiwS77gwbf3zmp/L90-SERIES.pdf

Screenshots: (Updated: 19.04)
RAYEX-L90
RAYEX-L90S
RAYEX-L90A
RAYEX-L90AS
RAYEX-L90B
RAYEX-L90BS
------------


poeschlr:
Will take a close look as soon as i have time. (As part of the renewed footprint review)

poeschlr:
A few of your symbols have pins not on the grid. Could you ensure they are for such a huge symbol. Furthermore would it be possible to ensure these symbols can be exchanged with each other where sensible. 
Meaning at least keep A1 and A2 at the same position for all of them. If possible also the ones for the contacts. (Example if two symbols have the pin number 12 in them then ensure that this one pin is always on the same position.)

Vlad-23:
I've changed everything a little, please review (see updated screeenshots on top). Not sure about L90B and L90BS, i've made an outline shorter to look more aesthetically but it's not centered around the origin now.

evanshultz:
So here's a question: why not use pin names instead of numbers? A1 and A2 for the coil (if you like), NO and NC, and C (or something else for common).

Then there isn't any confusion over pin numbers because there are logical names and it makes the symbols more generic and applicable to more footprints because we aren't using pin names from one specific datasheet. This takes advantage of the philosophy of KiCad where symbols and footprints are separate.

We have discussed this before and come to a conclusion that this works at https://github.com/KiCad/kicad-symbols/pull/592. For new symbols all should be well. I'll update pin names on the footprints if you agree.

Vlad-23:
You mean in fields "Pin number" enter names? Does it correct to use that fields made for numbers that way? I actually don't mind that but isn't it look like a kind of a 'dirty' hack? Also I'm a noobie to KiCad so maybe you can explain that philosophy of KiCad think? I mean I like that conception of reusable symbols for different footprints, but is seems not working. For example, there are bunch of relay symbols that look much the same but I need to create one more now. As I understand this concept, for example with relays, there should be general types of relays like SPST, SPDT, DPDT, with one ore two common pins and etc., not symbols with model names and current rating in descriptions. Also data sheet fields of symbols should be empty if symbol is generic for different manufacturers and models. 
Maybe you can discuss this with other experienced librarians and community. I could help clean-up and rearrange relay libs this way with your guidance.

Vlad-23:
Like this? 
Screen Shot 2019-04-01 at 9 02 37 PM

Vlad-23:
I think it may be good to use some kind of naming for relays like A and B for coil and NO NC COM. For two pairs of COM it could be COMA COMB. In case of few poles it could be NO1 NC1 COM1 (or COMA1 COMB1) and NO2 NC2 COM2.

Vlad-23:
I understand the concept and a fact that lib can contain specific symbols. I just want to say that for relay lib in our case we haven't any generic symbols at all, only specific. And a lot of that "specific" symbols are actually not so specific because the only difference is pin numbering scheme. 
From my observations, we have at least 12 SPDT relays, two SPST NC and 8 SPST NO that can be replaced with three generic relays. 

I also don't undrestand when we should create aliases for symbols and when not.

poeschlr:
Aliases can only be used if the only difference between two symbols is in the description or datasheet. It must have the same fooprint, pin numbers, graphics, ...

---

The relay lib could indeed go either way. We can think about a complex naming scheme, standardize pad numbers for footprints and use generic symbols.
Or we go the specific way as we do now and have one specialized symbol/footprint pair for them. This would be something for version 6 as i guess the new file format will give us more options here.

poeschlr:
Oh and even: Why invent a new naming scheme for symbol pins if there is a perfectly good standard out there? (These symbol pins are named after my suggestion in accordance with industry standards. Admittedly european standards but hey that is where i am from ;))

The only problem is that these have a second pin connected to the common pin for some of them. This is something you will not really find in many other parts and also not in the standard (why i suggested suffix a.)

evanshultz:
Ah. This relay pin naming is unfamiliar to me. I really didn't know where it can from so I was lost above.

IMO using pin names here is far more clear than these numbers, especially if someone isn't familiar with this nomenclature, as `NO` (for example) is going to be found in every datasheet even if the pin numbers do not adhere to the scheme you brought up above.

It appears EN 50005 is the standard you're talking about? I didn't find it online in a quick search to educate myself so does anyone know where it can be found? https://en.wikipedia.org/wiki/DIN_72552 is an alternate standard for relay pins.

Perhaps if you like to use EN 50005 you could write this up in KLC and from here on out use this in the library? At least then we have a specified methodology for generic or unnamed pins in relays.

Vlad-23:
I don't know much about this standards. I just proposed numbering scheme that looks logical to me. I also tried to find something and also didn't found. (EN 50005 by the way seems to be canceled and not relevant anymore.) I don't have much expirience using KiCad, i'm switching from Eagle. I decided to add some components and thought why not share. But in this case it's up to you and other librarians / expirienced comunity mebebrs to decide with strategy to use. I can cleanup and rearenge this ralay librarary under your guidance. I think it's a good idea to "rename" pins in my RAYEX footprints with easily understandable A1, A2, NO, NC, COM, etc. as I wrote earlier (I actually can't imagine that anybody, who know what relay is, could not undststand this pin naming). And i can create some generic relay symbols that will work with that footprints instead of adding another "specific" symbols that just duplicate existing one more time with different name and numbering. After that i can edit other relay footprints to ensure compatibility with that generic symbols if you and oters will feel this a good idea.

Can you share some links to find out about upcoming v6 release improvements/changes?

poeschlr:
@evanshultz the probosed naming scheme of COM,NC,NO breaks down if there is more than one contact and especially if the part has different contact configurations (3 power contacts, 1 NO, 1NC and one NO+NC)

There is a reason this scheme is developed in the way it is. (I never cared what it is called as all relays one can buy why i am from simply use this scheme on their housings. And it was the only standard i ever learned. But well i am from an european country.)

By the way the graphic style of these relays does not follow the standard either. But i would allow for now as it keeps these relays similar to the ones we already have. (We can switch to full standard compliance with v6)

---

The din standard you list is not for relays them selves. It is an automotive standard where the function of the contact is what gives the number (So a NO contactor can have different numbers depending on what it is intended to switch.)

---

Oh and 50005 is superseded by DIN EN 60947-1. Contact naming did not chang https://www.voltimum.de/sites/www.voltimum.de/files/fields/attachment_file/de/others/H/200411059746DINVDE0660-115.pdf (german source)

evanshultz:
Ignoring the above, what do you want to do? You said "there's a perfectly good standard" so you would like this PR modified to fit? Can you write it up into KLC since this is unfamiliar to me, and perhaps others, so we can all have something canon to use?

poeschlr:
I thought the original pin numbering was already what i suggested on the footprint side (namely the one from the standard)

But i am not sure which of the many screenshots represents the current symbols. (And don't have kicad access right now.)

evanshultz:
Here are the current symbols:
image

Does this match what you wanted?

Vlad-23:
So what's now guys? Maybe somebody can merge if no issues?

poeschlr:
I seem to remember that i requested the manufacturer to be included in the symbol names.

Vlad-23:
@poeschlr I can't remember such a request from you, but anyway, it's done now. Please check. (Updated screenshots on the top.)

Vlad-23:
@poeschlr no response from you within a week now. Are there any issues?

poeschlr:
I have a personal live. I can not be here all time.

Vlad-23:
Hello everybody. Two months passed, still nobody to review and merge?

evanshultz:
@Vlad-23 
We are generally short on manpower and time. @poeschlr assigned this case to himself, but he's out right now so it will take at least another week before you'll get a response from him. Sorry. Continuing to ping at a reasonable rate is the right way to do it when you're looking for a response.

But here are a few notes from me:
1. Let's use the official URL for the datasheet. It's https://5rrorwxhkjiliij.ldycdn.com/attachment/7jioKBjnRiiSrjrjknRiwS77gwbf3zmp/L90-SERIES-min.pdf. that's a horribly ugly URL, but it comes from https://www.relay-rayex.com/L90-SERIES-pd6792755.html.
1. All descriptions show `30A`, but that's not correct for all parts. Some even have different current ratings for different contacts or contact arrangements.
1. I'm confused about the pin 13. https://www.finder-relais.net/en/Finder-general-technical-information-en.pdf#page=2 doesn't mention a `.3` pin, so I would have thought the common pin would always be `11` for all symbols. The footprints were merged with pin 13, though, as Rene requested at https://github.com/KiCad/kicad-footprints/pull/1084#issuecomment-462118860. However, the link above is the one he just sent recently about relay pin naming. I'd much rather use pin naming or something commonly understood, rather than "like some spec except for this one thing". I'm not familiar with this naming convention so perhaps it's best to get his input before proceeding.
1. Since the A and B common pins are discrete terminals, shouldn't they have unique pins? If they're internally connected the user should be able to choose which pin to use, or both. Right now I see the non-S symbols have two stacked common pins with different names which is not good and means those get the same net connection which forces both common terminals to be connected on the board.

poeschlr:
@evanshultz the use of the .3 pin comes from the standard itself. (For pure NO contacts it uses .3 and .4) finder simply does not mention it as they do not show such a contact arrangement. I do not quite understand why the standard is defined that way.

Vlad-23:
Hi, @poeschlr. 
I don't understand what's next? Is it ready to merge or you want me to make changes?

poeschlr:
Looks good, thanks and sorry it took so long (It was a mixture of me kind of being burned out and confusion between us maintainers.)

Vlad-23:
Thanks everybody involved, it's been a pleasure to participate and create something useful. Hope to contribute more soon.

PR #1140, opened 2018-11-13
Converter_ACDC: Add isolation lines

antoniovazquezblanco:
Fixes https://github.com/KiCad/kicad-symbols/issues/429

Please, contribute with ideas for Vigortronix converters:
In the datasheet looks like FG is not connected (http://www.vigortronix.com/15WattSMPSPCBModuleAC-DC) but it is included. Can anybody tell me what is this used for? Should I delete that pin? I would avoid moving that pin if possible.

screenshot from 2018-11-13 17-02-22


evanshultz:
- It looks like you only added or removed lines to place a single dashed line along the isolation barrier. Is that it? Did I miss any other changes?
- What about putting the ref des on top? It will look more common (I think we almost always put it on top) if the user doesn't turn on auto-placement of text.
- I asked Vigortronix about the FG pin and will comment here if/when I get a response.

antoniovazquezblanco:
> It looks like you only added or removed lines to place a single dashed line along the isolation barrier. Is that it? Did I miss any other changes?

Yes. Only a style fix :)

> What about putting the ref des on top? It will look more common (I think we almost always put it on top) if the user doesn't turn on auto-placement of text.

Will do.

evanshultz:
Alright. Got the answer from Vigortronix. On this PN, the FG pin does nothing (no internal connection) and is only for structural support, so it can be made into an NC pin. (On some other PNs it's used as a zero crossing detector.)

antoniovazquezblanco:
Done!

evanshultz:
Merged!

PR #848, opened 2018-08-25
ATtiny: Add ATtiny series: x12 x14 x16 x17

dokthar:
ATtiny212 / ATtiny412: http://ww1.microchip.com/downloads/en/DeviceDoc/40001931A.pdf
ATtiny212 SOIC-8
Oops didn't saw this other PR: https://github.com/KiCad/kicad-symbols/pull/798 

---

ATtiny214 / ATtiny414 / ATtiny814: http://ww1.microchip.com/downloads/en/DeviceDoc/40001932A.pdf
ATtiny1614: http://ww1.microchip.com/downloads/en/DeviceDoc/ATtiny1614-data-sheet-40001995A.pdf
ATtiny214 SOIC-14

---

ATtiny416 / ATtiny816: http://ww1.microchip.com/downloads/en/DeviceDoc/40001933A.pdf
ATtiny1616: http://ww1.microchip.com/downloads/en/DeviceDoc/ATtiny3216_ATtiny1616-data-sheet-40001997B.pdf
Two packages SOIC-20 and QFN-20, not pin compatible.
ATtiny416 SOIC-20

---

ATtiny417 / ATtiny817: http://ww1.microchip.com/downloads/en/DeviceDoc/40001934A.pdf
ATtiny1617 / ATtiny3217: http://ww1.microchip.com/downloads/en/DeviceDoc/ATtiny3217_1617-Data-Sheet-40001999B.pdf
ATtiny417 VQFN-24

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1142, opened 2018-11-15
Add ATtiny406/806/1606 and 416/816/1616/3216

fauxpark:
Previously added in #848 (not merged), this PR also includes the x06 series, so those symbols can be made aliases instead.

Datasheets:
406: http://ww1.microchip.com/downloads/en/DeviceDoc/Microchip%208bit%20mcu%20AVR%20ATtiny406%20data%20sheet%2040001976A.pdf
806/1606: http://ww1.microchip.com/downloads/en/DeviceDoc/ATtiny806_1606_Data_Sheet_40002029A.pdf

416/816: http://ww1.microchip.com/downloads/en/DeviceDoc/40001913A.pdf
1616/3216: http://ww1.microchip.com/downloads/en/DeviceDoc/ATtiny3216_ATtiny1616-data-sheet-40001997B.pdf

------------


fauxpark:
It seems there is a missing VQFN footprint. [Here](http://ww1.microchip.com/downloads/en/DeviceDoc/20%20Lead%20VQFN%203x3x0_9mm_1_7EP%20U2B%20C04-21496a.pdf) is the Microchip packaging spec, it should be easy to add it to the script.

evanshultz:
Of course the PR I just merged broke this one. Can you please fix the branch conflict?

These symbols have power pins named VCC but the datasheet shows VDD. Maybe it's not a big issue now, but is this something to add to #387 and address across all symbols at some point?

When you submit multiple symbols, would you mind to include screenshots of all symbols? That makes it faster and easier to review.

evanshultz:
Oh, and should the description say "VQFN" instead of "QFN"? I understand https://github.com/KiCad/kicad-symbols/pull/1043#issuecomment-428743384 to mean the actual work of any footprint manipulation will come later, but should the descriptions be done now?

fauxpark:
I was expecting the conflict :P I've been juggling these symbols around through all of these PRs.

The power pin naming is annoyingly inconsistent with the majority of these AVRs. I wasn't sure what to do with that so I just stuck with VCC for now. I think it's still fairly obvious what is meant, but if you'd prefer to use the exact naming from the datasheet I can change it.

I didn't think it was necessary to show the SOIC package symbol since it's essentially the same, but I realise you might need to quickly check the pin numbering etc. since it is different. I'll remember that next time. Here it is just in case:

406s

As for the package description, it can be done now, but I prefer to do mass changes like that in one go to avoid fragmentation of work. In this case the footprint is not yet in the libs so I will change it to VQFN.

evanshultz:
This is all good.

On Monday I'll make the footprint if it hasn't been done already and add the two items to #387 if they aren't there. Thanks!

evanshultz:
If the footprint filter gets a leading "V" this should be green and can be merged.

fauxpark:
Whoops, sorry.

Issue #18, opened 2017-12-05
Memory symbols require extensive rework


SchrodingersGat:
- [ ] A number of old / outdated / useless symbols
- [ ] Many lacking good description
- [ ] Many lacking footprint information
- [ ] A lot of KLC issues to fix

bobc:
I will work on this, seems like a good candidate for symgen.

Memory_EEPROM has some MC68HC11 parts which should go into a separate library, eg. MCU_NXP_MC68

bobc:
Can we adopt a standard convention for memory sizes? Standalone memory is usually quoted in bits, sometimes as X x Y e.g. 128k x 8. However, MCU memory is usually quoted in bytes. Additionally, we have a good mix of kb, kB, KB, and also ko (kilo octets). All this makes it pretty ambiguous when looking at a description "128kB".

In some cases, we might also need to distinguish between decimal kilo (k=1000) and binary "k" (1024). Clock speeds and data rates are always quoted with decimal prefix, so 1 kHz means 1000 Hz.

Fortunately there is a standard, we don't need to invent our own! Namely  ISO/IEC IEC 80000-13:2008, described at https://en.wikipedia.org/wiki/Binary_prefix#kibi.

The prefix "k" is strictly reserved for decimal kilo (1000). For binary usages, the prefixes Ki, Mi, etc are used. Additionally, "B" is strictly reserved to mean "byte" (8 bits), and "bit" is used for bits. 

Therefore:
128 kB means 128,000 bytes
128 KiB means 131,072 bytes
128 Kibit means 131,072 bits

The standard is simple to adopt and completely non-ambiguous.

poeschlr:
The 1024 byte is not defined as k (kilo) but as Ki (kibi). And yes JDEC still uses uppercase K for 1024.
https://en.wikipedia.org/wiki/Kibibyte

bobc:
> The 1024 byte is not defined as k (kilo) but as Ki (kibi).

Yes, you understand correctly. :)

> JDEC still uses uppercase K for 1024.

Perhaps, but 

`The specification cites three prefixes as follows:

    kilo (K): A multiplier equal to 1024 (210).
    mega (M): A multiplier equal to 1,048,576 (220 or K2, where K = 1024).
    giga (G): A multiplier equal to 1,073,741,824 (230 or K3, where K = 1024).

[...]

PR #1144, opened 2018-11-16
Move CH376 from Memory_Controller to to Interface

antoniovazquezblanco:
See https://github.com/KiCad/kicad-symbols/issues/18


evanshultz:
Sorry about this, but since it's being touched I wanted to make sure this symbol was of appropriate quality. And no good deed goes unpunished. So would you mind making a few changes:
- Do not duplicate the description as keywords.
- Pin 16 should be Tri-state type.
- Pin 4 should be Bidirectional type.
- Pin 5 should be Input type.
- Pin 7 should be Passive type.
- Remove trailing tildes from pins 1 and 3.

antoniovazquezblanco:
Thank you! Fixed!

evanshultz:
And thank you!

PR #1145, opened 2018-11-16
Add Artesyn ATA Series SMT DC to DC converter symbols.

stambaughw:
Please consider this pull request of the [Artesyn ATA series DC to DC converter](https://www.artesyn.com/power/assets/ata_series_ds_01apr2015_79c25814fd.pdf) symbols.  The [pull request](https://github.com/KiCad/kicad-footprints/pull/1090) for the footprint has already been submitted.

artesyn-ata00aa18s-l
artesyn-ata00f18s-l

------------


evanshultz:
- These should just be named with the module PN, and not the vendor name (I'm going by this lib and the ACDC lib, and I did miss this on the XP Power modules you submitted earlier).
- Add "Artesyn" to the description so we know the vendor.
- "+" and "-" go in front of the pin name.
- Please place -VOUT for the single output where COM is for the dual output.
- ATA00H18S-L should be 9-36V input.
- ATA00CC36S-L should be 18-75V input.
- ATA00AA18S-L and ATA00F18S-L have the wrong datasheet.
- ATA00F18S-L is 2W, not 3W.
- Update the footprint and footprint filter to match the footprint name.

stambaughw:
I think I covered all of these fixes except the XP Power naming issue.
I will submit a separate pull request for that.

evanshultz:
- ATA00CC36S-Lhas the wrong output voltage and is a dual output.
- ATA00F36S-L should have stayed at 2W and ATA00F18S-L should change to 2W.
- The single output -VOUT pin has not been moved to where the dual output COM pin is located.
- ATA00AA18S-L and ATA00F18S-L are now missing a datasheet, and ATA00CC36S-L now has the wrong datasheet.
- Let's add `SMD*` to the end of the footprint filter since these PNs are only SMD packaged.

stambaughw:
I left the _SMD* off of the footprint filter because these symbols can
be reused with the THT versions of the same part numbers since the pin
outs are the same.  I didn't need the THT parts for my project so I just
implemented the SMT footprint and 3D model.  I don't have the time right
now to work on the THT footprint and 3D model at the moment.  I will add
_SMD* to the footprint filter if that is your preference.

evanshultz:
All PNs added end in "S", which means SMD package. And they have the SMD footprint assigned as default. So to allow either footprint the PNs need to be generalized and the default footprint filter removed. Since we don't have THT footprints, and the PNs were already specific to the SMD versions, I requested the footprint filter also match only the SMD footprint.

The SMD footprint is now merged and just the items above (which cover the CI failure) are outstanding.

stambaughw:
Not a problem.  We can always go back later and add the THT 3D model,
footprint, and symbol aliases and filter later.  I just pushed the fixes
so hopefully this will be good to go.

evanshultz:
Not quite all fixed yet. All are mentioned above or caught by CI.
- ATA00AA18S-L and ATA00F18S-L are missing datasheets.
- ATA00CC36S-L is a dual output so the keywords are wrong.
- The NC pins were moved to an off-grid location.

stambaughw:
There is definitely something wrong with the symbol editor in the dev
branch.  I know I fixed all of these issues.  I just pushed the fix
which I edited the dcm file by hand so all should be well.

evanshultz:
Fortunately I'm sure you can track it down! :)

Thanks for this submission and sorry it's been such a traumatic saga.

PR #1148, opened 2018-11-17
Comparator: Add LM397 and TL331

henryhallam:
Add TI's [LM397](http://www.ti.com/lit/ds/symlink/lm397.pdf) single voltage comparator and its pin-compatible cousin [TL331](http://www.ti.com/lit/ds/symlink/tl331.pdf)

lm397

These couldn't be aliases of existing SOT-23-5 comparators since the pinout doesn't match any that are in the library already.

------------

- [X] If there are matching footprint PRs, provide link(s) as appropriate - **n/a**


evanshultz:
Thanks for this nice contribution!

Please remove the duplicate "open drain" and/or "open collector" keywords since those are found in the description and then I will merge!

henryhallam:
Updated - I removed the keywords altogether since all relevant words were in the description. Let me know if that wasn't correct.
Thanks.

evanshultz:
Some keywords are required. Just "single comparator" or something simple is acceptable.

antoniovazquezblanco:
@henryhallam ping! :)

evanshultz:
@henryhallam 
Can you please fix the branch conflicts?

Also, we ask that you use incremental commits instead of force pushes (sounds like it should be from Star Wars!) so that's it's easier to keep track of what happens in a PR between starting and merging it.

evanshultz:
I fixed the branch conflict and apparently I felt things were OK in the past. Nothing jumped out at me just now, so I'll merge.

PR #1149, opened 2018-11-19
Remove vender name prefix from XP Power DC to DC converter symbols.

stambaughw:
This pull request was a fix for something that was missed in a prior submission.  The XP Power vendor name suffix has been removed from all of the DC to DC convert symbols and alias where it was previously used and "XP Power" was added to the descriptions in the .dcm file.  No other changes to the symbol were made so there is no need to submit a screenshot and footprint pull request as they are not applicable.

------------


evanshultz:
Thanks!

It's rather annoying that GH can't figure this out and show a nice diff...

stambaughw:
I think that is due to the fact the kicad sorts by name so all of the xp
power symbols were at the end of the file and the sorting made a mess of
things.  I could have changed it by hand but the next time someone added
a new symbol to the dc to dc converter library using the symbol editor,
there would have been a big delta which would have been confusing.

PR #1152, opened 2018-11-20
Add OPA847 as alias of OPA890

ferdymercury:
http://www.ti.com/lit/ds/symlink/opa847.pdf

------------


ferdymercury:
Add SO-8 package version
image

evanshultz:
Only two things:
1. Footprint should be `Package_SO:SOIC-8_3.9x4.9mm_P1.27mm`.
2. Footprint filter should then be `SOIC*3.9x4.9mm*P1.27mm*`.

Please take a look at footprints in the future, including other symbols in this library.

ferdymercury:
Fixed, thanks for the hint.

evanshultz:
Certainly! Looks good now. Thank you!

Issue #850, opened 2018-08-26
4000 series CMOS 4082 schematic symbol is incorrect


Agent24:
In library 4xxx_IEEE, the Dual 4-input AND gate, 4082 has an incorrect pinout.
For Unit A, what should be Pin 5 is a duplicate Pin 12 from Unit B. Unit B itself is correct.

antoniovazquezblanco:
Thank you very much @Agent24 !

Will you be willing to open a pull request in order to fix this?

antoniovazquezblanco:
PR submited at https://github.com/KiCad/kicad-symbols/issues/850

PR #1159, opened 2018-11-22
4xxx_IEEE: Fix 4082 pin (solves #850)

antoniovazquezblanco:
Fixes https://github.com/KiCad/kicad-symbols/issues/850


evanshultz:
(For all...) What is this all about? At the referenced issue here is the problem report:
"In library 4xxx_IEEE, the Dual 4-input AND gate, 4082 has an incorrect pinout.
For Unit A, what should be Pin 5 is a duplicate Pin 12 from Unit B. Unit B itself is correct."

Here is the old symbol:
image

And a datasheet https://assets.nexperia.com/documents/data-sheet/HEF4082B.pdf.

The pinout is fixed in this PR, but pins 5 and 12 are not Input type. Can you correct that too?

antoniovazquezblanco:
Done! Thanks!

evanshultz:
Thank you!

Issue #1040, opened 2018-10-06
KLC: Stacking of power pins


FabianInostroza:
A friend made a schematic and added an ATmega16A, this part has 3 VDD pins stacked and he added only one bypass capacitor. 
Would it be better to avoid stacking of power pins when the number of them is not high (not FPGAs nor large MCUs)?, this may prevent errors like this and also make the library consistent since other MCU libraries doesn't use stacking on power pins (LPC and STM32 at least).

PD: I'm aware that nothing beats reading the datasheet or that the _error_ would be detected in the layout phase.

evanshultz:
Sorry to hear about this and thank you for taking the time to post this message.

Our policy is to not stack power input (where the part draws power) pins if the datasheet calls out decoupling capacitors for each pin or gives other explicit reasons to keep the pins separate. For ground pins or other pins which do not have explicit instructions about any per-pin requirements, we stack pins.

Because the library is joined with the KiCad releases in an attempt to avoid breaking existing schematics unnecessarily, and also because some symbols are simply old, there are sure to be many symbols which violate the current guideline I mentioned above. Or the guidelines above may explain why some MCUs have stacked power pins and other don't.

Looking at the part datasheet (http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-8154-8-bit-AVR-ATmega16A_Datasheet.pdf), I just scrolled through it front-to-back and searched for a few key terms and I don't see anything about requiring a bypass cap for each VCC pin. Can you point me to the section of the datasheet or other documentation showing this requirement?

Since you used the word error in italic it makes me think this is a clear requirement from Atmel or just that you feel very strongly that not having a cap on each VCC pin is unacceptable. If there isn't any documentation requiring one cap for each VCC pin, then I assume that's your or your friend's "best practice" and so it's just what you do even if Atmel doesn't require it? In that case, it's hard to make an argument because a single cap might be perfectly fine in some or all applications, or perhaps two, or perhaps 3 are really required but not written down from Atmel. I don't honestly know, but if Atmel doesn't tell us what to do with VCC caps it falls into preference and applications-specific needs.

This has been discussed several times and there are valid reason for stacking and not stacking. We do expect the situation to be much better with an update to the symbol file format in a future KiCad release, but that doesn't change things now.

Thank you for your input since it does help guide how the official KiCad library is built. We are all volunteers trying to develop and maintain a powerful and useful library to support all KiCad users, like you and your friend.

[...]

PR #1160, opened 2018-11-23
Add 7705

lmoelleb:
Datasheet: http://www.ti.com/lit/ds/symlink/lm7705.pdf

image

Capacitor pins set as input (I used 7660 as an example), not sure if passive would be a better fit. The datasheet refers to them as "analog".

------------


lmoelleb:
Would it make sense to stack the GND pins? I did not do it as the KLC contains these two statements:

> Power and ground pins should be set to either Power Input or Power Output

> Special Case: Pins of electrical type [Output, Power Output, Power Input] are special cases. Connecting these pins together causes ERC (electrical rule check) errors and must be avoided.

But then I randomly browsed through issues and found this:
https://github.com/KiCad/kicad-symbols/issues/1040
containing the statement from @evanshultz (emphasis mine):

> Our policy is to not stack power input (where the part draws power) pins if the datasheet calls out decoupling capacitors for each pin or gives other explicit reasons to keep the pins separate. _For ground pins or other pins which do not have explicit instructions about any per-pin requirements, we stack pins._

DanSGiesbrecht:
GND pins should be stacked. The special case that you mention is illustrating that only **one** of the stacked group can be the correct type. That means that one of your two GND pins will be **visible, power input**, and the other will be **invisible, passive**. Having them both as **power inputs** would cause an ERC error in the stack. I usually go by lowest pin number as the dominant (visible) pin in a stack, unless there's an overriding reason (such as a pin being listed as the primary power pin).

Regulator_SwitchedCapacitor.dcm
```diff
@@ -24,6 +24,12 @@ K Low-noise inverting charge pump with both positive and negative LDO's
 F http://www.ti.com/lit/ds/symlink/lm27762.pdf
 $ENDCMP
 #
+$CMP LM7705
+D Switched capacitor inverter, inverts 3V to 5.25V to regulated negative 0.23V, VSSOP-8
```
DanSGiesbrecht: Please let me know what you think of this description: `Negative bias generator, low-noise, 3-5.25V input, -0.232V fixed output, VSSOP-8`

Regulator_SwitchedCapacitor.dcm
```diff
@@ -24,6 +24,12 @@ K Low-noise inverting charge pump with both positive and negative LDO's
...
+K switched capacitor voltage converter inverter negative bias regulator
```
DanSGiesbrecht: Along with the description suggestion above, here are keyword suggestions that are not present in the description: `switched cap capacitor converter inverter regulator supply`.

Regulator_SwitchedCapacitor.lib
```diff
@@ -74,6 +74,29 @@ X C- 9 -400 -600 100 R 50 50 1 1 P
...
+DRAW
+S -300 300 300 -300 0 1 10 f
+X CF+ 1 -400 0 100 R 50 50 1 1 I
+X GND 2 0 -400 100 U 50 50 1 1 W
+X SD 3 -400 200 100 R 50 50 1 1 I
+X VDD 4 0 400 100 D 50 50 1 1 W
+X GND 5 0 -400 100 U 50 50 1 1 P N
+X VOUT 6 400 200 100 L 50 50 1 1 w
+X CRES 7 400 0 100 L 50 50 1 1 I
+X CF- 8 -400 -100 100 R 50 50 1 1 I
```
DanSGiesbrecht: I think that you made a good suggestion: I think that `CRES`, `CF+` and `CF-` should be passive.

PR #1162, opened 2018-11-23
New component SN74LVC2T45DCUR

SkyEngineer:
This is Dual-Bit Dual-Supply Bus Transceiver With Configurable Voltage Translation in VSSOP-8

image

Footprint: Package_SO:VSSOP-8_2.4x2.1mm_P0.5mm from official footprint library.


SkyEngineer:
not correct

SkyEngineer:
not correct

PR #1169, opened 2018-11-25
add lt1963eq (alias lt1963aeq)

aewallin:
LT1963EQ and LT1963AEQ adjustable 1.5A voltage regulators in in TO-263-5 package.

Datasheets
https://www.analog.com/media/en/technical-documentation/data-sheets/1963fc.pdf
https://www.analog.com/media/en/technical-documentation/data-sheets/1963aff.pdf

Screenshot (symbol based on existing LT1962)
lt1963

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1171, opened 2018-11-25
add LT3015Q

aewallin:
LT3015Q adjustable negative low noise voltage regulator in TO-263-5 package.

datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/3015fb.pdf

Screenshot:
lt3015q

------------


myfreescalewebpage:

Looks great, maybe you can enlarge a little bit the symbol vertically between pins SHDN/VIN and ADJ/OUT. @antoniovazquezblanco @DanSGiesbrecht what do you think about that ? Else, ready to be merged.

Cheers,
Joel

aewallin:
We almost always use LT3015 (this one) together with LT1963 (PR #1169) - IMO best to make the symbols similar.
If the symbol shape is agreed on I can add these v-regulators in fixed-voltage output format and other footprints also.

myfreescalewebpage:
@aewallin I agree we should treat both PR at the same time, I have the same remark for PR #1169.

Just adding below an example (LT3010) of what I mean by enlarging a little bit the symbol. This is to get a better consistency between symbols.

reg

myfreescalewebpage:
@aewallin can you fix the footprint filter to `TO?263?5*` ?

evanshultz:
Footprint filters should not have pin count, so don't include the `?5` part. Furthermore, use asterisks instead of question makes wherever possible and that applies to all instances here. Note that I haven't looked at anything else but the last comment.

myfreescalewebpage:
@evanshultz ok. @aewallin so keep the footprint like that for that PR.

aewallin:
fp-filter changed.
box enlarged.
library seems inconsistent on negative voltage regulators: to have Vin/Vout at bottom and GND at top, or like positive regulators Vin/Vout at top and GND at bottom. now using latter approach.
lt3015q

aewallin:
> maybe you can provide a screenshot.

The two styles I was thinking about are below. Maybe it is best to have symbols KLC-compliant with GND down, and the user can flip with X/Y as appropriate for the schematic.

GND-down style on +v and -v:
dual-polarity-regulated-power-supply-1326579994

GND-up on -v, with a symmetryc drawing for a bipolar supply:
dual voltage

myfreescalewebpage:
@aewallin agreed with you I have forgotten X flip possibility and just made a try. Your proposition sounds good and it respects the KLC so that's also very nice !

@antoniovazquezblanco I think this PR is ready to be merged.

aewallin:
@myfreescalewebpage Looking at examples from the current regulator_linear library it looks like some ADJ/SENSE/FB pins are marked Passive, and some are marked Input.

Any opinion on this from a Librarian - I'll change the ADJ pin type here and in #1169, if there is consensus.

myfreescalewebpage:
Any librarian about ADJ pin type question ? Thanks.

antoniovazquezblanco:
@KiCad/librarians question above, please :)

Misca1234:
Even if older merges is different I would argue to follow the data sheet,
according to page 14, 
SHDN should be input 
SENSE should be input

DanSGiesbrecht:
I think that either passive or input is suitable for `ADJ`. I have seen it argued that it should be passive because you could feed it with a resistor-divider, but this wouldn't generate an ERC warning if the pin was an input. This could be suggested to be formalized in the KLC.

Misca1234:
yes, the ERC .....

aewallin:
after changes:
lt3015q

poeschlr:
Regarding pin types: we are working on setting up an algorithm for determening which type to use. See https://github.com/KiCad/kicad-website/issues/342

Misca1234:
I am happy with this little gem, any objections ?

poeschlr:
Not from my side. As there are already so many reviews i am going to assume this is ok and simply merge. Thanks @aewallin for your great work here.

PR #884, opened 2018-09-03
Fix some footprints in Isolator.lib

dsa-t:
Fixes #792.

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #1177, opened 2018-11-26
Fix some footprints in Isolator.lib

### LINKED ISSUE #792: Incorrect footprints in Isolator.lib
As reported by @alaut at https://github.com/KiCad/kicad-symbols/issues/520#issuecomment-409347596:

>It appears to me that TLP291-4 and others in the __Isolators.lib__ library are associated with the wrong footprints. I'm not nearly experienced enough to make any reliable attempt to correct through these yet... though I plan to contribute sooner than later.
>
>For example. I'm not sure if technically the footprint associated with LTV-847S is correct or within a certain tolerance. Datasheet indicated 9mm width, footprint appears to be 9.53mm. I'm inclined to believe I'm wrong here.
>
>TLP291-4 is definitely wrong though by a lot unless I'm missing something. According to the [datasheet](http://bit.ly/2Oyhfaj), the pitch is 1.27mm not 2.54 as associated in _Package_DIP:SMDIP-16_W7.62mm_

I agree that TLP291-4 is definitely wrong, and LTV-847S looks suspicious as well.

antoniovazquezblanco:
Fixes #792.

Re push of https://github.com/KiCad/kicad-symbols/pull/884.


evanshultz:
This all looks good to me. Are you going to create the missing footprints?

antoniovazquezblanco:
I do not have that much free time... I would prefer to leave that as a todo in a issue. Can that be done?

evanshultz:
In general we don't do that, because users will be stuck without a valid footprint. And when that user is Wayne something like #520 happens and it's a frenzy. If you'd be willing to review my footprint PR, let me try to squeeze it in today (tomorrow at the latest).

Wayne mentioned tagging 5.0.2 today and having the libs tagged with a week, so we should be able to get this done. :)

antoniovazquezblanco:
Let's do that. I will review your PR. ;)

evanshultz:
They are at https://github.com/KiCad/kicad-footprints/pull/1115.

evanshultz:
And script is at https://github.com/pointhi/kicad-footprint-generator/pull/227.

evanshultz:
Footprints are merged and we're all green. Thanks!

PR #1181, opened 2018-11-26
Update all files to eeschema library version 2.4

antoniovazquezblanco:
There are many broken PRs due to other PRs being merged updating the file version.  I have decided to update all the files to the latest format to avoid pinging people on the long term to fix those merge conflicts. This will break quite some opened PRs but I guess it will save time in the long term.


antoniovazquezblanco:
It would be nice to have the input of all the @KiCad/librarians because this can cause trouble with a couple of open PRs.

I would vote to try and squeeze this after 5.0.2 just to have time enough to solve conflicts. I am willing to perform conflict resolutions for users for already in review PRs.

evanshultz:
I'm on the fence but it feels like "make work" FWIW. I guess we should look at the tradeoff of doing this versus not. How many PRs are affected, if you had a look? With so many open PRs, is it easier to resolve them as needed rather than this? It seems you surveyed the status of things and figured this was the best way to go so how did you come to that conclusion? (I'm not being belligerent, just seeking understanding.)

Misca1234:
I am leaning as well to take them one by one, if the contributor does not come back we can take over the push and fix them separately.

antoniovazquezblanco:
My idea is that every time a PR is opened for a library that has not been updated changes are more difficult to review due to KiCad reordering the files. Also every PR that is merged on a older format will cause a conflict with other opened PRs. I wanted to merge this after 5.0.2 or so in order to avoid people collaborating on older formats so that no new PRs will see conflicts.

Average git users are not comfortable solving the conflicts in their PRs. It is even worse when the conflict comes from a reorganization of a whole file just because of a format change.

I understand that this will cause trouble in the short term. I can have a look if there are any PRs in progress that update a lib and work on trying to move those before merging this.

antoniovazquezblanco:
Should we go all in now with plenty of time before 5.1.0 or should I back out and close this issue?

I can even try to split this in batches if it makes you librarians feel better and less problem prone in the short term.

Thanks!

DanSGiesbrecht:
I think that breaking this into smaller chunks would be ideal. I have no problem with making these enhancements, but having it all in one PR is a lot.

Do you know why the additions/deletions are different numbers?
diff

antoniovazquezblanco:
> Do you know why the additions/deletions are different numbers?

That is a very good question but I don't have an answer.

I will be splitting this then.

PR #1182, opened 2018-11-27
Interface_USB: Add CH340T USB - TTL converter

henriqueqc:
Add the CH340T USB - TTL serial interface converter.

------------

https://cdn.sparkfun.com/datasheets/Dev/Arduino/Other/CH340DS1.PDF

screenshot from 2018-11-27 11-20-00

It uses the SSOP-20_5.3x7.2mm_P0.65mm footprint.

Footprint already present in the kicad-footprints repo.


henriqueqc:
Updated screenshot:
screenshot from 2018-11-30 10-00-31

evanshultz:
Thanks! Sorry for my delay.

Just a couple more things:
1. Push the NC pin to the right so that the end is on the symbol body.
1. I think you can make the symbol thinner like it was. No need to fatten it.
1. I see the changes on pins 1 and 2. Makes sense to change them to clock outputs, but should pin 2 be just `CKOL` on the T suffix?
1. Would you arrange the pins more like CH340G which is already in the library:
image

Closing and re-opening to trigger the CLA check to complete.

evanshultz:
Thanks! Can you push up the bottom? Making the symbol centered isn't a strong guideline in KLC and the GND pin doesn't need to hang down so low. Then it's ready! :)

henriqueqc:
> Thanks! Can you push up the bottom? Making the symbol centered isn't a strong guideline in KLC and the GND pin doesn't need to hang down so low. Then it's ready! :)

Yes, I had this in mind but gave priority to the symbol being centered, good to know. Thank you for the guidance.

Here is a updated screenshot:

screenshot from 2019-02-21 09-35-50

evanshultz:
Looks good, and you had the right idea. Thanks for taking the time to finish off this symbol.

This was your first PR and even with a delay on my side I hope you're happy to see your contribution in KiCad and you'll continue to grow and improve the library. Thank you!

PR #1184, opened 2018-11-28
Add D52M

DanSGiesbrecht:
Adding the "M8" form-factor D52M ANT module by Garmin Canada (Dynastream Innovations).

------------

  - [Datasheet URL (warning - download)](https://www.thisisant.com/assets/resources/Datasheets/D00001687_D52_Module_Datasheet.v.2.3.pdf)
#### Schematic Symbol:
symbol

  - [Footprint PR](https://github.com/KiCad/kicad-footprints/pull/1117).
  - **Violating S5.1**: Pending [Footprint PR](https://github.com/KiCad/kicad-footprints/pull/1117).


myfreescalewebpage:
hi @DanSGiesbrecht 

Comments I have during my review:

- [ ] The Documentation field should be https://www.thisisant.com/assets/resources/Datasheets/D00001687_D52_Module_Datasheet.v.2.3.pdf
- [ ] Where have you found the pad numbering (A1 to F6) ?

Cheers,
Joel

DanSGiesbrecht:
@myfreescalewebpage agreed, datasheet link fixed! Thanks.

A1 - F6 came from the footprint screenshot found in the module datasheet (figure 14, page 22), and the Altium footprint/symbol. I chose it because it aligns more closely with other footprints in the library. Unfortunately, this makes checking a bit tricky - when making this, I translated figure 7 (pin names) to figure 14.

#### Figure 7 (mirrored, bottom looking up)
fig7
#### Figure 14 (footprint, top looking down)
fp

myfreescalewebpage:
@DanSGiesbrecht thanks for the clarification. Footprint verified, no comment to do.

A single comment for me:
- [ ] Footprint filter should be `M8*9.8x14.0mm*P1.5mm*` (remove leading star).

Cheers,
Joel

evanshultz:
- Doesn't just `D52M` fully describe this symbol?
- See the note I made about the footprint name in that PR.
- Should this symbol be designed to support the 6 extra pins of the D52Q, and then this one is depopulated?
- GND pins should be named VSS.
- I'm wondering if the P0xx/AINx pins should have their names swapped around the slash so that the P0xx part is always near the symbol pin. What do you think?

There may be a more awful and unclear and confusing way to notate a pinout, but I can't possibly imagine what it would be.

myfreescalewebpage:
> Doesn't just D52M fully describe this symbol?

Yes probably.

> See the note I made about the footprint name in that PR.

Add of Dynastream right ? So part name should be `Dynastream_D52M` ?

> Should this symbol be designed to support the 6 extra pins of the D52Q, and then this one is depopulated?

I like this practice but in this case this will increase the height. Pins are P007, P017, P019, P023, P026 and P027, all on the left. @DanSGiesbrecht what is your opinion ? If you planned to propose the D52Q too, so you can make both symbol consistent, this is a good idea, so it's easy to swap them on the schematic.

> GND pins should be named VSS.

Agree.

> I'm wondering if the P0xx/AINx pins should have their names swapped around the slash so that the P0xx part is always near the symbol pin. What do you think?

That's a common practice and it is done by @DanSGiesbrecht for NFC pins, so I will vote yes.

DanSGiesbrecht:
@evanshultz , @myfreescalewebpage thank you for all your comments. Sorry that this has taken me a while to get to.

> Doesn't just `D52M` fully describe this symbol?

It could. However, it is possible that a module could be released that depends on the `M8` descriptor of the part number.

> Should this symbol be designed to support the 6 extra pins of the D52Q, and then this one is depopulated?

I see what you're suggesting. I think there's a way to accommodate both (symbol size and drop-in symbol compatibility): when making the D52Q symbol, start from the D52M, and enlarge downwards. This would mean that the D52M symbol would remain as small as possible. Making the D52Q symbol drop-in replaceable would only require moving GND downwards. Please let me know what you think of this, and whether or not what I'm describing is clear.

> GND pins should be named VSS.

Will fix.

> I'm wondering if the P0xx/AINx pins should have their names swapped around the slash so that the P0xx part is always near the symbol pin. What do you think?

I like it. Will fix.

> There may be a more awful and unclear and confusing way to notate a pinout, but I can't possibly imagine what it would be.

🤣 I agree. It was a huge pain to translate and double (triple) check.

> Add of Dynastream right ? So part name should be Dynastream_D52M ?

Dynastream Innovations has officially re-branded to Garmin Canada. The new [datasheet](https://www.thisisant.com/assets/resources/D00001687_D52_Module_Datasheet.v.2.3_(Garmin).pdf) reflects the name-change.

myfreescalewebpage:
@DanSGiesbrecht I would suggest `Garmin*D52M*9.8x14.0mm*` for the footprint filter. Other modifications okay for me.
Cheers,
Joel

evanshultz:
Thanks!

Only one question left for me. I do understand what you're saying about `M8` being the package descriptor so I think that should be included in the footprint name (which affects the symbol) as even with the dimensions using the term `M8` is helpful. But now I'm questioning the `D52M` part. Are there other modules which are physically compatible so we should do something like `Garmin*M8*9.8x14.0mm*`? And if this package isn't unique to Garmin it should be more generic.

DanSGiesbrecht:
I think you're right, the footprint name should be changed.
Here's some elaboration:
- The module form-factor and naming convention are wholly unique to Garmin's modules, so I think Garmin should be left in the name.
- M8 is a descriptor for form-factor (9.8x14.0mm), **not** pin land-pattern (pitch, number of pins).
- D52M is not the only module that has this pin land-pattern. The older [N5xxM8](https://www.thisisant.com/assets/resources/D00001598_N5_Module_Series_Datasheet_(Garmin).pdf) shares this land-pattern.

- To summarize, there is nothing in the part naming convention that actually defines the pin land-pattern (pitch, number of pins). If it's desirable to have a generic footprint that would cover D52M and N5, I suggest reverting to my original footprint name, with slight modification: ~~`Dynastream_M8-35_9.8x14.0mm_Layout6x6_P1.5mm`~~ -> `Garmin_M8-35_9.8x14.0mm_Layout6x6_P1.5mm`

evanshultz:
Thanks for the info. Hmmm... it is messy. Since you say this is Garmin-specific, though, we probably don't need to get too deep since this footprint won't spread across the library. So I'll just take your suggestion. That would mean a FP filter of `* Garmin*M8*9.8x14.0mm*Layout6x6*P1.5mm*` if I'm not mistaken.

DanSGiesbrecht:
Agreed, it's unfortunate that M8 doesn't also define the pin-out. Is the leading `*` needed in the footprint filter, ahead of `Garmin`?

myfreescalewebpage:
For me it is not required.

DanSGiesbrecht:
I believe the leading `* ` was a copy-paste error. It happens if you copy-paste the string from above: `* Garmin_M8-35_9.8x14.0mm_Layout6x6_P1.5mm`

evanshultz:
Nope. There was a time when a requirement for a leading asterisk was put into the code (between releases, only affecting nightlies), but it was fixed shortly thereafter and should not be done anywhere in the library.

myfreescalewebpage:
Just made a global review of your part, for me it's okay, just need to wait the footprint to merge this.

DanSGiesbrecht:
✔️ Good to go!

myfreescalewebpage:
Thanks, merging!

PR #1188, opened 2018-11-28
Restore 3.3V HO8/15/25 symbols

### LINKED ISSUE #1170: Was there a reason for removing the 3.3V version of the LEM HO-8 (and its sisters)
I seem to remember that we had the 3.3 versions of these in our lib. (I have been the one adding it a while back) But i can no longer find them. Where they removed for some good reason on accident or am i just blind?

For reference we now have LEM_HO8-NP but not LEM_HO8-NP/SP33 (The / in the part number would need to be replaced by some other character) Both can reference the same footprint but would need a different datsheet link.

The same applies to the 15A and 25A variants.

evanshultz:
Fixes #1170.

Apparently the SP33 versions were removed (by me) and this PR restores them. The symbol is a simple copy of the non-SP33 symbol so no screenshot.

There's a bit of noise in the commit but nothing else was changed.

------------


antoniovazquezblanco:
Looks fine to me

antoniovazquezblanco:
Thanks!

PR #1189, opened 2018-11-29
Add ATmega3209/4809

fauxpark:
Almost done adding new symbols 👍 One more ATmega to go after this.

3209-a
3209-m

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/40002016A.pdf

------------


evanshultz:
Yay! Almost there!

This is a rather annoying datasheet. No ordering information and it doesn't have both package drawings.

1. I'm not familiar with the UPDI interface, but http://ww1.microchip.com/downloads/en/AppNotes/00002519A.pdf validates it as Bidirectional type.
1. Is there a reason to wrap the ports clockwise in alphabetical order instead of only top down?
1. Is the excess height to match other, existing symbols?
1. Where did you find info about the UQFN package? I can see it's 6x6mm on page 2, but https://www.microchip.com/quality/packaging-specifications?packageFamily=UQFN shows 3 UQFN-48s of this size. All of them are 0.4mm pitch with a 4.6mm square EP so they can probably fit the same footprint, but I wonder if I'm missing a clear link to the package drawing for this family.
1. That being said, we don't have exactly the right footprint. I'll make it. For the last family of these symbols to submit, could you check if there are also any missing footprints so I can do them together?

fauxpark:
1. That's what it looks like to me too. I just followed along with the block diagram.
2. The clockwise port arrangement was already in place in many symbols when I started, so I just stuck with it. Since they're bidirectional (+ hi-Z) either side is fine, and to save space they can be placed on both.
3. Yes, but I may as well trim that since it's not pin compatible with anything else AFAIK.
4. [Here](https://www.microchipdirect.com/product/ATMEGA3209) is the ordering information, but it's not much more informative about the QFN package. The M4 and PTX drawings seem to have the exact same land pattern but the MV drawing (the currently assigned footprint) is different. I'm not sure which one these parts use.
5. The remaining ATmega is a single symbol in LQFP-48 package, which footprint is already in the libs 👍 there's also a WLCSP-12 package for the ATtiny20, and that should be it.

evanshultz:
4. I'll ask Microchip which package drawing is for the UQFN package of this part.
5. Does the WLCSP-12 footprint already exist?

fauxpark:
Not yet. I have it in my tree and I'll review it and submit when these PRs are done.

evanshultz:
Would you mind letting me know the package info so I can generate both footprints in one go?

fauxpark:
There is no entry in the Microchip packaging specs, but the package drawing is [here](http://ww1.microchip.com/downloads/en/devicedoc/atmel-8235-8-bit-avr-microcontroller-attiny20_datasheet.pdf#page=208).

evanshultz:
The WLCSP footprint we will need in the future is submitted at https://github.com/KiCad/kicad-footprints/pull/1179.

evanshultz:
Once I get the answer from Microchip about the correct package drawing for the UQFN used by this symbol, this can be merged if the existing footprint is OK. Everything else looks great to me.

evanshultz:
The drawing for the correct package isn't on Microchip's website that I can see, but here is an image:
image

It doesn't appear that we have a footprint in the library already (I didn't check any open PRs). Are you up for making one?

fauxpark:
I can, but it won't be scripted :P

poeschlr:
Scripting such footprints is extremely easy. You only need to enter the information found in the dimensions table into an appropriate yaml file. Then simply run the script.

I bet you would be faster scripting this compared to creating it by hand.
If you are not prepared to look into this then i could help you out there.

If you need help with the script then contact either me personally (via the forum for example) Or post the question here.

fauxpark:
Footprint and script PRs submitted, see the above links.

evanshultz:
Looks good now. Yay!

PR #1190, opened 2018-11-29
Analog_Switch: Add ADG1414

antoniovazquezblanco:
https://www.analog.com/media/en/technical-documentation/data-sheets/ADG1414.pdf

screenshot from 2018-11-29 12-07-10

------------


evanshultz:
- Please add the switch on resistance to the description.
- Remove pin count from the FP filter.
- It seems like pin 23's main function is the SPI interface power supply. This is because a hardware POR is provided and it may not ever be used for a reset function, but it will definitely be required to be used as a power supply. I know that messes up the entire symbol design, but calling is `~RESET~/VL` and making it Power Input and putting it on the top seems like the proper thing for this pin. What do you think?

antoniovazquezblanco:
All done. Sym is now less visually apealing though.

screenshot from 2019-01-08 11-21-42

evanshultz:
Yeah. I agree it looks worse. What about using two units, one for the switch and one for power? Do you have other ideas?

Travis caught an error with the moved pin.

I also wonder if there's a better arrangement of pins to make daisy-chaining easier? Maybe DIN above all the other digital pins? Or SDO on the right side?

antoniovazquezblanco:
> Yeah. I agree it looks worse. What about using two units, one for the switch and one for power? Do you have other ideas?

I don't really like the two unit idea (just a matter of opinion). Another option would be to drop the reset function from the pin. It looks like a reset operation consists on dropping power on the digital side of the uC. That pin looks like it is only a power supply... What do you think?

> I also wonder if there's a better arrangement of pins to make daisy-chaining easier? Maybe DIN above all the other digital pins? Or SDO on the right side?

SDO can be used to read data back or for daisy chaining, given this, I would prefer to keep it in the left side. In order to improve chaining I would prefer to move DIN up.

Do you agree with the changes?

evanshultz:
Sure. Two units was just an idea.

It appears RESET is not the typical use of the pin, but it does have that function if needed. To clarify you are talking about leaving the location alone and just calling it `VL`, right? That makes me feel uneasy because, except for MCUs where we decided to limit pin names and triangle-shaped (comparators and opamps) symbols with limited area and common pins, we usually stick closely to the datasheet pin names. I dislike doing random deviations here and there which are hard to document. That being said, it would clean this up. Let's get Rene's input; it feels like a small thing but if he feels strongly that the full pin name on the datasheet should be shown, except for the few defined cases above, then that is that.

Oh. Gotcha. Yeah, moving the digital pin sounds good. I just noticed that SDO is open drain so please change the pin type.

antoniovazquezblanco:
> To clarify you are talking about leaving the location alone and just calling it `VL`, right

Yes

evanshultz:
Now looks like this:
image

The only remaining question is what to do with `VL`. Sure. It will look like this then:
image

If you do that I'm good to merge. Also, the footprint text is floating way above the symbol and it could scoot down.

antoniovazquezblanco:
Done!

evanshultz:
Great. Let's go with this. Thanks!

antoniovazquezblanco:
Thank you!

PR #1192, opened 2018-11-30
Replace semigraphical 0+E288 with minus character

carlo-bramini:
I discovered that I could not find the VREF- of MSP430G2755 pin because the minus "-" character was in reality a semigraphical character with code 0+E288, that I have no idea on which kind of keyboard it exists. Perhaps, it has been copied and pasted somewhere from the datasheet, but as a CAD symbol, I think that it would be better to use the simple minus, otherwise I have no idea how someone could type it. It's a quite simple fix hopefully.


poeschlr:
Thanks

Issue #550, opened 2018-04-26
Splitting up large symbols into units


evanshultz:
@poeschlr @jkriege2 @Ratfink 
We haven't done much splitting up of large symbols into multiple units that I see or can recall. How do we want to handle it?

See https://github.com/KiCad/kicad-symbols/pull/457 for an example of an active PR with a split symbol.

I personally like splitting for some cases floating in my mind, but there are a lot different facets to this. If it's allowed it seems we should have rules or at least a guidelines. How should they be split up: number of pins, size of symbol, etc.? Are there only certain type(s) of symbols that could be split? Should we allow split and non-split versions of the same symbol as two users might have opposing views?

whitequark:
Ping. It's been 9 days with no replies...

Ratfink:
There are already a few symbols like this in the library.  I think they're mainly in the CPU libraries, e.g. CPU_PowerPC:MPC8641D.

The way it's grouped in #457 seems okay to me, but I feel like the placement of power pins should be just like in any other symbol.  Also, I'm not sure if a 49-pin symbol is big enough to really need to be split, but I wouldn't say it has to be not split either.

As for making guidelines for splitting large symbols, gosh, I wouldn't know where to begin.  If I had any suggestions, I would have brought them up when I reworked the STM32 library.

evanshultz:
How about this:
```
- If a symbol does not fit within an A4 size page with power pins on top and bottom and signal pins on the sides, it may be split into multiple units.
- Each unit should have a clear and defined grouping of pins, for example by function (GPIO B, ACPI, etc.)
- All power pins are to be placed on the last unit unless they have a specific relation to another unit.
- Note that KiCad does not generate a warning if not all units are placed!
<provide an example symbol image>
```

https://github.com/KiCad/kicad-symbols/pull/457 is one I'm planning to merge as a split symbol unless somebody objects.

See https://github.com/KiCad/kicad-symbols/pull/622 for another example. It's just barely short enough to fit on an A4 page, but because of obvious pin groups it might also be a candidate to split.
![image](https://user-images.githubusercontent.com/1936989/41056982-b17b2012-697a-11e8-8d61-ffb006786906.png)

Ratfink:
That's a reasonable-sounding set of rules you have.  I think this phrasing might be better for the third bullet, using RFC 2119 language:

> * All power pins **must** be placed on the last unit unless they have a specific relation to another unit, in which case they **may** be placed on that unit instead.

[...]

PR #1195, opened 2018-12-01
MCU Module: create VisionSOM boards

myfreescalewebpage:
Creation of VisionSOM-6UL and VisionSOM-6ULL modules from SomLabs available at https://somlabs.com/category/products/system-on-module/arm-cortex-a7.

sl_som_6ul

Symbol preview:

capture

These modules have SODIMM-200 form factor. Several connectors can be used, however, I will greatly appreciate if pull request https://github.com/KiCad/kicad-footprints/pull/375 is reviewed/accepted.

Info: these symbols have many GND pins. I have not put them at the bottom of the symbols, thus I'm currently violating S4.2. However, size of the symbol will be unacceptable at my point of view if all GND are of the bottom.

Capture of checklib result:

capture

Thanks for review,
Joel


evanshultz:
I support breaking up big symbols as a general statement. I proposed an update to KLC for splitting up symbols at https://github.com/KiCad/kicad-symbols/issues/550. It wasn't ratified but perhaps this is a good opportunity to see if there's a reasonable set of rules we can come up with for big symbols.

GND pins are always stacked unless there's a really weird requirement and I don't see anything here that would prevent stacking them. It's power pins that are not always stacked and I didn't catch any reason they can't be stacked with a quick read through of the datasheet. Is anybody able to take a look at the Altium symbol? If it has stacked pins that's probably a good indication they can be stacked.

myfreescalewebpage:
@DanSGiesbrecht thanks for your review.

One remark, the KLC S3.1 says:

> Exception: For non-symmetrical symbols, or symbols where this requirement would move pins off the 100mil grid, then the symbol should be placed as close to the origin as possible while observing the pin grid alignment requirements.

That's typically my case here and that's why the violation is in white characters. If this is absolutely to be respected, I would suggest to modify the KLC rule and the check script to make it has an error,  displayed in red.

=> What should I do ? I'm actually conformed with the KLC...

Regarding the stacked pin, I will propose you a screenshot with 5V/3.3V and GND pin stacked at the top and the bottom, just to see what it looks like.

The symbols proposed by SomLabs is not stacked, see the schematic of the evaluation board at https://somlabs.com/wp-content/uploads/VisionCB-STD-v2-0_schematics.pdf page 3. However, the pins are not grouped at all and that's not very convenient as you can see, particularly if you add a screen and a camera...

I will make the other fixes this evening.

Cheers,
Joel

DanSGiesbrecht:
While centering the symbol is a KLC warning, I would still recommend centering the part, as there is no reason not to.

Please don't stack the 5V/3V pins, but still place them on the top. The GND pins should be stacked. This allows for clearer indication of decoupling on the 5V/3V pins from the designer, while making it easy to connect the GND pins.

Additionally, how would you feel about creating multiple units for this part, grouping pins by port/function?

myfreescalewebpage:
@DanSGiesbrecht thanks for your response

> While centering the symbol is a KLC warning, I would still recommend centering the part, as there is no reason not to.

There is one, all pins are no more on the 100mil grid when centering the symbol, and that's a real KLC error.

> Additionally, how would you feel about creating multiple units for this part, grouping pins by port/function?

About the units, I think this very useful for NAND / NOR / OpAmp but totally useless and unreadable for this kind of symbols. If we go this way all microcontrollers should be split in several units.

----

As previously discussed, here is a screenshot of the module with GND stacked at the bottom.

stacked

The rendering with symbol not stacked and stacked is provided below on a A3 page (portrait):

v1

v2

My point of view is that not stacking is better, also see the evaluation board schematic I have quoted before. However, that's the Kicad strategy to check and decide.

DanSGiesbrecht:
It's possible to size the symbol so that it's centered, and also keep pins on a 100mil grid. With regards to making units, I think doing this for all [high pin-count] MCUs is what @evanshultz is suggesting, similar to what is done for the FPGA libraries here.

I think there's some confusion: the screenshots that you've shown above do not have any pins stacked. Please see [this screenshot](http://kicad-pcb.org/img/klc/S4.6.png) for an example of pin stacking, and [S4.3](http://kicad-pcb.org/libraries/klc/S4.3/) for rules on how to pin stack.

myfreescalewebpage:
> It's possible to size the symbol so that it's centered, and also keep pins on a 100mil grid. 

Don't know if this is useful but I agree.

---

Sorry for the confusion stacked vs pins at the bottom. In my opinion, pins should never be stacked, that's leading to a big confusion. A schematic should describe everything, not permitting confusion or doubt. You know what you are doing but another guy looking at your schematic will be totally lost if pin are not visible / missing / stacked.

Well, regarding to this symbol I propose, what the other librarians think about ? GND on the left and right sides as proposed initially, all GND at the bottom and stacked into a single pin ?

DanSGiesbrecht:
Pin-stacking has been a controversial topic in the past. You have a valid point as why not to pin stack, however, currently that is the format that the KiCad libraries employ, which requires new contributions to match.
@poeschlr, can you think of a past PR with a similar discussion?

evanshultz:
First, I absolutely understand what you mean, Joel, about having all pins numbers visible and making the schematic self-documenting. I'm not against you or saying you're wrong. Nobody is.

I would not like to get into a "stack or not stack" discussion again right now. There are valid points for both styles, and we hope KiCad can satisfy the desires of folks on both sides in v6. However, in the middle of v5 my strong opinion is we need GND pins on symbols to be stacked to accept them.

Second, regarding multi-unit symbols, can you explain why splitting up is "totally useless and unreadable for this kind of symbols"? I would guess LCD, USB/Ethernet, GPIO/JTAG, and Power may be one way to divide the pins up into four units which are logically grouped.

myfreescalewebpage:
> First, I absolutely understand what you mean, Joel, about having all pins numbers visible and making the schematic self-documenting. I'm not against you or saying you're wrong. Nobody is.

No worries it's ok :)

> I would not like to get into a "stack or not stack" discussion again right now. There are valid points for both styles, and we hope KiCad can satisfy the desires of folks on both sides in v6. However, in the middle of v5 my strong opinion is we need GND pins on symbols to be stacked to accept them.

I agree, we should conclude now, if the symbol need to wait for v6, I will wait.
I haven't check v6 content but it seems it will be a nice release with many new functionalities isn't it ?

> Second, regarding multi-unit symbols, can you explain why splitting up is "totally useless and unreadable for this kind of symbols"? I would guess LCD, USB/Ethernet, GPIO/JTAG, and Power may be one way to divide the pins up into four units which are logically grouped.

You can imagine various possibilities:
- Using a GPIO to control Ethernet behavior
- If you are not using the USB OTG VBUS you can turn it back to a simple GPIO to control whatever you want on another part of the schematic, in this case it's totally unreadable because you create a unit "USB" but the pins in this unit are not USB only....
- CSI and LCD interface can also be used for SPI, UART, I2C... etc etc.
- a single pin function can also be routed to various possibilities, the default name of the pin is only the "Alternate 0" function. Example "UART1-TX" can be routed internally to pin 1, 4, 8 and 11 (for example). So you can cross one UART and one Ethernet pin for example to simplify the PCB.

Micro-controllers and microprocessors can be programmed the way you want, the pins are not fixed. Ordering by blocks makes no sense when you think about this like that.
That's my point of view but we should think to Kicad globally to define the strategy to review this and future devices the same way.

evanshultz:
Splitting up symbols hasn't been discussed much so this is good info. And this parts seems quite flexible. 

Also note that the units don't have a name, so I was only thinking of functional grouping of pins. I'm not familiar with the part so I was just guessing based on a quick look at the schematic you linked above.

The Lattice symbols were split up by port because that made sense. All pins were universal, as you mentioned with MCUs, so there wasn't any functional grouping to consider there.

Lastly on splitting symbols, as there's no consensus and little discussion so far I would say finish this PR and then it can be merged in any fashion (and not just this one, but in general). Waiting out a decision before merging is OK too. There's no guidelines for splitting symbols so do it either way. :)

DanSGiesbrecht:
Agreed that the discussion on splitting can wait. I think that with pin-stacking done on this, we can get this merged in the short term.

poeschlr:
Every-time pin stacking comes up there are two options. The contributor is happy about our decision or they really hate it. 
I therefore feel there are two camps with about equal size. (maybe a third one with "i don't care) There is no  way to make both happy so we simply needed to decide for one of the camps. 

This decision was made a long time ago, i am not questioning it now. In part because consistency is a very important aspect of any library. But in part it is also because i personally prefer pins to be stacked.
My reasoning is not really because it avoids clutter. (This is a positive side effect) But because it moves responsibility for correct connections away from the circuit designer to the symbol designer. (And in this case to the whole library team.) Think of stacked pins as the equivalent to eagles one to many connection option for connecting symbol pins to footprint pads.

myfreescalewebpage:
I'm not aware of all the things discussed in the past, but maybe we need a new functionality in Kicad that permit to create several drawing of the device. I mean not creating X times the same device in the library, but creating multiple "views" in the same symbol definition so people creating schematic can choose the best view they want. With a single right click available in the schematic editor to switch from one view to another one, it can be very cool !
For example, view 1: all pins on a single block ; view 2: pins stacked ; view 3 : several blocks.

Or maybe another possibility with something looking like the hierarchical sheets. You create 1/2/3... block(s) representing the same device and you can arrange the pins the way you want on the different blocks the same way you can import the hierarchical pins from a sheet...

Any place somewhere people can discuss/propose about futur functionalities of Kicad ?
I have also just read https://github.com/KiCad/kicad-symbols/issues/1040#issuecomment-432547760 that I think is very nice proposition.

myfreescalewebpage:
So what about the symbol now ? Power on top and bottom ? Pins stacked ? Only GND ? Datasheet of the device don't specify individual capacitor for 5V and 3.3V pins, they are grouped at the same place on the SODIMM200 connector.

poeschlr:
You read that wrong. The exception means that the invisible pins of such a stack should not be of type powr input as these would be global labels. So a stack of power input pins would be one visible power input pin with invisible passive pins.

myfreescalewebpage:
@poeschlr ok understood

myfreescalewebpage:
@poeschlr what about the ERC testing ? How passive pins are validated ?

myfreescalewebpage:
Also just seen the footprint is proposed in anew PR : https://github.com/KiCad/kicad-footprints/pull/1158

evanshultz:
There is empty space at the top and bottom of this symbol that can be removed. Personally, I still would like breaking it up; at least move the power pins to a separate unit.

Regarding ERC, Passive pins can be connected any way you like without creating an error (I'm talking about the default ERC settings):
image

Is there a particular pin or pins you were thinking about?

myfreescalewebpage:
@evanshultz The empty space is to get the symbol vertically symmetrical with KLC S3.1 validated.

About power pins, my position will remain this one. Without any KLC to guide this, it is the responsibility of the submitter, and I assume this design. Separating the power pin is also far away of the part from SomLab (Altium lib).

evanshultz:
It looks like you could take off on both the top and bottom to keep it centered. Also, centering is probably the weakest guideline and is often ignored for aesthetic reasons.

It's quite true that KLC should contain all library requirements, and after any discussion that results in a change KLC should get an update quickly. Unfortunately, that didn't happen in this case (or at all lately).

Here is the statement about stacking pins from Rene where he laid out how the library should be done:
image
image

Ground pins are stacked above so I understand this is only in regards to power pins. Is there a requirement for this board to have caps on each power pin that fits into the statements above?

KiCad symbols also don't copy other symbol styles for the sake of doing so, but rather use a logical (connectivity-based) design that explicitly doesn't represent the physical part.

myfreescalewebpage:
@evanshultz thanks for this. I understand Rene comment no only for power pins, but all that need the device to work. So if two signals with the same name are connected, they need to be stacked. Seems this is a guideline that is easy to understand and control. As he points out, we need in the future a way to indicate in the schematic some pins are stacked (and which ones) so the schematics reviewers can easily checked.

Going back to the symbol of this PR, the +5VIN pins needs to be connected together with only a single group of caps and the +3V3OUT pins are simply all connected together with no need for caps. This is how it is designed on the evaluation board of this module : https://somlabs.com/wp-content/uploads/VisionCB-schematic-1-4.pdf (page 3).

So, we agree that I need to stack 5V and 3V3 pins, right ?

evanshultz:
Yes, I think a single ground, a single +5VIN, and a single +3V3OUT pin would work. Unless there is documentation that it shouldn't be this way. You didn't find anything in the reference design and I looked through the datasheet and didn't see anything either. So yes, I think stacking all should be good and the most direct route to getting this merged.

myfreescalewebpage:
@evanshultz it's done with the last commit. Thanks.

DanSGiesbrecht:
Hi @myfreescalewebpage. Sorry for the delay. Thank you for all your patience with making these changes. Please orient the power pins on the top of the part to align with the KLC.

I haven't had a chance to validate the pin-out of this yet. We'll also need to wait for the footprint to be merged before merging this PR.

myfreescalewebpage:
@DanSGiesbrecht no worries :) just changed the position of the power supply pins

Cheers,
Joel

DanSGiesbrecht:
@myfreescalewebpage the pin location change looks good!
Unfortunately, I don't have time to check the pin-out until after Christmas. If anyone else has time to go through it, feel free. I also noticed that because there are so many GND pins, they must be checked in a text editor.

To-Do:
- [x] Check description, footprint filter, keywords
- [x] Check Travis output
- [x] Check datasheet links
- [x] Check alias: `-6ULL` has the same pin-out as the original `-6UL`
- [ ] Check pin number/type
- [ ] Check footprint (wait for PR merge)

evanshultz:
- Can you add some differentiating info to the descriptions so they're not identical?
- This one TE PN can't be the only possible footprint, can it? I would think no default footprint is best. The footprint filter is nice, though.
- Since the pin number needs 3 characters, I believe pin length should be 150mil.
- Should `+3.3VOUT` be on the side? Not to obviate any of the discussion above, but we do this on linear regulator output pins. I'm fine if we consider that library an aberration for aesthetic reasons and use top placement here. Just asking.

evanshultz:
I verified all top and bottom pins are good. Now moving on to the side pins by pin number...
- Pin 6 is missing the `-3V3` part.
- Pin 8 is missing the `-3V` part.

Only got through pin 10 before I have to go for now...

evanshultz:
- Pin 11 should be Power Output as it appears it can source current.
- Page 13 of the datasheet doesn't show a slash in pin 14's name.
- How did you know the pin type for pin 31?
- I don't see that pin 48 is a GPIO; it only mentions it's an input.
- NC pins, like 83, must be present in the symbol. Place their end along the edge of the symbol body.
- Pin 163 has "ENABLE" written out.
- Pin 196 is type Unspecified. Can this be more specific? Even if not, I don't recall this type being used on any other symbol but I do see Passive used even in VREF pins that can be input or output.

That's all the pins and their types reviewed.

DanSGiesbrecht:
Thanks, Evan! I know that's not a small amount of pins to review...

I agree with your comments.

myfreescalewebpage:
Thanks @DanSGiesbrecht  and @evanshultz , will check all your comments soon and try to respond one by one :)

myfreescalewebpage:
Comment by comment:

> Can you add some differentiating info to the descriptions so they're not identical?

Already done, first one is i.MX6-UL, second one is i.MX6-ULL. I add the max clock speed which is different but this is the only difference between the two modules.

> This one TE PN can't be the only possible footprint, can it? I would think no default footprint is best. The footprint filter is nice, though.

Several connectors can fit the module, yes. I will remove the footprint but keep the filter it's ok for me.

> Since the pin number needs 3 characters, I believe pin length should be 150mil.

It is 200mil today, which is confortable. Need to change ?

> Should +3.3VOUT be on the side? Not to obviate any of the discussion above, but we do this on linear regulator output pins. I'm fine if we consider that library an aberration for aesthetic reasons and use top placement here. Just asking.

I was also wondering but seen other module with power output on top (Arduino_Leonardo), and also this place all filtering caps on the top of the schematic which is nice.

> Pin 6 is missing the -3V3 part.
Pin 8 is missing the -3V part.

Yes, removed when stacking the power pins because the name was very long. If need to be conformed with the datasheet (which is better I agree), I will need to enlarge the symbol vertically. Please let me know.

> Pin 11 should be Power Output as it appears it can source current.

It can't. This is to drive a 5V switch to supply the USB interface. Example at https://somlabs.com/wp-content/uploads/VisionCB-schematic-1-4.pdf page 5.

> Page 13 of the datasheet doesn't show a slash in pin 14's name.

Agree. Changing.

> How did you know the pin type for pin 31?

Not indicated in the datasheet of the modules. Come from the datasheet of the iMX6 itself: https://www.nxp.com/docs/en/data-sheet/IMX6ULIEC.pdf page 112.

> I don't see that pin 48 is a GPIO; it only mentions it's an input.

Yes in the datasheet it's right. The pin is default to JTAG RESET behavior but can be converted to GPIO. Usage of this pin to control LCD EN on the evaluation board of the module: https://somlabs.com/wp-content/uploads/VisionCB-schematic-1-4.pdf page 3. This is also confirmed in the definition of the pin in the linux device tree of the iMX6 https://github.com/torvalds/linux/blob/master/arch/arm/boot/dts/imx6ul-pinfunc.h (lines 68-74).

> NC pins, like 83, must be present in the symbol. Place their end along the edge of the symbol body.

Done.

> Pin 196 is type Unspecified. Can this be more specific? Even if not, I don't recall this type being used on any other symbol but I do see Passive used even in VREF pins that can be input or output.

As already discussed in other PR, I do not agree with Passive, which lead to "do whatever you want with it it does not have any impact". This is not the case. On the VisionSOM module I think this is a Power Output for the CSI interface but it is not specified anywhere. If we only refer to the datasheet, "Unspecified" is obviously the most appropriate word for this. I'm in touch with SOMLabs, I will ask the question.

Cheers,
Joel

evanshultz:
The CI warning about the footprint filter is fine.

Yes, pin length should be 150mil.

Let's go with the full datasheet pin names.

Regarding pin 11 (and pin 12), those USB-OTGx-VBUS pins are the USB VBUS voltages on J200. Since they are OTG ports, they can be master or slave. If power for the downstream USB devices comes from VCC-5V0 through U200 in master mode, then are pins 11 and 12 current sinks to power the board in slave mode?

Is pin 31 a GPIO but without an internal pullup? Can it only be an output (understand open drain and not push-pull), or can it be an input too? Page 112 of the IC datasheet shows it as "GPIO" type.

Thanks for checking on pin 196.

evanshultz:
With all the pin length changes, can you post a screenshot for quick reference? I find this to always be handy if there a notable visual changes during a review.

Are pins 11 and 12 just a voltage monitor in slave mode?

Alright. Let's go with Open Collector type for pin 31.

Let us know about pin 196 when you get a response.

myfreescalewebpage:
> With all the pin length changes, can you post a screenshot for quick reference? I find this to always be handy if there a notable visual changes during a review.

Sure, sorry I have forgotten to post it.

capture

> Are pins 11 and ~~12~~ 13 just a voltage monitor in slave mode?

I just realize I made a mistake, I was thinking to pins VBUS-EN on the evaluation board schematic, but it is not correct. I try to summarize:
-> The pins we speak about are USB-OTG1-VBUS and USB-OTG2-VBUS. I have set them as Output but I'm wrong.
-> The datasheet of the module do not give details, the evaluation board shows the pins connected to the +5V of the USB connectors (you are right).
-> The datasheet of the processor says they are used to detect valid USB voltage, in Host mode when the VBUS-EN is effectively driven (that's where I'm making a confusion), and in Slave mode to detect cable is attached.
-> The datasheet of the processor indicates the max current sink by the pin is 50mA, so it's only detection of voltage, and possibly supply of USB part in the processor only ; it is not possible to supply the whole device with only 50mA, its need more or less 500mA in running mode.

=> However, we don't know how the module uses it exactly. I asked the schematic of the module (don't know if it is possible) but in any case it is an Input and I would suggest Input only, not Power Input, except if 50mA is enough to consider the pin as a Power Input ? What is your opinion on this value ?

This explain the role of the pin 31, CHD, CHarger Detect, which is probably driven in Slave mode when valid OTG1-VBUS is detected.

That also means I have to review my current project I'm working on with this module...... :(

> Let us know about pin 196 when you get a response.

SomLabs just answers: "CSI-VREF is connected on module by default to +3.3V voltage level (using 0R jumper, alternatively can be connected to internal +1.8V".
So the pin is connected to internal 3.3V regulator output and provide 3.3V. This behavior can be changed and I understand that it is intended to provide the output voltage level used on CSI interface. I propose Output or Power Output, what do you think ?

Cheers,
Joel

myfreescalewebpage:
@evanshultz what about my previous response ?

Cheers
Joel

evanshultz:
Got it. Pins 11 and 13 are Input type. That works for me.

Should pin 31 also be Input type?

It sounds like pin 196 is used to set the logic hi level for the CSI interface. It's not clear to me if this pin actually sources current for the CSI pins (internal to the module) when they're driven high, but I think Input type should be OK if you're fine with that.

myfreescalewebpage:
> Got it. Pins 11 and 13 are Input type. That works for me.

Great!

> Should pin 31 also be Input type?

No definitely is an Output.

> It sounds like pin 196 is used to set the logic hi level for the CSI interface. It's not clear to me if this pin actually sources current for the CSI pins (internal to the module) when they're driven high, but I think Input type should be OK if you're fine with that.

Yes that's how I understand too. This kind of processor have multiple power pins, and some are reserved to power the internal peripherals. For example you can power one CSI interface at +1.8V and an other one at +3.3V on the same CPU at the same time, depending what is the best for you. Also works for SD card interfaces, HDMI, etc etc.
On this module, I have understand that the CSI-VREF pin is for this purpose, but that there are options on the module with 0ohm resistors to connect directly to 3.3V (populated by default) or 1.8V. By removing the default populated resistor, it is possible to set CSI-VREF to the voltage user wants and this sets the voltage level of the CSI interface.
Input is fine, but by default, it will be left floating (thanks to the 0ohm resistor installed by default). In this case, I think Passive is the best isn't it ? Yes I definitely changed my point of view about this word :D :D

evanshultz:
Sorry. Pin 31 should be Open Collector. Typo. And it is, so that's good.

Hmm... pin 196 is tricky. Sure, Passive is fine then. It probably makes little difference in practice.

If you push those last couple things this can get merged.

myfreescalewebpage:
So lets go with that, I made the modifications right now :)

evanshultz:
Thanks! If nobody objects I'll merge this around 24 hours from now.

evanshultz:
@DanSGiesbrecht 
Is that OK with you? If everything is good with you, merge when you please.

myfreescalewebpage:
Thanks @evanshultz and @DanSGiesbrecht for the job you have done on this PR :)

DanSGiesbrecht:
@myfreescalewebpage no problem! Thank you to @evanshultz for taking this over in my absence. Glad to see that this got through!

PR #1196, opened 2018-12-02
Connector: create Mini-PCIe

myfreescalewebpage:
Creation of Mini-PCIe according to Mini-PCIe specifications (https://s3.amazonaws.com/fit-iot/download/facet-cards/documents/PCI_Express_miniCard_Electromechanical_specs_rev1.2.pdf).

Screenshot:

symbol

I have just pull requested the footprints for cards and sockets at https://github.com/KiCad/kicad-footprints/pull/1124.

The symbol have many GND pins. I have not put them at the bottom of the symbols, thus I'm currently violating S4.2. However, size of the symbol will be unacceptable at my point of view if all GND are of the bottom.

Thanks for review,
Joel

PS: the editor also fixed a typo in the symbol Bus_PCI_32bit_5V probably coming from previous Kicad version?


evanshultz:
Unless the ground pins must be unique pins for some reason, stack them and it won't take up so much room. I don't see any indication that the GND pins can't be stacked.

The changed line outside this symbol is because of a new way to note spaces in text strings in v5. No problem there.

Review notes:
- Pin stacking, as noted above.
- I think we should dispense with brevity and name this symbol `Bus_Mini_PCI_Express` or `Bus_PCI_Express_Mini`. I somewhat prefer the latter since it will group all PCIe symbols together.
- Is this document legal? Is it owned by some foundation or consortium that requires payment to access?
- Is there only a single footprint that can use this symbol? I would think no default footprint and a catch-all footprint filter would work best.
- Use overbars instead of pin types other than Line.
- Name pins 2, 24, and 52 `+3.3Vaux`. Same with the `+1.5V` pins. And all those pins are supplying power to the card, right?
- The UIM, if provided, is on the system board, right? That's the only way I can read page 49 and make sense of all the pin directions (including non-UIM pins). I'm not an expert on PCIe so I'm not sure, thus asking so I can understand.
- All the `Reserved` and `UIM_C*` pins should probably be Passive type. I don't see that Unspecified type is disallowed by KLC but I haven't seen it used before and Passive should be just as safe. Passive type is definitely used for a similar pins. Same with pins 3 and 5. Thoughts?
- Would grouping the pins as done in Table 3-1 be nice? Right now the symbol is designed as the physical bus is laid out, but that's explicitly not how KiCad symbols are designed.

myfreescalewebpage:
Thanks @evanshultz I try to respond your questions below before making some modifications depending what you will answer:

> Pin stacking, as noted above.

This is a real question, today the symbol reflect the position of the pins on the mini pcie connector. Sure this will eb easier to wire if pin are stacked, and moved to the right positions (GND on the bottom, ...) We agree that the position of the pins on the connector is not a priority on the symbol ?

> I think we should dispense with brevity and name this symbol Bus_Mini_PCI_Express or Bus_PCI_Express_Mini. I somewhat prefer the latter since it will group all PCIe symbols together.

Why not.

> Is this document legal? Is it owned by some foundation or consortium that requires payment to access?

It isn't but I haven't found the access to the official specifications. I think it is not free. I have crossed several sources to ensure the content is valid

> Is there only a single footprint that can use this symbol? I would think no default footprint and a catch-all footprint filter would work best.

No, see https://github.com/KiCad/kicad-footprints/pull/1124, agreed with your suggestion.

> Use overbars instead of pin types other than Line.

OK.

> All pins should be 100mil long.

OK

> Name pins 2, 24, and 52 +3.3Vaux. Same with the +1.5V pins. And all those pins are supplying power to the card, right?

They all provide supply to the card. I suppose you say to add "aux" because you refer to the doc page 39 ? In this case, no addition to +1V5 ? I have just written +3V3 and +1V5 to not overload the symbol but I can change, particularly because I will stack the pins so it will be easier to do it.

> The UIM, if provided, is on the system board, right? That's the only way I can read page 49 and make sense of all the pin directions (including non-UIM pins). I'm not an expert on PCIe so I'm not sure, thus asking so I can understand.

Your understanding is correct.

> All the Reserved and UIM_C* pins should probably be Passive type. I don't see that Unspecified type is disallowed by KLC but I haven't seen it used before and Passive should be just as safe. Passive type is definitely used for a similar pins. Same with pins 3 and 5. Thoughts?

Will do the way Kicad usually address this kind of pins, okay for Passive according to me. You have to know some mini pcie card manufacturers uses those pins, specified in their datasheets, for some extra functionalities they provide to the system board (like for reserved pins 45-51). In this case this is not Passive, but used for any purpose, Input, Output, Bidir, etc etc. However, probably it's better to follow the doc for UIM_C* and COEX* pins ?

> Would grouping the pins as done in Table 3-1 be nice? Right now the symbol is designed as the physical bus is laid out, but that's explicitly not how KiCad symbols are designed.

This is also what we written above about stacking, I agree if Kicad prefer to group by functionality, and this definitively simplify the design of the schematic (I'm working on a project with mini pcie and the wiring is horrible !)

myfreescalewebpage:
@aewallin just another comment after looking at the libs: PCI 32bits connectors in Kicad are not grouped by function/have no pins stacked. They are designed according to positions of the pins on the connector itself. However, I agree this is not convenient at all !

If other librarians can give their comments that should be very nice :)

Joel

evanshultz:
Perhaps I should have itemized things. I will do that above and below.

1. Yes, the symbol does not represent the physical part in the official KiCad library. The PCI symbols are very old and a bad reference. BTW, did you ping the right person in your comment above?
2. "Why not?" Because "PCI Express" is the full name as given in the document you linked and I also explained why I think the order of the terms above is nice.
3. We should not use a link of this nature, if my experience is a guide. Perhaps a link to the spec page (if there is a legal one) or just use a `~` for the datasheet field.
6. If the number of characters in the pin number is <100, as it is here, use 100mil long pins.
7. Let's name the power pins as in the spec.
9. Right. Passive will not cause any errors if these optional pins are used.
10. If you have a better grouping that's fine to propose, but if you like the Table 3-1 grouping use that.

poeschlr:
The symbol side should never care about where the pin really is on the part. So group by function and stack if appropriate.

myfreescalewebpage:
@evanshultz I have reworked the symbol according to your comment and vision of Rene. Let me known what you think about it. I have only keep the spec link like that because it is very useful and I haven't found any better documentation.

Cheers
Joel

myfreescalewebpage:
@evanshultz any returns ?

Cheers
Joel

myfreescalewebpage:
> This looks so much better to me. Basically what I was hoping to see. Nice!

Thanks :)

> I trust you didn't change pin numbers or types, which leaves on the power pins. Do they need to be separated? Since 3.3V and 1.5V are going into the connector (right?) it's unclear to me why each of the pins would have any special requirements. Is there specific capacitance on each power pin or some other reason they need to be separate?

The power pins are along the connector, they are not closed all together, so I would like to let the possibility to indicate there is several decoupling capacitors. MiniPCIe specification indicates modules can draw up to 2.7A. I'm currently working on such project and will probably add several caps for this reason.

> I'll hold off on merging also to go through the footprints.

Thanks!

Joel

evanshultz:
@poeschlr 
Can you weigh in on the doc link here and also with regards to pin stacking? I favor changing the link and stacking pins.

evanshultz:
Thanks Rene!

@myfreescalewebpage 
Can you see if there is other documentation freely available that includes the items mentioned above? It appears that's the only thing to wrap up before merging.

myfreescalewebpage:
Thanks @evanshultz and @poeschlr.

Well, that's really the best link I have found for this. It's possible to find various information but usually not at the same place (wikipedia + google image + forums...). The document can also be found on other websites, but only an older version.

Joel

myfreescalewebpage:
@evanshultz ping ? Thanks !

evanshultz:
I do not like an unofficial spec, and it's also an old one. Searching online and turns up a host of documents so it's not imperative that we have something here IMO. I would allow this one to have `~` for a datasheet (no doc but avoids Travis errors) and then merge this symbol.

myfreescalewebpage:
Ok, not a problem for me, removing the doc.

evanshultz:
Thanks! This should be good for now.

myfreescalewebpage:
Thanks @evanshultz !

Issue #381, opened 2018-03-14
What pin type for analog pins?


evanshultz:
@poeschlr @jkriege2 @Ratfink @Misca1234 
A few discussions recently have been about pin types. And I'm not clear what the KiCad library should do with analog pins (pins that do not generate or recieve a logic level or digital signal) which don't have an obvious function of input or output.

https://github.com/KiCad/kicad-symbols/pull/322 is one example regarding the MULTOUT pin of UC3854 (http://www.ti.com/lit/ds/slus336a/slus336a.pdf). This pin is an output of some internal circuitry and also the input to some internal circuitry. It does not send or receive signals to any other ICs.

It also only nominally connects to resistors or maybe a capacitor. (Of course some user may have a creative design and not adhere to the typical application of this IC.)

Is Bidirectional the right type for this pin? Information can be read on this pin using an oscilloscope, and the circuitry surrounding the IC does input a signal on the same pin. So there is some sense of bidirectionality. However, if Bidirectional pins in KiCad are meant to be used for logic level pins for the purposes of ERC, then this pin may best be set to Passive type.

What should the KiCad library require in this case? What pin type should be used for this, and similar, pins?

Ratfink:
I can see valid reasoning for calling it either Bidirectional or Passive.  I'm not sure which would be better, so I'll leave it to our more experienced librarians.

evanshultz:
The differences in KiCad should be captured by the ERC options:
![image](https://user-images.githubusercontent.com/1936989/37434078-a1c2f92c-279b-11e8-9a54-4ed0db909b40.png)

A pin of type Bidirectional generates a warning when connected to a pin of type Power Output or Open Emitter. A Passive pin will not generate any warnings in those cases. Otherwise they are the same to the ERC engine.

Otherwise, I think it's a philosophical question of how the librarians want pin types to be defined/used.

Shackmeister:
I think the passive type makes the most sense.

Its a slightly unlucky title when checking the emails on my phone :P
![11610cb88a98b33a61a4139afba823d5c9ab71f3](https://user-images.githubusercontent.com/5279723/37436133-b846e69a-27e6-11e8-9571-bf42aa4f1081.jpg)

PR #1197, opened 2018-12-02
Analog ADC: create ADS8684 and ADS8688

myfreescalewebpage:
Creation of TI ADC ADS8688 and ADS8644.

Datasheet : http://www.ti.com/lit/ds/symlink/ads8688.pdf

Screenshot:

ads8688

No error with checklib.py, using existing TSSOP38 footprint.


evanshultz:
- Since your symbols are individual, we need to make the datasheet description singular. Change `ADCs with Bipolar Input Ranges` to `ADC with Bipolar Input Range`.
- It might be nice to add `, SPI Interface` after that in the description, but you have `spi` in the keywords so it's not required.
- Add `, TSSOP-38` to the end of the description.
- Use only Line style for the pins (no bubbles). Put an overbar on the letters instead.
- I'd like to see SDO at the bottom of the SPI pins and DAISY at the top for easiest signal routing.
- Maybe scoot pin 2 closer to the SPI pins since it will go the host. Same with pin 4.
- Pin 5 isn't logic so it doesn't matter what pin type it may be connected to, and it may only connect to a capacitor, thus we've used Passive in the past for this type of pin.
- Pin 7 can be Passive type also.
- Pin 35 should be named DNC and push it in so the terminus is on the edge of the symbol body.
- AGND pins 28, 29, and 32 can be stacked. I would argue that page 58 (and elsewhere) is clear enough to stack all AGND pins; the description giving pin numbers is actually just telling the user to keep the analog decoupling paths as ideal as possible.
- These 2 parts need unique symbols since the `84 version has NC pins for inputs 4-7.

myfreescalewebpage:
@evanshultz thanks for the review, I agree with most of your points (I have lean a lot since this PR by reviewing and reading the KLC !) I will do the modifications asap.

Joel

myfreescalewebpage:
Comment by comment:

> Since your symbols are individual, we need to make the datasheet description singular. Change ADCs with Bipolar Input Ranges to ADC with Bipolar Input Range.

Agree. Was copy/paste from datasheet.

> It might be nice to add , SPI Interface after that in the description, but you have spi in the keywords so it's not required.

OK, no modification.

> Add , TSSOP-38 to the end of the description.

Done.

> Use only Line style for the pins (no bubbles). Put an overbar on the letters instead.

Done.

> All pins should be 100mil long.

Done.

> I'd like to see SDO at the bottom of the SPI pins and DAISY at the top for easiest signal routing.

Agree. New arrangement of the pins looking at datasheet page 35.

> Maybe scoot pin 2 closer to the SPI pins since it will go the host. Same with pin 4.

Agree. I keep an extra space because that's not the same block.

> Pin 5 isn't logic so it doesn't matter what pin type it may be connected to, and it may only connect to a capacitor, thus we've used Passive in the past for this type of pin.

> Pin 7 can be Passive type also.

Still the same behavior and still my point of view, you can't do anything you want with this pin so it can't be Passive.

> Pin 35 should be named DNC and push it in so the terminus is on the edge of the symbol body.

Agree, done.

> AGND pins 28, 29, and 32 can be stacked. I would argue that page 58 (and elsewhere) is clear enough to stack all AGND pins; the description giving pin numbers is actually just telling the user to keep the analog decoupling paths as ideal as possible.

Agree for pins 8, 28, 29, 31, 32.

> These 2 parts need unique symbols since the `84 version has NC pins for inputs 4-7.

Agree, splitting in two devices.

Cheers,
Joel

evanshultz:
Please change "Ranges" in the description to "Range". Also please update the screenshot in the first post or add fresh ones with all the visual changes done above.

I don't understand your objection to Passive type, which is what we've done elsewhere in the library. Especially for pin 7 since it only goes to a cap and should not really factor into any ERC checking. The ERC engine is extremely basic and making these pins Passive will avoid false positives (such as REFIO being an input coming from a linear regulator output which is of type Power Output). Input, Output, and Bidirectional types can catch a few issues with logic pins but it doesn't help with analog pins like this with the current capability of the KiCad ERC engine. I'm making suggestions so this PR conforms to the method that's been used on other recent symbols and thus will be consistent and require less discussion. https://github.com/KiCad/kicad-symbols/issues/381 and https://github.com/KiCad/kicad-website/issues/342 are a couple of discussions about pin types which would allow conversation about pin types to stay in one place and also avoid bogging down a PR with a bigger topic.

myfreescalewebpage:
@evanshultz I understand the reason and I agree it will be better for ERC checking and consistency with other devices if already used like that.
I still have doubt the term "Passive" is correctly chosen, but for the above reason I agree to change my symbol.
Joel

myfreescalewebpage:
@evanshultz additional comments ?

Cheers
Joel

evanshultz:
Here is the current screenshot of the "big" one:
image

The position of SDO and DAISY should be flipped for the easiest signal routing.

I suppose you could shove the power pins to the right and squeeze down the symbol if you wanted to make it shorter, but I'm happy how it is already.

myfreescalewebpage:
> The position of SDO and DAISY should be flipped for the easiest signal routing.

Depends where you place the first/last device of the chain. Here is an example that's works great with current symbol, with the first device at the top of the chain:

capture

Cheers,
Joel

evanshultz:
@myfreescalewebpage 
Sorry that I'd forgotten about this. It looks like this is ready to merge. Yes?

myfreescalewebpage:
@evanshultz no worries :) I have just checked again, on my side this is ready to merge.

evanshultz:
Sounds good. Thanks and sorry for the delay!

PR #1202, opened 2018-12-02
Add the LED5000 LED driver from ST

Doeme:
This adds the marvelous ST LED5000 chip symbol to the Driver_LED library
https://www.st.com/resource/en/datasheet/led5000.pdf

I'm not sure whether adding the thermal pad (pin 9) was the right thing to do, but it was included in the assigned footprint and the datasheet mentions that it should be connected to GND:

> Exposed pad to be connected to GND to increase the package thermal performance and the device noise immunity 

led5000

Edit: Addressing the travis-issues...
Edit2: Should all be fixed now


myfreescalewebpage:
Hi @Doeme , thanks for contributing,

- [ ] Description should ends with the package type, please add ", HSOP-8" at the end
- [ ] Capital letters for the keywords is not mandatory, would like to suggest "LED buck driver"
- [ ] You are currently violating S3.6 (see travis output), pin offset should be 20mil
- [ ] Footprint filter should be `HSOP*1EP*3.9x4.9mm*P1.27mm*`
- [ ] Length of the pins should be 100mil (symbol<100pins)
- [ ] I would say BOOT (pin 1) should be passive
- [ ] COMP (pin 4) should be moved to the left of the symbol, bottom is better for GND power supply only
- [ ] The exposed pad should be renamed GND (the datasheet recommends to do it) and stacked with the GND pin (same position, not visible, type Passive)
- [ ] The device seems very large, you can probably reduce it a little bit

Cheers,
Joel

Doeme:
Hi Joel.

Thanks for the Feedback. Force-pushed the changes (rebased onto master) now.

led5000

myfreescalewebpage:
Hi @Doeme thanks for the modifications, no more comment to do and merging.
Joel

PR #1204, opened 2018-12-03
Add AP7361C-33E and aliases

gcormier:
LDO linear regulator

https://www.diodes.com/assets/Datasheets/AP7361C.pdf

image

This was a copy of existing symbol AP2204R.

S3.1 - I think this should be exempt, it is slightly off, and aligns better with input/output. I looked at a dozen other symbols doing the same thing for linear regulators.
S4.2 - This matches other regulator footprints and as it's only function is power regulation, having the input power on the left side makes sense. Again, dozens of other symbols doing the same.


gcormier:
Also think 10mil offset is better, and is also consistent with other linear regs. Keeps the symbol tight.

image

Can revert to this if needed.

myfreescalewebpage:

- [ ] It's better to use an existing symbol and add wanted aliases. This reduce library size and maintenance (even if it is not always done!) Please use SPX2920M3-3.3_SOT223 for example, which has the same package and pinout and add the aliases.
- [ ] Your keywords are good but the description is not correct. For consistency with the other regulators in the library, use something like `1A Low Dropout regulator, positive, 1.5V fixed output, SOT-223` (this is an example, to be modified for each part).

Cheers,
Joel

PR #1205, opened 2018-12-03
Add MAX40200

DanSGiesbrecht:
Adding MAX40200ANS, MAX40200AUK.

- **[Datasheet](https://datasheets.maximintegrated.com/en/ds/MAX40200.pdf)**
- **[Footprint PR - ANS variant only](https://github.com/KiCad/kicad-footprints/pull/1129)**
- **[Script PR](https://github.com/pointhi/kicad-footprint-generator/pull/231)**

#### MAX40200ANS
wlp

#### MAX40200AUK
image

------------


DanSGiesbrecht:
@antoniovazquezblanco for my own knowledge, what gets added to the backlog milestone? Should I be applying it to PRs as well?

Analog_Switch.dcm
```diff
@@ -377,14 +377,14 @@ K Video Multiplexer analog switch
...
+D Above- and Below-the-Rails Low On-Resistance Analog Switches, max 2 Ohms On-Resistance, TDFN-8
 K analog switch low resistance above below rails
 F https://datasheets.maximintegrated.com/en/ds/MAX14759-MAX14763.pdf
 $ENDCMP
 #
-$CMP MAX14759
-D Above- and Below-the-Rails Low On-Resistance Analog Switches, max 2 Ohms On-Resistance, TDFN-8
+$CMP MAX14761
+D Above- and Below-the-Rails Low On-Resistance Analog Switches, max 2 Ohm On resistor, TDFN-10
 K analog switch low resistance above below rails
```
DanSGiesbrecht: Note: lines 380-381 and 386-387 got reordered by KiCad, to be in order.

evanshultz:
AUK doesn't need the pincount in the FP filter, so just `SOT?23*`.

Otherwise I think this can be merged, but I gotta ask why round the WLP package size? The drawing shows 0.728mm side lengths so why use 0.73mm  in the footprint name? Especially because the footprint sides are actually 0.728mm long. Is there a rounding guideline I've missed?

DanSGiesbrecht:
Good catch, thanks! The footprint filter has been fixed.

I rounded the WLP package size because I haven't noticed other parts with more than two places after the decimal, and because the datasheet also rounds the package dimensions in the description on the cover page:

datasheet-cover

Is there a maximum acceptable position after the decimal? Should this footprint be changed?

evanshultz:
Thanks! I'll merge then.

We have had higher precision, but I can only recall it on BGA and CSP footprints. For example, see https://github.com/KiCad/kicad-footprints/blob/master/Package_BGA.pretty/Texas_DSBGA-5_0.822x1.116mm_Layout2x1x2_P0.4mm.kicad_mod. This has been unregulated in the past and changing it now would be tough. While it doesn't specifically mention a case where more than two sig digs past the decimal are required to capture the actual dimension, IPC 7351's naming convention says this:
- All dimensions are in Metric Units
- All Lead Span and Height numbers go two places past the decimal point and “include” trailing Zeros
- All Lead Span and Body Sizes go two place before the decimal point and “remove” leading Zeros
- All Chip Component Body Sizes are one place to each side of the decimal point
- Pitch Values are two places to the right & left of decimal point with no leading Zeros but include trailing zeros

PR #1216, opened 2018-12-04
add Mini-Circuits LRPS-2-1 power splitter/combiner

aewallin:
https://ww2.minicircuits.com/pdfs/LRPS-2-1.pdf

lrps-2-1

Footprint: https://github.com/KiCad/kicad-footprints/pull/1135

------------


myfreescalewebpage:

- [ ] We usually put reference in top left corner and name in top right corner - Actually you have reference at a strange position which not very convenient...
- [ ] The footprint is not correct, you should update it (but footprint filter don't need modification)
- [ ] You should add NC pins 2 and 5 (Type Not Connected, not visible) because the pin count is not specified in the footprint (KLC S4.5 http://kicad-pcb.org/libraries/klc/S4.5/)
- [ ] You can center GND pins, but maybe you want consistency with other MiniCircuit symbols ?
- [ ] Type of pin 6 is Output (in my understanding of the device)

Cheers,
Joel

aewallin:
centered GND-pins would just obscure readability of symbol and port-pin IMO.
after revision:
lrps-2-1

myfreescalewebpage:
@aewallin, missing the impedance and you are violating S4.1 with the NC pins.

Joel

aewallin:
same question as elsewhere: if hidden NC-pins should be on the box-boundary, do I have to change the box to be on 100mil grid?

myfreescalewebpage:
Right, but this will violate S4.6 which absolutely requires that the NC pins must terminates on the outline box of the symbol. So I prefer violating S4.1 and keep your symbol like that.

No other comment, merging.

Joel

Issue #1004, opened 2018-09-25
Refernece Designator for TL431DBZ


leoheck:
I believe that the reference designator for the `Reference_Voltage:TL431DBZ` should be `U` because it is an IC. Right know it has the diode `D` designator.

evanshultz:
This goes for all symbols in this library with a D refdes prefix now:
- CJ432
- LM285*
- LM403*
- LM404*
- TL43*

antoniovazquezblanco:
PR with fix sent in https://github.com/KiCad/kicad-symbols/pull/1220

PR #1220, opened 2018-12-05
Reference_Voltage: Fix RefDes in multiple devices

antoniovazquezblanco:
Fix https://github.com/KiCad/kicad-symbols/issues/1004


antoniovazquezblanco:
Thanks!

PR #1222, opened 2018-12-05
Add ATmega406

fauxpark:
This one's got a little bit more going on. Not 100% sure about the pin types and positioning.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-2548-8-bit-AVR-Microcontroller-Battery-Management-ATmega406_Datasheet.pdf

------------


evanshultz:
- Pin 30 should be Open Collector type from the top of page 6.
- Pin 23 should be Input type.
- Do you think pin 34 should be Power Input type and on the top? Is the primary function a power input for the regulator (I think so) or a voltage level monitor?
- Can a cap fit between pins 43 and 44?
- Perhaps a pin layout like page 16 of http://ww1.microchip.com/downloads/en/AppNotes/doc2598.pdf, but with only power pins on top and bottom and other stuff to match KLC, would work well? It's hard to know what really might work without seeing the symbol, but that seems like a good starting point.

fauxpark:
- The block diagram claims PC0 is bidirectional, but I think you are probably right there
 - Changed SCL to input
 - Moved VFET to the top and changed to power input
 - Moved VREFGND down 100mil, hopefully that should be enough

evanshultz:
Thanks!

What about other pin placements like in the app note? I thought the FET drives on top left and battery pins below it, with communication and GPIO on the left would work nicely.

fauxpark:
How does this look?

evanshultz:
That looks nice to me. A couple of ideas:
1. Flip the GPIO so it's A -> D, top to bottom.
1. Spread apart VREF and VREFGND to fit a cap symbol there.
1. Perhaps I2C and RESET grouped together, if anything, without the XTAL in between.
1. Can pins 38-42 be spaced apart, consuming dead space, for cap symbols to fit there?

None of this has to be done, just thinking out loud.

fauxpark:
- The reverse ordering is consistent with both the app note and the rest of the ATmega symbols (surprisingly). I think it should stay like this
 - Done
 - I moved RESET above the XTAL pins but didn't group it with the I2C pins
 - I can't space them all out evenly but I did move PI/NI away from each other:

evanshultz:
Alright then. Thanks for the time to experiment with pin placement. If you push this I'll merge. :)

Issue #957, opened 2018-09-18
LED: S5.1 and S5.2 errors


antoniovazquezblanco:
KLC Rules:
http://kicad-pcb.org/libraries/klc/S5.1/
http://kicad-pcb.org/libraries/klc/S5.2/

Splitting issue #28 I have found the following violations:

```
Checking symbol 'HLCP-J100':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'HDSP?48*' does not match the footprint 'Display:HLCP-J100' set for this symbol.
     - could not match 'HLCP-J100' against filter 'HDSP?48*'
     - could not match 'Display:HLCP-J100' against filter 'HDSP?48*'
checking ruleS5.2
checking ruleS5.1
Checking symbol 'HLCP-J100_2':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'HDSP?48*' does not match the footprint 'Display:HLCP-J100' set for this symbol.
     - could not match 'HLCP-J100' against filter 'HDSP?48*'
     - could not match 'Display:HLCP-J100' against filter 'HDSP?48*'
```

PR #1228, opened 2018-12-06
#957 HLCP-J100 footprint filter

aewallin:
simple fix for #957 
footprints do seem to contain many copies of the same footprint with different names...

------------


myfreescalewebpage:

Many modifications by the editor but it's okay.
Footprint modified, ok for me.

@antoniovazquezblanco can be merged according to me, and #957 can be closed.

Cheers,
Joel

antoniovazquezblanco:
@aewallin and @myfreescalewebpage  you seem to be contributing a lot lately so I will try to give you my best pointers whenever I can.

In this case, it would be awesome if you were able to split the library update and the changes made into two different commits for reviewing. The way I usually do that is: I open any component in the lib, I open the properties dialog of the component and click OK without making any changes and it will allow me to save the lib with the new format. I will then commit that and proceed to make the changes. I know it is much more tedious but makes it easier later down the review process.

Now, about the changes. You have used `*HLCP?J100*` as the new filter. I don't usually prepend an `*` before the filter. Is there any reason to do that?

Thank you very much!

myfreescalewebpage:
@antoniovazquezblanco 

> you seem to be contributing a lot lately so I will try to give you my best pointers whenever I can.

Yes I have discussed with other librarians to propose my reviews. Many PRs pending and I think it's good to help when it's possible. I  (but ma,ny I'm try to practice with simple PRs actually.

> In this case, it would be awesome if you were able to split the library update and the changes made into two different commits for reviewing.

I agree with the method, that's always better but we usually see the "bad effects" of the edition after when making the commit so it's too late... If you are not making any change, does it seriously modify the library to the new format ??

> Now, about the changes. You have used *HLCP?J100* as the new filter. I don't usually prepend an * before the filter. Is there any reason to do that?

I have seen with and without in the libraries and I do not know what is the best practice for that, I have not seen any KLC about this.

antoniovazquezblanco:
> If you are not making any change, does it seriously modify the library to the new format ??

Yes. Not a bug, it's a feature :rofl:.

> > Now, about the changes. You have used HLCP?J100 as the new filter. I don't usually prepend an * before the filter. Is there any reason to do that?

> I have seen with and without in the libraries and I do not know what is the best practice for that, I have not seen any KLC about this.

I think this could be related to ancient behaviour when library name was also matched. This is no longer like that but this should be confirmed by more experienced librarians like @evanshultz or @poeschlr .

Thanks

myfreescalewebpage:
> I think this could be related to ancient behaviour when library name was also matched. This is no longer like that but this should be confirmed by more experienced librarians like @evanshultz or @poeschlr .

And conclude with a KLC and a travis test, because this will be redundant question in the future !

aewallin:
Thanks for the guidance.

I agree that ideally the KLC-scripts should check validity of the fp-filter.

myfreescalewebpage:
@antoniovazquezblanco can be merged now following modification of the footprint, and #957 can be closed.

Cheers,
Joel

poeschlr:
Having a * at the beginning can sometimes still be useful. Especially if one might expect fitting footprints to start with a different prefix. 

This was a bit more common in the past especially for things like chip packages. There are for example qfn, vqfn, tqfn and pqfn which are very similar or even equivalent and (seemingly) only named differently because some marketing guy wanted their company to be different. 
Having a * at the beginning allowed us to reuse not quite fitting footprints where we did not have the exactly correct footprint. 
Now that we have scripts for these footprints we can be a bit more strict about that. (We now require the correct footprint to be made and pre assigned.)

myfreescalewebpage:
@poeschlr thanks for clarification

antoniovazquezblanco:
Thank you!

evanshultz:
The leading prefix was also requested by the devs for a time before v5 was released when the footprint filtering behavior was being sorted out. I don't know if these footprints were affected.

VQFN, TQFN, QFN, etc. are often unique in package height. T = thin, V = very thin, etc. So the 3D model is unique even if the footprint is the same. And the 3D models are scripted too, making it easier to be strict, but at the cost of more footprints and 3D models which are very similar.

In this case, why would there be a leading asterisk? Not that I can see it really matters, but the "strictness" would seem to apply here since a display won't have variation like the QFNs I mentioned above. Also, I only recall seeing a question mark used where we know allowing a march against multiple characters will result in non-matching footprints being displayed in CvPcb. For example `SOT*23` will show SOT-23 and SOT-223 so we can use `SOT?23` for a SOT-23 footprint filter but still use `SOT*223` for a SOT-223 package. This is a colloquial rule to be as flexible as possible (so using asterisk wherever possible) but perhaps it's time for a formal rule?

aewallin:
have the travis-scripts print out a list of matched footprints (or the 10 first, if there are more than 10). at least that allows for reasonably fast manual review of the filter results.
then have the KLC dictate wether, and to what extent, false positives and false negatives should be allowed.
too much (manually checked) syntax-rules on the actual filter-string just gives me headache..

PR #1229, opened 2018-12-06
Add Symbols for TJA1043T and TJA1043TK CAN FD Transceivers

HubertD:
Add support for current NXP CAN-FD transceivers with Sleep Mode.
Datasheet link: https://www.nxp.com/docs/en/data-sheet/TJA1043.pdf

Since there seem to be no CAN transceivers with sleep capability yet in this library, I had to make some choices regarding pin order etc.

My rationale was as follows:
- Put all pins that interact with the CAN Controller / Microcontroller to the left
- Put external CAN part (CAN_H, Split, CAN_L) to the top right
- Group the power control / sleep-mode related pins (Inhibit, Wake) to the bottom right.

Would be glad to hear better ideas :-)

tja1043t
tja1043tk

------------


myfreescalewebpage:
Hi @HubertD, thanks for contributing,

- [ ] VIO is a Power Input and should be moved to the top of the symbol according to KLC S4.2
- [ ] WAKE is an Input and should be moved to the left of the symbol according to KLC S4.2
- [ ] According to the datasheet, pin 8 should be called ERR_N instead of ERR
- [ ] ERR_N is an Output and should be moved to the right of the symbol according to KLC S4.2
- [ ] According to the datasheet, pin 14 should be called STB_N instead of STB
- [ ] I would also suggest to move pin 15 of the DFN symbol 100mil on the right, so the symbol is vertically symmetrical (also move a little bit the name to achieve this)

Cheers,
Joel

myfreescalewebpage:
@HubertD up ?

HubertD:
Thank you @myfreescalewebpage ,
I agree to most of your points, and modified the symbol accordingly:

tja1043t
tja1043tk

The ERR pin, however, should be grouped with the enable / standby pin, since these three are used to control the transceiver state from a microcontroller, see e.g. the state diagram in the datasheet.
I would argue that KLC 4.2 (1) should overrule KLC 4.2 (6) here.

myfreescalewebpage:
@HubertD  thanks for the fixes, looks great !

Agreed with your comment about ERR.

Two last comments for me:
- [ ] According to the datasheet, INH should be open emitter instead of open collector
- [ ] Can you move the name of the device in the top right corner, for consistency with other devices, see for example TJA1042 or TJA1049 ?

Thanks for your contribution,
Cheers,
Joel

myfreescalewebpage:
@HubertD ping! Are you still alive on this subject ?

HubertD:
sorry, I'm not able to keep up with my personal projects atm.
Will make these changes now.

HubertD:
tja1043tk
tja1043t

myfreescalewebpage:
Thanks @HubertD for the fixes !
I'm wondering about the size of the device, I have just looked at TJA1049T in the lib, the width is smaller and the height too.
Can you reduce at least the width of the device to look like the TJA1049T ?
Thanks,
Joel

HubertD:
tja1043tk
tja1043t

HubertD:
tja1043tk
tja1043t

myfreescalewebpage:
Well done @HubertD no more comment to on this PR, thanks for the job you have accomplish here ! Merging.

PR #1087, opened 2018-10-22
Add LTC6994-2

ferdymercury:
https://www.analog.com/media/en/technical-documentation/data-sheets/699412fb.pdf

Use with Package_DIP:DIP-6_W7.62mm_LongPads

------------

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the official footprint library
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1231, opened 2018-12-07
add LTC6993 pulse generator

aewallin:
https://www.analog.com/media/en/technical-documentation/data-sheets/69931234fc.pdf

questions:
- This comes in four variants -1, -2, -3, -4. do we want a symbol-variant for each? The pins and footprints are the same, just slightly different functionality.
- There are two footprint options, I would prefer atomic symbols, i.e. a symbol or variant for TSOT-23 and separate for DFN - but what is the guideline for KiCad libraries?
- depending on the choices above there would be either one symbol or eight symbols in the library for this part!

ltc6993

------------


aewallin:
possible variant is also LTC6994
https://www.analog.com/media/en/technical-documentation/data-sheets/699412fb.pdf
(pin1 is called IN there, while it is called TRIG here..)

aewallin:
revised screenshot
ltc6993

diegoherranz:
> This comes in four variants -1, -2, -3, -4. do we want a symbol-variant for each? The pins and footprints are the same, just slightly different functionality.

See http://kicad-pcb.org/libraries/klc/S2.2/ where we recommend not using variants for non-functional variations (e.g. temperature). However, in this case they are functional variations, so I think we should have aliases for them. I think it is similar case to voltage regulators with a few pre-defined voltage options, and for those, we normally use aliases. The symbol could be called `LTC6993-1` and the other 3 as aliases of that one. Could you arrange it that way?

> There are two footprint options, I would prefer atomic symbols, i.e. a symbol or variant for TSOT-23 and separate for DFN - but what is the guideline for KiCad libraries?

That is OK. See http://kicad-pcb.org/libraries/klc/S2.3/

> depending on the choices above there would be either one symbol or eight symbols in the library for this part!

Actually it would simply be 2 symbols (1 per footprint) and a few aliases :)

> possible variant is also LTC6994
https://www.analog.com/media/en/technical-documentation/data-sheets/699412fb.pdf
(pin1 is called IN there, while it is called TRIG here..)

Hmm, they are very similar, but actually, they serve to different purposes, and hence the different pin name. I think they should be different symbols unless we called that pin `IN/TRIG`? Librarians?

Thanks!

aewallin:
re-named to LTC6993-1 and added the others as variants -2,-3,-4. Description-text has falling/rising-edge and non-retriggerable/retriggerable descriptions.

diegoherranz:
Thanks for the changes which look OK.
A few more comments:
- Can you add the footprint to the end of the descriptions? See http://kicad-pcb.org/libraries/klc/S6.3/
- On the footprint filter, I think that the initial `*` should be removed unless there is a specific reason for it?
- "Filters should not contain the pin count if the pin-count in the symbol matches the pin count in the footprint." (http://kicad-pcb.org/libraries/klc/S5.2/). Could you remove the `6` from it? 
- I think that the disposition of the `DIV` pin on #1087 can work better since it is necessary to connect a cap from `DIV` to `V+`. Could you do the same?

> LTC6994 (#1087) has two footprint-options, so maybe it can be another symbol?

Each footrprint needs a different symbol. Given the LTC6994 has a different pin name, I'm more inclined now to keep them separated. It'd be convenient, though, to make both symbols as similar as possible.

Thanks!

aewallin:
Pins and look is now similar to #1087:
ltc6993-1

diegoherranz:
- Thanks for the rearrange. I think it looks OK.
- I've seen that you have changed the SET type to "passive". I think that probably better than input in this case. So I think it is OK.
- This is a picky one, but, can you separate the footprint from the rest of the description with a comma instead of a dot? ("Description field contains comma-separated device information" on http://kicad-pcb.org/libraries/klc/S6.3/)
- Footprint filter looks OK.

Thanks!

diegoherranz:
Thanks for all the changes. In my opinion this is ready to be merged.

myfreescalewebpage:
Thanks @diegoherranz for the job, really great.

An additional comment: the device also exists in DCB package, the parts names must include something specific to the package. I suggest renaming the parts to `LTC6993xS6-1`, `LTC6993xS6-2`, `LTC6993xS6-3`, `LTC6993xS6-4` according to ordering information pages 2/3 of the datasheet.

Cheers,
Joel

diegoherranz:
How did I miss the other package??!!!

I agree with the suggested names given by @myfreescalewebpage, could you @aewallin fix the part numbers?

Thanks!

diegoherranz:
Thanks, @aewallin, for all the changes. It looks good to me.

myfreescalewebpage:
For me too, thanks you both for the job ! Merging the PR.

PR #1232, opened 2018-12-07
LVCMOS 1:8 clock-buffer IDT 5PB1108PGG in TSSOP-16

aewallin:
https://www.idt.com/document/dst/5pb11xx-datasheet

(there are 5 different fan-out variants of this, in two different packages - a candidate for scripting perhaps?)

idt5pb1108pgg

------------


aewallin:
After edits, it looks like this.
GND pins could maybe be combined into one?
Vdd pins should not be, probably want separate decoupling caps on each Vdd
idt5pb1108pgg

aewallin:
stacked and hidden GND pins (only pin4 visible)
idt5pb1108pgg

myfreescalewebpage:

- [ ] Footprint filter should be `TSSOP*4.4x5mm*P0.65mm*`
- [ ] Maybe you can move CLKIN 100mil to the top so CLKIN and 1G have an extra space between them

Cheers,
Joel

aewallin:
changes made as suggested:
idt5pb1108pgg

myfreescalewebpage:
Thanks,

@antoniovazquezblanco this PR can be merged according to me.

Cheers,
Joel

poeschlr:
Might i suggest either to remove the manufacturer name from the symbol name or at least separate it with an underscore from the part number.

The second G denotes Pb free. So either do not include that or wildcard it. If you wildcard it i would also suggest to add a second x for wildcarding the temperature range.

myfreescalewebpage:
@poeschlr and @aewallin agreed with the manufacturer name.

@poeschlr However, I do not found at all "5PB1108PG" only. octoparts don't know it and the page 2 of the datasheet only shows "5PB1108PGG". Where is the information from ?

poeschlr:
Under the list of part numbers (page 20) it tells us that the "G" after the package code (PG) stands for Pb free (=ROHS complaint). 
You will also not find the part number used right now as it is missing either an I or an K for the temperature range and maybe an 8 for tube packaging. Meaning we should use something like `5PB1108PGxx` as the symbol name.

myfreescalewebpage:
@poeschlr agreed with `5PB1108PGxx`. Thanks to point me out this point. I will pay more attention to this kind of details in the future.

poeschlr:
Thanks @aewallin for your contribution and thanks @myfreescalewebpage for the review.

aewallin:
- description: OK - hopefully these checks can be automated so that there is immediate feedback from travis on simple things. package and pin-count could in principle be automated to some extent?
- pin length seems to comply with S4.1. I don't understand your 20mil suggestion.
- VDD pins have individual decoupling-caps for each pin, following good design practice for this kind of device. The separate pins remind the designer about this. Stacking is possible but then you have to anticipate the number of real VDD pins when drawing the schematic. The other parts in this family have between 1 and 4 pcs VDD pins.

Where do we go from here, can I push more to my branch, although this is now merged into master? Or do I create a new PR?

antoniovazquezblanco:
> Where do we go from here, can I push more to my branch, although this is now merged into master? Or do I create a new PR?

I am not sure if we can merge if you push to the same branch. It would be better to open a new PR pushing the fix. 

Thanks!

aewallin:
> I am not sure if we can merge if you push to the same branch. It would be better to open a new PR pushing the fix.
> 
> Thanks!

let's continue there then: #1273

PR #1238, opened 2018-12-08
Add THS3491 op-amp in SOIC-8-1EP

aewallin:
http://www.ti.com/lit/ds/symlink/ths3491.pdf

Op-amp with standard pins and additionally:
- exposed pad
- PD
- ref (reference for PD pin)
also in VQFN package, but this symbol only for soic-8-1ep

ths3491idda

------------


diegoherranz:
Thanks for submitting this symbol.
A few comments:
- Regarding the exposed pad, the datasheet says the following which makes me think that the electrical type should be "Passive" and not "Power input". Could you change it?

> "The thermal pad can be connected to a heat spreading plane that can be at any voltage because the device die is electrically isolated from this metal plate. The thermal pad can also be unused (not connected to any heat spreading plane or voltage) giving higher thermal impedance."

- The dimensions of the exposed pad on the selected footprint are quite different from the recommended one on the footprint. Should a new footprint be created? Or you considered that it shouldn't be a problem?

- "opamp" is not mentioned on either the description or the keywords which may make it hard to find. Could you add "single" and "opamp" to the keywords?

Thanks!

aewallin:
Good question about the EP layout.
The datasheet has two figures, page 36/52 has the copper-layer+vias, and page 52/53 has the soldermask. I don't understand the via-diameter spec which seems to be 0.254mm vs. 0.33mm in the drawings?
The soldermask opening/pad suggestion is 2.40 mm x 3.10 mm on the datasheet.
The footprint `SOIC-8-1EP_3.9x4.9mm_P1.27mm_EP2.35x2.35mm` has a 2.35mm x 2.35mm pad which is probably OK?

I would personally use the symbol with the existing footprint although it doesn't conform to the datasheet recommendations exactly.
I can create the new footprint and submit a PR if Librarians think a new footprint is required before the symbol can be merged. If there are many more symbols that also specify "TI DDA PowerPAD" then this makes sense.

ths_layout

ths3491_cu

aewallin:
datasheet suggests connecting pin5 to GND. However it is an NC pin.
Maybe it should be included in the symbol, named NC, so that we have the option to follow the datasheet suggestion?

aewallin:
pin5 added. EP/NC names made as small as possible, 20mils.
ths3491idda

myfreescalewebpage:

- [ ] Part name should be `THS3491xDDA`, the I is for industrial temperature and doesn't matter in the lib
- [ ] Footprint to be proposed as a footprint PR, and to be update in the symbol definition
- [ ] Description should ends with package name (according to previous modification)
- [ ] Pin 5 should be Passive

Cheers,
Joel

myfreescalewebpage:
Hello @aewallin any news here ? Thanks for the update, Joel

aewallin:
sorry, not working on this part actively - the datasheet seems to be ambiguous on the footprint, and I am not familiar with the scripts used for SO-footprints.
If the right footprint appears the changes needed here to the symbol are minor I guess - and not hard to do.

myfreescalewebpage:
> If the right footprint appears the changes needed here to the symbol are minor I guess - and not hard to do.

Yes sure!

aewallin:
footprint PR is now https://github.com/KiCad/kicad-footprints/pull/1541

aewallin:
remaining KLC-error is about the NC-pin:
"NC pins are not correct pin-type:   - Pin NC (5) @ (100,300) should be of type NOT CONNECTED, but is of type PASSIVE"

not sure what is best? 
- leave it visible, named NC, type passive
- alternatively invisble named NC and type not-connected.

myfreescalewebpage:
@aewallin 
To answer your last question, the best option is "invisble named NC and type not-connected."

One comment during a new review here:
- [ ] Footprint filter should be `SOIC*1EP*3.9x4.9mm*P1.27mm*`

Cheers,
Joel

evanshultz:
The footprint exists already:
image

myfreescalewebpage:
@aewallin seems people have created footprints for you :) I agree it seems to be the right one, can you confirm too and in this case update the symbol ?
Thanks @evanshultz for this !
Joel

aewallin:
not sure about the fp-filter, now that footprint switched to the "Texas" one...

myfreescalewebpage:
Looks great ! Thanks @aewallin . Merging :)

PR #1247, opened 2018-12-10
Valve: Update to latest file format

antoniovazquezblanco:
See https://github.com/KiCad/kicad-symbols/pull/1181


evanshultz:
OK. Looks like all is well and the CI errors aren't affected by this.

evanshultz:
Does this break any outstanding PRs? Do we have a good way to determine that?

antoniovazquezblanco:
I have looked for valve related prs using GH search function and found nothing.

evanshultz:
Alright then. Thanks!

FWIW, doing more than one lib at a time is fine but taking care of libs with no pending PRs, if there's some way to determine that, is a good place to start.

antoniovazquezblanco:
Thanks!

Issue #1249, opened 2018-12-11
ESP32 Symbol


gcormier:
GPIO 34/35 are INPUT in the current symbol, but should be BIDIRECTIONAL as they are GPIO.

https://www.espressif.com/sites/default/files/documentation/esp32-wroom-32d_esp32-wroom-32u_datasheet_en.pdf

![image](https://user-images.githubusercontent.com/8388983/49770477-8cb74780-fcb2-11e8-85b5-89964f81c89d.png)

![image](https://user-images.githubusercontent.com/8388983/49770460-79a47780-fcb2-11e8-95f3-5ae44885f155.png)
![image](https://user-images.githubusercontent.com/8388983/49770464-7e692b80-fcb2-11e8-8931-2a7f30bb05cb.png)

DanSGiesbrecht:
When I merged this, I defaulted to the datasheet listing those pins as inputs:
![datasheet](https://user-images.githubusercontent.com/7242621/49771505-a5b9fb80-fca6-11e8-8ca9-1539f5469c26.png)

I agree that they appear to be GPIO (bidirectional). If this is a datasheet error, I think that we should change the pins to bidirectional. However, I'm not sure if it's a mistake on Espressif's part, or not.

DanSGiesbrecht:
Not a problem! Glad we got it sorted out.

PR #1250, opened 2018-12-11
Fix two pins on ESP32-WROOM-32

gcormier:
Please see #1249.


gcormier:
Bah, messy. Will fix.

PR #1251, opened 2018-12-11
ESP32-WROOM-32 - Change two pins to BIDIRECTIONAL

gcormier:
Please see #1249.


DanSGiesbrecht:
Hi @gcormier . Why did you create a new PR, rather than pushing to #1249?

gcormier:
To be honest, I wasn't aware you could push to an issue on github...

DanSGiesbrecht:
@gcormier I don't think you can, sorry. That was my mistake. I thought that your issue was another PR...

PR #1252, opened 2018-12-11
Photomultiplier Tube library

ferdymercury:
Here the first schematic symbol of a PMT. It follows this datasheet, but it is made abstract, so not atomic:
https://www.hamamatsu.com/resources/pdf/etd/R9800_TPMH1298E.pdf

------------


ferdymercury:
image

myfreescalewebpage:
Hi @ferdymercury , thanks for contributing,

Librarians, not sure this to be added in a new library ? According to me it's just to be added in Sensor_Optical. What do you think ?

@ferdymercury you made a generic symbol, but I think it's better to make an atomic part (called R9800), isn't it ? Are all photo-multipliers the same number of pads etc ? Also, I think the design should be inspired of the devices in the Valve library.

Cheers,
Joel

ferdymercury:
There are many photomultiplier of different brands with 8 pins, thats why I preferred not to do an atomic part.

Because this sensor is kind of special and big, I thought it could go into a special library. But if librarians prefer, it could be moved to Sensor Optical lib

myfreescalewebpage:
Just want to say that the physical size of the part is not a criteria to add it in a new library. Librarians, your opinion ?

ferdymercury:
Fixed according to your comments, thanks.

myfreescalewebpage:
@ferdymercury thanks!

Additional comments:
- [ ] Remove the question mark in the footprint filters, `*`is enough : `PMT*25.4mm*` and `PMT*50.8mm*`
- [ ] Move reference in top left corner

I'm still in doubt regarding to generic/atomic decision. The FP filter added let thinking it should be atomic... To any librarians (@poeschlr ?) what do you think about that ?

Cheers,
Joel

ferdymercury:
Fixed both things, thanks.

There are several 8-stage PMTs, some with a 1 inch diameter, some with 2 inch diameter, from different vendors, etc.

myfreescalewebpage:
@ferdymercury Thanks, no other comment on my side, just let other librarians have a look regarding my previous question:

> I'm still in doubt regarding to generic/atomic decision. The FP filter added let thinking it should be atomic... To any librarians (@poeschlr ?) what do you think about that ?

Cheers,
Joel

ferdymercury:
I also have the question if we should incorporate the electrostatic shielding in this symbol, or as an external symbol? Or as a shielded variant of this one?

image

And what about the Valve library consistency, should I remove the filled background?

ferdymercury:
To add some info to the discussion, this was the symbol style many years ago (I find it cool)

image
image

myfreescalewebpage:
@ferdymercury I wasn't aware of this kind of device but the photos are nice and very helpful to progress. I will do the review asap.

Joel

myfreescalewebpage:
@ferdymercury 

I see two possibilities today:
- You want to keep it generic symbol, I have no comment on it, it is okay for me and can be merged
- You plan to submit some footprints for this kind of symbols, in this case pad numbers are important and must match the datasheets, and in this case it should be modified and duplicated for each part with correct pin numbering, datasheet lin, footprint, etc

Please let me know your option.
Joel

ferdymercury:
I do not plan to submit footprints of this symbol in the near future.

ferdymercury:
> I also have the question if we should incorporate the electrostatic shielding in this symbol, or as an external symbol? Or as a shielded variant of this one?
> 
> image
> 
> And what about the Valve library consistency, should I remove the filled background?

What's your opinion about the optional shield?

myfreescalewebpage:
You can have the two versions, shielded and not shielded. However, the shielded version must have Cathode+Dynode+Anode+Shield. Because it's only device, it is not possible to accept a separated shielding in another device.

ferdymercury:
Added shielded version, thanks for the review.

diegoherranz:
If we eventually start having many PMTs, it may make sense to have a separate library. In the meantime, I guess it's OK to have it in Sensor_Optical.
Thanks!

ferdymercury:
fixed both, thanks for the review.

Yes, the magnetic pin can be wired, see http://www.fusor.net/board/viewtopic.php?t=6185#p41029

myfreescalewebpage:
> the magnetic pin can be wired

Sure, I was just surprised by the previous design of the design.
Current cone is okay for me, no more questions or comment, merging.
Thanks for your job !

Joel

zegarac:
Ehh i need PMT component symbol to document something (R374 with 11 dynodes version) i have made with it, and it is not submitted (don't see it in standard lib), so ill have to draw my own 8-/ Cant believe you have MPPC's (very recent parts) and not PMT's which are being used for many decades now...

evanshultz:
I'm obviously late to the party here, but just saw this PR when it was re-surfaced...

A few notes, and bear in mind this type of component is outside my field of expertise:
1. We use `_Shielded` as a suffix for other part types with a shield. Like connectors.
1. Where there are internal lines connecting, a 10mil dot should be placed.
1. There are text and lines crashing in the shielded version.
1. The symbol is huge.
1. Is there a reason not to capitalize all pin names?
1. Pin lengths are longer than required. Usually the pin will only reach the boundary of the symbol so the pin name, if shown, is centered on the line.
1. There are lines behind the pin lines on all 8 dynode pins, and they're not even consistently placed.
1. If the symbol design makes a pin name redundant, then the pin doesn't need a name. This is understood on FET and BJT symbols because they're common, but folks using PMT symbols should be well-versed in how to apply them as well. Therefore, just a pin number may suffice.
1. For a generic symbol, pin names should be generic. Anode, Cathode, etc. ought to be clear. Since PMTs appear to contain a variable number of dynodes, does DY1 give more clarity than just D1?
1. Underscores could be applied to separate the part type, exact part info (dynode quantity in this case), and then unique info (shielded or not).
1. Does electrostatic and magnetic shielding always come together? If not, maybe there's a naming scheme that would accommodate all symbols with variations in shielding? From the quick reading I've done, magnetic shielding is by a high-perm metal which is a separate piece from the tube. Again, this isn't my strength in the electronics field and maybe there are integrated magnetic shields.
1. NC pins should be considered because CvPcb will need them to match with the pin count of the footprint. Since these are generic symbols we probably don't care, but in general if there are standard packages the pin count matching must be considered.
1. Does IEEE 315-1975 mention the ref des prefix? In this case, all tubes I see expect rectifier tubes should use V.
1. Does that same doc give guidance to the symbol design? This is a good check for any new symbol. In this case, I see a couple of potentially-relevant sections:
image
image

I don't mean to sound negative. Please don't take anything that way. We went around a few times about the Geiger-Muller counter when that symbol was introduced. So in light of that, and a full symbol redesign that occurred after merging an initial one, I wanted to write down some thoughts when I saw this new type of symbol as well.

Here's a quick hack job to make another version of this symbol for possible discussion:
image
```
# PMT_x08_Shielded
#
DEF PMT_x08_Shielded V 0 20 Y Y 1 F N
F0 "V" -250 550 50 H V C CNN
F1 "PMT_x08_Shielded" 50 550 50 H V L CNN
F2 "" 0 100 50 V I C CNN
F3 "" 0 100 50 V I C CNN
$FPLIST
 PMT*
$ENDFPLIST
DRAW
A 0 0 447 634 1166 0 1 0 N 200 400 -200 400
A 225 -400 25 -899 899 0 1 0 N 225 -425 225 -375
A 225 -300 25 -899 899 0 1 0 N 225 -325 225 -275
A 225 -200 25 -899 899 0 1 0 N 225 -225 225 -175
A 225 -100 25 -899 899 0 1 0 N 225 -125 225 -75
A 225 0 25 -899 899 0 1 0 N 225 -25 225 25
A 225 100 25 -899 899 0 1 0 N 225 75 225 125
A 225 200 25 -899 899 0 1 0 N 225 175 225 225
A 225 300 25 -899 899 0 1 0 N 225 275 225 325
C 0 -450 10 0 1 0 F
C 0 450 10 0 1 0 F
T 0 0 -50 50 0 0 0 Shielding Normal 0 C C
S 300 -500 -300 500 0 1 10 f
P 4 0 0 0 -250 50 -200 50 -200 -150 -250 -150 N
P 2 0 1 5 0 -500 0 -450 N
P 2 0 1 5 0 500 0 450 N
P 2 0 1 0 200 -450 -200 -450 N
P 2 0 1 0 300 -400 250 -400 N
P 2 0 1 0 300 -300 250 -300 N
P 2 0 1 0 300 -200 250 -200 N
P 2 0 1 0 300 -100 250 -100 N
P 2 0 1 0 300 0 250 0 N
P 2 0 1 0 300 100 250 100 N
P 2 0 1 0 300 200 250 200 N
P 2 0 1 0 300 300 250 300 N
X ~ D1 400 300 100 L 50 50 1 1 P
X ~ D2 400 200 100 L 50 50 1 1 P
X ~ D3 400 100 100 L 50 50 1 1 P
X ~ D4 400 0 100 L 50 50 1 1 P
X ~ D5 400 -100 100 L 50 50 1 1 P
X ~ D6 400 -200 100 L 50 50 1 1 P
X ~ D7 400 -300 100 L 50 50 1 1 P
X ~ D8 400 -400 100 L 50 50 1 1 P
X ~ E -400 -100 100 R 50 50 1 1 P
X ~ K 0 600 100 D 50 50 1 1 P
X ~ M -400 0 100 R 50 50 1 1 P
X ~ P 0 -600 100 U 50 50 1 1 P
ENDDRAW
ENDDEF
```

Again, I'm not saying anything needs to change. Just trying to contribute even though I missed out on the previous discussion, with the ultimate goal of having the best possible library for KiCad users that choose to use this library.

ferdymercury:
Thanks @evanshultz for the thorough review. No offense taken, I think it is helpful to discuss this. I did the first symbol of this type here, so it is still a draft and I am happy to change and standardize it, before we make copies with different number of dynodes, as already requested by other users.

>     1. We use `_Shielded` as a suffix for other part types with a shield. Like connectors.
Fine for me to change the name.
 
>     2. Where there are internal lines connecting, a 10mil dot should be placed.
Ok

>     4. The symbol is huge.
I agree, but the component is also huge with respect to any other PCB parts. Also, the most typical use of the PMT is with a voltage divider network, see picture https://user-images.githubusercontent.com/10653970/51792015-d7fdba80-2179-11e9-8b55-93fed8243213.png
I designed it in such a way that a resistor fits exactly between each two pins. I think this way it would be much more helpful to users than having a very compact version where you need to put extra global labels (more effort). Also, it is the typical way of drawing by manufacturers, see https://www.hamamatsu.com/resources/pdf/etd/PMT_handbook_v3aE-Chapter5.pdf
http://lmu.web.psi.ch/docu/manuals/bulk_manuals/PMTs/Photonis_PMT_basics.pdf
http://et-enterprises.com/images/data_sheets/L25D19.pdf 

>     5. Is there a reason not to capitalize all pin names?
Fine for me to capitalize.

>     6. Pin lengths are longer than required. Usually the pin will only reach the boundary of the symbol so the pin name, if shown, is centered on the line.
Good point.

>     7. There are lines behind the pin lines on all 8 dynode pins, and they're not even consistently placed.
My bad.

>     8. If the symbol design makes a pin name redundant, then the pin doesn't need a name. This is understood on FET and BJT symbols because they're common, but folks using PMT symbols should be well-versed in how to apply them as well. Therefore, just a pin number may suffice.
Agreed.

>     9. For a generic symbol, pin names should be generic. Anode, Cathode, etc. ought to be clear. Since PMTs appear to contain a variable number of dynodes, does DY1 give more clarity than just D1?
Both could work. I think we could follow the conventions taken from the one of the biggest manufacturers: https://www.hamamatsu.com/resources/pdf/etd/PMT_handbook_v3aE-Chapter5.pdf
F would be the focusing grid, if any.

>     10. Underscores could be applied to separate the part type, exact part info (dynode quantity in this case), and then unique info (shielded or not).
Agreed. Although probably at some point it is better to just do atomic parts because each PMT tube will have a custom footprint, see e.g. the many number of variations of pins and layout https://www.hamamatsu.com/resources/pdf/etd/PMT_TPMZ0002E.pdf.

>     11. Does electrostatic and magnetic shielding always come together? If not, maybe there's a naming scheme that would accommodate all symbols with variations in shielding? From the quick reading I've done, magnetic shielding is by a high-perm metal which is a separate piece from the tube. Again, this isn't my strength in the electronics field and maybe there are integrated magnetic shields.
Not sure about this, but I think magnetic is an optional but very frequent addition https://www.hamamatsu.com/resources/pdf/etd/PMT_120_en.pdf

>     12. NC pins should be considered because CvPcb will need them to match with the pin count of the footprint. Since these are generic symbols we probably don't care, but in general if there are standard packages the pin count matching must be considered.
Yes, this should be considered when designing atomic parts.

>     13. Does IEEE 315-1975 mention the ref des prefix? In this case, all tubes I see expect rectifier tubes should use V.
Don't understand the question.

>     14. Does that same doc give guidance to the symbol design? This is a good check for any new symbol. In this case, I see a couple of potentially-relevant sections:
>         image
This symbol defined by IEC is nice and compact, and is used mostly for describing the footprint of the PCB. See https://www.ebay.com/i/322918399784?chn=ps https://www.ebay.com/i/322918399784?chn=ps
For a schematics sheet, where you will connect it with a voltage divider network, I prefer the bigger one. See here https://www.hamamatsu.com/resources/pdf/etd/PMT_handbook_v3aE-Chapter5.pdf
See also point 4.

>         image
Good point, we could change the dynode shape to match IEC. 

> I don't mean to sound negative. Please don't take anything that way. We went around a few times about the Geiger-Muller counter when that symbol was introduced. So in light of that, and a full symbol redesign that occurred after merging an initial one, I wanted to write down some thoughts when I saw this new type of symbol as well.
No problem at all, thanks for your input.

> Here's a quick hack job to make another version of this symbol for possible discussion:
> image
I would still prefer the huge version (or we make one with _Small). Otherwise I am fine with your design otherwise. Maybe you could change the dynode shape to match IEC, and do a pull request. Thanks!

evanshultz:
The size of the physical component is unrelated to the symbol size. I see where you're going with the voltage divider, but wires from a small PMT symbol to a string of resistors works fine. So does rotating resistors on their side to reduce their height.

If and when there are footprints we may have to revisit these topics. But ATM the pins are all using names, not numbers, so it should be fairly rugged to support various footprints.

I'll update the dynode pin style and submit a PR.

PR #1258, opened 2018-12-12
Add SE5004L.

matthuszagh:
Adding SE5004L
http://www.skyworksinc.com/uploads/documents/SE5004L_202393B.pdf

image

------------


myfreescalewebpage:

- [ ] I would suggest to order GND pin from the left to the right pin 2 to pin 17, for consistency with the other devices in the libraries
- [ ] According to the datasheet, VREF is a Power Input, not an Input. It should also be moved to the top of the symbol. I would suggest to order the positive power inputs pin from the left to the right : VCC1/VCC2/VCC3/VREF. And also to center the entire group horizontally on the symbol.

Cheers,
Joel

matthuszagh:
Thanks for the feedback Joel. Since VREF is optionally connected to SP1 and SP2 via a capacitor, think it makes sense to keep that on the top left? Everything else makes sense.

matthuszagh:
I've updated the commit as well as a screenshot of the component. Let me know what you think.

myfreescalewebpage:
Agreed, looks great. Can you also align horizontally reference and name fields ? Moreover, we usually put reference above.

@antoniovazquezblanco @DanSGiesbrecht The VREF pin on this symbol makes the KLC S4.4 error, do you know why ?

Joel

matthuszagh:
Updated. I believe this is what you want for reference and name fields, but let me know if not.

antoniovazquezblanco:
- VREF as far as I know is not a power input. It does not require a high power source to operate. It seems to be an input as only requires a reference voltage. If I am right, that pin should be of type Input.

- I would place VREF at the left (if it is not a power input). Just a mater of opinion I think.

- The symbol could use some pin stacking in the GND pins. Maybe VCC pins also (I don't know if the chip requires any decoupling capacitors or similar.)

myfreescalewebpage:
@matthuszagh yes.

@antoniovazquezblanco VREF is to be discussed, I agree it does not require lots of power, but when submitting micro-controller, we put VDDA / VSSA on top and bottom and that's exactly the same behavior... Also, reference voltage output pin in the Kicad library are of type "Power Output" so it is logically connected to a Power Input in my opinion.

I'm not part of the people recommending pin stacking, so today I have adopted a neutral position which consist saying nothing about that ! I let the submitter doing the way he want.

matthuszagh:
No worries, @myfreescalewebpage. I've made the changes and stacked the ground pins as per @antoniovazquezblanco's recommendation. I haven't done this for the VCC pins since they require bypass caps.

poeschlr:
Stacking pins means they are placed on top of each other. So please move the invisible pins to the same position as the visible one. (This forces the pins to be always connected together, transferring responsibility from the circuit designer to the library designer.)

matthuszagh:
@poeschlr apologies for that. Updated version is now up.

myfreescalewebpage:
Can you move reference and name close to the box of the symbol as many others ? Example below.

49876155-08380680-fe23-11e8-9854-024e04b928df

matthuszagh:
Done. Note that the reference and name are now off the 100mil grid, but maybe that's only an issue for pin positions?

myfreescalewebpage:
> Note that the reference and name are now off the 100mil grid, but maybe that's only an issue for pin positions?

Yes it is, it's not applicable for reference and name.

@antoniovazquezblanco to my point of view it's ready to be merged.

Joel

antoniovazquezblanco:
@myfreescalewebpage 

> I'm not part of the people recommending pin stacking, so today I have adopted a neutral position which consist saying nothing about that ! I let the submitter doing the way he want.

I don't like it very much. In the future, if we are lucky enough, we will see a feature in KiCad allowing to have both options (stacked and unstacked), but for the time being, the general librarian consensus is to provide stacked pins when possible in order to make symbols smaller and more convenient.

I will have a look at this in a minute.

PR #1267, opened 2018-12-15
ADP5063

EvilMav:
"Linear LiFePO4 Battery Charger with
Power Path and USB Compatibility in LFCSP"
Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/ADP5063.pdf

I don't like the Vin position at top as I prefer treating the USB input as an input, placing it to the left, but other PMICs seem to do so too. Otherwise the checker script complains or I'd have to remove the "power input".

Feels a bit tight, let me know if it is a problem.

kicad_screenshot_tmp


Misca1234:
There a number of devices that not follow "power ontop" rule, one example is power regulators

For this component, unless people oppose, I can not see any problems to move VIN to the left side

Then i have the following suggestions which is up to you.
Because the LED is connected to VLED, I suggest you place them above the ISO, BAT THR then you wont get crossing lines for ILED and THR (which want to GND)
Same goes for SYS_EN.

Move CBP down below DIG_IO for same reason

Because connection of the EP pin is not required I suggest you make it invisible, NC (no connection) and place the 'X' on the symbols body line.

If you want to connect it, stack it on AGND

EvilMav:
@Misca1234 
Thank you for your comments. This makes sense, I will do as suggested. I tend to disagree about the EPAD though: you do not need to connect it to anything, but you might want to connect it to a plane to improve the power dissipation. Typically it is GND, but not necessary (e.g. you might have a V+ opposite PCB layer to dissipate into). Having this as a separate pin give the designer freedom to choose where the heat goes.

EvilMav:
Commited version:
kicad_screenshot_tmp

Misca1234:
Ok, then we are good to go

PR #1268, opened 2018-12-15
 LT3010-5/ADJ: fixes NC pins (#1259) and other fixes

### LINKED ISSUE #1259: Move No-connect (NC) pins (LT3010)
`LT3010-ADJ` and `LT3010-5` both should have their NC pins moved to align with the KLC, to be on the symbol box edge. While this is a pin-movement change, it is non-breaking, because the designer should never connect these pins.

#### Incorrect:
LT3010-5
LT3010-ADJ

#### Correct Example:
correct

diegoherranz:
Move NC pins inside the box. Fixes #1259.

Taking advantage of the pull request, fixing a couple more things:
- Pin name offset was 30 mil instead of the recommended 20 mil.
- Datasheet links weren't working anymore.

Other things spotted but which may be more disrupting if changed. I can submit changes if requested:
- Not sure why the ADJ pin is an input and the SENSE pin is passive. They both seem to be inputs to the error amplifier.
- The part numbers could possibly be changed. For instance the adjustable one is LT3010 and not LT3010-ADJ which is a MPN that doesn't exist.

**LT3010-5**
lt3010-5

**LT-3010-ADJ**
lt3010-adj

Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/30105fe.pdf

Thanks!

------------


antoniovazquezblanco:
> Not sure why the ADJ pin is an input and the SENSE pin is passive. They both seem to be inputs to the error amplifier.

It would be nice to standardize this. I have seen comments about this kind of pins in another issue but I can't locate it.

> The part numbers could possibly be changed. For instance the adjustable one is LT3010 and not LT3010-ADJ which is a MPN that doesn't exist.

Renaming would mean breaking the lib for a minor release which is not ok. Please, don't do this. I will open an issue to be resolved for v6.0.0.

Thanks!

diegoherranz:
> It would be nice to standardize this. I have seen comments about this kind of pins in another issue but I can't locate it.

If you find it, please let me know and I'll fix it.

> Renaming would mean breaking the lib for a minor release which is not ok. Please, don't do this. I will open an issue to be resolved for v6.0.0.

Yes, thanks for rising #1272.

poeschlr:
There will not be a standard rule to fit them all. Some parts use the same name for slightly different functions so there will always be some thinking needed.

However i would assume that ADJ and SENSE are input in most cases. (They are after all "analog input" pins for the "controller")

evanshultz:
If ADJ was connected to another Input pin directly, then the ERC engine would say there's no driver for the net. But I don't really see this pin having that situation. I suppose while both ADJ and SENSE connect to passive components in the typical case, they are inputs so that would be fine and more specific than Passive (according to Rene's pin type hierarchy at https://github.com/KiCad/kicad-website/issues/342#issuecomment-426986726).

Nobody is objecting above and this is a nice fix, so I'll merge. Thanks!

diegoherranz:
From what I read, there's an agreement to make both `input` but it's been merged without fixing that (one of them is still passive)?

evanshultz:
I think so, but I'm not clear if any more discussion is outstanding about pin types. I for one would like to review and confirm a final decision on pin types (of course Rene gets the final word). And then see pin types, among other things, in KLC.

As the referenced issue and this PR is dealing with just one regulator family, and pin type updates affect many regulator symbols (and symbols which are not regulators, too) I wanted to get this resolved without delay and close the open issue.

Does that make sense? Was I too hasty?

diegoherranz:
I thought it might have been overlooked. But now that you've explained, it makes sense.
Thanks.

PR #1270, opened 2018-12-16
Added ATMega-0-Series Devices

dominformant:
* ATMega 3208-SSOP (http://ww1.microchip.com/downloads/en/DeviceDoc/40002018A.pdf)
atmega3208-x

* ATMega 3208-TQFP (http://ww1.microchip.com/downloads/en/DeviceDoc/40002017A.pdf)
atmega3208-a

* ATMega 3208-VQFN (http://ww1.microchip.com/downloads/en/DeviceDoc/40002017A.pdf)
atmega3208-m

Aliases:
* ATMega 4808-SSOP (http://ww1.microchip.com/downloads/en/DeviceDoc/40002018A.pdf)
* ATMega 4808-TQFP (http://ww1.microchip.com/downloads/en/DeviceDoc/40002017A.pdf)
* ATMega 4808-VQFN (http://ww1.microchip.com/downloads/en/DeviceDoc/40002017A.pdf)

Series-0 Datasheet:
http://ww1.microchip.com/downloads/en/DeviceDoc/40002015A.pdf

------------


fauxpark:
Thanks for the contribution!

Could you make these symbols more like the existing ones? For example:
 - The refdes/name/footprint fields should be positioned and aligned like the rest of the symbols
 - Since the IO pins are tri-state, you can save some space and place them on both sides of the symbol in a clockwise direction (but with pin 0 always at the top)
 - The lowest-numbered stacked pin should be the visible one, so pin 19 for the VQFN and TQFP GNDs and pin 15 for the SSOP GNDs
 - Place AVDD on the right of the VDD pin
 - Instead of -TQFP, -SSOP, and -VQFN suffixes, these symbols should have -A, -X, and -M respectively (see [this page](https://www.microchipdirect.com/product/search/all/atmega3208)) The FR refers to lead finish and tape & reel which we don't care about
 - PF6 is also the /RESET pin
 - The UPDI pin I believe can be Bidirectional instead of Passive

dominformant:
Thank you very much for your hints for improving the symbols. I tried to fix all the points you mentioned and hope i did not forget any.
I furthermore added a pin for the exposed pad for the device in QFN package with the last commit.

atmega3208-a
atmega3208-m
atmega3208-x

fauxpark:
Looks good :) Just a couple more things:
 - I think you've accidentally changed the datasheet URLs to all use the SSOP one
 - PF0 and PF1 should have the TOSC labels on the inside so the Pxn labels all align
 - You don't need the trailing ~ on PF6, it's only necessary if there's something after the overline (eg. if the reset pin was on the other side of the symbol)

dominformant:
I'm sorry, I totally overlooked these problems. Fixed them now.

evanshultz:
@fauxpark 
Thanks for the review!

1. I only find "EXTCLK" on these three symbols in the entire lib? Is there a reason to show that pin function? Have you shown it elsewhere?
1. Did you mean to wrap the GPIO around to the left side instead of keeping it all on the right? Such as on the SSOP package putting ports D and F on the left? And do all pins of a port go together, even if some bits in the port are skipped (SSOP symbol again)?

@dominformant 
Thank you for including new screenshots when there were notable visual changes (updating screenshots in the original post is OK too). I do see that PF6 is not ~RESET on the TQFP/QFN symbols and it's always nice to catch all fixes in the screenshot.

1. As noted above, please flip the pin functions so that the Px# is to the body edge (see pins 16-17 on the SSOP symbol).
1. Please include pin pitch in the footprint filters with a "P" before it, like `P0.65mm` (SSOP only).
1. Do not include the thermal pad size in the footprint filter (QFN only).
1. Does the datasheet say the QFN thermal pad must be connected to ground? Usually it does just below the pinout diagram, but I don't see it in this datasheet.

fauxpark:
Considering these devices have no XTAL pins to speak of I figured EXTCLK could be left there, but I guess the TOSC pins serve that function? I'm not sure what the difference is between XTAL and TOSC, especially as there are some symbols with both.

In order to keep the symbols from getting too tall I wrap GPIO pins clockwise, but with pin 0 always at the top of the group. Port D might be able to squeeze in there on the left.

In symbols where there are missing GPIO pins from a single package (like this one) I keep the remaining ones in their original positions in case the user changes their mind and decides to use a different package. Otherwise they are grouped together.

dominformant:
@evanshultz Thank you for your comments. I now updated the symbols according to your hints. Regarding the thermal pad, I'm was not sure how to name/number it, the datasheet does not say anything about the connection of the pad, but I thought it always gets connected to ground. If not, how do I place and number it? The updated screenshots are now in the first post to keep the thread short.

@fauxpark / @evanshultz 
Regarding the missing pins in port F for the SSOP device it was also my idea to be able to easily replace the device without changing the connections in the schematic.
Regarding the EXTCLK and TOSC pins, I took a look into the overall Series-0 datasheet ([http://ww1.microchip.com/downloads/en/DeviceDoc/40002015A.pdf](http://ww1.microchip.com/downloads/en/DeviceDoc/40002015A.pdf)) and (if i understood correctly) it says, that the TOSC pins are the only pins that may be used for a crystal, but may only be used with a 32kHz crystal. So they are no real replacement for the common XTAL pins. In contrast, the EXTCLK can be used with an arbitrary external clock source.
Speaking of the overall datasheet, is there an option to reference two datasheets in a symbol?

fauxpark:
Generally EP pins are numbered pincount + 1, except of course where the pin numbering is strange eg. dual row QFN which uses A1, B1 etc. in which case it is usually "EP".

For all but one of the AVR symbols the EP can (not must) be connected to ground and thus simply named GND and stacked. Mostly it is just for thermal and mechanical relief.

dominformant:
Okay, that's what I did, so I do not have to change it, right?

fauxpark:
Yes, all is well there :)

evanshultz:
@fauxpark 
Help us out with the style you'd used in the lib already. Does just port F go on the left side, or port D and F to make the symbol more balanced?

fauxpark:
D and F could go on the left here if UPDI is moved up, with PF0 opposite PA0 and PD0 opposite PC0:

pins

evanshultz:
@dominformant 
In the last commit some symbol got pushed to the Sensor_Touch library. After fixing that I'll review again.

Thanks for updating screenshots when you've made significant visual changes!

dominformant:
@evanshultz Thank you for your review. I'm sorry I did not care for that, I've reset the library now. Is it correct that way, or should I have force-pushed an erlier commit to keep the mistake out of the library?

evanshultz:
I'll squash merge so it's fine to leave it as is. :)

And I missed your question above about datasheet quantity. I only recall seeing a single datasheet, and with two present I don't think the UI button to open the datasheet works. So let's leave it at just one.

Thanks for the fixes. A few items mentioned are not included in this branch:
1. ATmega4808-A and -M are still using the SSOP datasheet.
1. Please include pin pitch in the footprint filters with a "P" before it, like `P0.65mm` (SSOP only).
1. Do not include the thermal pad size in the footprint filter but do add `1EP` (QFN only).

@fauxpark 
Are you good with everything if the above three items are done?

fauxpark:
@evanshultz yep, I don't see anything else (but your eye is probably sharper than mine!)

dominformant:
@evanshultz @fauxpark  I'm sorry, I was sure I had already included these fixes, but apparently I missed to commit or so... However, now they are included in the branch. Thank you very much for your reviews and support with my first contribution.

evanshultz:
Thank you! You've contributed some nice symbols and we hope to see you again soon!

PR #713, opened 2018-06-30
Add symbols for all Microchip SAMC20/C21 microcontrollers

HubertD:
### Overview
This PR adds symbols for the whole Microchip (ex Atmel) SAMCx family of microcontrollers.
All footprints except the WLCSP-56 are covered in the current kicad-footprints;
PR for the WLCSP-56 is here: https://github.com/KiCad/kicad-footprints/pull/648

### Datasheet
http://ww1.microchip.com/downloads/en/DeviceDoc/DS60001479B.pdf

### SAMC2xE1xA-A
samc20e15a-a

### SAMC2xE1xA-M
samc20e15a-m

### SAMC2xJ1xA-A
samc20j15a-a

### SAMC2xJ1xA-M
samc20j15a-m

### SAMC2xN1xA-A
samc20n15a-a

### SAMC2xN1xA-U
samc20n15a-u

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #1271, opened 2018-12-16
Add ATSAMC2xE and ATSAMC2xG series MCUs

EvilMav:
Added MCU_Microchip_SAMC.lib with ATSAMC2xG (48-pin) and E (32-pin) microcontrollers. Both  have C20 and C21 versions, multiple memory sizes and TQFP (-AU) as well as QFN (-MU) packages, which results in a zillion aliases. I did NOT add the -AN/-MN aliases (high temperature versions) as I'm not sure which are available off-the shelf and I did NOT add the -AUT aliases as those are same chips in different tapes.

For the QFN versions I've put the thermal pad on GND - the datasheet does not say anything about the pad, probably not connected, but according to forum topics it seems not wrong...

Datasheet: https://www.microchip.com/mymicrochip/filehandler.aspx?ddocname=en598829

kicad_screenshot_tmp_samc-e
kicad_screenshot_tmp_samc-e-qfnfoot
kicad_screenshot_tmp_samc-e-tqfpfoot
kicad_screenshot_tmp_samc-g
kicad_screenshot_tmp_samc-g-qfnfoot
kicad_screenshot_tmp_samc-g-tqfpfoot

------------


EvilMav:
Travis seemingly died because the library is new...

I've been thinking about moving PA00/01 and PA14/15 to the other side on the G version at least, as those are the pins which could also be used as oscillator connections if I read the pinout right. May or may not... What do you think?

EvilMav:
Googling and looking through my own projects, seems like external crystal and SWD debug is used more often the not. In the spirit of making the common case cute, I've moved the oscillator and debug pins to the left and grouped them further apart to ease placing crystal schematics in between. Probably a matter of taste, but I would prefer it this way.


kicad_screenshot_tmp_samc-e
kicad_screenshot_tmp_samc-g

EvilMav:
Oops, just found PR https://github.com/KiCad/kicad-symbols/pull/713 providing those and even more parts...

myfreescalewebpage:
Hello @EvilMav 

I have not checked all parts, but yes seems it's a Duplicated PR. Do you confirm all your part are covered by #713 ? Are you agree to close this PR so we can be concentrated on the previous one ?

Cheers,
Joel

EvilMav:
@myfreescalewebpage 
yes, all parts seem to be covered in the other PR.  Closing.

myfreescalewebpage:
@EvilMav ok, great, will focus on the other one so.

PR #1274, opened 2018-12-17
Amplifier_Audio: Add SSM2167.

mgottschlag:
Like SSM2165, but with shutdown pin and configurable noise pin, and in a
different package (MSOP-10). I tried to mirror the SSM2165 symbol as closely as possible.

------------

[https://www.analog.com/media/en/technical-documentation/data-sheets/SSM2167.pdf](https://www.analog.com/media/en/technical-documentation/data-sheets/SSM2167.pdf)
2018-12-17-222839_762x821_scrot
(seems to select the correct footprint)
(no new footprint)


mgottschlag:
Note that, like for SSM2165, the symbol checker complains about VOUT not being Power Output. "Output" is correct given that this is an audio output.

myfreescalewebpage:
Hi @mgottschlag , thanks for contributing,

- [ ] The package at the end of the description should be "MSOP-10" not "MSOP10"
- [ ] The footprint filter should not contains the number of pads: `MSOP*3x3mm*P0.5mm*`

Even if the symbol does not fit exactly the datasheet regarding to pin names, etc it is consistent with SSM2165 already in the lib, so it's okay for me. KLC S4.4 warning accepted for this symbol.

Cheers,
Joel

mgottschlag:
I modified the commit, both should be fixed now.

mgottschlag:
Ok, sorry, other projects (e.g. Linux) prefer a new set of clean patches after review.

PR #1279, opened 2018-12-18
Add AD630ARZ in SOIC-20W

aewallin:
https://www.analog.com/media/en/technical-documentation/data-sheets/ad630.pdf

AD630ARZ in SOIC-20W (other packages available also)
ad630arz

Location of -V pin off-center is a bit ugly, but if we want a compact symbol this seems to be one solution. Alternatively make the offset-adjust pin texts smaller?

Functional diagram:
ad630funct

------------


aewallin:
Questions:
- SEL A, SEL B ordering as in datasheet? SEL B on top?
- ADJ pin type? Suggested connection is to a potentiometer.
- RB RA RF pin order as in datasheet?

aewallin:
library version change touches a lot of files :(

updated symbol based on self-review ;)
ad630arz

jneiva08:
can you shrunk the pins to 100mil (KLC S3.1)

EvilMav:
May I suggest considering creating a symbol with extensive inner diagram embedded into the symbol? It is a lot of work and quite big on the printout, but it would make a schematic made with this component way more readable. I'm happy to help if it is of interest...

poeschlr:
Adding graphics to the inside is always a tradeof. We generally do not really require it for anything as complex as this component. (Kicad is not really well suited to create such graphics)

I leave it up to @aewallin. If they are interested in investing time into this then it is ok as long as the symbol does not get overly large and the pin names are still readable.

EvilMav:
I generally try not to overuse the detailed symbols, but in this particular case wrapping my head around this chips application just by looking at the pin names is tricky, as it is basically a set of building blocks for a synchronous detector to assemble by external connections..

Here is just a draft of what this could look like (quick shot, I know the offset pins etc are missing - just to see what it might look like)
kicad_ad630_detail

If you would like this design, I can continue on this path.

aewallin:
@EvilMav that looks like a good start. Any idea on how to make room for standard pin numbers and names?
- the output op-amp triangle looks different from the input ones?
- the switch-symbol (between A and B, controlled by SEL) could maybe be improved?

if you send a PR to my branch or make available the drawing some other way I will try to merge.

- KLC S3.1. I'd like the KLC to more clearly state if 100mil pins are in fact to be preferred. The way I read it anything 100-300 is OK now.
- agree on the open-collector for pin 7.

aewallin:
@EvilMav many thanks for the artwork.
I have made some modifications and my suggestion is below. Maybe one day kicad will allow for an offset in the pin-name so that the 'wire' inside the symbol doesn't overlap with the name.

- SEL_A, SEL_B pins were swapped
- COMP pin was not on 100mil grid
- Standard pin-labels 30mil size
- Moved output resistors 10k and 5k a bit to make room for pin-name texts
- switch-symbol has circles on all three contacts (they are actually different sizes... noticed this just now..)
- Text to guide the user in pin-names rather than in the symbol itself.
- Vs- and Vs+ symmetrically in the middle.
ad630arz

EvilMav:
Thank you! I like the switch better now =) Well, what we can still consider when it comes to pin names is hiding them and placing the name texts manually above wires instead. This changes the font color, but it is probably better for a black-an-white printout. Maybe we can move the 2k5 resistors a little bit too... Otherwise I like it very much, pin assignments seem right, so I think it would be ready to be pulled then =)

EvilMav:
Found a way of hiding names only for certain pins by setting their font to 0 size... I've branched your repository and will send you a PR shortly.

aewallin:
This now eliminates standard pin-names, in favor of "hard coded" text in the symbol, which is more legible - but a bit non-standard.
minor adjustments:
- switch circles should now be the same size
- transistor arrow more pointed. (top arrow-'wing' was vertical previously)

ad630arz

EvilMav:
As far as I'm concerned it is perfect =) KLC warnings are IMHO ok to ignore in this case. @poeschlr is this solution OK with you?

@aewallin I'm a bit confused with this pull as Travis and "Files changed" view indicate multiple components have been changed, not just AD630. I'm not sure if it is safe to pull ...

aewallin:
AFAIK the changes in other symbols has to do with a change in the library editor version.
"EESchema-LIBRARY Version 2.4"
In principle that version change should be submitted as a separate PR, and then new symbols after that..

poeschlr:
The graphical style looks ok to me. I did not check anything else.

myfreescalewebpage:
@EvilMav and @aewallin very nice work you have done, very great ! :)

I have made a complete review, changes made by KiCad on other symbols are okay (no loss of information).

Regarding to the AD630ARZ:
- [ ] The footprint filter is not correct, it should be `SOIC*7.5x12.8mm*P1.27mm*` (no pin count allowed)
- [ ] The name of pin 4 should be "DIFF_OFF_ADJ2"

Cheers,
Joel

aewallin:
after revisions:
ad630arz

myfreescalewebpage:
@aewallin thanks, no more comment to do, merging.

Cheers,
Joel

aewallin:
Thanks again to @EvilMav for the help.

PR #1280, opened 2018-12-18
add MC12093D

aewallin:
https://www.onsemi.com/pub/Collateral/MC12093-D.PDF

Did not find a good place for a prescaler/frequency-divider, so put it in the general RF-library.

mc12093d


myfreescalewebpage:

- [ ] The symbol is a little bit small and pin are very closed together, maybe it is better to enlarge a little bit horizontally ? I'm not sure, I do not realize where you are using this device, but fill free to provide a schematic screenshot to give us a better understanding.
- [ ] footprint filter should be `SOIC*3.9x4.9mm*P1.27mm*`

Cheers,
Joel

aewallin:
after revision, a little larger, more symmetric, with 150mil pins.
fp-filter changed.
mc12093d

myfreescalewebpage:
@aewallin thanks for this symbol too.

@antoniovazquezblanco PR ready to be merged.

PR #1283, opened 2018-12-18
add MC12080

aewallin:
https://www.onsemi.com/pub/Collateral/MC12080-D.PDF

mc12080


myfreescalewebpage:

A few comments I have during my review (basically the same as #1280):
- [ ] The symbol is a little bit small and pin are very closed together, maybe it is better to enlarge a little bit horizontally ? I'm not sure, I do not realize where you are using this device, but fill free to provide a schematic screenshot to give us a better understanding.
- [ ] footprint filter should be `SOIC*3.9x4.9mm*P1.27mm*`

Cheers,
Joel

aewallin:
after revision we have:
mc12080

myfreescalewebpage:
@aewallin thanks, great work.

@antoniovazquezblanco this PR is ready to be merged according to me.

aewallin:
grr.. these are no fun to resolve by hand. a more git-friendly library format would definitely be needed!

poeschlr:
You will get used to it ;)

Also v6 should bring us the sweet format.

myfreescalewebpage:
Thanks @aewallin

PR #1284, opened 2018-12-18
74xx: Add 74LV8154

m-laniakea:
74xx: Add 74LV8154 2x16-bit binary counter IC
------------

- [x] There is two variants of the device, TSSOP and PDIP. You should remove the footprint, and add a second footprint filter DIP*W7.62mm*


m-laniakea:
[74LV8154 Datasheet](http://www.ti.com/lit/ds/symlink/sn74lv8154.pdf)
Schematic

diegoherranz:
Thanks for your contribution.

On schematic symbols, we group the pins by function and we arrange them so that the schematic is the easiest to draw and understand from a logical poinr of view. That means that the arrangment of the symbol doesn't need to follow the actual pinout/package.

Could you rearrange following http://kicad-pcb.org/libraries/klc/S4.2/ ?

Thanks!

m-laniakea:
@diegoherranz Updated.
74LV8154 Schematic
Current symbol has background color.

Note that ~CLKBEN is often connected to ~RCOA

m-laniakea:
See https://github.com/KiCad/kicad-footprints/pull/1197.

diegoherranz:
Thanks for the changes.

Footprint chosen is not identical to recommendation on the datasheet but it should work OK since they are very similar as mentioned on https://github.com/KiCad/kicad-footprints/pull/1197.

Regarding the footprint filter, see point 3 (http://kicad-pcb.org/libraries/klc/S5.2/):

> Filters must not contain the pin count if the pin-count in the symbol matches the pin count in the footprint. In those cases the footprint is matched by KiCad’s pin count filter.

Also, the P for pitch is normally included too (see [an example](https://github.com/KiCad/kicad-symbols/blob/master/74xx.lib#L50)).
So, I think it should be `TSSOP*4.4x6.5mm*P0.65mm*`.
Could you fix that? I think everything else is OK now.

Thanks!

myfreescalewebpage:
Hi @m-laniakea , thanks for contributing,

- [ ] There is two variants of the device, TSSOP and PDIP. You should remove the footprint, and add a second footprint filter `DIP*W7.62mm*`

Position of RCOA if consistent with other devices in the library so probably it's better like that even if RCOA and CLKBEN are often connected.

Cheers,
Joel

diegoherranz:
I've just noticed as well, shouldn't it be called `SN74LV8154`?

diegoherranz:
Another question

> There is two variants of the device, TSSOP and PDIP. You should remove the footprint, and add a second footprint filter DIP*W7.62mm*

If the aim of the libraries is having atomic symbols (last sentence on [KLC S2.2](http://kicad-pcb.org/libraries/klc/S2.2/)), shouldn't we create both `SN74LV8154N` (PDIP) and `SN74LV8154PW` (TSSOP). Or at least rename this one as `SN74LV8154PW` and create the other one when/if needed?

Thanks!

myfreescalewebpage:
@diegoherranz yes and no. The 74xx library is plenty of symbols that have footprint variations, especially SOIC/DIP. Just look at 74CBTLV3257 for example.

Maybe the bets solution is that footprint should be defined in the dcm file instead of in the lib file, so it is possible to modify it for each alias. But it's not made like that today.

Cheers,
Joel

myfreescalewebpage:
Thanks. Merging.

Issue #1157, opened 2018-11-22
c96abc and all other  DIN41612  connectors  pins dont match sch lib  v footprint pin numbers 


achiestdragon:
the pins on the symbol in the schematic are wrong ,, the pin names 1a though 32c thats correct but the pin numbers are numbered  a1 though  c32  that dont match the footprint numbers  that are numbered like the pin names  1a though 32c  so none of the pins produce a usable netlist connections 

the same problem is also true of all the schematic symbols associated with all the DIN41612 footprints

antoniovazquezblanco:
Does the standard specify any preference on whether to use letter-first or number-first pin numbering?

achiestdragon:
actually no , but consistency across the schematic libs/footprint lib's
matters , when it comes to net lists

antoniovazquezblanco:
Would you please be willing to contribute this issue by elaborating a list of affected symbols or even creating a PR that fixes the issue?

Thank you!

evanshultz:
I don't think anything needs to be done.

Here are our current footprints:
![image](https://user-images.githubusercontent.com/1936989/48910194-2b732580-ee24-11e8-890f-d737e404ffb9.png)

They are letter-number.

(https://github.com/KiCad/kicad-footprints/pull/1076 will be added a ton more which are also letter-number.)

The symbol library have both letter-number and number-letter:
![image](https://user-images.githubusercontent.com/1936989/48910339-89a00880-ee24-11e8-8f9a-e02c974e52d7.png)

So while we provide symbols that don't match with the DIN41612 footprints, this is a convenience for the user if they want to add footprints which are number-letter. I agree that it would be easy to make a mistake since the footprints are only letter-number and the user needs to pick the right symbol to make them match with our DIN41612 footprints, but this report is wrong because there are no specific DIN41612 symbols and the symbols needed to match our DIN41612 footprints are provided.

If there is no action on this soon I think it can be closed.

achiestdragon:
on the other footprints for the 64 ab and 96abc  i get  as attached

dave

evanshultz:
@achiestdragon 
Thanks for the reply.

I see the schematic symbols, but they don't have a footprint assigned.

I'm not seeing these footprints in the current repo. Can you tell me the library where you found those footprints? Also, are your libraries up-to-date? The old Connectors footprint repo (https://github.com/KiCad/Connectors.pretty) had some C96 connectors but they were of low quality and were not ported to the current footprint libraries.

[...]

PR #1286, opened 2018-12-19
Added connector 41612 A-C

AndreaQ77:
New symbol, sorry I'am a beginner, first time.

------------


antoniovazquezblanco:
> New symbol, sorry I'am a beginner, first time.

Welcome and thanks for the contribution. Things around here sometimes go slow so don't expect us to answer your questions inmediatelly, but we will also not forget about your PR.

In order to help us, please look at the instructions in your first post and try to follow them (share a link to the datasheet, post an screen capture of the symbol...).

Thanks!

AndreaQ77:
Component data sheet:
https://www.mouser.com/datasheet/2/418/NG_CD_5650951_O1-668739.pdf

I have split it in two parts, one per row.
image
image

AndreaQ77:
Footprint PR:
https://github.com/KiCad/kicad-footprints/pull/1193

evanshultz:
Unless I'm mistaken, we already have scripted connector symbols that are suitable:
image

As mentioned at https://github.com/KiCad/kicad-symbols/issues/1157, we need to ensure capitalization matches the footprints (expected to come in https://github.com/KiCad/kicad-footprints/pull/1076) but I think on the symbol side we're good either way.

lorem-ipsum:
Great that you create these symbols, I did some work on the footprints (not quite finished yet, but already usable). As there are so many they definitely should be scripted, but having some is already a start.

The currently existing symbols are not suitable as they don't match the notation used by the DIN41612 connectors. The examples you showed use `a` and `b`, but the connector uses `a` and `c`. This becomes more problematic with skipped pins if only even numbered rows are populated.

I prefer the approach of splitting the rows into different components as this scales to more rows. There are connectors with three and up to five columns. For those splitting them is the only solution and then it would make sense to use the splitted approach for all of them as it would allow to "upgrade" from a 2 column version to a fully populated 3 column version.

poeschlr:
As already written in the issue about this:

- We might want to create a separate symbol library holding only symbols for din connectors
- specialized symbols are required for many of them so i might suggest to include specialized symbols for all of them (will get confusing for users otherwise)
- with a bit of adaption the symbol generator in https://github.com/KiCad/kicad-library-utils/tree/master/schlib/autogen/connector could be used to generate symbols for these. I would opt for a multi unit approach with these as din connectors come in 3 and 4 row variants which are simply not feasible otherwise.)
- The graphical style should be similar to our normal connector symbols. I would even suggest to use similar pin numbering (meaning use lower case letters) to increase cross compatibility. (it seems both lower and upper case letters are used by manufacturers of such connectors.)

AndreaQ77:
I'd like to add an information...
In the symbols library there are 3 symbols of connectors 41612 that (in my opinion) are not usable, too long and with pin name inverted A1 becames 1A.
I show you...

image

evanshultz:
@AndreaQ77 
Agreed. The existing symbols are bad an should be replaced.

@poeschlr @fauxpark @lorem-ipsum 
I haven't used the symbol generator before and had other things I planned to spend my time doing. Are any of you up for this task?

evanshultz:
But back to this...

We could still merge a one-off symbol if desired. Here are the things I gather above that we want to do:
1. Use lower case letters.
1. Name the pins letter-number.
1. I would prefer symbols exactly like the ones I drew above since those are done with the symbol generator now. Since the connector you mention above has only rows A and C populated, the question is then to put A on the left of unit one with C on the left of unit two, or A and C on the same unit? I ask because the version with rows A, B, and C is available and we do want to support that. My preference is A on unit one and C on unit two with the B column missing from unit one.

evanshultz:
With #2935 merged this can be closed. It includes a scripted `DIN41612_02x32_AC` symbol which should suffice.

It's still an outstanding point to remove the old and busted DIN41612 connector symbols. Perhaps it's best to do that here before moving to GitLab. @chschlue ?

chschlue:
👍🏼

evanshultz:
Thank you @AndreaQ77 ! We appreciate you taking the time and initiative to create this symbol and make a PR regardless of eventually closing it. And I apologize for taking so long to resolve this.

PR #1291, opened 2018-12-19
4xxx fixed open collector output pins

feichtinger:
Hi

I found out that the pin type of the output pins of 4023 (NAND) and 4025 (NOR) was set to 'open collector'. As far I know there is no part in the 4xxx family with open collector output...?

The 4xxx library should be free of this issue now.


evanshultz:
Good catch. Thank you.

Please sign the CLA and I'll merge. The most likely cause of this issue is when the GitHub account is not using the same email address as the commit.

PR #1294, opened 2018-12-20
Sensor_Optical: Add APDS-9306 and APDS-9306-065

antoniovazquezblanco:
https://docs.broadcom.com/docs/AV02-4755EN

imagen

imagen

Missing footprints. Will contribute when I make them.
------------


evanshultz:
- The package name doesn't seem to be given, but I think `LGA-8` and `DFN-6` would be fine for the description.
- This is a bit pedantic, but I like to see ` with` replaced with a `,` in the description for brevity.
- Can you put ref des on the left and value on the right?
- Unless these packages are specific to these exact PNs, perhaps a more generic footprint name is best? It can still start with `Broadcom_` but maybe use the package description and not the MPN?
- As the INT pin is open drain, can you make it Open Collector type?

antoniovazquezblanco:
> Unless these packages are specific to these exact PNs, perhaps a more generic footprint name is best? It can still start with Broadcom_ but maybe use the package description and not the MPN?

I will do as you ask me to but I am not sure on how to name them or wether theese are used for any other part.

evanshultz:
We'll continue the footprint naming at that PR.

All other changes are good. Thanks!

antoniovazquezblanco:
FP Merged, re triggering travis

antoniovazquezblanco:
Thanks!

PR #1296, opened 2018-12-20
Assign correct (new) footprint to OPA2695xRGT

poeschlr:
footprint PR: https://github.com/KiCad/kicad-footprints/pull/1199

datasheet: http://www.ti.com/lit/ds/sbos354a/sbos354a.pdf

PR #1297, opened 2018-12-20
Correct footprint assignment for MAX5139

poeschlr:
This symobl had the wrong footprint assigned (wrong qfn type and wrong
ep size)

fooptrint pr that adds the newly scripted footprint:
https://github.com/KiCad/kicad-footprints/pull/1199

datasheet: https://datasheets.maximintegrated.com/en/ds/MAX5138-MAX5139.pdf (link to package drawings is on page 16)


myfreescalewebpage:
@poeschlr just one comment about the footprint filter, I think it should contain EP, isn't it ? I not, no comment during the review, okay for me and can be merged as soon as the footprint will be accepted.

Cheers,
Joel

poeschlr:
No we typically do not include the EP size in the filter definition. The default footprint should be setup for the normal usecase. using the filters means that the user either wants a different ep size, add or remove thermal vias, handsolder version, ....

And yes this should be documented in the KLC

myfreescalewebpage:
@poeschlr I was pointing this because you have put it in https://github.com/KiCad/kicad-symbols/pull/1296 ;-)

myfreescalewebpage:
So now can be merged as soon as the footprint is accepted.

Joel

PR #1308, opened 2018-12-21
Fix footprint for TPA5050

poeschlr:
We already had a better fitting (scripted) footprint for this part in the lib.

Datasheet http://www.ti.com/lit/ds/symlink/tpa5050.pdf#page=20


myfreescalewebpage:
Hello @poeschlr 

No comment for this.

Cheers,
Joel

PR #1314, opened 2018-12-22
MCU_ST_STM8: Add STM8L152x

piotr-wiszowaty:
Datasheet: https://www.st.com/resource/en/datasheet/stm8l152r8.pdf
stm8l152x-sym

------------


myfreescalewebpage:
Hi @piotr-wiszowaty, thanks for contributing,

- [ ] You are currently violating KLC S3.6, you should change pin name position offset to 20
- [ ] Pins length should be 0.1 (symbols<100pins) (KLC S4.1)
- [ ] Your descriptions are a little bit long, we usually use "k" only instead of "kB", you can remove some peripherals, for example just keep USART, I2C, SPI, ADC, DAC, LCD and all without "s", the "°" for the temperature is to be avoided and "C" for Celcius should be after the temperature with no space
- [ ] Datasheet of R6 series is "https://www.st.com/resource/en/datasheet/stm8l152r6.pdf" (seems the same but probably it's better to separate)
- [ ] Type of pin 30 is Power Input, not Bidirectional
- [ ] You can rename VSS1, VSS2 and VSS3 to GND and stack them
- [ ] You can rename VDD1, VDD2 and VDD3 to VDD (do not stack) 

Cheers,
Joel

piotr-wiszowaty:
Stacking GND pins makes checklib.py complain. Tried setting invisible pins' type to passive - checklib.py complains even more. In the current state the (whole) output looks like this:
```
Checking symbol 'STM8L152R8T6':
  Violating S4.3
    Rules for pin stacking
```
I'm not sure what to do with this.

myfreescalewebpage:
@piotr-wiszowaty thanks for the quick fixes !

To perform pin stacking:
- all pins stacked must have the same name (okay - GND)
- one pin remain visible and it's type is Power Input
- all other pins are not visible and type set to Passive

In your situation, just turn type of pin 10 and 56 to Passive and it's okay with this KLC.

Joel

piotr-wiszowaty:
Done.

myfreescalewebpage:
@piotr-wiszowaty the remaining KLC error is raised because the pin is not properly defined. This should be solved by checking the two Sharing boxes in the Pin Properties:

capture

Cheers,
Joel

myfreescalewebpage:
Hi @piotr-wiszowaty 

Cheers,
Joel

PR #1318, opened 2018-12-23
Fix footprint assignment for SX127x

poeschlr:
This one had the wrong footprint assigned (wrong EP size)


myfreescalewebpage:
Hello @poeschlr 

No comment during the review. Ok with change of datasheet link and package EP size.

Just adding the link to the PR of the footprint: https://github.com/KiCad/kicad-footprints/pull/1207

Cheers,
Joel

evanshultz:
Footprint merged. Would you mind changing the pin name offset to 20mil?

poeschlr:
Same comment as with the other pull request. I have enought to do with fixing all these bad qfn footprint assignments. I really have no time looking into anything else.

PR #1319, opened 2018-12-23
added LM5001MAx

rs-development:
For all the Flyback fans out there, LM5001. I created a part for the SOIC Version, if this one is fine i will also add the WSON version.

The checks are not happy with some rules:

- S4.2: I now power pins on top but i took reference on similar devices and they all have the input pin on the left top side of the component, if i should change, where to put the name and designator
- S5.1 and S5.2: I am not sure how to create an atomic part, may somebody will assist me please.
 

[http://www.ti.com/lit/ds/symlink/lm5001.pdf](http://www.ti.com/lit/ds/symlink/lm5001.pdf)

screenshot from 2018-12-23 15-38-06


rs-development:
> S5.1 and S5.2: I am not sure how to create an atomic part, may somebody will assist me please.

I think i got this one done.

aewallin:
Datasheet has also a SOIC package part number LM5001IDRQ1, but maybe that (non lead-free?) is not available anymore?
Your screenshot doesn't show pin types - unfortunately they are hard (for me) to decode from the fileformat...

poeschlr:
> S5.1 and S5.2: I am not sure how to create an atomic part, may somebody will assist me please.

under atomic i think the KLC simply means a fully specified part (one where you have the footprint already defined)
I tried to document that better but did not find a wording that was fine with everyone. Maybe i will be motivated beginning next year to give it another try.

rs-development:
> Datasheet has also a SOIC package part number LM5001IDRQ1, but maybe that (non lead-free?) is not available anymore?
> Your screenshot doesn't show pin types - unfortunately they are hard (for me) to decode from the fileformat...

On the TI website only the SOIC MA type and the WSON packages are listed
screenshot from 2018-12-23 19-15-45
[http://www.ti.com/product/LM5001/samplebuy](http://www.ti.com/product/LM5001/samplebuy)

This should help :)
screenshot from 2018-12-23 19-14-52

aewallin:
GND is usually type Power Input - but I haven't checked how it is done in other switched-mode parts

poeschlr:
Yes gnd should be power input. It is shared by the input and output side. But input takes precedence. (After all if it is not connected to outside power than it can not supply its output side.)

myfreescalewebpage:

- [ ] In the description, we usually dont insert space before the amp unit: `1A, High Voltage (75V) Switch-Mode Regulator, Adjustable Output Voltage, SOIC-8`
- [ ] Pin 1 (SW) is "Open Collector"
- [ ] Pin 3 (VCC) should be "Passive" to avoid ERC errors
- [ ] Pin 5 (RT) should be "Passive" to avoid ERC errors

KLC S4.2 accepted for pin VIN only.

Cheers,
Joel

myfreescalewebpage:
Hi @rs-development are you still alive on this PR ? Thanks.

rs-development:
If you don't mind, please keep it open. I will finish it asap.

myfreescalewebpage:
No worries @rs-development that was only to get news from you ! We keep the PR opened !

evanshultz:
We don't need to copy somebody else's symbol design as our main concern is all KiCad symbols looking nice (as best we can...), so you can move the power pin to the top.

The trailing "x" in the symbol name is for the packaging, so leave it off.

Lastly, in the schematic above C51 has a value of "56k". That's a rather rare value for a cap.

@myfreescalewebpage 
It does look weird. Since this is a very flexible IC, the feedback can be done in various ways. See figure 19 in the datasheet for a drawing and explanation of the flyback application.

myfreescalewebpage:
@evanshultz 

> so you can move the power pin to the top.

Usually power regulator have in on the left and OUT on the right and similar devices to this one are already like this.

evanshultz:
For linear regulators, yes. But this isn't a linear regulator.

Our check scripts only flag power pins on top and bottom as a warning, not an error, but here I don't see any reason why VCC can't move to the top to eliminate the warning.

myfreescalewebpage:
I agree, but many ICs should be reworked in this case. This trouble me a little bit because design will be different if it is a linear or a switching regulator...

evanshultz:
We defined the word "regulator" to mean a symbol where voltage comes in and goes out of. A "controller" would be the IC that controls a power supply, like this one, even though it has an included switch. See the various Regulator_ libraries to see the ways we segregated symbols in v5.

A buck converter controller IC won't necessarily always have the input on the left and output on the right. Even if there is a lot of old cruft, new contributions will hew more closely to KLC. I typed "buck" into the symbol chooser in Eeschema and grabbed the first symbol that showed up:
image

It has power pins on the top and the bottom. And it also has an included switch.

IMO, if there's no reason to violate KLC why not adhere?

Are you thinking of something else? Maybe we're not talking about the same thing.

myfreescalewebpage:
I have a clear view of how many parts are violating the rule for no reason, but that's clear for me. The screenshot you have chosen is a good example !

myfreescalewebpage:
@evanshultz complemntary question, are battery charger circuits with input on the left and output on the right accepted ?

PS : fixed required modification to @rs-development in my above list

evanshultz:
@myfreescalewebpage 
I just searched for "buck" and picked the first one. Honestly, it supported what I recalled and expected so I went with it. It may well be a bad example.

We have generally taken the stance to make symbols as usable as they could be. Keeping a bus together regardless of signal flow (like MOSI and MISO on a SPI bus) wins. And if some signals all go to a controller IC, perhaps in the battery charger example you gave, they can be grouped on either side to make the symbol most visually pleasing. But if there's no other factors involved, I goes on the left and O on the right. This is a case-by-case area and it's hard to capture with words. I think general statements about what the library is trying to achieve and taking feedback from contributors who actually are making and using the symbols are the important parts to balance. Maybe Rene has better ideas?

myfreescalewebpage:
@evanshultz thanks, agree with you. I'm not thinking to a specific example, just say I have already seen several practices.

evanshultz:
@myfreescalewebpage 
No doubt! There is still a lot to do...

myfreescalewebpage:
Yes there is!

rs-development:
All points should be fixed, checks fails with:

```
     - Pin VCC (3) @ (400,-300) is of type PASSIVE
```

>     * Pin 3 (VCC) should be "Passive" to avoid ERC errors

myfreescalewebpage:
New review after the last modification, ok for me, merging.

PR #1323, opened 2018-12-23
Fix fp assignment for XPT2046QF (corrected ep size)

poeschlr:
The datasheet assigned to the symbol is not english. I am not really motivated enough to search one that is.
http://www.xptek.com.cn/uploadfile/download/201707171401161883.pdf


evanshultz:
The footprint change looks good. Would you mind changing the pin name offset to 20mil?

poeschlr:
I would really like these fixes to stay focused on fixing the footprint stuff. I have so many of these to do i doubt i will even find time to do them all as is. If we add additional workload on them then i will never get them done.

evanshultz:
Got it.

PR #1328, opened 2018-12-26
TPS7A7200RGW should be in Regulator_Linear.

matthuszagh:
I made a mistake with this earlier. It is an LDO regulator not a switching regulator and should really be in the Regulator_Linear library, not the Regulator_Switching library.

www.ti.com/lit/ds/symlink/tps7a7200.pdf

I understand that this breaks existing designs. However, since the original branch was merged only 7 days ago the damage should be minimal. Let me know of any thoughts/concerns.

------------


myfreescalewebpage:
@matthuszagh thanks for this, we have missed the regulator type in the previous PR.

No comment to do, can you just fix the branch conflict (due to another merge) so I can merge the modification in Regulator_Linear ?

Joel

poeschlr:
This would even be acceptable if the original symbol has been part of kicad for a long time. (Reason: This fixes a bug that could lead to users misinterpreting what the symbol stands for.)

myfreescalewebpage:
Agree with @poeschlr

matthuszagh:
Should be fixed now. Thanks for explaining that, @poeschlr.

matthuszagh:
@evanshultz done.

PR #1333, opened 2018-12-27
Zilog Z80 PIO Parallel I/O DIP-40 

b1ackmai1er:
Zilog Z80 PIO Parallel I/O DIP-40 

Second pass with Travis fixes.
http://www.zilog.com/docs/z80/um0081.pdf
Regards Phil
------------


b1ackmai1er:
Datasheet extract vs schematic

image

aewallin:
- your commit has changes to .github/PULL_REQUEST_TEMPLATE.md, which is not desirable. The comments here in the github discussion are not in the actual git-repo. could you revert those changes?
- Interface_PIO doesn't say much to someone who isn't a Zilog-enthusiast - could these symbols go into an existing MCU-library instead?
- if you take a screenshot from the symbol editor it would show pin-types also, which would help review.

aewallin:
There is an existing symbol CPU:Z80CPU (DIP-40) - would these go into CPU also?

b1ackmai1er:
Hi, CPU:Z80CPU (DIP-40) is in the correct catagory in my opinion unless a new catogory for Zilog Z80 family is created. In which case the Z80, PIO, CTC and DMA chip would all be included.

After looking further I can see similar devices by other manufacturers (8255,6821 ) are in interface.lib. 

So PIO could go in there as well. Ideally a new catagory interface_parallel (instead of PIO) could capture these devices. However, if there is a desire to not extend the catagories then I think the way forward is to put PIO in interface.lib.

Let me know and I will submit update. Thankyou.

b1ackmai1er:
image

myfreescalewebpage:
Hi @b1ackmai1er thanks for this contribution and @aewallin thanks for the first steps review.

@b1ackmai1er can you update your branch, you have a conflict to solve, then will be able to make the review for this PR.

Thanks
Joel

b1ackmai1er:
Hi Joel, Try now. Made some mistakes along the way - still learning. Regards Phil

myfreescalewebpage:
@b1ackmai1er thanks for the update.

I'm troubled by the name of the symbol. The datasheet don't seems to be the right one. Looking at Z8420 and Z84C20, I have found this document: https://www.zilog.com/appnotes_download.php?FromPage=DirectLink&dn=PS0180&ft=Product%20Specification%20(Data%20Sheet)%20%20&f=YUhSMGNEb3ZMM2QzZHk1NmFXeHZaeTVqYjIwdlpHOWpjeTk2T0RBdmNITXdNVGd3TG5Ca1pnPT0= which is better I think. Is it ?

Also, in this case, the device should be named Z8420 with a single alias called Z84C20. Z80PIO is not a known reference from Zilog according to me.

I also agree with that this device is betetr in Interface.lib. Please, move it there.

Joel

b1ackmai1er:
All requested changed completed. Yes your reference data sheet is better. Thank you.

myfreescalewebpage:
Thanks, let's start a full review now:

- [ ] Name of pins A0 to A7 should be PA0 to PA7
- [ ] Name of pins B0 to B7 should be PB0 to PB7
- [ ] Graphic Style of pin 25 should be Line
- [ ] The height and width of the symbol should be reduce a bit, there is too much blank space lost

Else, seems ok for me.

Thanks,
Joel

b1ackmai1er:
Hi Joel, think this is done now. Made update with Kicad 4.07 and I think it messed stuff up so had to restore so left a bit of a mess in between. Regards Phil

myfreescalewebpage:
@b1ackmai1er something is not correct, you have now twice the definition, each with the aliases, please look at the git diff at https://github.com/KiCad/kicad-symbols/pull/1333/files.

Yes I strongly recommend you to install KiCad 5 !

Joel

b1ackmai1er:
Thanks Joel, fixed. Sorry yes I am running both 4.07 and 5.x at the moment. Will do any further work on 5.x branch. Once again, thanks for your patience.

myfreescalewebpage:
No worries @b1ackmai1er thanks for the fixes ! No more comment to do, merging.

PR #1335, opened 2018-12-27
Fix QFN-16-1EP 4x4 footprint assignments in regulator_switching

poeschlr:
Fixes footprint assignments in the regulator switching library.

New footprints are in https://github.com/KiCad/kicad-footprints/pull/1210


poeschlr:
footprint merged

PR #1336, opened 2018-12-27
Add TUSB320, TUSB321, TUSB322I

EvilMav:
A series of USB Type-C Configuration Channel Logic ICs. TUSB320 is especially popular to easily convert USB2 applications to Type-C.

Depends on X2QFN package [PR 1221](https://github.com/KiCad/kicad-footprints/pull/1221)

screenshot_tusb320

Datasheets:

* www.ti.com/lit/ds/symlink/tusb320.pdf
* www.ti.com/lit/ds/symlink/tusb321.pdf
* www.ti.com/lit/ds/symlink/tusb322I.pdf

------------


poeschlr:
New footprint pull request (scripted): https://github.com/KiCad/kicad-footprints/pull/1234

myfreescalewebpage:
Hi @EvilMav, thanks for contributing,

TUSB320 / TUSB320I:
- [ ] You should remove leading star of the footprint filter: `X2QFN*1.6x1.6mm*P0.4mm*`
- [ ] CC1 and CC2 should be Bidirectional
- [ ] SDA/OUT1 and SCL/OUT2 should be Bidirectional
- [ ] Reference should be in the top left corner, and name in the top right corner of the symbol for consistency with other devices
- [ ] The positions of the pins is globally strange, something like the schematic page 21 is probably better

TUSB321:
- [ ] You should remove leading star of the footprint filter: `X2QFN*1.6x1.6mm*P0.4mm*`
- [ ] CC1 and CC2 should be Bidirectional
- [ ] Reference should be in the top left corner, and name in the top right corner of the symbol for consistency with other devices
- [ ] The positions of the pins is globally strange, something like the schematic page 14 is probably better

TUSB322I:
- [ ] You should remove leading star of the footprint filter: `X2QFN*1.6x1.6mm*P0.4mm*`
- [ ] CC1 and CC2 should be Bidirectional
- [ ] SDA/OUT1 and SCL/OUT2 should be Bidirectional
- [ ] Reference should be in the top left corner, and name in the top right corner of the symbol for consistency with other devices
- [ ] The positions of the pins is globally strange, something like the schematic page 21 is probably better

Cheers,
Joel

EvilMav:
Hey @myfreescalewebpage,
thank you for taking the time to review and welcome to 2019 ;)

Many of the things you mention were done on purpose, this was my reasoning:
[] The assigned default footprint is TI-specific and therefore begins with Texas_X2QFN. I included the leading star to support the prefix, or does KiCad imply a leading star by itself?
[] CC1/CC2  - agree, I/O would work here too
[] SDA/OUTx - I2C is always an open-drain in output mode, and the datasheet specifies them to be open-drain in output mode too on page 4
[] Reference position - OK, will do
[] KLC S4.2 does not allow me to do as on page 21, as only power pins are supposed to be on the bottom. I have placed microcontroller-facing pins to the left and USB-facing pins to the top right. PORT, CURRENT_MODE and ADDR pins will be shorted to either ground or V+ for most if not all use cases, so I distributed those on the lower positions on both right and left - they could be all placed on the left but it would make the symbol unnecessary large, considering I do have a lot of space to the right bottom. ~EN is shown tied to ground on page 21, but it is to be expected it will be driven by the microcontroller in some applications, so I did always place it to the left just below I2C pins.

EvilMav:
kicad_screen_tusb320_v2

myfreescalewebpage:
Hi @EvilMav 

Thanks and best wishes for the new year too!

> The assigned default footprint is TI-specific and therefore begins with Texas_X2QFN

Right, thanks to point me that, I have look at other devices and corrected my previous comment, the footprint filter should be `Texas?X2QFN*1.6x1.6mm*P0.4mm*`

> SDA/OUTx - I2C is always an open-drain and the datasheet specifies them to be open-drain in GPO mode too on page 4 in the description texts

Agree with GPO output when I2C is not used, but the pin is also an input when using I2C. So it is an Input and and Open Drain Output, this makes it Bidirectional.

> Reference position - OK, will do

TUSB321 has not the same reference position compared to the two others. It's better to get consistency between similar devices. Just need to move a little bit to the left please.

And finally regarding to pins position, agree that only power pin can be placed on top and bottom of the symbol. However, the others are placed a little bit randomly, for example C1/C2 are not always at the same place etc. that's not very convenient if I want to change a symbol by another one.
Can you choose one position for each pin, i.e. for example SDA and SCL and INT the three first pin in top left corner, then another then another etc etc. PORT always on the left also for example.

Thanks,
Joel

EvilMav:
You're right about swapping those parts - it would have been a pain. Tried to keep pin positions as consistent as possible in this version. Changed I2C to bidirectional. Included "Texas?"  into the filters.
kicad_screen_tusb320_v3_1
kicad_screen_tusb320_v3_2
kicad_screen_tusb320_v3_3

myfreescalewebpage:
@EvilMav looks really great, thanks for this new and very useful symbols !

No more comments, merging.

Cheers
Joel

diegoherranz:
I'm curious why we have included the "I" variants which, unless I've read it wrongly, only differ on the (industrial) temperature range. Shouldn't that be, for example, `TUSB320` alone or `TUSB320x` according to [KLC S2.2](http://kicad-pcb.org/libraries/klc/S2.2/)?

By the way, in this case where the temperature suffix can be blank (commercial) or "I" (industrial), we could make KLC clearer to specify what we prefer, e.g. `TUSB320` or `TUSB320x`. I can update it if we reach a consensus.

Thanks!

myfreescalewebpage:
@diegoherranz you're right, however, both TUSB320 and TUSB320I exist while only the TUSB322I is available because this one is only Industrial temperature. Even the datasheet is "tusb322i.pdf".

KLC S2.2 says it "should" be replaced with wildcard so in my understanding a part of freedom is given to the submitter.

EvilMav:
@myfreescalewebpage 
One other thing is there is no other letter instead of I, which might make the X a little confusing. Nevertheless, if an X is preferred, I do not mind to revive the branch and make a change.

PR #1337, opened 2018-12-28
Adds TI AM26LV31E line driver symbol

RobertBaruch:
Datasheet: http://www.ti.com/lit/ds/symlink/am26lv31e.pdf


myfreescalewebpage:
Hi @RobertBaruch , thanks for contributing,

- [ ] Pins length should be 100mil (symbol < 100 pins)
- [ ] "Z" output pins should use standard graphic lines, with `~` at the beginning of the name to get the over-bar
- [ ] Footprint filters are not correct (see also travis output), however, I have a more general problem with the packages, because the current name of the device cover more than only SOIC/TSSOP packages. A QFN version also exists and it will be a problem when added to the lib... I would like to suggest adding devices separately without using aliases. Names can be `AM26LV31ExD` for SOIC version, `AM26LV31ExPW` for TSSOP version and will be `AM26LV31ExRGY` for future QFN version, according to device config page 3 of the datasheet.

Cheers,
Joel

myfreescalewebpage:
Hi @RobertBaruch are you still alive on this subject ? Thanks.

myfreescalewebpage:

Issue #1331, opened 2018-12-27
EMI_Filter_LL: KLC issues


aewallin:
- has a datasheet link - should generic Device-parts have datasheets?
- no footprint defined (a datahsheet link would tell me a specific part, with a footprint, is meant?)
- footprint field visible
- datasheet field visible
![emi_ll](https://user-images.githubusercontent.com/909509/50488953-b7710200-0a0d-11e9-93fb-d416c0c5f10f.png)

poeschlr:
You are right about all 4 points.

PR #1168, opened 2018-11-24
add murata bnx025 symbol

aewallin:
add one symbol: Murata BNX025 EMI-filter (LC-type)

Datasheet:
https://www.murata.com/en-us/products/productdetail.aspx?cate=luNoiseSupprFilteBlockType&partno=BNX025H01%23
Catalog, page 223/245:
https://www.murata.com/~/media/webrenewal/support/library/catalog/products/emc/emifil/c31e.ashx?la=en-us

Screenshot:
bnx025_symbol

Footprint pull-request is here: 
https://github.com/KiCad/kicad-footprints/pull/1106

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1338, opened 2018-12-28
add Bourns common-mode EMI filter

aewallin:
https://www.bourns.com/docs/Product-Datasheets/SRF0905.pdf

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1220
Symbol based on EMI_Filter_LL

emi_filter_bourns_srf0905

------------


myfreescalewebpage:
Hello @aewallin 

Thanks for this part. However, I wonder about the place of it in the library, it is probably better to use a generic symbol for it. What do you think ? Question also for more experimented librarians, @poeschlr @evanshultz 

The library has L_Core_Ferrite_Coupled in the Device lib, but the pin numbering is not the same. Maybe something like it is done for LED_Dual_AACC / LED_Dual_ACAC etc is nice for this EMI filter too ? We have to think about the naming of the device.

Cheers,
Joel

aewallin:
I could use a generic symbol like EMI_Filter_LL, but it has issues #1331 , and the pin-numbering is wrong relative to this Bourns part.

evanshultz:
I'd prefer generic symbols in Device.lib, too. We could have a handful of generic symbols or a nearly-infinite number of specific ones. Exactly the reason to have generic symbols...

Since the ends of the windings need to match but the pin numbers, don't matter, we could just have two generic symbols (one with windings using pins 1-2 and 3-4 and the other with windings on 1-3 and 2-4). However, in practice it's probably easier if we had 4 generic symbols (1-2 and 3-4, 1-3 and 2-4, 2-1 and 4-3, and 3-1 and 4-2) to consider the dot placement as well. This should give us good coverage of parts on the market so those generic symbols could match with many parts and their unique footprints.

Other thoughts? Have I missed something?

For naming, I suppose we could suffix with the pin numbers `_1234`, `_1324`, etc. There's probably something better but it's not coming to me right now.

@aewallin 
The points you raised about the existing symbol need to be corrected. Good catches.

aewallin:
revised as suggested.

What do you think about #1168 - it is similar but more complex "LLCLC" - if you name it after the internal parts.

evanshultz:
Alright. :) Let's see what everybody else says.

IMO once we get to the level of parts integrated into #1168 we should probably go with a specific part and not generic. There must be far fewer integration component filters with that specific configuration and simple common mode chokes.

myfreescalewebpage:
@evanshultz 

> Other thoughts? Have I missed something?

No I think it's okay.

> For naming, I suppose we could suffix with the pin numbers _1234, _1324, etc. There's probably something better but it's not coming to me right now.

Alright for me too, I don't find a better solution.

@aewallin 

> I could use a generic symbol like EMI_Filter_LL, but it has issues #1331 , and the pin-numbering is wrong relative to this Bourns part.

Also seen this issue. If you want to submit the different symbols @evanshultz suggested, this will also permit to close it (8 symbols but easily created by copy/paste). Have you got time to do it ?

> What do you think about #1168 - it is similar but more complex "LLCLC" - if you name it after the internal parts.

Same position of @evanshultz , very specific and difficult to say it's generic. I agree this kind of device should be specific.

Regarding to your submission:

- [ ] You are currently violating S6.3, you can fix it by setting datasheet link to `~`

Else sounds good.

Cheers,
Joel

aewallin:
Thanks (again) for review.

With the new symbols I would prefer a pragmatic approach: when I come across a part that I want to use that has different pin-ordering from this Bourns part or the existing #1331 part I will create a new one.

myfreescalewebpage:
> With the new symbols I would prefer a pragmatic approach: when I come across a part that I want to use that has different pin-ordering from this Bourns part or the existing #1331 part I will create a new one.

Ok!

No more comment on your submission, merging.

Joel

PR #1340, opened 2018-12-28
Addition Of Traco DC-DC Converters

jamesrlucas:
g)
Added the following ranges of Traco DC-DC Converters.

    Traco TMR2 Range of DC-DC Converters - https://www.tracopower.com/products/tmr2.pdf
    Traco TMR2WI Range of DC-DC Converters - https://www.tracopower.com/products/tmr2wi.pdf
    Traco TEC2WI Range of DC-DC Converter - https://www.tracopower.com/products/tec2wi.pdf

TMR2WI Range of converters are awaiting a footprint (submitted under request https://github.com/KiCad/kicad-footprints/pull/1222 on kicad-footprints)

screenshot from 2018-12-28 16-47-55

------------


jamesrlucas:
>     * [ ]  These parts should be moved to Converter_DCDC library
> 
>     * [ ]  The name of the footprint has "TMR-", so I suggest to use "-" instead of "_" in the part names
> 
>     * [ ]  We now represent the isolation barrier with a single dash line, can you remove one ?

Will do :)

> TMR_0510:
> 
>     * [ ]  You have forgotten On/Off Input. You can set name of the pin to "ON/OFF" only, this will permit to center the isolation barrier, you can have a look to TEN20-2421WIN to copy.

Whoops. my mistake!

>     * [ ]  Pins 5 and 8 should be added to the symbol, type Not Connected, Not Visible

Could this be added to the library conventions on the Kicad website please? I've been using the recommendations there, and it says either miss them, or place NC/NV pins
Just to avoid further confusion!

> TMR_2-2410WI:
> 
>     * [ ]  The footprint "Converter_DCDC_TRACO_TMR-2xxxxWI_THT" do not exist, I haven't find any footprint PR to submit it. Have you planned to do it ?

Footprint is under PR KiCad/kicad-footprints#1222

> 
> TEC_2*:
> 
>     * [ ]  They can be all alias of the TMR_0510, this remove one symbol definition
> 

Can I do this when they have different footprints/pin numbers (I've not worked out how to do it if there is)

I'll try and get the changes made over the weekend.

James

myfreescalewebpage:
> Could this be added to the library conventions on the Kicad website please? I've been using the recommendations there, and it says either miss them, or place NC/NV pins

In fact there is http://kicad-pcb.org/libraries/klc/S4.5/, but footprint recommandation prefer to not put the number of pads in the footprint filter. So need to put all NC pad, invisible.

> Footprint is under PR KiCad/kicad-footprints#1222

Thanks for the link !

> Can I do this when they have different footprints/pin numbers?

No this applies only if pad and footprint are the same. Here TEC_2* and TMR_0510 have the same footprint and same pin number according to the datasheets. Let me known if I'm making a mistake.

Joel

jamesrlucas:
> In fact there is http://kicad-pcb.org/libraries/klc/S4.5/, but footprint recommandation prefer to not put the number of pads in the footprint filter. So need to put all NC pad, invisible.
I'll get that done as soon as I get a chance then!

> No this applies only if pad and footprint are the same. Here TEC_2* and TMR_0510 have the same footprint and same pin number according to the datasheets. Let me known if I'm making a mistake.
Nope, I understand now. Should I move the existing symbol to the Converter_DCDC library as well? (And any other DC-DC Converters in that library), or will that cause problems?

I'm more than happy to contribute, I feel I should give back something to the community after making use of KiCad in a few of my own projects. I've a huge number components I have yet to add...

myfreescalewebpage:
> Should I move the existing symbol to the Converter_DCDC library as well? (And any other DC-DC Converters in that library), or will that cause problems?

Only move the device you submit for this PR. IF you found other devices that should be moved, please do it in another PR, to get it easier to review for us.

> I'm more than happy to contribute, I feel I should give back something to the community after making use of KiCad in a few of my own projects. I've a huge number components I have yet to add...

That's a pleasure to heard that :) We are very busy so this takes sometimes a long time to get the PR merged, but we are happy to know people want to contribute !!

jamesrlucas:
All done for you Joel, hopefully I've got it right this time!

James

myfreescalewebpage:
@jamesrlucas very good job done here, no more comment to do on these symbols ! Merging.

PR #1345, opened 2018-12-28
Fix fp assignement for PIC18 MCUs (qfn-44)

poeschlr:
These had a package with 6.6x6.6mm ep assigned but need one with 6.45x6.45mm

This was an oversight in my first round of scripting footprints.


evanshultz:
`PIC18F4331-IML`
EP should be 6.8x6.8mm.
http://ww1.microchip.com/downloads/en/devicedoc/39616b.pdf#page=380

`PIC18F4455-IML`
6.55x6.55mm
https://ww1.microchip.com/downloads/en/devicedoc/39632c.pdf#page=409

`PIC18F4458-IML`
Where did you find the package info? It's not in the datasheet (http://ww1.microchip.com/downloads/en/devicedoc/39887b.pdf) nor specified exactly at https://www.microchip.com/wwwproducts/en/PIC18F4458.

poeschlr:
The datasheets you link above are different than the once assigned to the symbols. Does that mean the symbols are out of date in this regard?

PIC18F4458-IML is indeed strange. if you click on the package name on its order page then you get to a list of available qfn footprints. (https://www.microchip.com/quality/packaging-specifications?packageFamily=QFN) There is only one 44 pin package listed in this list. That is what i used for this part. (Yes not really the best option. We can create an issue that this one needs more research.)

evanshultz:
Huh. I just Googled those PNs and opened the datasheet. But you're right that the date is newer and also the letter at the end of the document number (probably the revision) is later also. So the first two are all good with me then. Thanks for sorting that out.

For `PIC18F4458-IML`, http://ww1.microchip.com/downloads/en/DeviceDoc/44L_QFN_8x8_ML_C04-0103D.pdf is the ML package so I'll merge it.

PR #1350, opened 2018-12-29
Add bq27750 (finish #471)

poeschlr:
This PR finishes the work started in #471


poeschlr:
SDA/SCL are normal i2c slave pins. input is the best pintype here that gives the least amount of false positves. See https://github.com/KiCad/kicad-website/issues/342 (And the discussion it links to)

Changed BAT and VCELL position. changed VCELL/INT type.

myfreescalewebpage:
Hi @poeschlr 

- [ ] I would say footprint filter should be `Texas?S*PDSO*` according to some other symbols in the lib
- [ ] We usually put SDA above SCL
- [ ] We usually put reference in the top left corner and name in top right corner of the symbol

Cheers
Joel

poeschlr:
The reference and value fields are all over the place in the lib. We also do not state any preference in the KLC. (The alignement used by this symbol is present in some of the screenshots of the KLC) Meaning we can not hold any contributor to a particular rule.
We can make a new rule for that but i am not sure that we should as these fields are auto placed anyways.

We rarely use "?" in footprint filters. We have 4 filters starting with Texas? Nearly 100 with Texas*
So i chose to overrule you and use Texas*

Exchanged SDA and SCL (I did not care to double check you on that as it can not be automated.)

PR #1352, opened 2018-12-30
Add ST VNH5019A-E motor driver  (finish #797)

poeschlr:
Finishes #797 

datasheet: https://www.st.com/resource/en/datasheet/vnh5019a-e.pdf

screenshot from 2018-12-30 14-09-18


myfreescalewebpage:
Hi @poeschlr 

- [ ] Description should end with the package name: `, MultiPowerSO-30`
- [ ] You are violating S4.3, VCC pin 13 should be invisible (it is stacked)
- [ ] EN/DIAGA and EN/DIAGB are Open Collector
- [ ] I would probably swap group INA/INB/PWM and group EN/DIAGA/EN/DIABG so it's easier to wire the pull up of EN/DIAGx poins.
- [ ] I would also probably swap groups GNDA and GNDB and make the symbol a little bit larger to have GNDA/GNDB centered

Also regarding CP, even if its a KLC error, can we consider moving to the top of the symbol between VBAT and VCC ? Have a look at the schematic in the datasheet page 15.

Finally, I realize heat slug 2 and 3 have been crossed in the footprint (Heat Slug 2 is pin 33 and Heat Slug 3 is pin 32). Don't know if you want to change it ? Currently your symbol is consistent with the footprint but in the future I think other symbols can be incorrect if we don't pay attention to this. Only one other symbol is using this footprint : VNH2SP30, and it is also consistent with the footprint.

Cheers,
Joel

poeschlr:
The two parts that use the package use different ways to number the heatsink. (page 6 for both of them)

- https://www.st.com/resource/en/datasheet/vnh5019a-e.pdf
- http://www.st.com/resource/en/datasheet/vnh2sp30-e.pdf

myfreescalewebpage:
> The two parts that use the package use different ways to number the heatsink. (page 6 for both of them)

Not seen this, I was afraid of possible future error but in this case, I haven't said anything :)

poeschlr:
I do not agree with making the symbol larger for centering power pins.
Switching en/diag groups with the other inputs would mean they collide with VBAT. Moving them there would mean we need to make the symbol even larger towards the left (going even more against your comment about centering gnd.)

myfreescalewebpage:
> I do not agree with making the symbol larger for centering power pins.
Switching en/diag groups with the other inputs would mean they collide with VBAT. Moving them there would mean we need to make the symbol even larger towards the left (going even more against your comment about centering gnd.)

I haven't try to see the resutl but I agree if this is too large it does not make sense to have plenty of lost space.

myfreescalewebpage:
@poeschlr what about the description regarding to the package name ?

poeschlr:
description updated

myfreescalewebpage:
@poeschlr okay for me, merging.

PR #1261, opened 2018-12-14
Add TPS5420D.

matthuszagh:
TPS5420D
http://www.ti.com/lit/ds/symlink/tps5420.pdf

image

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1363, opened 2019-01-02
Added symbol for Silicon Labs Si512/513 dual frequency oscillator

fhw72:
Added symbol for Silicon Labs Si512/513 dual frequency oscillator with 6 pin (3.2x2.5mm²) package.

------------

- [x] https://www.silabs.com/documents/public/data-sheets/si512-13.pdf
- [x] https://github.com/KiCad/kicad-footprints/pull/1261

screenshot 2019-01-02 22 36 59
screenshot 2019-01-02 23 00 11


fhw72:
Hi Joel,
thanks for taking your time to review.

I have applied the suggested fixes except the naming. I'm not sure if renaming the part
to (in this case) 512C/G/KxxxxxxxxAxxx makes much sense!?
(C,G,k -> Single CLK output, everything else -> differential output).

Neither the voltage, nor the temp. range should play a role for the part. The only
important things I see are:

- Single or differential CLK output
- Package "Housing" size

What do you suggest?

PS: The same problems partially apply to Si570/Si571 which I took as template... I guess there's no
easy solution? (Si57x also got Vdd/VDD wrong, naming etc).

fhw72:
I just added "Single-ended CMOS output" to separate the part submitted from other possible variants.

myfreescalewebpage:
Hi @fhw72 

The other Si57x devices have the issue too, but variants must be clearly separated to avoid confusion and the best way to achieve this is using the full part name of the devices.

Voltage is usually indicated, just look at voltage regulator for example ;) however I agree temperature is not, but this device has all variant with the same temperature range.

I think we need the experience of other librarians, @poeschlr @evanshultz if you can give your opinion about the name of this part. is "Si512AxxxxxxxxCxxx" something valid ? How is it possible to shorten it ? Ordering info of the manufacturer available in the datasheet https://www.silabs.com/documents/public/data-sheets/si512-13.pdf page 13.

Cheers,
Joel

fhw72:
Hi Joel,
just my two cents:

I think that the naming scheme needs/seems to
be a compromise as we have two extremes for naming a device:
The first extreme would be just the name of a device (i.e. Si512) and the other extreme is including all related information (e.g. Ordering info).
Both extremes are not practicable or making sense - obviously.

I agree that including the output voltage of a regulator is certainly a good idea as
this is an essential part of it's specified functionality/purpose.
However I don't think it makes too much sense with an simple oscillator as the
target voltage is "just" a technical requirement of its surrounding target circuit.

So what makes sense to be included in the naming in the case of Si512/513 then?
Let's sum up, what we need to know least to successfully draw a schematic and route a PCB using
this specific oscillator form Silicon Labs:

- Single-ended or differential output? (2 options)
- Three different housings/footprints (3 options)
- Two different options for "FS" (Frequency select) and "OE" (output enable)

The last two options are separated by the device number: Si512 <-> Si513.
The other 6 permutations need to be distinguished by the name.

So my suggestion would something like:

Si512_Single-ended_output_3.2x2.5mm
Si512_Differential_output_3.2x2.5mm
Si513_Single-ended_output_5x7mm
etc.

I we really need to include the voltage too:

Si512_Single-ended_output_3.3V_3.2x2.5mm
Si513_Differential_output_1.8V_5x7mm

Anyway: I'll gladly do whatever needs to be done to contribute that IMHO nice part.
I hope we'll find  a workable compromise. :-)

Cheers,
Frank

fhw72:
Changed footprint reference to:
Scripted fp for Silicon Labs Si512/513 oscillator #1261

fhw72:
Oops... simply an oversight. Just pushed it again.

myfreescalewebpage:
@poeschlr thanks to react. I have quoted your name about the symbol name:

> is "Si512AxxxxxxxxCxxx" something valid ? How is it possible to shorten it ? Ordering info of the manufacturer available in the datasheet https://www.silabs.com/documents/public/data-sheets/si512-13.pdf page 13.

Joel

poeschlr:
we could leave out everything after "C" meaning Si512AxxxxxxxxC
or we use cleartext Si512A_2.5x3.2mm

I think KLC would point to the former with the 2.5x3.2mm note at the end of the description.

myfreescalewebpage:
Thanks @poeschlr looks this is a good intermediate solution.

@fhw72 just made a complete review and you will find below the comments I made:

- [ ] "Si512" should be renamed "Si512A_2.5x3.2mm"
- [ ] "Si513" should be renamed "Si513A_2.5x3.2mm"
- [ ] Footprint filter should be `Oscillator*SMD*Silicon*Labs*LGA*2.5x3.2mm*P1.25mm*` for both devices (no pin count in fp filter)
- [ ] KLC Error S4.1 to be fixed for both devices

Cheers
Joel

fhw72:
Hi Joel,
I've changed the fp filter accordingly and also renamed the device. Please let me know if I it's correct that way.

However I cannot fix KLC S4.1 as I don't have access to the footprint.

@poeschlr Could you please look into this?

fhw72:
Fixed KLC S4.1... just realized that it's not dependent on the footprint.

myfreescalewebpage:
Yes, thanks. Just need to wait the footprint now.

Joel

fhw72:
Hi Joel,
there's a new footprint in a new PR #1261.

fhw72:
Hi Joel,
I (once again checked https://github.com/KiCad/kicad-footprints/pull/1261

and @poeschlr is right. The actual land pattern of the device is on **page 19** and the created footprint
is identical. On page 20 it's just a **suggested** land pattern for this device.

Cheers,
Frank

fhw72:
Hi Joel!
I've now checked four different devices with the same package (6pin 3.2x2.5mm²).
Please see my comment here:

https://github.com/KiCad/kicad-footprints/pull/1261

What do you suggest to get that part into the library?

Cheers,
Frank

myfreescalewebpage:
Hi @fhw72 

See my comment in the footprint PR thread, my position is also yours I think. For this kind of part we should follow the recommandation of the manufacturer. Let see what Rene will answer.

Joel

fhw72:
Thanks! :+1:

myfreescalewebpage:
@fhw72 , seen your response. I have not enough experience to say if it's better to have one of the other solution. I'm clearly convinced that a standard footprint is better but I also agree with your comment that it's good to follow the manufacturer recommendations....

fhw72:
Yes. Therefore I'd recommend to have two footprints anyway. As long as Silicon Labs is supplying parts according to an older standard I'd better stick to that.

myfreescalewebpage:
Footprint has been merged. Merging the symbol.

PR #1367, opened 2019-01-04
Add LTC4156 (USB/LiFePo4 Charger/PMIC)

EvilMav:
[Datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/4156fa.pdf)

I've left a little more space inbetween BATSNS/CHGSNS/GATE because I had it and it would make placing a MOSFET as seen in the datasheet easier.

kicad_ltc4156

**Checker bug?**
I get a KLC S4.4 warning from the checker, which seems to come from passive hidden Vout from the Vout stack. But it is required to be set passive by KLC pin stacking rules. Is it a bug in the checker or did I get something wrong?

------------


myfreescalewebpage:
Hi @AUTHOR, thanks for contributing,

- [ ] You can use capital letters for all pin names
- [ ] I will say VBUS is Passive to avoid ERC error when the battery supply VBUS (depending of ID state).
- [ ] I will say pin CHGSNS and VOUT should be Passive to avoid ERC error. They are both power input and output depending when the battery charges or discharges.
- [ ] Maybe you can cross VOUT and VOUTSNS as shown on the schematic page 15 of the datasheet ?

Cheers,
Joel

EvilMav:
Hi @myfreescalewebpage ,

implemented all the changes. It's a bit of a pity that the implicit power_flag is lost on Vout but it is probably safer to declare it passive indeed.

Ilya

myfreescalewebpage:
Thanks. No more on this PR, merging.

Joel

PR #1373, opened 2019-01-04
RLP-30

aewallin:
old PR was a git-mess. here is a new one.

https://ww2.minicircuits.com/pdfs/RLP-30+.pdf

rlp-30

------------


myfreescalewebpage:
@aewallin footprint and footprint filter not correct here, should be the GP1212 one (description is ok)

Joel

aewallin:
There's an existing GP731_LandPatternPL-176 footprint. This is the same footprint/land-pattern we want here.

The package GP1212 has the same footprint, but it is 0.150" high, instead of 0.100" for GP731
https://ww2.minicircuits.com/case_style/GP731.pdf

Are you suggesting to copying the exact file-content of footprint GP731_LandPatternPL-176 and renaming it GP1212_LandPatternPL-176 - so that the packages match the footprint naming, and so that eventually two 3D models with 0.050" height difference can be used?

OR should we re-name the footprint with two package-names?!  **Mini-Circuits_GP731_GP1212_LandPatternPL-176** ?

myfreescalewebpage:
> so that the packages match the footprint naming, and so that eventually two 3D models with 0.050" height difference can be used?

Yes, I think it's better.

myfreescalewebpage:
> so that the packages match the footprint naming, and so that eventually two 3D models with 0.050" height difference can be used?

Who knows ? Even if the difference of height is very small, this 50% of the package. Duplicating the footprint will avoid confusion in the future, so I want to say yes, it's the best option.

aewallin:
Now waiting for: https://github.com/KiCad/kicad-footprints/pull/1258
After that footprint is merged, footprint and fp-filter can be changed here.

myfreescalewebpage:
No more comment, merging.

Joel

PR #1374, opened 2019-01-04
Added 5400BL15B050E.

matthuszagh:
https://www.johansontechnology.com/datasheets/5400BL15B050/5400BL15B050.pdf

image

Associated [footprint PR](https://github.com/KiCad/kicad-footprints/pull/1259). This fails the tests because the footprint PR has not been accepted yet.

Looking at this again, it seems like this may fit better in the Transformer lib and look more like ADT1-1 below

image

The one thing that's confusing to me about that symbol is it's not clear to me which pins are which (balanced, unbalanced and ground). Maybe the pin that connects to ground should include a small ground "downward triangle" as in the picture below to signal it is the ground pin. The rest of the pins then would also be evident.

image

Let me know what you think is best and I'll amend the PR to reflect that.

------------


matthuszagh:
I went ahead and made the changes mentioned at the end of the original post. The new component looks like this:

image

aewallin:
AFAIK the mini-circ parts are symmetric and it doesn't matter which side is grounded.

Your part clearly has GND pins. I would suggest naming them "GND" also - the downward arrow is OK but to be sure the user understands it the GND-name is motivated I think.

matthuszagh:
@aewallin, I've amended the ground pins and taken out the graphic, which seems superfluous at this point. This violates the origin requirement, but since the symbol is no longer symmetric, I think it makes sense.

myfreescalewebpage:
@matthuszagh thanks, no more comment here, merging. @aewallin thanks for the review, great job here!

PR #1378, opened 2019-01-05
SiT9121 Differential Oscillator.

awygle:
This PR adds the symbol for the SiT9121 differential oscillator, 3.2mmx2.5mm variant.

Footprint PR can be found at https://github.com/KiCad/kicad-footprints/pull/1260.

I followed the lead of the other oscillator symbols here and made this a generic symbol, although the footprint specified is the 3.2mm by 2.5mm variant. I'm not really sure how this is intended to work, but I just did what everybody else seemed to be doing.

Datasheet: https://www.sitime.com/datasheet/SiT9121

Screenshot:
image

P.S. the Python script yelled at me so I made the pin text spacing 20 instead of 40, which doesn't match a single other oscillator symbol as far as I can tell. I'll revert this if you like, or I'll make the change for all the other symbols if you prefer that - the inconsistency annoyed me only marginally less than being wrong.

------------


awygle:
On further consideration I made the names of the pins match the datasheet. Updated screenshot:

image

aewallin:
Yes ref/name placement would be a good guideline to add to the KLC. Not sure how KLC is developed, maybe just a new issue in this repo?

myfreescalewebpage:
Hi @awygle , thanks for contributing,

- [ ] The device can be smaller, something like the ABLNO symbol for example
- [ ] Pin "OE/~ST" is pin number 1, pin NC is pin number 2

Else, just need to wait the footprint.

Cheers,
Joel

myfreescalewebpage:
@awygle any news here ? Thanks for the update, Joel

myfreescalewebpage:

evanshultz:
Undoubtedly these are old symbols and making pin name offset 20mil on all would be nice.

The same thing for the footprint. If you want to add a suffix or some other detail so that the symbol name specifically ties to the default footprint assigned that would be great. This change can only be done on new symbols and not existing ones.

PR #1379, opened 2019-01-05
Add Si53307 clock mux/buffer.

awygle:
Add symbol for Si53307 clock mux/buffer.

I drew out the function on the symbol because it's semi-complex but also quite simple to draw and barely makes the symbol any larger (assuming inputs on the left).

Datasheet: https://www.silabs.com/documents/public/data-sheets/Si53307.pdf
Screenshot: 
image
------------


myfreescalewebpage:
Hi @awygle , thanks for contributing,

- [ ] Leading star and EP size are not added in the footprint filter, it should be `QFN*1EP*3x3mm*P0.5mm*`
- [ ] I do not agree with the selected footprint, it should be `Package_DFN_QFN:QFN-16-1EP_3x3mm_P0.5mm_EP1.7x1.7mm`. This one do not exist. @poeschlr I know you are familiar with the script to generate this kind of footprint, can you submit it for us or indicate how @awygle can achieve this? Thanks.

No other comment, really nice symbol :)

Cheers,
Joel

awygle:
Can you elaborate on the issues with the selected footprint? I chose the current footprint because the datasheet specifically calls out a 1.75mm square thermal pad (see section 6, page 29) and the other pads were also the correct size as specified in the datasheet. What do we gain by going to this other footprint?

awygle:
Also for footprint filter, I added the leading * because there's no meaningful difference between a QFN and a UQFN or a VQFN or... from the perspective of the symbol. I can remove it of course but I feel it's valuable in this case. The EP size I agree with removing and I'll do that soon.

myfreescalewebpage:
Hi @awygle 

That's to not create unwanted question in the future, if you think they are the same, duplicate the footprint, submit it and we will review. But we have always create VQFN/WQFN etc variants, look at the footprint repository.

The best is to create them with scripts available at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_DFN_QFN.

Joel

myfreescalewebpage:
@awygle any news here ? Do you have submitted the new footprint ? if so, can you make a link tp the footprint PR here ? Thanks, Joel

myfreescalewebpage:
@awygle ping :)

evanshultz:
Just to clarify about the footprint type...

The footprints are mapped 1:1 with a 3D model. The various flavors of QFN often vary in package height (V = very thin, etc.) and so having unique footprints will allow accurate 3D models to be generated and displayed. Because QFN footprints and 3D models are generated by script they're easy to make and maintain so we create all flavors as needed.

myfreescalewebpage:

PR #1381, opened 2019-01-06
Add MAX5215 and MAX5217

antonblanchard:
Add symbols for the MAX5215/MAX5217 14-/16-Bit, Low-Power, Buffered Output,
Rail-to-Rail DACs with I2C Interface

https://datasheets.maximintegrated.com/en/ds/MAX5215-MAX5217.pdf

image


myfreescalewebpage:
Hi @antonblanchard , thanks for contributing,

- [ ] Pin 1 (REF) should be Power Input, and move to the top of the device (I know other devices do not respect this rule)

Cheers,
Joel

antonblanchard:
Thanks Joel, that makes more sense to me. How does this look?

image

myfreescalewebpage:
@antonblanchard thanks, sounds good for me. No more comment, merging.

Joel

PR #1384, opened 2019-01-06
fix part spelling for TPS79333-EP

frantony:
According to datasheet (http://www.ti.com/lit/ds/symlink/tps79333-ep.pdf)
the correct name for 3.3V version is TPS79333-EP.

Signed-off-by: Antony Pavlov <antonynpavlov@gmail.com>


frantony:
@myfreescalewebpage 
Done!

myfreescalewebpage:
Thanks, merging.

PR #1387, opened 2019-01-06
Added ADA4940-2.

matthuszagh:
https://www.analog.com/media/en/technical-documentation/data-sheets/ADA4940-1_4940-2.pdf

image

image

The footprint for this is in [a PR](https://github.com/KiCad/kicad-footprints/pull/1269).

------------


matthuszagh:
image

image

aewallin:
pin25 EP (passive) should maybe be visible here, datasheet leaves the option of either -V or GND connection?

matthuszagh:
You're right, it should be. Fixed now.

image

matthuszagh:
No problem. Should I also move this to Amplifier_Operational like LMH6551?

matthuszagh:
I've left the schematic in Amplifier_Difference unless you think it should be moved.

image

image

image

image

evanshultz:
I'm good with that library.

I played around with the symbol and bit, trying to make it fit with the smaller triangle symbol, but I can't find a way to make it work.

I also don't find a way to stack the logic and power pins. It could probably work out OK without the EP pin, but we need it visible. Do you see any way to do it?
image

I'm thinking about the pin type for the FB pins. I guess Output works. It could be other types but it likely won't make any difference in the real circuit so let's leave it.

evanshultz:
Yeah. And the DISABLE pins between the inputs are annoying. But we gotta decide this symbol good enough at some point if inspiration doesn't strike.

Correct, I meant the FB pins. They appear to be internally connected to the OUT pins, which are Output type, so that's why I was thinking they'd be the same type. But again, I'm not sure the type used here will have any effect in practice.

So... are you ready to merge?

PR #1388, opened 2019-01-06
Fix some invalid NXP URLs

antonblanchard:
A number of URLs no longer resolve after the NXP to Nexperia
change. Fix them.

### DIFF
--- a/4xxx.dcm
+++ b/4xxx.dcm
@@ -33,7 +33,7 @@ $ENDCMP
 $CMP 40106
 D Hex Schmitt trigger inverter
 K CMOS
-F http://www.nxp.com/documents/data_sheet/HEF40106B.pdf
+F https://assets.nexperia.com/documents/data-sheet/HEF40106B.pdf
 $ENDCMP
 #
 $CMP 4011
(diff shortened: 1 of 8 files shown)


antonblanchard:
The BC337 URL still doesn't resolve, looks like they removed it from the datasheet:

https://assets.nexperia.com/documents/data-sheet/BC817W_SER.pdf

evanshultz:
Thanks for this nice fix! I assume this was all search and replace within the text files?

The link in your last comment confuses me because BC337 is in a TO-92 package. How about using the same datasheet for BC338: https://diotec.com/tl_files/diotec/files/pdf/datasheets/bc337.pdf? If you can update BC337's datasheet this looks good to me.

antonblanchard:
@evanshultz Yes, it was a search and replace. I only modified the URLs that were obviously bad (they return a one liner "File not found.")

Update pushed with your suggestion to use the Diotec bc337 datasheet. Thanks.

evanshultz:
Thanks! Looks good!

In the future, we would prefer incremental commits so that it's easier to track what happened commit-by-commit.

antonblanchard:
Thanks @evanshultz, will do so next time.

PR #1389, opened 2019-01-06
Fix 4046 pinout

antonblanchard:
The current symbol has the RefIn and SigIn pins reversed.

The datasheet:

https://assets.nexperia.com/documents/data-sheet/HEF4046B.pdf


myfreescalewebpage:
Thanks @antonblanchard for this fix ! Merging.

Joel

PR #1392, opened 2019-01-07
Move UL865 from RF_Module to RF_GSM

jneiva08:
Move SARA-U2xx, SIM7020x, SIM800C, SIM900 From RF to RF_GSM

------------


myfreescalewebpage:
@jneiva08 footprint library of SIM900 not updated. Else sounds good.

Joel

jneiva08:
@myfreescalewebpage have already changed

jneiva08:
thanks

myfreescalewebpage:
This is not a major change. I agree with you there is still a risk, but some enhancements will always need to be performed, blocking them is not a solution.

antoniovazquezblanco:
Yes, I agree with you.

Bugs should be fixed straight away without caring if an existing design will be broken but this is only a reordering and may cause trouble in designs that are expected to remain stable. I personally prefer to go for the break everything option as long as the library improves but the line is drawn by people upstairs.

Just saying that in doubt it is best to ask upstairs just in case this will need to get reverted for a release.

Thanks for your work!

PR #1393, opened 2019-01-07
Create aliases for LP5907MFX-1.8.

matthuszagh:
http://www.ti.com/lit/ds/symlink/lp5907.pdf

Created aliases for the various output voltages.

I also removed the leading asterisk from the footprint filter.

------------


evanshultz:
- The footpilter filter should also not have the pin count. Can you change it to `SOT?23*`?
- At the end of each description please add the voltage and package, like `, 1.8V, SOT-23`.
- And one more silly thing: would you mind making the lowest voltage the symbol name and higher voltages as ALIASes?

matthuszagh:
@evanshultz, all suggestions fixed, let me know if anything else.

evanshultz:
Thanks! :)

Would you mind to order the ALIASes in increasing voltage? That means moving the 1.8V from the end.

matthuszagh:
Done

evanshultz:
Thanks for doing more then the minimum! :)

matthuszagh:
Absolutely! Might as well do it right.

PR #1396, opened 2019-01-08
Add ICS525-01/-02 User configurable clock with pll

fhw72:
Added: ICS525-01RILF and ICS525R-02ILF
"User configurable clock" to Timer_PLL library.

Device is using standard SSOP-28 package.

------------

- [x] https://www.idt.com/document/dst/525-01-02-datasheet

ics525r-02ilf
ics525-01rilf


fhw72:
Several version of the device exist. However all version without "I" (consumer temp. range) are marked obsolete (Last buy: mid 2019). Therefore I only created the two version with "I" (industrial) marking.

myfreescalewebpage:
Hi @fhw72, thanks for contributing,

- [ ] Regarding to the name, "LF" is Pb-Free which doesn't matter in the library, Temperature doesn't matter too. I would suggest that the part should be simply named "ICS525-01R" and "ICS525R-02".
- [ ] Description can include the frequency range for each device to make them different
- [ ] The two symbols are the same, isn't it ? Define only one and the second should be an alias.
- [ ] We don't put manufacturer in the keywords, maybe you can write `configurable clock`, that's a little bit redundant with the description but I do not know what we can indicate here...
- [ ] You have not selected the right footprint, it should be `Package_SO:SSOP-28_3.9x9.9mm_P0.635mm` (take care to the dimensions). Footprint filter to be modified too: `SSOP*3.9x9.9mm*P0.635mm*`.
- [ ] Pin length should be 100mil (symbol<100pins)
- [ ] You can order R0-R6, S0-S2, V0-V8 from up to bottom, I think its better
- [ ] CLK and REF are Output, not Tri-State
- [ ] The symbol is globally very large, you can reduce the width.

Cheers,
Joel

fhw72:
Hi Joel,
I've changed the symbol according to your review.
As you can see it's now less wide; I could even reduce a little bit more but readability gets worse then.
Cheers,
Frank

fhw72:
Just a short question on top:

According to the manual both outputs are tri-stated when PD# (Pin19) is low.
Should they be "output" nonetheless? I'm not entirely sure.
Cheers,
Frank

myfreescalewebpage:
> According to the manual both outputs are tri-stated when PD# (Pin19) is low.
Should they be "output" nonetheless? I'm not entirely sure.

Not seen this previously, yes keep Tri-State in this case.

fhw72:
Done. Reverted back to tri-state.

myfreescalewebpage:
Regarding to the stacking of the GND pin, you have a KCL error (see travis output), pin 20 should be Passive.

Else, everything is okay for me, great job !

Joel

fhw72:
Cool... learning with every commit. Now I know how to stack pins correctly.
Cheers,
Frank

myfreescalewebpage:
Thanks for being responsive ! No more comment to do on this PR, merging.

Joel

PR #1401, opened 2019-01-09
Add ADF4158.

matthuszagh:
https://www.analog.com/media/en/technical-documentation/data-sheets/ADF4158.pdf

image

------------


myfreescalewebpage:
Hi @matthuszagh 

No comment on the definition of the symbol.
I only wonder about the position of hte pins, I have seen several schematics (https://ez.analog.com/cfs-file/__key/communityserver-discussions-components-files/649/e65c3ca36e0cabcd8bc83d0e3297585e.png for example), and would like to ask you if it is not better to:
- group TXDATA/CE/CLK/DATA/LE
- move MUXOUT which prevent to wire CP with SWx easily

Let me known, if you have another schematic sample to share, that's also very nice.

Cheers,
Joel

matthuszagh:
What do you think about this?

image

I like your suggestion about moving muxout. given how this is hooked up, that just gets in the way. in the same vein i think it makes sense to move sw2 as well even though it means ungrouping the sw pins.

I like leaving ce and txdata out of that grouping. ce just enables the device and txdata is for fsk or psk whereas the rest of the grouping is for loading data into the configuration registers, so they're not really related. In general I like most of the way it's laid out in the image you sent. that reflects a lot of what I saw in the datasheet and seems to make sense.

matthuszagh:
I guess the other complaint that could be made with this is it feels a bit compressed. Any thoughts on that? Maybe if I pulled out each side by 100mil?

matthuszagh:
That would look like

image

I'm inclined to think that's a bit better... What do you think?

myfreescalewebpage:
Last version looks great :)

I'm confident with you about the positions of the pins, if you say it's better for the wiring, let's go with this solution.

Merging your PR.

Joel

PR #1405, opened 2019-01-10
Add LTC229x.

matthuszagh:
https://www.analog.com/media/en/technical-documentation/data-sheets/229321fa.pdf

image

footprint pr: https://github.com/KiCad/kicad-footprints/pull/1285

------------


matthuszagh:
I added a little spacing between the REF pins. They need to be bypassed together and the additional space makes it easier to do that.

image

myfreescalewebpage:

- [ ] In my point of view, "LTC229x" is not enough detailed. I would prefer to put all aliases. This will have several advantages: description will be more accurate, (resolution and sps) and datasheet links will be available for all parts. Moreover, maybe another package will be available in the future, so should add somethign to detail the package type. According to ordering info page 2, let says the symbol can be "LTC2291xUP" with aliases "LTC2292xUP" to "LTC2299xUP" and "LTC2282xUP" and "LTC2284xUP"
- [ ] Remove leading star of footprint filter
- [ ] Pin 1 name is "AINA+"
- [ ] Pin 2 name is "AINA-"
- [ ] REF pins are all Inputs
- [ ] VCMx should be Passive to avoid ERC errors when not used
- [ ] REFHB should be above REFLB
- [ ] AINB+ should be above AINB-
- [ ] VCMx can be connected to SENSEx, so it's better on the left of the symbol
- [ ] VDD and OVDD pins should not be stacked and should be all Power Input
- [ ] The symbol is globally very large, you can reduce the width a lot.

Globally I wonder about the position of the pins. Looking at the datasheet, I think it's better to group analog pin together, and to have upper part of the symbol for port A and bottom part of the symbol for port B. A proposition is given below. let me known.

capture

Cheers,
Joel

matthuszagh:
@myfreescalewebpage, I think these suggestions make a lot of sense. Thanks for figuring all that out! As for OEA and SHDN pins being on the right side, is that to provide a bit more room for the pins on the left side?

matthuszagh:
Here's what it looks like now, let me know what you think.

image

myfreescalewebpage:
@matthuszagh really great job on this device!

A single remaining comment now:

- [ ] REF pins should be all Inputs, even if stacked in this configuration (see travis result)

Joel

myfreescalewebpage:
Thanks, no more comment on the symbol, waiting the footprint.

aewallin:
pin65 is now "GND" but I wonder if it should be called "EP" (although hidden/passive stacked, so user will not see it...)

matthuszagh:
Is it permissible to stack pins with different names? I do think they should be stacked because the datasheet requires that ep is soldered to ground ([page 12](https://www.analog.com/media/en/technical-documentation/data-sheets/229321fa.pdf)).

myfreescalewebpage:
> I do think they should be stacked because the datasheet requires that ep is soldered to ground (page 12).

@matthuszagh yes so should be stacked

> Is it permissible to stack pins with different names? Whenever I do that I get a checklib violation s4.3.

That's right.

> pin65 is now "GND" but I wonder if it should be called "EP" (although hidden/passive stacked, so user will not see it...)

 @aewallin when the EP need to be stacked, to comply with the above KLC, it's name is set to GND (or other name depending...)

myfreescalewebpage:
Footprint has been accepted, merging the symbol.
Thanks @matthuszagh  and @aewallin !

PR #1094, opened 2018-10-26
Driver_Motor: Add TI DRV8837/37C/38

florolf:
- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
 - http://www.ti.com/lit/ds/symlink/drv8837.pdf
 - http://www.ti.com/lit/ds/symlink/drv8837c.pdf
- [x] An example screenshot image is very helpful
drv8837
drv8838
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #1408, opened 2019-01-10
Driver_Motor: Add DRV883x

antoniovazquezblanco:
My symbol for DRV883x. This PR should substitue https://github.com/KiCad/kicad-symbols/pull/1094 although my symbols were made from scratch.

http://www.ti.com/lit/ds/symlink/drv8837.pdf
http://www.ti.com/lit/ds/symlink/drv8837c.pdf

screenshot from 2019-01-10 12-18-08
screenshot from 2019-01-10 12-18-19

------------


myfreescalewebpage:
Hi @antoniovazquezblanco 

- [ ] Footprint filter do not contain EP size, should be `WSON*1EP*2x2mm*P0.5mm*` for all devices
- [ ] DRV8837C, pins 2 and 3 should be Power Output

I globally prefer the previous arrangement of the pins, with SLEEP in bottom left corner, and VCC/VM symmetrical. Maybe it is easier to wire like that ? Let me known.

Cheers,
Joel

antoniovazquezblanco:
FP filter fixed

About pins 2 and 3, from the previous PR @evanshultz wrote:
> Pins 2 and 3 should just be Output type.

We should find consensus here.

> I globally prefer the previous arrangement of the pins, with SLEEP in bottom left corner,...

I prefer sleep at the top because in certain applications it will be wired to VCC through a resistor to be permanently enabled. It cleans up the schematics. I can move it down if you are not convinced by this.

> ... and VCC/VM symmetrical...

If I make them symmetrical then the symbol will have to become taller while keeping the same distance between OUT1 and OUT2, which corresponds to the height of a DC motor symbol in the lib.

myfreescalewebpage:
> We should find consensus here.

It doesn't matter for me, we can be confident with @evanshultz comment I think, but in this case the DRV8838 pins 2 and 3 need to be turned to "Output" ;)

> I prefer sleep at the top because in certain applications it will be wired to VCC through a resistor to be permanently enabled. It cleans up the schematics. I can move it down if you are not convinced by this.

No this is also a good argument. Keep like that.

> If I make them symmetrical then the symbol will have to become taller while keeping the same distance between OUT1 and OUT2, which corresponds to the height of a DC motor symbol in the lib.

Ok, good point, simplify the wiring with DC motor so accepted

Joel

antoniovazquezblanco:
> It doesn't matter for me, we can be confident with @evanshultz comment I think, but in this case the DRV8838 pins 2 and 3 need to be turned to "Output" ;)

I missed that one! Sorry :sweat_smile: . Fixed

myfreescalewebpage:
No more comment to do, merging.

Joel

antoniovazquezblanco:
Thank you very much!

PR #1409, opened 2019-01-10
Relay: Add RT3 family

antoniovazquezblanco:
https://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=RT1_bistable&DocType=DS&DocLang=English

screenshot from 2019-01-10 16-33-02

------------


evanshultz:
- Because the IEC rating is non-functional, shall we remove it?
- This DRAW part of this symbol can be simplified with exactly the same look by doing this:
```
S -400 200 400 -200 0 1 10 f
S -325 75 -75 -75 0 1 10 N
S -320 -40 -260 -70 0 1 10 F
S -320 70 -260 40 0 1 10 F
S -280 -40 -260 40 0 1 10 F
P 2 0 1 0 -200 -200 -200 -75 N
P 2 0 1 0 -200 200 -200 75 N
P 2 0 1 10 -75 0 -50 0 N
P 2 0 1 10 -25 0 0 0 N
P 2 0 1 10 25 0 50 0 N
P 2 0 1 10 75 0 100 0 N
P 2 0 1 10 125 0 150 0 N
P 2 0 1 20 200 -100 125 150 N
P 2 0 1 0 200 -100 200 -200 N
P 4 0 1 0 100 200 100 100 125 125 100 150 F
P 4 0 1 0 300 200 300 100 275 125 300 150 N
X ~ 11 200 -300 100 U 50 50 1 1 P
X ~ 12 100 300 100 D 50 50 1 1 P
X ~ 14 300 300 100 D 50 50 1 1 P
X ~ A1 -200 300 100 D 50 50 1 1 P
X ~ A2 -200 -300 100 U 50 50 1 1 P
```

antoniovazquezblanco:
> Because the IEC rating is non-functional, shall we remove it?

Do you mean remove it from naming?
I thought about that but the only models in 3.3v and 24v are IEC rated (no other part exists that is non rated) and those two are just an alias. Also those two models differ in height as far as I know which may lead to different 3D models in the future.

> This DRAW part of this symbol can be simplified with exactly the same look by doing this...

Done

evanshultz:
Thanks for updating the symbol drawing.

I was talking about totally removing the IEC-rated versions. It looks like the 24V coil only comes in an IEC rating but 3.3V coils are available with and without IEC rating, right? Where does it show the height is different?

antoniovazquezblanco:
> I was talking about totally removing the IEC-rated versions. It looks like the 24V coil only comes in an IEC rating but 3.3V coils are available with and without IEC rating, right? Where does it show the height is different?

I found time to have a look at it again. Sorry, I confused two relay models, there is no height differencence between IEC and non IEC rated relays.
I will be changing this in a minute.

Thanks!

antoniovazquezblanco:
@evanshultz any other changes to perform?

Thanks!

evanshultz:
Just one thing: Would it be better to change `RT31xA03` to `RT314A03` for consistency? That matches the MPN in the datasheet and may be more clear than having a wildcard in just this one symbol.

antoniovazquezblanco:
Done!

evanshultz:
Thanks!

antoniovazquezblanco:
Thank you!

PR #1412, opened 2019-01-10
Add SKY65404.

matthuszagh:
http://www.skyworksinc.com/uploads/documents/SKY65404_31_201512K.pdf

image

footprint: https://github.com/KiCad/kicad-footprints/pull/1289

------------


myfreescalewebpage:

- [ ] The package is "QFN", not "DFN". To be changed in footprint, footprint filter and description.
- [ ] Name of pin 1 should be "V_ENABLE"

Cheers,
Joel

matthuszagh:
@myfreescalewebpage, this confused me because although the datasheet says qfn there are only pins on two sides.

As for V_ENABLE, I also changed RF_IN and RF_OUT to RFIN and RFOUT to make them fit on the symbol. What do you think here? Do I need to go with the exact names on the datasheet? Otherwise I will have to either:
1. make the symbol bigger (probably quite a bit bigger because it's hard to make triangles bigger while keeping the 100mil pin grid)
2. move the input lower
3. or something else that will probably also look bad.

myfreescalewebpage:
> this confused me because although the datasheet says qfn there are only pins on two sides.

Agree, but I think this makes confusion if not following the datasheet.

> As for V_ENABLE, I also changed RF_IN and RF_OUT to RFIN and RFOUT to make them fit on the symbol. What do you think here? Do I need to go with the exact names on the datasheet? Otherwise I will have to either:

That's right, haven't cached the RF pins. That's always better to avoid confusion. However here I think this is not a priority, I don't see any possibility to get confused. You can keep current names.

matthuszagh:
I made the footprint changes.

myfreescalewebpage:
No worries.
Just need to wait for the footprint now here.
Joel

matthuszagh:
Changes made.

myfreescalewebpage:
@matthuszagh thanks, no more comment to do here, merging.

PR #1416, opened 2019-01-11
Add TPS7A91.

matthuszagh:
http://www.ti.com/lit/ds/symlink/tps7a91.pdf

image

footprint: https://github.com/KiCad/kicad-footprints/pull/1290

------------


myfreescalewebpage:

- [ ] "1-A" in the description should be "1A".
- [ ] Name of pin 5 should be `~PG` and type should be Open Collector
- [ ] You can move OUT pins in front of IN pins.
- [ ] It is possible to make the symbol a little bit smaller by reducing the height. Have a look to other regulators.

Cheers,
Joel

matthuszagh:
Are you sure PG should be designated active low? It's open collector so obviously it can only be driven low, but it indicates power good with a high voltage (via high impedance and external pull-up). What do you think? I made all the other changes.

I also compressed the height a bit. This violates the symbol being at the origin. I could compress it further to make it symmetric again, but that would require grouping all left-side pins. What're your thoughts?

image

myfreescalewebpage:
Centering is only a "info" in travis, you can keep like that.
Sorry for PG, you are right, I have made a wrong reasoning before (the text of the datasheet is not so clear sometimes, looking at the schema page 15 helps :) ).

Now just need to wait the footprint.

Joel

matthuszagh:
Made the changes. Is it ok to leave the end of the description as DFN-10?

myfreescalewebpage:
Correct, you can change in the description to `Texas S-PDSO-N10`

PR #1417, opened 2019-01-11
Add W25Q32JVZP

matthuszagh:
http://www.winbond.com/resource-files/w25q32jv%20revg%2003272018%20plus.pdf

image

footprint: https://github.com/KiCad/kicad-footprints/pull/1291

------------


matthuszagh:
Good catch @myfreescalewebpage. Do you think it's really worthwhile keeping that one? I'm not even sure if a DFN exists with those specs.

myfreescalewebpage:
> Do you think it's really worthwhile keeping that one?

Are you speaking of the footprint ?
If yes, can be kept but we will not review with a high priority.

matthuszagh:
Ok, I've added the fp [here](https://github.com/KiCad/kicad-footprints/pull/1293).

I ended up removing the other footprint because I don't think it's worth the extra work it will take in review, especially since I don't know if any packages use it. If you really think it's worth keeping I can bring it back.

myfreescalewebpage:
Thanks. No more comment on the symbol, need to wait the footprint now.

Joel

myfreescalewebpage:
@matthuszagh no more comment here, footprint was merged, merging the symbol. Thanks for your contribution.

PR #1420, opened 2019-01-11
Add CP2102 QFN20

lmoelleb:
Based on the QFN-28 symbol merged to master a few minutes ago.

Datasheet:
https://www.silabs.com/documents/public/data-sheets/cp2102n-datasheet.pdf

image

------------


myfreescalewebpage:
Thanks @lmoelleb.
Understood you will not have time now to work on this. No comment except the footprint is not correct, but do not exist today in the library.
Keep on going.

Joel

diegoherranz:
A few comments:
- RSTb and SUSPENDb have a double negation currently (the `b` and the bar above it). See http://kicad-pcb.org/libraries/klc/S4.7/
- The temperature information on the part number should use a wildcard. See http://kicad-pcb.org/libraries/klc/S2.2/

Thanks!

myfreescalewebpage:
@diegoherranz there is also hw and sw revision not so relevant, but I have not ask modification to keep consistency with other CP2102 variants already in the lib, which is better I think at this moment.

Cheers,
Joel

lmoelleb:
I can the QFN-28 symbol to remove the extra "b" on the "negative" pins as well. I don't mind changing the existing QFN-24 to use identical pin names, the correct pin length etc.

But the missing footprint is not something I am likely to see at the top of my "todo" list anytime soon.

myfreescalewebpage:
Exactly, the QFN24 was in the lib before @lmoelleb proposed QFN28 and QFN20 variants and I have preferred at this time keeping the same pin names, at least for SUSPEND and RESET.

I would suggest to raise an issue to fix the consistency, to be solved later (milestone 6.0.0) because modifying the QFN24 variant will break existing schematics.

> But the missing footprint is not something I am likely to see at the top of my "todo" list anytime soon.

Don't worry for that.

Joel

cpresser:
In the meantime the footprint `QFN-20-1EP_3x3mm_P0.45mm_EP1.6x1.6mm` has been added.
We are also nearing 6.0 release, so this would be a good time to revive this PR and perhaps commonalize all CP2102 symbols at the same time.

myfreescalewebpage:
@lmoelleb footprint has been integarted, are you able to fix branch conflict here so we can merge the symbol too ?
Joel

lmoelleb:
@myfreescalewebpage Sorry Joel, it seems this PR suffered the "open too long death".

At this moment I do not have a running KiCad, and there are validation issues after the merge. Fixing these will require more time than I am able to spend in the foreseeable future.

myfreescalewebpage:
Thanks!

To simplify the work, the exact list of point to fix is:
- [ ] Footprint : `Package_DFN_QFN:QFN-20-1EP_3x3mm_P0.5mm_EP1.8x1.8mm` (but in fact seems to not exist today)
- [ ] Footprint filter : `QFN*1EP*3x3mm*P0.5mm*`
- [ ] VDD should be Power Input

Joel

myfreescalewebpage:
@lmoelleb ping

cpresser:
This is a duplicate of https://github.com/KiCad/kicad-symbols/pull/2364
I plan to get 2364 merged soon. I think #2364 is better since it also makes the symbols drop-in compatible.

lmoelleb:
@myfreescalewebpage As I mentioned I consider this PR dead due to the long time it has been opened. I tried to see what would happen with a simple merge, and when that was not enough I tried starting KiCad to see if I could quickly edit it. The library editor was grayed out which is where I stop - sorry. I spend way too long on getting tools to run at work, and with limited time available in my spare time I have to use it carefully - and preferably not on the exact same problems I solve at work.

So I recommend using #2364

myfreescalewebpage:
Haven't seen before a new PR cover this one, in this case it's great !

PR #1421, opened 2019-01-11
Sensor_Optical: Kingbright KPS-3227 ambient light sensor.

mgottschlag:
Symbol for the Kingbright KPS-3227 ambient light sensor.

The center of the sensor was copied from the other phototransistors, but the KPS-3227 is a 4-pin package (actually, would fit a 2-pin LED footprint as well, but I thought the symbol should better match the datasheet).

The device uses a standard 1210 RGB LED footprint, so I selected that generic footprint.

------------

http://www.kingbright.com/attachments/file/psearch/000/00/watermark00/KPS-3227SP1C(Ver.9).pdf
2019-01-11-220054_671x656_scrot


aewallin:
- if the pins 1-3 and 2-4 are always used together then they can be stacked pins in the symbol?

mgottschlag:
Yeah, that would work... I guess the pin numbers need to be invisible then?

aewallin:
yes, one pin visible, the other stacked pin passive/invisible, with the same pin-name as the visible pin (here the pin names are empty I guess).

mgottschlag:
2019-01-12-143852_515x413_scrot

myfreescalewebpage:
Hi @mgottschlag 

Thanks for your contribution, no comment to do following the update.

@aewallin thanks for your review, good catch.

Cheers,
Joel

PR #1423, opened 2019-01-12
Add MMSD4148

JacobEFO:
Fix S5.2 on MMSD4148 

**Please provide a description of your pull request*:**

Add switching diode MMSD4148, 100V, 200 mA SOD-123.

------------

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing:
https://www.onsemi.com/pub/Collateral/MMSD4148T1-D.PDF

- [ ] An example screenshot image is very helpful:
image

Done

None whatsoever.

I used the library-utils -> schlib -> checklib.py
Currently violating S5.1, I have no idea why. This is the first time I am trying to do a pull-request and quite new at using kicad in a collaborative manner.

It seems as if the "check individual component" function is not working. I tried:
python checklib.py MMSD4148 path-to-file.lib

I'd appreciate your feedback.

EDIT:
python checklib.py MMSD4148 path-to-file.lib
should have been including a -c argument.

Regarding violation S5.1 I have remained unable to solve the issue. It seems it can't find the footprint. When including it in the pcb view, it finds the correct footprint as set per my path.
But is the utils looking for footprints in kicad/share/kicad? If so, it won't find it since I have deleted the footprint content in there and solely use another location for footprints (path set accordingly).


myfreescalewebpage:
Hello @JacobEFO 

I think it is possible to add this device as alias of MMSD914, isn't it ?

Cheers,
Joel

PS: travis complains because of the footprint filter, not matching the footprint (which is the right one). Anyway if you define alias of MMSD914, this will no more be an issue.

JacobEFO:
> Hello @JacobEFO
> 
> I think it is possible to add this device as alias of MMSD914, isn't it ?
> 
> Cheers,
> Joel
> 
> PS: travis complains because of the footprint filter, not matching the footprint (which is the right one). Anyway if you define alias of MMSD914, this will no more be an issue.

Except for the fact, that I don't know what alias means, sure it should be possible. MMSD914 is almost hte same.

myfreescalewebpage:
In MMSD914 properties, you can add an "alias" with name "MMSD4148". This recover the symbol, footprint etc for this symbol and you only have to fill datasheet link, description and keywords for the MMSD4148.

For better understanding, look at MMSD914, it has already aliases 1N4148W, 1N4448W and BAV16W :)

Joel

myfreescalewebpage:
Yes you are correct.

JacobEFO:
Ah, excellent.

I'll figure out how to get my stupid git on Ubuntu on Windows working and update my pull request.

JacobEFO:
I am pretty convinced, it should be solved now.

Not entirely sure, why the commit is performed by "root" and not my user.

myfreescalewebpage:
@JacobEFO yes it is! Thanks for your contribution and your work :) Merging this PR.

Joel

PR #1424, opened 2019-01-12
RF Module: Add CMWX1ZZABZ

myfreescalewebpage:
Add CMWX1ZZABZ module from muRata (LoRa / SIGFOX).

------------

https://wireless.murata.com/RFM/data/type_abz.pdf

capture

https://github.com/KiCad/kicad-footprints/pull/1296

No error at this time, except that the footprint is not found (see linked PR).


myfreescalewebpage:
@evanshultz just seen you assigned yourself to my PR several days ago :) Any comment on this symbol ?
Thanks, Joel

evanshultz:
- Can you remove the "TM" from the description?
- Should it be "LoRa" with that specific capitalization in the description?
- Pins 8,9,11,12,13 are missing the leading "DBG_" in their name.
- Pins 44 and 45 have a hyphen before "OSC", not a slash.
- Is there a pin-by-pin description somewhere to check the pin types? The doc attached above doesn't cover it and I don't see any other docs on Murata's website.

myfreescalewebpage:
Thanks for the reply.

> * Can you remove the "TM" from the description?
> * Should it be "LoRa" with that specific capitalization in the description?
> * Pins 8,9,11,12,13 are missing the leading "DBG_" in their name.

Agree, fixed in the last commits.

> * Pins 44 and 45 have a hyphen before "OSC", not a slash.
> * Is there a pin-by-pin description somewhere to check the pin types? The doc attached above doesn't cover it and I don't see any other docs on Murata's website.

There isn't better documentation, even when subscribing to murata website support. The module has a MCU + SX1276 inside. There are:
- several supplies for the various parts
- PA, PB, PH are the ports of the MCU, I have keep the "/" for PH0 and PH1 because it looks like an error on the schematic in the doc (this is the PH0 and PH1 pins of the MCU and it is more logical to keep the same naming scheme of the other ports)
- MCU_RST is reset pin
- TCXO_OUT is internal SX1276 clock output
- SX1276_DIO are SX1276 pins available for various purpose (mainly debug except DIO4 which is used to make a loopback to the MCU in some configurations)
- ANT is the antenna attached to the SX1276
- BOOT0 is probably the boot pin of the MCU (not sure, but it is an input according to the example schematic)
- I don't know what are DBG_CRFx and STSAFE_RST pins, set to Passive, never seen connected on the several app note available on internet

Hope this respond your questions,
Joel

evanshultz:
Sorry for the delay. All changes look good.

I'd really like to know something about the DBG_ and STSAFE_RST pins. Would you mind checking again? Anything with "RST" in the name makes me think it's an Input, but I really don't know.

myfreescalewebpage:
I'm working with this module since few month now and really don't know. Already searched in Murata documentation several times. Will still be possible to update the pin type later if the information can be found :)

PR #1428, opened 2019-01-16
HSMC Card Symbols

calebreister:
This pull request contains symbols for [Intel/Altera HSMC cards](https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/ds/hsmc_spec.pdf). Two versions of the same symbol have been provided:

1. `HSMC_Card_ASP` for the Samtec ASP-122952-01 connector
2. `HSMC_Card_QTH` for the Samtec QTH-090-01-x-D-A connector

HSMC_Card_ASP
screenshot_20190115_193032

HSMC_Card_QTH
screenshot_20190115_192858

For now, the new symbols are in their own `Connector_HSMC` library. They should probably be moved into `Connector.lib` before being merged.

**KLC Notes:**
- Pin name offset not preferred value: the offset was set to 50mil to make room for clock pins.
- Pin types should match pin functions: they do, with the exception of stacked power pins.
- NC pins were included in the `HSMC_Card_QTH` symbol, since the HSMC specifications recommends that they be grounded.

**See also:** [pull request](https://github.com/KiCad/kicad-footprints/pull/1304) for the corresponding connector footprints.

------------


calebreister:
A few additional (minor) revisions have been made, and the symbols were moved into Connector.lib

myfreescalewebpage:
Hi @calebreister thanks for this contribution,

I have not made a full review yet, but from the screenshot:

- [ ] Graphic style should be line for all pins
- [ ] All pins should be 150mil (symbol >= 100pins and < 1000pins)

Can you also resolve the branch conflict to permit the next step of the review ? Thanks.

PS : Connector is the right lib for that.

Cheers,
Joel

calebreister:
Hi @myfreescalewebpage,

I've implemented the changes you suggested. Note that Travis now fails because the footprints have not yet been merged. Below is an image showing one of the updated symbols.

image

myfreescalewebpage:
> Should I also remove the < from SCL and CLKIN0?

Yes, that's a possibility of KiCad but not allowed in the official library. Let me known when you have done it I will continue the review.

calebreister:
@myfreescalewebpage: Done.

myfreescalewebpage:
Thanks @calebreister for the first fixes. Making a full review from this point.

- [ ] You should group the pins by functions, not depending of the physical location of the pins on the connector. For example SDA goes with SCL, JTAG pins all together, etc
- [ ] I think SDA should be Bidirectional and SCL should be Output
- [ ] We do not use `~` in the name of the complementary pins. You should remove all of them in this symbol, except for pin 160, it's name should be `~PSNT`
- [ ] PSNT should be on left or right side. Bottom is reserved for GND pins.
- [ ] JTAG TCK and JTAG TMS should be Input

Joel

calebreister:
I'll implement the suggested changes as soon as I have the time.

> I think SDA should be Bidirectional and SCL should be Output

This could be a problem, since I2C is a multi-master bus, it is possible that other devices will need to drive SCL. Since this connector is intended for FPGA development, it is possible that another master (such as an external microcontroller) could be connected on the same bus. This would create an ERC error. Perhaps both SCL and SDA should be bidirectional? However, I'm not entirely sure why the output/bidirectional convention was adopted in the KiCad libraries, since I2C is inherently an open-collector bus.

> PSNT should be on left or right side. Bottom is reserved for GND pins.

The HSMC specification states that PSNT should always be connected to GND. Should it still be placed on a side?

> You should group the pins by functions, not depending of the physical location of the pins on the connector...

What about the unconnected pins on `HSMC_Card_QTH`? The specification recommends (but does not require) that they be tied to GND or one of the supply rails. In larger designs, there may be reason to connect these pins to other potentials, or even to route signals through them.

myfreescalewebpage:
> However, I'm not entirely sure why the output/bidirectional convention was adopted in the KiCad libraries, since I2C is inherently an open-collector bus.

Me too! But as you said it is a convention. If you connect a MCU to the I2C bus, the MCU pins are bidirectional and there will be no errors. The direction of the pins is not always the reality of the devices but the way the ERC checker should consider the pin.

> The HSMC specification states that PSNT should always be connected to GND. Should it still be placed on a side?

Yes, it's not a power pin and we should think to all strange possibilities user will need where it will not be directly connected to GND.

> What about the unconnected pins on HSMC_Card_QTH? The specification recommends (but does not require) that they be tied to GND or one of the supply rails. In larger designs, there may be reason to connect these pins to other potentials, or even to route signals through them.

Not yet reviewed, I have make a concentration to the main points at the moment. I haven't spoken of NC pins in my previous comment, but if the specification recommend to connect them to GND, then they should be stacked with GND. Moreover I think this is a good practice.

calebreister:
I've just performed the revisions you requested. A few comments:
- The pins are named relative to the host. The JTAG signals originate on the host board, thus the directions of TMS and TCK are correct.
- I've tried a compromise with the NC pins in the QTH variant: the odd and even pins have been stacked separately, on either side of the ground plane.
  - This allows for different sides to be tied to different potentials (for instance, one side can be tied to 12V and the other to 3V3), or both to GND.
  - I used the NC pin style. Is this ok?
- Transmit signals have been moved to the right, receive to the left.

Here's a screenshot of the new QTH variant used in a design I'm currently working on.
image

poeschlr:
Regarding pin types see: https://github.com/KiCad/kicad-website/issues/342

myfreescalewebpage:
Thanks @calebreister however, all Dxx pins should be all together on the unit.

calebreister:
@myfreescalewebpage, wouldn't that make the symbol too tall? Currently, the sections (roughly) correspond to the three physical pin banks.

myfreescalewebpage:
The policy in the library is to group by function, not depending of the physical place of the pins. I don't think this will be too much tall. You can keep Rx pins on the left and Tx pins on the right. We have SOM for example with about 200 pins all on the same unit :)

calebreister:
@myfreescalewebpage 

I'm still a little confused. To clarify, this is what you want:
- Group D0..3 together
- Move all pins to a single unit?

Also, note that the LVDS pins are actually dual-purpose, and can serve as bidirectional single-ended channels. Currently, I have odd (D1,3,5,7...) on the left and even (D0,2,4,6,...) on the right. Unfortunately, I can't group all of the Dxx pins because it would mess up the LVDS groupings.

myfreescalewebpage:
No I just mean grouping units B and C into a single unit. Keep unit A unmodified. You can keep odd and even on each side it's ok.

calebreister:
So something like this?

image

image

image

calebreister:
Perhaps it would be best to keep both versions (appending _Split to the old version).

myfreescalewebpage:
Yes something like this !! You can reduce the width of the units also.

> Perhaps it would be best to keep both versions (appending _Split to the old version).

The policy of the library is to group by functions, the first proposition has not a place in the library.

Thanks,
Joel

calebreister:
Done. I've applied the suggested changes and made the units as narrow as I can while keeping both units the same width.

calebreister:
@myfreescalewebpage I thought we already discussed the NC pins, since the HSMC spec. recommends they be connected to GND. I can still move them, but can they still be connected when invisible?

calebreister:
@myfreescalewebpage I thought we already discussed the NC pins, since the HSMC specification recommends connecting them to GND. I can still move them, but can they still be connected when they are invisible?

evanshultz:
Just a couple comments:
1. @calebreister Does https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/ds/hsmc_spec.pdf show the grounding of NC pins? I searched for "nc pin" and "unconnect" and didn't find it. No that I doubt you, but you mentioned this a few times so perhaps pointing out the exact location of this requirement will make it more clear.
2a. Should symbol names be changed to `HSMC_Card_Samtec_QTH` from `HSMC_Card_QTH`? That will make it clear the vendor being used and, if there are other potential HSMC vendors, the new symbols can be named clearly.
2b. If the symbols are otherwise the same, no default footprint can be provided but then multiple footprint filters or a simpler "catch all" footprint filter (like `*HSMC*`) could be used. In this case, the footprint `Samtec_QTH-090-01-x-D-A_P0.5mm_Vertical` doesn't include the word `HSMC` so that wouldn't work; however, I believe that's an oversight since the connector specification should be given in the footprint name (see KLC). If `HSMC` existed in both connector footprint names then this would work; if the footprints were merged recently, so there's little risk of name change, they could potentially be re-named now but it would need buy-in from the librarian team. This may also affect the NC pins referenced above (I didn't check myself) but perhaps this would still allow the same symbol for either connector family? Just an idea...
3. Footprint filters should use asterisks wherever possible. I think all question marks here can be replaced with asterisks.
4. The footprint filter doesn't include the library, so remove `Connector?Samtec:` from `HSMC_Card_QTH`'s filter.

calebreister:
@evanshultz 

1. See Section 2-3 of the HSMC spec (p.10)
    > QTH family connector with three banks. The difference between this connector and the ASP version is that all pins are populated in bank 1 of the connector. This difference equates to slightly lower signal integrity in that bank which typically carries multi-GHz clock-data-recovery signals. It is recommended that you do not leave the undefined pins in bank 1 floating; ground them or tie them to power. This connector has 192 physical pins.

2a. The specification gives Samtec part numbers, with no alternatives mentioned.
- Since HSMC is an Altera (now Intel) specification, things get even more confusing.
- If we were to make any changes, I recommend appending the full connector part number to the symbol name. I used the shortened names during development, but I just realized that there could be problems if the QTH-060-01-L-D-DP-A version were to be added later.

2b.
- The pin numbering on the ASP and QTH versions of the symbol are different. Thus, it is probably best to leave them as separate symbols.
- When I was initially creating the symbol(s), I thought of using a unified pin numbering scheme. However, the specification addresses this issue directly (p.11):
    > ASP-122952-01 is the recommended connector defined by the HSMC specification. If any of the compatible connectors from the above table [QTH-090-01-L-D-A or QTH-060-01-L-D-DP-A] is used, then the pin numbering will change accordingly due to different pin counts. Hence the example pinouts must be adjusted accordingly for the selected connector part.
- The omission of `HSMC` from `Samtec_QTH-090-01-x-D-A_P0.5mm_Vertical` was not an oversight. The QTH-090-01-L-D-A is a member of the Samtec QTH family of connectors, and is by no means specific to the HSMC specification. On the other hand, the ASP-122952-01 is used solely for HSMC cards (ASP stands for *Application-Specific*).

calebreister:
I've implemented the following changes suggested by @evanshultz so far:
- Rename symbols to include full connector part numbers
- Replace `?` with `*` in footprint filters, remove library from QTH footprint filter

I'm still not sure what to do with the NC pins, and there seems to be some disagreement. Additional clarification would be helpful.

calebreister:
I've made some additional revisions: the NC pins on the QTH version of the symbol are no longer stacked.
image

myfreescalewebpage:
You can put NC pins inside the symbol itself to respect the 100mil grid position.
Joel

calebreister:
@myfreescalewebpage I still need to be able to connect to them. Perhaps it would be best to extend all pins by 50mils and shrink the rectangle by 50mils on each side?

calebreister:
I've just committed a new revision that re-aligns all of the pins on a 100mil grid and centers the symbols correctly.

Minor cosmetic question: since the 3V3 and 12V pins are located in banks 2 and 3 (part B of the symbol), should they be moved there, along with the ground pins?

myfreescalewebpage:
> I still need to be able to connect to them.

You lean you need to connect NC pins ? Maybe in this case you can make a unit with all NC pins, visible, so it's still possible to use them for specific application.

> Minor cosmetic question: since the 3V3 and 12V pins are located in banks 2 and 3 (part B of the symbol), should they be moved there, along with the ground pins?

The current design is great I think.

About pins length: 3 digits for the pin number => length of the pin must be 150mil

calebreister:
Here is the section of the [HSMC specification](https://www.intel.com/content/dam/www/programmable/us/en/pdfs/literature/ds/hsmc_spec.pdf) that suggests connecting the NC pins (p.10).

> QTH-090-01-L-D-A: QTH family connector with three banks. The difference between this connector and the ASP version is that all pins are populated in bank 1 of the connector. This difference equates to slightly lower signal integrity in that bank which typically carries multi-GHz clock-data-recovery signals. It is recommended that you do not leave the undefined pins in bank 1 floating; ground them or tie them to power. This connector has 192 physical pins. 

A third unit is an option, but would not be ideal. Where does the KLC specify that the pin length must be 150mils with 3-digit pin numbers? 200mils certainly fits within the constrains of [S4.1](http://kicad-pcb.org/libraries/klc/S4.1/), which is the only rule I can find that restricts symbol pin lengths.

myfreescalewebpage:
So I suggest in this case to stack NC pin with GND, it you agree with that.

> Where does the KLC specify that the pin length must be 150mils with 3-digit pin numbers?

It is not written, it is a convention we have, every thing can't be written in the KLC but should be added...

Joel

calebreister:
@myfreescalewebpage Stacked NC pins is what I had before, but @evanshultz seemed to think they should *not* be stacked, so I created the current arrangement.

poeschlr:
In this case as the pins are suggested to be either left open or connected to one defined potential for increased signal integrity i would suggest to make the NC pins visible but not stack them. (I would guess the user has the freedom to decide for every single of these pins which potential if any to connect it to.)

myfreescalewebpage:
Yes @calebreister NC pin should not be stacked !

calebreister:
In that case, I see three options:
1. Leave the NC pins where they are, invisible
2. Leave the NC pins where they are, visible
3. Move the NC pins to a third unit, visible

Opinions?

myfreescalewebpage:
The 3rd option, definitively. If the user do not want to connect the NC pins, then he should simply don't add the 3rd unit and it's fine.
Thanks,
Joel

poeschlr:
i personally would possibly prefer either option 2 or 3 (option 2 with the change that i would move them out from the outline. Only invisible pins need to be within the symbol outline to reduce the chance of accidentally connecting them.)

calebreister:
I've just committed another revision:
- Revert to 150mil pin lengths (and corresponding rectangle dimensions)
- Add 3rd unit to `HSMC_Card_QTH-090-01-x-D-A`

Here is what the new unit looks like:
image

myfreescalewebpage:
That's great for me !! Thanks for this job.
Just need to wait the footprints (https://github.com/KiCad/kicad-footprints/pull/1304) before merging here.
Joel

calebreister:
Quick question: should I make Unit C of `HSMC_Card_QTH-090-01-x-D-A` narrower, so that it looks more like the generic `Conn_02x*` symbols?

myfreescalewebpage:
If you want, yes.

myfreescalewebpage:
Thanks @calebreister it's nice to have it this way too !

evanshultz:
Coming back to some older PRs I had bookmarked. I'll take a look at the footprints.

calebreister:
I've just pushed a new revision that updates footprint references and adds a new HSMC_Card_QTH-060-01-x-D-A symbol. Since the appropriate footprint(s) now exist, I also created a  new HSMC_Host symbol for the ASP-122953-01 footprint.

The addition of host symbols brings up a naming issue. Up to this point, I've been using the HSMC host pin names for the card symbols. For instance, the XCVR_RXp0/XCRV_RXn0 are card-to-host channels are inputs on the card and outputs on the host. This is ok for the transceiver pins, but potentially confusing for the bidirectional GPIO/LVDS pins. Should the TX/RX names on the HSMC_Card symbols be swapped so that TX on the host connects to RX on the card and vice-versa?

myfreescalewebpage:
> Question: Should the TX/RX names on the HSMC_Card symbols be swapped so that TX on the host connects to RX on the card and vice-versa?

You should use the name of the spec only. If pin 1 is "Tx", it is "Tx". Else this is confusing.

You have some travis errors to check.

Joel

calebreister:
@myfreescalewebpage The issue is that the official spec. only provides pin names for the host connector (ASP-122953-01), not the card (ASP-122952-01). It seems like the TX pins on the host should connect to RX pins on the card, and vice versa. However, one can argue either way, since both symbols would never be used in a single schematic.

myfreescalewebpage:
I vote to have the same names on both sides, else this is confusing.

calebreister:
@myfreescalewebpage I took a look in the library to see what the current convention is for TX and RX on bidirectional pins on connectors. I found that `USB3_A` and `USB3_B` the names are swapped for the SSTX±/SSRX± pins. However, this is not the case for the `USB_C*` symbols (they also seem to be less developed).

I think it may be worth clarifying this TX/RX issue in the KLC. Perhaps @evanshultz or @poeschlr could weigh in.

calebreister:
Any updates on what I should do from here?

myfreescalewebpage:
@calebreister sorry I'm back from vacations. Seems correct for me, just need to wait the footprints now.
Joel

PR #1431, opened 2019-01-17
Add EP to SE5004L.

matthuszagh:
SE5004L should have an EP pin.

datasheet: http://www.skyworksinc.com/uploads/documents/SE5004L_202393B.pdf

image

------------


myfreescalewebpage:
Good catch @matthuszagh was forgotten at https://github.com/KiCad/kicad-symbols/pull/1258.

Seems the land pattern indicate that the EP must be connected to GND, isn't it ? Datasheet page 7. I would like to suggest that the EP pin need to be stacked with GND.

Joel

matthuszagh:
I'm not seeing the part that says it should be connected to GND, I'm just seeing the dimensions and thermal via info. Mind sending a quote or screenshot of where you're seeing that?

myfreescalewebpage:
You're right, no information. Just the pattern below:

capture

If you prefer keeping EP separated, let me known.

matthuszagh:
Yeah I think we probably should keep them separate. I'm not sure why you wouldn't want to connect it, but if it's not explicitly required by the datasheet I think we should give people the option.

myfreescalewebpage:
> I think we should give people the option.

OKay for me.

Merging.

PR #1432, opened 2019-01-17
Add BQ24392

EvilMav:
[Datasheet](http://www.ti.com/lit/ds/symlink/bq24392.pdf)

kicad_bq24392

~~Requires footprint: https://github.com/KiCad/kicad-footprints/pull/1310~~
Footprint has been merged.

I've kept spacing and order on the left side to exactly match those of the USB-connectors in the library.

------------


aewallin:
datasheet link above has some github-url at the front...

myfreescalewebpage:
Hi @EvilMav , thanks for contributing,

- [ ] Footprint filter never contains number of pin, I suggest `Texas*R*PUQFN*` (format which is also consistent with other Texas devices)
- [ ] VBUS should be moved to the top of the symbol (that's the main supply) and I do not understand the position of the pins on the symbol. I'm referring to typical applications chematic page 10 of the datasheet and would like to suggest the following:

capture

Else, no comment to do.

Cheers,
Joel

EvilMav:
Hi @myfreescalewebpage, thank you for taking the time.

The reasoning behind putting VBUS to the left was to make it fit directly to an usb connector, as DP/DM and VBUS will always be just connected there. Vbus will be usually connected to a power bus too, so having it on top still works if it is preferred.

I've fixed the footprint filter and changed the layout exactly to the one you've suggested.

myfreescalewebpage:
Thanks @EvilMav 
Yes power pin on the top is preferred.
Merging.

PR #1434, opened 2019-01-18
Added symbol for ICM-20948

luca-della-vedova:
Added symbol for Invensense ICM-20948 MEMS sensor (upgrade of previous MPU-9250).

The ICM-20948 is an "almost" pin compatible upgrade of the MPU9250, with the main difference being that a pin has been changed from Reserved - GND to Reserved - NC, packaging is also the same.
Datasheet is here http://www.invensense.com/wp-content/uploads/2016/06/DS-000189-ICM-20948-v1.3.pdf , while the MPU for comparison can be found here http://www.invensense.com/wp-content/uploads/2015/02/PS-MPU-9250A-01-v1.1.pdf

This is a screenshot of the symbol:

image

By comparison this is the MPU-9250 that I basically copied and changed slightly.

image

Since it's a "replacement" chip I made it as similar as possible to the MPU-9250 to ease transition, but I still have a few reservations towards the MPU-9250 that I would like to check if I should fix:

- SCL is not marked as clock (did it in my part).
- RESV-GND is not in the bottom of the symbol.
- RESV-VDDIO is not marked as power input and is not in the top of the part.
- No unconnected pins are present, I actually don't know how to deal with this one since there are a lot of unconnected pins (9 in MPU and 10 in ICM) and they would crowd the symbol a bit. I was thinking of skipping them (as in http://kicad-pcb.org/libraries/klc/S4.5/ ), however following the guideline the number of pins should be specified in the footprint filter, which it is currently not since the filter for the MPU is *QFN*3x3mm*P0.4mm*

Any ideas? Since I'm introducing a new part might be a good time to fix these issues.

Cheers!
/Luca

------------


myfreescalewebpage:
Hello @luca-della-vedova thanks for this contribution,

I understand your point of view that it is nice to get it looking to the MPU6050 but this symbol violates many KLC and we can't encourage that.

Here are my comments bellow to get this symbol conformed with the new guidelines:

- [ ] Description should ends with the package name: `, QFN-24`
- [ ] Footprint filter should be `QFN*3x3mm*P0.4mm*`
- [ ] Pins length should be 100mil (symbol < 100pins)
- [ ] Pin 7 name should be `AUX_CL`, graphic style should be Line
- [ ] Pin 9 name should be `SDO/AD0`
- [ ] Pin 21 name should be `AUX_DA`
- [ ] Pin 23 graphic style should be Line
- [ ] Pin 24 name should be `SDA/SDI`
- [ ] I would suggest to order SPI pins in this order : 9, 24, 23, 22
- [ ] Pins 1/2/3/4/5/6/14/15/16/17 should be added to the symbol, name NC, type Not Connected, not visible, placed on the border outline of the symbol
- [ ] I would suggest that pin 20 should be renamed GND and stacked with pin 18 because the datasheet indicate this pin need to be connected to GND
- [ ] Pin 19 should be added to the symbol, name RESV, type Not Connected, not visible, placed on the border outline of the symbol
- [ ] I would suggest that INT1 pin should be moved to the left of the symbol because it is intended to be connected to a MCU or such kind of device, already connected to the I2C/SPI interface

Cheers,
Joel

fauxpark:
> The left side of the symbol is a bit crowded, I can move one NC to the other side if necessary, I chose that convention because pin 1-6 are on left side of the chip, while 14-19 are on the right.

NC pins will not show up in eeschema, it doesn't really matter where they're placed. I tend to bunch them up around the bottom from right to left in pin order (cf. 100-pin ATmega symbols).

As for the footprint filter - it should in fact be `InvenSense?QFN*3x3mm*P0.4mm*`. Just take the footprint name, replace dashes and underscores and the pin count with `*` and add one at the end (for `_ThermalVias` variants etc.).

Since this is a manufacturer-specific QFN package, `InvenSense_QFN` is all one thing, so we replace the underscore there with a `?` (matches 0 or 1 characters).

luca-della-vedova:
Hi!

Thanks for the quick review! I fixed the footprint to be vendor specific.

/Luca

myfreescalewebpage:
Thanks @fauxpark yes, this is an error when writing the review, The footprint filter is correct now.
@luca-della-vedova all modifications are okay for me. I think it is possible to reduce a little bit the size of the symbol, currently very large and tall. At least 100mil on the left, 100mil on the right and 100mil on the top.
Cheers,
Joel

luca-della-vedova:
Hi Joel,

I shrinked the symbol by 100mil on each side, I believe some additional space could be saved if there was a way to save on the NC pins but not sure how to do it.
This is what it looks like now:

image

Cheers,
/Luca

myfreescalewebpage:
Hi @luca-della-vedova 
No worries, this looks good like that!
Merging.
Thanks for this job,
Joel

Issue #1433, opened 2019-01-18
Guidance on new symbol


gcormier:
So a quick question. I'd like to submit something for the LMH7324

1) Should I just do a trivial pull request to start a thread? I will probably need some assistance to pull this one off. But while we're at it... 

2) Thoughts on the best way to handle Vee pins? Stacked with one hidden?

![image](https://user-images.githubusercontent.com/8388983/51360532-0dc7e280-1a9a-11e9-937b-a10d4484944d.png)

![image](https://user-images.githubusercontent.com/8388983/51360551-1b7d6800-1a9a-11e9-9c2a-99e41088d272.png)

Sample schematic from the TI evaluation board:
![image](https://user-images.githubusercontent.com/8388983/51360613-52537e00-1a9a-11e9-8aa6-339def774e2c.png)

poeschlr:
Vee stacked with one hidden is the right way to go here.

If it has more than one such unit then the power pins should be placed on their own unit.

PR #1436, opened 2019-01-18
Add LMH7324

gcormier:
First draft, I will surely need help.

http://www.ti.com/lit/ds/symlink/lmh7324.pdf

image

- [ ] I'm unsure what to do with pad 33. It should be soldered to a thermal pad but otherwise unconnected. First multi-unit part, so I don't know how to add it in.
- [ ] I made the font 1.0mm for some text rather then growing the size of the symbol, unsure if acceptable, did not find reference to font size in KLC for this and I saw some other comparators with a rather small LATCH font vertically so I assumed playing with font-size is allowed for compactness.

------------


diegoherranz:
For future reference, a discussion about this symbol happened at https://github.com/KiCad/kicad-symbols/issues/1433

myfreescalewebpage:

- [ ] Description should ends with the package name, `, WQFN-32`
- [ ] Footprint filter should be `WQFN*1EP*5x5mm*P0.5mm*`
- [ ] Pin 29, 12, 20, 28 should be Passive (stacking of pins, see travis error S4.3)
- [ ] All Vcco pins should be Power Input

Cheers,
Joel

evanshultz:
This is a 4-unit symbol, but on other multi-unit symbols we have split up the power pins and the logic pins. That makes a rather messy eight units since each gate has unique power supply pins, but it is matching with the current library design. Split off the Vcc* and Vee pins from each triangle symbol into a separate unit.

Also, why not put Q above ~Q? This seems natural to me and it's the way the first page of the datasheet has the output pins placed.

gcormier:
Agree RE: Q. I copied another symbol so was unsure if this was in KLC somewhere I missed.

For description with package name, why are we putting this information in the description? Is it covered in KLC?

evanshultz:
Consider a part that is identical except it comes in two different packages. Putting the package info in the description allows searching by this parameter since the root of the symbol name and the description is otherwise identical. The other other difference if the suffix of the symbol name and that's a burden to recall or understand versus being able to clearly see the package in the symbol chooser. Play around in Eeschema and you'll see why this info is valuable when searching for symbols.

gcormier:
But if the part is only available in one package, this seems unnecessary?

evanshultz:
It still allows searching by package type. And consistency.

gcormier:
6ff8b33 addresses some of Joel's feedback and Q from evan.

> The KLC are a set of guidelines, rather than rules. Electronic component libraries are diverse and complex, and exceptions can be made at the discretion of the library team.

Perhaps to be considered for this :) Splitting this into 8 units.... would it not make the design more prone to a simple mistake?

evanshultz:
I would prefer consistency in this case. All other multi-unit triangle-like symbols (comparators, opamps, etc.) have separate power units, so the power units can be clustered on the triangle or placed separately. That ship has sailed and the whole library is like this now.

KiCad doesn't currently have a method to detect if the power gate is missing, which is an oversight, but it is slated for v6. That will help to catch one possible mistake. Are there others you're concerned with?

This is the most obtuse case I have come across, with 4 logical units in the same package, but I would like to maintain the same paradigm of separate power pins.

gcormier:
Okay I'll take a look this and make some changes.

myfreescalewebpage:
@evanshultz seems Rene validated the current design at https://github.com/KiCad/kicad-symbols/issues/1433#issuecomment-455444116 for this symbol, that's why I have not react on this. Separating the pins will maybe introduce some confusion on the schematic I think because power voltage can be different from one unit to the other ones. Also, single op amp have their power pin on the same unit, so I don't think it is a trouble for this one too.

I have no preference, just want to point this.

About the previous modifications @gcormier I have listed some Vee pins that must be set Passive, but the other Vee pins must remain Power Input. You should change pins 4, 5, 13 and 21 back.

Joel

evanshultz:
Maybe. But I'd confirm with Rene. I interpreted his comment as being that the power pins need to be separate from the logic pins, and this applies to as many comparators are in the physical device (so 8 units here). I read that issue and didn't comment because I thought Rene was saying what I would have said. And also, your comments are welcome.

Either way works, and naively I'd put them together here, but since we already did all the splitting up for other multi-unit symbols I lean towards doing that everywhere else.

So from my recollection (and there should be issues canonizing this)...

For single-gate symbols power pins weren't split off because other folks liked the simplicity of a single unit.

For multi-gate symbols, they were split for a couple reasons:
- All gates cannot have power pins because that could result in shorted nets. Yikes! (This was the way our libraries were before v5, and once we found this issue we made the change to split off power pins.)
- If only a single unit has the power pins, the user may not place that unit. (It's good practice to tie all inputs pins to a known voltage, but it could happen). Say the power pins are on unit B of a dual op amp but only unit A is placed so the power pins aren't on the schematic and have no connectivity. Oops.
- If the power pins are a separate unit (we chose the last unit) they can be grouped together away from the other units. But that means the power pins can't be grouped together for any single-gate symbols. Also, KiCad does not know if any units are required so there's a potential gotcha there.

Phew...

And that's how we got here. Since I was the one who personally combed through the libraries and did the splitting up, I may be biased.

Sorry if this sounds overblown, and truly this is a bit of a wacky part, but maybe we've found a reason _not_ to do what has been done. I'm OK with that as long as we talk about it and aren't doing things arbitrarily. It was a LOT of work (with still a lot more to go!) to get the library into this state for v5!

gcormier:
I'm no expert, and a mere hobbyist. 

I'm not sure if this would fit existing parts, but perhaps the deciding factor is common/shared power pins?

In this case, Vee could be split off in part 5, and apply to all 8 pins. Vcci Vcco would still remain with Part 1-4, as they can be different.

However, more fun... it is recommended to only connect pins in-use.

https://e2e.ti.com/support/amplifiers/f/14/t/516029?Power-Pins-for-unused-channels-in-LMH7324-comparator

> From an operation point of view, only connect power and ground (VCCI, VCCO, VEE) to the channel you are using, and leave the unused channel pins (input, output, power, ground) unconnected.

With that in mind, I support only 4 parts. Each can be viewed as an isolated unit, sharing the same package, but otherwise nothing is shared.

gcormier:
@myfreescalewebpage I can't change the power pins as you'd like. If I set Pin 4 to Power Input, it changes Pin 29 to the same. If I set it to Passive, same behaviour.

poeschlr:
Are you still on version 4? This should have been fixed in version 5.

If this is still present in version 5 then you can use the same workaround we used in version 4. Back than we moved the pins apart, set them up correctly and only then stacked them.

gcormier:
Yes, v5. Okay, I'll wait until we clear up the direction before finishing touches :)

myfreescalewebpage:
As said above, I have no preference and agree with @evanshultz 

@poeschlr your position ? For this part, do you think power supply pins must be in a separated unit or attached to each single op amp they are linked in the device ?

Thanks,
Joel

gcormier:
@poeschlr Any thoughts?

poeschlr:
Power pins specific to a unit can be placed on that unit.
Global power pins must be in a separate unit. It is not allowed to duplicate any pin or set the pin to "show on all units".

If pins seem to be specific to a unit but must be connected to the same net then they must still be placed on a separate unit and stacked.

gcormier:
In that case I would propose to proceed with only 4 parts based on TI's recommendation that unused channels be left unconnected. The device effectively has no global power pins.

myfreescalewebpage:
@gcormier the design is great now with the 4 units.
Regarding to Vee pin stacking, you need to change type of visible Vee pins to "Power Input" (only the visible pins, the invisible pins remains Passive !)
Joel

evanshultz:
Looks good to me too, then, with regards to pins and layout. Only suggestion is to move the ref des and value both to the right of the power pins so the ref des text won't run over the power pins (if the user has disabled text auto-placement).

gcormier:
- [x] Move ref des/value as per @evanshultz 
- [x] Visible Vee set to power input (@poeschlr - it was indeed synchronized pin edit mode giving issues :) )

gcormier:
20 mil offset doesn't keep things spaced enough.

Pin name text size adhering to KLC would increase symbol size and follows other symbols going with smaller text when needed.

evanshultz:
Most of the triangle symbols are 5mil or 10mil. 20mil is for typical "black box" symbols with a rectangular body. With 10mil the text is somewhat floating away from the symbol outlines, and with 5mil the text just kisses the outline but I think it still looks OK.

gcormier:
If 5mil is acceptable, I agree it looks better.

Hopefully final commit?

image

evanshultz:
@myfreescalewebpage 
The Vee pins are corrected and this looks OK. It's been almost a day so I'm going to merge to ensure this is in 5.1.

@gcormier 
Thanks for this contribution and for participating in and allowing the discussion to settle on the power pins.

myfreescalewebpage:
@evanshultz yes I'm happy with @gcormier fixes too ! Thanks for merging to get it in 5.1 :) Thanks @gcormier for the job on this PR.

PR #1441, opened 2019-01-20
Fix KLC errors on Memory_RAM:628128

fauxpark:
sym

Datasheet: http://www.futurlec.com/Datasheet/Memory/628128.pdf

No idea if this should have a more specific name, it appears to be a somewhat jellybean part.

------------


evanshultz:
This seems to be an old part that is no longer in production. I don't find a direct manufacturer datasheet so I guess we aren't going to get better than this. It does seem to be a generic name so I'm OK leaving it alone.

You've totally redone this symbol, but it was terrible before. Here it is:
image

Let's hold off until after the final 5.1 library is released to merge this so it will go into v6. But as the original symbol was so bad, and it's very likely to not be used much, the risk of breaking a few schematics is worth this major improvement.

Notes:
- It looks like there are other access times besides 55ns. Could you add 70ns or remove that from the description?
- Only the DIP and SOP packages use this pinout. TSOP is different.
- Can you move the ref des and value down 50mil to be just above the part body?

Thanks as always!

fauxpark:
Added 70ns and moved the refdes/value down. As for the pinouts - I could have sworn they were all the same...

fauxpark:
According to the datasheet there is another TSOP package for this part with a "reversed" pinout - looks like reverse mount judging from the drawing. There's no footprint for it but since as you said this is a pretty old part I don't feel particularly motivated to add it.

poeschlr:
We are not obliged to offer symbols (and therefore footprints) for every part. Not even all available alternatives to parts in the lib. The question about the production state of the part does not even come into play here. (But well it makes even less sense to put work into maintaining obsolete or soon to be obsolete partts.)

evanshultz:
- Use underscores to separate the part name from the package where there are multiple vendors with conflicting MPN suffixes for the same part and package combination. See `Analog:LF398_DIP8` and `Regulator_Linear:LT1175-5_SO8_DIP8` for examples
- How is one supposed to know the TSOP symbol is for the normal pinout? Maybe add something in the description? Without the reversed symbol also there's no way to tell by the description. I'll  merge as is but if you think of something clear you can add it. As the original symbol was only for DIP/SSOP you haven't removed any symbols even if the TSOP isn't present, but since you've already made it I don't at all object to keeping it.

fauxpark:
I originally thought the reverse TSOP package was mounted inside the board, but that is actually not the case. According to [this question](https://electronics.stackexchange.com/questions/407703/what-is-a-reverse-type-tsop), it allows for placing the part on both sides of the board with "a minimum of wiring congestion". The pinout is exactly the same, but it is mirrored horizontally.

So the footprint is pretty easy to create - it might not be scriptable though.

evanshultz:
Great. Is there any reason to make the description more clear that this is normal/non-reverse?

fauxpark:
There is no difference as far as the symbol is concerned, the pinout is the same. I'm submitting a PR now for the reverse mount footprint with the suffix _Reverse, so unless that needs to be changed the current footprint filters should pick it up.

evanshultz:
Thanks!

PR #1447, opened 2019-01-20
Add LTC4055 and LTC4055-1 (USB Power Controller and Li-Ion Charger)

EvilMav:
[Datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/4055fb.pdf)

kicad_ltc4055

- IN1 and IN2 are combined into IN as they should always be connected.
- OUT is bidirectional as it is allowed to feed back 
- Pad and GND are combined as pad must always be connected to GND
- CHRG and ACPR are on the left as the would frequently just feed a status LED near the port and that makes the symbol more symmetric. Arguably they could be moved to the right near the inputs - please let me know if this is preferred.

------------


myfreescalewebpage:
Hi @EvilMav , thanks for contributing,

- [ ] Footprint filter should be `QFN*1EP*4x4mm*P0.65mm*`
- [ ] Pin 5 (WALL) is Input, not Power Input (block diagram page 10 of the datasheet)
- [ ] I would like to suggest IN should be in front of OUT. Move IN to the top left corner, and WALL just below.
- [ ] On another symbol we have set OUT and BAT Passive to avoid ERC errors. I suggest to do the same.

Cheers,
Joel

EvilMav:
Thank you for taking time to review! Agree with all points made, new version:

kicad_ltc4055

Issue #1442, opened 2019-01-20
Footprint-filter for Mini-Circuits symbols should use Case-code (not LandPattern code)


aewallin:
As noted in https://github.com/KiCad/kicad-footprints/issues/1308 the footprint filter for Mini-Circuits parts should use the case-code, not the land-pattern code.
This is done so that the filter will match both the "plain" footprint (no GND-zones/vias) and the LandPattern footprint.

The symbols affected are (more to be added, if found):

- [x] ADE-6
- [x] SXBP-140
- [x] LRPS-2-1

myfreescalewebpage:
Issue solved in attached PR.

PR #1448, opened 2019-01-21
Fix for #1442: Mini-Circuits footprint filters.

aewallin:
Discussion in #1442 

These symbols were using the land-pattern code PL-XXX in the footprint-filter, but this does not match the plain footprint which uses only the case code.


aewallin:
KLC error is about a hidden NC pin on the border of the symbol - not on 100mil grid.

myfreescalewebpage:
Thanks @aewallin for this work of consistency on those devices.
No comment to do, merging your modifications.
Cheers,
Joel

PR #1453, opened 2019-01-24
add MAX394 Quad SPDT CMOS switch

lorem-ipsum:
alias to MAX333A which is explicitly pin compatible. Extendends its footprint filter to contain the match the SOIC-20W variant used by the MAX394.

[Datasheet](https://datasheets.maximintegrated.com/en/ds/MAX394.pdf)

there is a extra component for the MAX333 and the MAX333A. Boths are pin compatible with each other, the only difference appears to be the footprint filter. In my opinion they could be the same part as one has to check for the footprint anyways. If this is not the case this should be an independent part too.


myfreescalewebpage:
Hi @lorem-ipsum thanks for this contribution,
No comment to do, merging.
Joel

PR #1454, opened 2019-01-24
add Micro-Chip SY58031U clock buffer

hvraven:
[Datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/sy58031u.pdf)
screenshot_20190124_092728


aewallin:
VCC pins could be stacked?

poeschlr:
The datasheet specifically asks for bypass caps at the VCC pins. Meaning they can not really be stacked. (My assumption would be that an ideal pcb design would have a separate cap for every VCC pin. To communicate this between the schematic designer and the layout guy you need every VCC pin separately in the symbol.)

myfreescalewebpage:

- [ ] I sugegst to have pins VT and VREF-AC type set to Passive to avoid ERC error in all possible configuration (datasheet page 9).

Cheers,
Joel

lorem-ipsum:
Hi @myfreescalewebpage,
sounds useful, I changed them to passive. I also added two aliases which are the same chip, but with different output signaling standards.

lorem-ipsum:
i've just realised that the two aliases should not be aliases, but new components as the signaling standard is PECL which uses open emitters for output. So the output pins have a different function. I will make the changes later this day.

myfreescalewebpage:
> i've just realised that the two aliases should not be aliases, but new components as the signaling standard is PECL which uses open emitters for output. So the output pins have a different function. I will make the changes later this day.

Yes exactly. Waiting for the change back then I will review the PR.

Joel

lorem-ipsum:
I split the aliases into separate components, should be good now.

PR #1460, opened 2019-01-26
Fix typo in dual opamp aliases list and add another opamp

attilakinali:
This fixes a small bug in the list of aliases for dual opamps, the currently listed LT6233 is a single opamp, not a dual one. The correct dual variant would be the LT6234. 
Additionally add another dual opamp to the list of aliases, the LT6237

Not done: Both the LT6234 and LT6237 come also in an DFN8 3x3mm with 1.7x2.43mm EP.
As that particular footprint isn't yet in the library, I didn't add it.

Datasheets:
LT6233/LT6234: https://www.analog.com/media/en/technical-documentation/data-sheets/623345fc.pdf
LT6237: https://www.analog.com/media/en/technical-documentation/data-sheets/623637fb.pdf


JacobEFO:
If you like @attilakinali I can make a PR resolving the two remaining travis violations ( as well as orienting part C (V+ and V-) on the y-axis, instead of having it fly around.

attilakinali:
Sorry, my bad. I missed that footprint somehow. And now travis runs successfully too.

I'll leave the fix up of the symbol to you, as I am not well versed in the rules that KiCAD uses.

JacobEFO:
> Sorry, my bad. I missed that footprint somehow. And now travis runs successfully too.
> 
> I'll leave the fix up of the symbol to you, as I am not well versed in the rules that KiCAD uses.

I'm very new to it as well, and I only offered you the PR since I randomly tried to resolve the travis issues, which turned out somewhat easy.

On another note, can I suggest you look into squashing your commits? I recently learned its functionality, and it seems very suitable on this PR.
I myself also had quite some commits for one PR, and its much easier to gain an overview, if commits are squashed beforehand, so one PR contains one commit.

Shackmeister:
@jacob
Commits gets squashed when merged :)

JacobEFO:
> @jacob
> Commits gets squashed when merged :)

@Shackmeister 
Simple and elegant. Nice.

poeschlr:
This has the benefit that we see the history while reviewing but can have a clean history in the master branch.

JacobEFO:
> This has the benefit that we see the history while reviewing but can have a clean history in the master branch.

I assume it's worth it. Didn't know of GitHub's merge functionality. I recall GitLab does not squash my commits :)

evanshultz:
@attilakinali 
I plan to split up the opamp/comparator symbols after 5.1 is out so this is fine for me. This is a valid fix and I'll merge it if you can resolve the branch conflicts.

evanshultz:
@myfreescalewebpage 
I fixed an error and a typo in this PR and merged in master to fix branch conflicts. Can you take a look at this? Since I made (actually fixed) the changes I probably shouldn't merge.

Not sure why the SAMD lib is showing up here. Seems to be identical to what's already on master from what I can see...

myfreescalewebpage:
@evanshultz just looked at this. Ok for the typo error, but I disagree with the added fp filter `DFN*8*1EP*3x3mm*P0.5mm*EP1.66x2.38mm*`should be `DFN*8*1EP*3x3mm*P0.5mm*` (no EP size in fp filter).

I have compare the modifications on SAMD library with Beyond Compare, with binary compare, no modifications. Don't know why git is showing a diff, but can be merged safely according to me.

capture

evanshultz:
Oh yeah. I noticed that and forgot to do it. Earlier I was thinking it's not a big deal since many possible matching footprints could have EPs, but then when I fixed the branch conflicts I realized this symbol doesn't have pin 9 so that FP filter doesn't work anyway. That's why somebody else needed to check this! Thanks for taking a look and should be fixed now.

myfreescalewebpage:
Thanks @evanshultz merging !

PR #1462, opened 2019-01-26
Diode: add BAT54W series

diegoherranz:
BAT54W, BAT54SW, BAT54AW and BAT54CW
https://assets.nexperia.com/documents/data-sheet/BAT54W_SER.pdf

See commit message for more details.

Regarding BAT54W, I wasn't sure how to proceed. It is a SOT-323 package but only pins 1 and 3 are in use. It's not easy to add a hidded pin on a diode symbol, and given that the SOT-323 package is 3 pin by default, I can't do something like `SOT-323-3` to force 3 pins on the footprint filter to comply with [KLC S5.2](http://kicad-pcb.org/libraries/klc/S5.2/). Please let me know if it needs any changes.

bat54w

Thanks!

---


Misca1234:
For the BAT54W
One suggestion could be like this

The "important" thing hidden NC pins is that the x is "far away" so the designer, when drawing lines, is not auto connecting to it.

bild

evanshultz:
For two-terminal diodes in SOT-23, we did this:
image

I'm not saying it's the right way to go, but it was one of the first things I contributed so I recall it.

I like having the unique diode footprints in a dedicated diode footprint library, but using a NC pin is common in the symbol library now and avoids having almost-identical footprints. I'm sorta on the fence but leaning a bit towards an NC pin as Stefan is showing above.

Misca1234:
Is it the same?
It is NXP special version of SOT, the so called SC-70, I have a faint memory that we added that one you and I

evanshultz:
Unless I misunderstand, the question is how to map a two-pin symbol to a three-pin footprint. I showed one way the library is using now above, but I think adding a NC pin is best (less maintenance of footprints and the package is identical anyway).

diegoherranz:
Thanks for all the replies. I have added the NC pin since I also think that it is better. I'm not that fond either of having/maintaining several footprints for what it is effectively the same thing.

Thanks!

Misca1234:
Should I add a SOT-323_SC-70-2 foot print instead ?
(there are several variants of the SOT-323_SC-70 already)

Misca1234:
The resulting symbol look like this

bild

evanshultz:
@diegoherranz 
Thanks!

Could you add a dot where the lines intersect around the center of the dual diode symbols?

@Misca1234 
I placed all the symbols on a schematic and through CvPcb and Pcbnew all the footprints and pinouts look good. Once there are dots go ahead and merge.

diegoherranz:
Dots/circles added.
I had copied these symbols from the existing BAT54S, BAT54A and BAT54C, so we should add the same circles there (and possibly to other similar symbols too). I'll try to submit another PR.

Updated screenshot:
bat54w

Thanks!

evanshultz:
Yes, indeed, all diodes should have this dot at line intersections.

Can you make the dot 10mil in size? That's what I see elsewhere in our library. Sorry for not checking that earlier.

diegoherranz:
10 mil now. I've forced-pushed, replacing the previous commit to have a cleaner history.
I had used 18mil trying to match the same size as used on schematic junctions, but if we use 10mil on other places, let's stick with that. I'll use 10mil for the PR for the rest of diodes too.

Thanks!

evanshultz:
Stefan was OK with this earlier so I'll merge. Thanks!!

PR #1463, opened 2019-01-26
Add XLR6 connector

RockyTV:
This is a pinout I've found by googling:

And I think it's worth noting that, while XLR cables were designed for audio systems, people can use these cables for anything other than audio.


myfreescalewebpage:
HI @RockyTV thanks for this contribution

No comment to do, merging.

Joel

PR #1466, opened 2019-01-26
Added STP08CP05 and STP16CP05

dominformant:
With these commits I added the STP08CP05 and STP16CP05 LED drivers.
STP08CP05 [Datasheet](https://www.st.com/resource/en/datasheet/stp08cp05.pdf)
stp08cp05
STP08CP05-XT
stp08cp05-xt
STP16CP05 [Datasheet](https://www.st.com/resource/en/datasheet/stp16cp05.pdf)
stp16cp05
STP16CP05-XT
stp16cp05-xt

With the -XT variants I was not sure about the correct package, as the datasheet names it TSSOP with exposed pad, while it is named HTSSOP in the ordering information ([Link](https://www.st.com/content/st_com/en/products/power-management/led-drivers/led-array-drivers/stp08cp05.html#samplebuy-scroll)). I decided to follow ordering information for the description and the datasheet for the footprint name and hope its the correct ways to solve this.

There are some more variants (e.g. [STP08DP05](https://www.st.com/resource/en/datasheet/stp08dp05.pdf), ...) which look similar to me, but use slightly different pin names in the datasheet. So, for now, I decided not to implement them, as I'm not sure, whether they should be aliases or individual symbols.


dominformant:
I just recognized that the CI build failed due to a missing footprint, for STP16CP05, I will add a pull request for that soon.

dominformant:
Footprints are fixed now, pull request for new footprints is here: [KiCad/kicad-footprints/pull/1348](https://github.com/KiCad/kicad-footprints/pull/1348)

dominformant:
I'm just thinking about moving the SDO pin to the left side of the symbol, so the shape would become smaller and it would be even easier to replace the eight bit driver with the 16 bit version. Any comments/hints on that?

myfreescalewebpage:
Hi @dominformant , thanks for contributing,

STP08CP05:
- [ ] Pin 15 type must be Passive to avoid ERC errors
- [ ] No need of comma in the keywords
- [ ] Can you also add DIP footprint filter to the list ?

STP08CP05-XT:
- [ ] According to ordering information, name should be "STP08CP05XT"
- [ ] Pin 15 type must be Passive to avoid ERC errors
- [ ] No need of comma in the keywords

STP16CP05:
- [ ] Pin 23 type must be Passive to avoid ERC errors
- [ ] No need of comma in the keywords

STP16CP05-XT:
- [ ] According to ordering information, name should be "STP16CP05XT"
- [ ] Pin 23 type must be Passive to avoid ERC errors
- [ ] No need of comma in the keywords

Cheers,
Joel

myfreescalewebpage:
No more comment to do, merging. Thanks for the fixes !

dominformant:
Thanks for the merge :-)

PR #1468, opened 2019-01-27
Adding data sheet and variants of the MC3306D and other power supervi…

Misca1234:
Well, we could not let evanshultz and ratfinks comments on the push 
https://github.com/KiCad/kicad-symbols/pull/386
be for nothing so I took it over with this push

MAX6355
https://datasheets.maximintegrated.com/en/ds/MAX6351-MAX6360.pdf
Question, shall we fix this up, Alignment of pins is not on 100 and the pins are way to long ?

MAX691
This was splitted in a DIP and SOIC package
https://datasheets.maximintegrated.com/en/ds/MAX6351-MAX6360.pdf
MAX691-SO   SOIC
MAX691-D     DIP
Question, shall we fix this up, Alignment of pins is not on 100 and the pins are way to long ?

TPS3831
http://www.ti.com/lit/ds/sbvs193d/sbvs193d.pdf
Just changed the FP filter

TPS3839DQN
http://www.ti.com/lit/ds/sbvs193d/sbvs193d.pdf
Just changed the FP filter

MC33064D
https://www.onsemi.com/pub/Collateral/MC34064-D.PDF
Made the SOIC and TSOP variants
Do anyone recognize the Micro8 capsule?
Shall I make one?

bild

bild

bild

bild

bild

bild

bild

bild

bild

---


Misca1234:
@KiCad/librarians 
I "salvage" evan and ratfinks review of the original push with this push, but many of the components have additional problems (see travis), should we fix those as well ?

evanshultz:
I would fix all of these by changing them to 100mil pins and adjusting the body size so that the pins are at the same location. Perhaps in the future they can be shrunk, but for now I wouldn't move pin locations since they are existing symbols.

MAX6355:
- FP filter should be `SOT?23*`.
- I wouldn't put an overbar on pin 5. It's setting a threshold for other pins and not actively responding to the level on this pin. And also, the datasheet doesn't give it an overbar.
- You can remove `/Triple` from the description because this specific part is only a dual.

MAX691:
- Since only MAXIM makes this IC, I think we can use the MPN suffixes. `xWE` is the SO package and `xPE` is for DIP. The trailing `E` perhaps can be removed?
- The DIP package is 16 pins.
- Perhaps put VBATT and VOUT near the bottom since they may have a cap or connect to ground?

*3x064:
- NC prefixes are for automotive so I'd remove this symbol. MC33064 are for high temp applications, so I'd remove them or rename the symbol to `MC3x064`. Since these symbols exist already, removing all but the two MC34064 symbols sounds best to me. Then remove temp range from the description.
- Pin 1 of the SOIC should be `~RESET`.
- GND is pin 1 of the TSOP-5.
- Don't stack NC pins unless it's called out in the datasheet and I missed it. Because these parts won't have any thermal issues I think all NC pins should be floating.
- We have an MSOP footprint that is compatible with ON Semi's Micro8. If you want to add that symbol it's fine with me.

Misca1234:
Some the pins not located on 100 grid either, should we adjust those as well or leave them where they are placed?

Misca1234:
**MAX6355:**
Fixed

**MAX691:**
> Perhaps put VBATT and VOUT near the bottom since they may have a cap or connect to ground?

If we change the location of one pin we could change them all ?

**3x064**
Keept the MC34064 in 4 variants, the 3x064 might not be so good, if the user enter MC33064, MC33064 or NCV33064 nothing will show up



bild

bild

bild


bild

bild

bild

bild

evanshultz:
Because these symbols already exist, let's leave the pins alone for 5.1. At 6 we can put more pins on grid.

MAX691:
- Remove the hyphen in the symbol names.
- You're right, let's not touch the pin locations.

34064:
- Leaving just the 34064 works for me.
- The DM suffix can have the package described as `MSOP-8` or `SSOP-8` or maybe `Micro-8`.
- The P suffix should be described as `TO-92`.
- The TSOP-5 still needs GND changed to pin 1.

Lastly, I don't see why a couple things were touched:
1. `MAX811TEUS-T` symbol description.
1. TPS.... pin name offset.

Misca1234:
MAX691
Removed hyphen in symbol name

34064DM
Changed the description to
"Undervoltage Sensing Circuit, MSOP-8"

34064P
Changed the description to
Undervoltage Sensing Circuit, TO-92

MC34064SN
Changed the pin number to 1 for GND pin

'MAX811TEUS-T` symbol description.
I was probalby changing the info about 34064 in an text editor and touched MAX811, change is reversed


TPS3839DQN
Changed it from 10 to 20 to get ride of travios complain, when I updated the FP, travis found it 
so i changed it so it would be no errors

evanshultz:
The description changes to 34064DM and 34064P aren't in your commit. Other things look good.

Misca1234:
It have been updated

evanshultz:
Thanks!

PR #1471, opened 2019-01-29
Diode: add/fix circles to dual diode junctions

diegoherranz:
Follow up from https://github.com/KiCad/kicad-symbols/pull/1462#issuecomment-458702070
See commit messages for details.
Thanks!


evanshultz:
All of them I found look good.

This doesn't affect the similar symbols in Device. lib that start with "D" and contain "x2".

Issue #1242, opened 2018-12-09
Analog_ADC: ADS1298_TQFP missing pin 28 (VCAP1)


gmsotavio:
The ADS1298_TQFP symbol from Analog_ADC library is missing pin 28 (VCAP1).

![screenshot_2018-12-08_23-38-27](https://user-images.githubusercontent.com/10669626/49692469-ded95b00-fb42-11e8-9b01-7fb8f9f38f06.png)

antoniovazquezblanco:
Thanks for the report and the PR!

PR #1472, opened 2019-01-30
Add VCAP1 pin to ADS1298 and update symbol

evanshultz:
Fixes https://github.com/KiCad/kicad-symbols/issues/1242.
Replacement of https://github.com/KiCad/kicad-symbols/pull/1243.

image

http://www.ti.com/lit/ds/symlink/ads1298.pdf

Changes:
-Rename symbol
-Change pin length to 100mil
-All pin styles can be Line.
-Pin name position offset should be 20mil.
-All DGND, AVSS, AVDD, and DVDD pins can be stacked. Pages 96-97 show that not each individual pin, but power pins grouped by name, need capacitance.
-VREFP should probably be Input type (like VREFN).
-VREFP and VREFN could be spread apart to let a cap fit in there.
-RLDINV might be best as an Input type. From the perspective of the RLD circuitry, from which it is named, it's an input.
-The VCAPx pins should to be on the left (analog) side.
-It might be nice if RLDINV and RLDOUT were spaced apart a bit for the feedback parts to fit there.
-I would group DAISY_IN with the other SPI signals, putting it at the top and DOUT at the bottom of the group to allow the easiest signal routing for devices that are daisy-chained.
-Shrink the symbol vertically and horizontally once that is all done.

---


evanshultz:
Is anybody able to review so this fix can go in 5.1?

antoniovazquezblanco:
> -VREFP should probably be Input type (like VREFN).

imagen

imagen

imagen

I am a bit lost with all the pin type stuff in this device 😆. From the datasheet I would not set the VREFP to input as it can work as output. Either bidirectional or passive would be my candidates and I agree more with passive.

> -RLDINV might be best as an Input type. From the perspective of the RLD circuitry, from which it is named, it's an input.

imagen

imagen

imagen

Again this one is tricky. There could be a signal in the input line comming from the PGA amplifiers as an output 🤔. Although that seems not to be the intended use so lets leave it as an input.

Just documenting my thought process. Everything else I agree on first sight. I will have a deeper look in a minute and will merge if nothing extrange is found.

Thanks!

antoniovazquezblanco:
Looks good to me ;)

Thank you very much!

evanshultz:
Thanks!

Pins that are not logic level are much more difficult to select in some cases. I favor Passive for these, in general, because otherwise we could have false negatives with the ERC engine. But it's an edge case.

Issue #805, opened 2018-08-08
Inadequate SPICE symbols


evanshultz:
There are a few issues I see with this library:
- Only a smattering of symbols are present
- Many symbols are very unlike the standard symbols in other libraries
- The size/shape/style of some symbols stands out (in a bad way)
- It would be nice to have some pre-configured voltage/current source symbols.

I will work on this.

@poeschlr 
This will require adding relevant fields to the symbols which would seem to violate KLC (see https://github.com/KiCad/kicad-symbols/issues/464#issuecomment-411449086), but KLC was written without SPICE simulation in mind and this is a special case so I think an allowance just for the simulation library is fine. Please let me know if this is not acceptable.

poeschlr:
As long as this stays (for now) within some specialized libs i am ok with that. I have absolutely no expirience with regard to spice and i really do not intent to learn it within the next few months. (Have too much other stuff to do)

poeschlr:
Just to add to this: please keep the current symbols alive in a way such that updating the lib would not break exisiting projects. (meaning there should be symbols with the same name and pin positions/pin numbers in the lib after your change)

evanshultz:
Yes, I understand. Honestly, I don't know who made these symbols or what they were thinking. They're so unlike the rest of the library, do not cover all SPICE primitives that the ngspice engine supports, and most are just plain ugly. And I already brought up my concern with the library name.

poeschlr:
From the git history of this lib in the old repo i gather that this lib existed since the beginning of the library here on github. Most likely it existed long before that. My guess would be it was used to test some early spice interface and never got updated as nobody needed it. Now that kicad supports simulation people will start to use it and notice how bad that lib is.

poeschlr:
Just to keep all information centralized here a repeat of my answer over at the mailing list:

Instead of simply renaming the existing pspice lib i would suggest to declare it deprecated. (Remove it from the default sym-lib-table). We might also want to add a "deprecated libraries" file that tells travis (and users) that some library is left out of the lib table on purpose.

I would then create a new lib with symbols that use all kicad 5 simulation specific features. It might be possible to make these symbols in such a way that they do not contradict our footpritns (pin number assignment for diodes for example. Here the spice pin number remap feature should be used instead of assigning the pins in the wrong order for footprints)

pgrudzinski:
I would be glad to help with that. Is there anything done on this subject right now? I do not want to duplicate something that is ready/being worked on.

[...]

PR #1473, opened 2019-01-30
Add new simulation symbol lib

evanshultz:
The first step in fixing https://github.com/KiCad/kicad-symbols/issues/805.

As mentioned at https://github.com/KiCad/kicad-symbols/issues/805, there are the many useful SPICE elements which are missing. The existing SPICE library's symbols are, frankly, just plain old butt ugly. Using the list of primitives at http://ngspice.sourceforge.net/docs/ngspice-manual.pdf, I have created a new set of symbols. This PR is only for voltage and current sources.

While KiCad doesn't currently have GUI control for all sources, I elected to created them all as they work if the SPICE parameters are entered manually. I have not implemented the external source since that may be platform dependent and would just all-around be better with a GUI.

Schematic for testing (also attached):
image
[new_spice_models.zip](https://github.com/KiCad/kicad-symbols/files/2814581/new_spice_models.zip)

Load this schematic and add the library in this PR as a new library named `spice`. You can see the sources are working by noticing changes in the random sources on each run.

And the resulting waveforms from voltage sources:
image

And current ones:
image

Questions:
1. The library is named `spice.lib`. Some of the libs have a lowercase name but most are uppercase. Which should this be?
1. The default parameters for the sources are selected to make them show up nice in the simulation above. If these symbols are OK, let me know and I'll push some typical default source settings (like 1kHz 1V for the sine voltage source and 1A for the DC current source, for example).
1. I've removed all extraneous symbols from `pspice.lib`. Since the existing transistor symbols simulate fine I don't see a reason to have them again, especially when they're big and ugly compared with the same symbols in our "main" libraries.
1. The diode symbol here is flipped (cathode to the right) relative to the old simulation lib and also relative to the normal library symbols. OK?

I'm marking this for 5.1 since I'd really like to make these available at that time. Not necessarily for RC1, but the final 5.1 release.

Edit: All Travis errors are OK with the belief that putting the source info on the schematic is good for simulation symbols. This is common in other simulation software. And it sure is helpful!

---


evanshultz:
To make this library I removed parts from the existing `pspice.lib` that I don't believe are needed. The previous conversation (linked above) indicated that when this was merged and ready, the old lib could be removed from sym-lib-table so only the new one remains. For that reason, I'm letting everybody know that I didn't bring over symbols which are duplicated elsewhere in our library and work fine.

I would prefer to capitalize the library name then. And instead of `Spice.lib`, which could be OK since the engine currently used is ngspice, I suppose we could also consider something more generic like `Simulation.lib`?

poeschlr:
I am still not certain what you mean with the removal comment. There are only two changed files listed by github. the pspice.lib and pspice.dcm are not changed by this pull request. Is this as it should be? 

(Nothing could have been removed by this pull request from that lib as it is not touched by this pull request. Meaning for your comment making any sense the removal must have already happened in a past pull request. Or you forgot to add or push these changes. The later option is why i ask as it something that can easily happen.)

poeschlr:
Ok i get it now.

Regarding name: I am not so sure if a more generic name is a good idea as these symbols kind of rely on the spice syntax. 
Assuming this syntax is even portable between different spice dialects. If not then we might want to really name it Ngspice (or ngspice)

Another possibility is to allow for future addition of symbols with other dialects. then we could include the prefix Simulation to the symbol name followed by the used dialect. In this case Simulation_spice or Simulation_ngspice.

evanshultz:
The sources, at least, use a general spice syntax. Other simulation tools copied this syntax for their sources since it was common. I'm sure there are cases where that isn't true, but I've used a number of simulators and they all specify sources this way.

For more specialized symbols it's not all common, but for now I think let's stick with the basics and provide a nicer alternative to the existing pspice lib. We can expand on this later and I like your idea.

poeschlr:
Ok in that case lets go with Simulation_spice for the sources and if we later add something that is specialized to ngspice we can still create a Simulation_ngspice lib.
Make sure the sym lib table comment makes it clear that the symbols are compatible with ngspice (and other spice simulators)

poeschlr:
Additionally: don't forget to add the library to the sym lib table.

evanshultz:
Alright. Renamed the lib and added the entry in the symbol table. I also changed the default source parameters to something generic (instead of the special ones I used to make the simulation above look nice).

Is there a way to link to http://ngspice.sourceforge.net/docs/ngspice-manual.pdf that would be helpful and convenient? Maybe in sym-lib-table? If users want to use these sources but need help they will find there.

poeschlr:
Hm so does this mean the sources are pre setup for a specific usecase?
`am(1 0 100k 1k 1n)` looks to me as if this is indeed the case. might it be better to have them in the lib with "named fields" that the user can replace without needing to read the manual? (and getting an error message if they do not replace it.

Or can the models read from normal kicad symbol fields? (In that case one could pre fill the field with sensible values)

evanshultz:
They are set up with some generic parameters. The text you show above is visible on the schematic canvas so the user can plop it down and then go right to work adjusting the source for their circuit without having to dive into any dialogs.

evanshultz:
This is exactly what you get when placing symbols from the latest commit:
image

fauxpark:
@evanshultz just a small thing, since SPICE is an acronym it should be capitalised in the library name :)

evanshultz:
@fauxpark 
Yep, you are right. Thanks for catching that. Fixed now.

poeschlr:
Thanks for your work. This will make simulation much easier for our users.

PR #1475, opened 2019-01-31
add AD8475 differential amplifier

lorem-ipsum:
This adds two symbols for the different housings of the AD8475 ([datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/AD8475.pdf). I tried squeezing everything into the triangle, please comment if thats all right. I dropped the numbers as they looked weird. The second variant is the same without the EPAD.
The LFCSP variant requires a footprint with different EP size which can be found in [this PR](https://github.com/KiCad/kicad-footprints/pull/1353).

screenshot_20190131_212349


myfreescalewebpage:

AD8475ACPZ:
- [ ] Pin number to be displayed
- [ ] Top/bottom/right pins can be longer to give more place to pin numbers
- [ ] Pin 17 should be named "EP" for consistency with other device and type should be Passive (not required absolutely to be connected to ground)

AD8475xRMZ:
- [ ] Pin number to be displayed
- [ ] Top/bottom/right pins can be longer to give more place to pin numbers

Cheers,
Joel

lorem-ipsum:
Hi @myfreescalewebpage,

thanks for the review. I've integrated your feedback. The number of the `+OUT` is still colliding with the triangle, but I would need to make it ridiculously long to fix this. I've also changed the pad as suggested. The datasheet states it should be soldered to ground, but its probably nothing the DRC has to complain about.

ad8475xrmz

myfreescalewebpage:
Hi @lorem-ipsum thanks for the fix.

AD8475ACPZ:
- [ ] Pins 7/8 should have the same length than pin 6 I think. Same comment for pins 14/15.

Else, ok for me.

Joel

lorem-ipsum:
Hi @myfreescalewebpage,

then these pins will no longer be on a 100mil, but on a 50mil grid. If thats ok I will change it. And most likely I will shorten them a bit in the process.

Cheers,
Hendrik

lorem-ipsum:
Ah, sorry. Messed up the figures. Your comment is about the invisible pins, right? Will fix it.

myfreescalewebpage:
Yes it was only invisible pins ! Thanks, no more comment to do for this PR, just need to wait for the footprint to merge.

evanshultz:
I have a couple thoughts:
- Is there a reason the descriptions are different? It looks like gain is selectable in both packages.
- The triangle symbols for opamps and comparators usually have 5mil pin name offset.
- I think we can simplify the pin naming a bit and shrink the text sizes to make things fit. There was a decision to leave no triangle-shaped symbol pin unnamed somewhere, which is done here, and shrinking pin text was the way to make that happen without blowing up the symbol size. See https://github.com/KiCad/kicad-symbols/pull/1038 as one example. This one has four input pins so I don't see how we can use the "standard" size triangle but at least it can be smaller. You can play around with it more, especially if there's a better way to clear the triangle outline and pin numbers (maybe just shrink them?), but here is my work up of this symbol in a similar fashion:
image
```
# AD8475ACPZ
#
DEF AD8475ACPZ U 0 5 Y Y 1 F N
F0 "U" 200 250 50 H V C CNN
F1 "AD8475ACPZ" 50 -200 50 H V L CNN
F2 "Package_CSP:LFCSP-16-1EP_3x3mm_P0.5mm_EP1.6x1.6mm" 0 0 50 H I C CNN
F3 "" 400 200 50 H I C CNN
$FPLIST
 LFCSP*1EP*3x3mm*P0.5mm*
$ENDFPLIST
DRAW
P 4 0 1 10 -300 300 -300 -300 300 0 -300 300 f
X +IN0.4x 1 -400 200 100 R 50 30 1 1 I
X + 10 300 -100 200 L 50 50 1 1 O
X - 11 300 100 200 L 50 50 1 1 O
X NC 12 -300 0 100 R 50 50 1 1 N N
X -Vs 13 -100 -400 200 U 50 30 1 1 W
X -Vs 14 -100 -400 200 U 50 30 1 1 P N
X -Vs 15 -100 -400 200 U 50 30 1 1 P N
X +IN0.4x 16 -400 200 100 R 50 30 1 1 I N
X EP 17 0 -400 250 U 50 30 1 1 P
X +IN0.8x 2 -400 100 100 R 50 30 1 1 I
X -IN0.8x 3 -400 -100 100 R 50 30 1 1 I
X -IN0.4x 4 -400 -200 100 R 50 30 1 1 I
X -IN0.4x 5 -400 -200 100 R 50 30 1 1 I N
X +Vs 6 -100 400 200 D 50 30 1 1 W
X +Vs 7 -100 400 200 D 50 30 1 1 P N
X +Vs 8 -100 400 200 D 50 30 1 1 P N
X VOCM 9 0 400 250 D 50 30 1 1 W
ENDDRAW
ENDDEF
```

myfreescalewebpage:
Proposition of @evanshultz looks great too. I have not reviewed a lot of op amp, I trust in you about these decisions. @lorem-ipsum can you address the comments above to get a symbol more consistent with the other devices of the lib ? Thanks.

lorem-ipsum:
I've played around a bit and got to this result: The font size is 0.035, but to make it fit like this I had to place the pins on top and bottom on a 50mils grid. Is this acceptable or do pins always need to be on the 100mils grid? Also I don't think there is any way except a ridicusly long pin to make the number of the positive output not collide with the outline. This is also true for most opamps already in the library.
screenshot_20190221_224134

evanshultz:
Thanks! Let's stick with a 100mil grid for pins. Can you look at the descriptions as I mentioned above?

lorem-ipsum:
Ok, will try again fitting everything onto 100mils grid. I will also fix the descriptions, first wanted to get your feedback on the version above.

lorem-ipsum:
I've updated the symbols to fit onto a 100mil grid. The size is now somewhere between the first attempt and yours. I find your version a bit to crammed, so I tried to find a compromise. Is this fine for you?
screenshot_20190222_205822

evanshultz:
Looking nice to me.

Two thoughts:
1. Would it be good to have VOCM between the + and - inputs? We have done that on other symbols? Certainly not required, and since you still need an EP pin on the bottom in one symbol, it may not buy you anything. If you want to play around that's fine, but I'm OK as it is.
1. If pin name offset is lowered to 10mil it gives a bit more space around the text inside the triangle. I don't find it compromises readability at all, either. (All, I think, of the smaller triangles are 5mil and KLC allows this).

lorem-ipsum:
I am not sure whether it helps anything to put VCOM in between + and -. It definitely would make sense in terms of its meaning, but as you said it will not save any space. And usually it will be connected to some Vref of an ADC which is also not located between the + and - pins. However having it in line with other symbols is a good argument, so I can change it if you want.

The pin name offset I increased back to 20mils on purpose as it prevents the + and - from clipping into the border of the triangle. And in this case the smaller distance does not gain anything. The problem is that on the 100mil grid one needs to create a lot of extra space to shrink the symbol by one step.

evanshultz:
OK. That's fine with me. Just asking. :)

There is no LFCSP footprint in the library. We don't want to merge symbols without a footprint, so are you up for making the footprint? We prefer scripted footprints if you can.

Secondly, there are a number of CI warnings about pin length because they are not increments of 50mil. While visually I find the symbol you've committed here to look fine, I used pin lengths in 50mil increments to avoid any of these warnings and I'd rather have more consistent symbols in the library if you can find a good way to use only 50mil pin increments.

lorem-ipsum:
I already made the footprint then posting this PR, it can be found [here](https://github.com/KiCad/kicad-footprints/pull/1353).

I will try to fit everything onto 50 mil increments, but its probably going to be difficult. I've seen the warnings, but chose to ignore them in favour of a better style.

lorem-ipsum:
Making the triangle slightly "flater" actually gave the desired result. I had to decrease the pin name distance to 10. Now the + slightly collides with the triangle, but it is still visible.
ad8475acpz

evanshultz:
@lorem-ipsum 
Got it. Yeah, there's some geometry that seems to work out nice. If all the triangle vertices are either zero or the same value (such as [0,-300] and [0,300] for the left corners and [300,0] for the right corner) then it's good.

Sorry if I missed the footprint link above. Thanks for sharing it.

@myfreescalewebpage 
Joel, sorry to butt in.  Do you thing. If nobody has looked at the footprint by Monday I'll do that.

myfreescalewebpage:
@evanshultz no worries, thanks for helping on this subject ! As indicated above, I have not a lot of experience with this kind of design (triangle) and I have carefully followed your exchanges, it was nice to learn several details :)

About:
> such as [0,-300] and [0,300] for the left corners and [300,0] for the right corne

It is not actually, but are you sure of the first two sets of coordinates ?

Joel

evanshultz:
@myfreescalewebpage 
I am sure... sure that it is wrong now that you caught me. The X coordinate should have been `-300`.

myfreescalewebpage:
BTW the symbol is not aligned with on the "300mil grid".

evanshultz:
Because this one has more pins it's bigger. So `400` and `0` or `500` and `0` work too to keep pin lengths in 50mil increments and we'll go with the smallest triangle size that works.

myfreescalewebpage:
Sure!

evanshultz:
I checked the footprint and it looks fine. Re-ran Travis here and all is green now.

myfreescalewebpage:
Yes, I think we can merge

evanshultz:
Alright. I'll do it then. :)

PR #1477, opened 2019-02-01
Remove the misplaced floating dot in the bottom right of the NRF24L01…

gbrown128:
Remove the misplaced floating dot in the bottom right of the NRF24L01_Breakout symbol

Nothing real changing here, just that accidental dot.


gbrown128:
Please excuse the rather crude drawing! The errant dot is outside the symbol. Sorry I should've included a screenshot, its not obvious against the background grid. This is the result I get under version `5.0.2-bee76a0~70~ubuntu18.04.1, release build`, using the latest version of the symbol on this repo.

Didn't deliberately change the HMC431 part, but there's a chance I opened it in the editor. I can close this and redo it if need be. (Process is new to me!)

Thanks

selection_063

gbrown128:
I've just started from a fresh clone of the repo, and was very careful not to go anywhere near the HMC431, but it still gets changed. Strange, perhaps a bug in my version?

evanshultz:
Likely not. The file writer in KiCad sometimes varies from release to release. I didn't look closely enough to see if it's a simple transform of the arc which result in the graphic. However, I already checked the symbol and there's no visual difference.

Perhaps the GAL renderer doesn't display squares with start and end points at the same location.

At any rate, it seems to harm to merge and the files open nicely so let's go with it. Thanks!!

gbrown128:
Thank you very much! Hopefully I can be of more help in the future!

PR #1478, opened 2019-02-01
Added SIM28ML

kajusK:
Adding SIM28ML GPS module

https://simcom.ee/documents/SIM28ML/SIM28ML_Hardware%20Design_V1.01.pdf

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1355
2019-02-01-122904_1920x1080_scrot


aewallin:
- you should probably add the NC pins (hidden, on box boundary)
- any reason for not stacking the GND pins?

kajusK:
@DanSGiesbrecht 

Sorry for the direction issue on the Vcc_RF, I read the description and obviously ignored 'O' in the direction column :)

Regarding the open-collector, I assumed these pins are open-collector as they are supposed to be I2C in the future version of the IC/firmware. But that assumption was most likely wrong as these pins are uart for now. Thanks for pointing this out, fixed.

DanSGiesbrecht:
No problem! That's why we review. 👍 
I'll have more time to review this later today/tomorrow.

---

#### As an aside (refer to https://github.com/KiCad/kicad-symbols/pull/1495):
My understanding is that I2C isn't only an open-collector, so to me it makes more sense to describe an I2C-only bus as: input/output (SCL of master or slave), and bidirectional (SDA).
image
(from [this TI app-note](http://www.ti.com/lit/an/slva704/slva704.pdf))

DanSGiesbrecht:
The changes you made look great, thanks! One more thing (sorry): even though I2C isn't supported yet, could you please change pin 16 to be bidirectional? This way, if/when it's implemented, this symbol will correctly reflect the labels.

poeschlr:
Footprint is merged. So only the I2C stuff is missing now.

myfreescalewebpage:
@kajusK are you still alive on this subject. Seems only one pin direction is not properly set, then it is possible to merge your contribution. Thanks Joel

kajusK:
I'm sorry for a long delay, the pin direction has been fixed.

myfreescalewebpage:
@kajusK thanks for the last fix !
@DanSGiesbrecht thanks for the review, I merge the commit right now to get it in 5.1 release which is soon.

PR #1479, opened 2019-02-02
Update fp assignement for TPS7A7200RGW

poeschlr:
This is to aling the assignement to the correctly named scripted
footprint.
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1363

---

See www.ti.com/lit/ds/symlink/tps7a7200.pdf#page=36


myfreescalewebpage:
Hello @poeschlr 
No comment to do, just need to wait for the footprint.
Cheers,
Joel

PR #1481, opened 2019-02-03
Fix footprint assignment for HIP2100

poeschlr:
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1366


poeschlr:
wrong EP size (again)

PR #1482, opened 2019-02-03
Fix footprint assignment for A4950E

poeschlr:
http://www.allegromicro.com/~/media/Files/Datasheets/A4950-Datasheet.ashx#page=8

footprint PR: https://github.com/KiCad/kicad-footprints/pull/1367


myfreescalewebpage:
Hello @poeschlr 
I suppose you want to fix footprint at the moment for the next release, and not the other KLC errors, so no comment to do, just need to wait for the footprint.
Cheers,
Joel

PR #1484, opened 2019-02-03
add "Heatsink:*" fp-filter to both heatsink-symbols.

aewallin:
as discussed in https://github.com/KiCad/kicad-footprints/pull/1217
the heatsink-symbols can use a fp-filter **Heatsink:*** so that the user sees the appropriate footprints.
just fp-filter changed.

---


myfreescalewebpage:
Hello @aewallin 
Thanks for this enhancement, no comment to do and merging.
Joel

PR #1496, opened 2019-02-06
Add Circuit Breakers

jremillard:
Add Generic circuit breaker symbols.

IEC
image

US/IEEE

image

---


jneiva08:
For remove the S3.6 error you should change 40 to 20 like in following image

screenshot

Still have more errors

myfreescalewebpage:
@jremillard thanks for this contribution. However, the dcm file should be modified too, I do not see it in the  git diff on the PR....

jremillard:
I updated the dcm files, pin name offset it now 20, I centered the 2P symbol, and included images of what the symbols look like.

fauxpark:
Per [KLC G1.6.3,](http://kicad-pcb.org/libraries/klc/G1.6/) the correct naming scheme for these symbols would be `CircuitBreaker_*`.

Also, keywords are separated by spaces, not commas, and I would make the symbol descriptions Sentence case, not Title Case :)

myfreescalewebpage:
Hi @jremillard , thanks for contributing,

- [ ] For this kind of very generic devices, no footprint is better, see the ampmeter for example. This will permit to use any connectors for example, that cannot be selected currently with you filter
- [ ] The length of the pins should be 100mil only (symbol < 100pins)
- [ ] Link to wikipedia not allowed, for this kind of generic device, just use "~" to indicate there is no datasheet
- [ ] All pins should be passive to avoid ERC errors in any configuration

Else design is nice :)

Cheers,
Joel

jremillard:
@myfreescalewebpage  - Thank you for reviewing the symbol. I made all the changes you suggested.

myfreescalewebpage:
Thanks @jremillard, just about the documentation link, you have left it empty, this should be fill with `~`

jkriege2:
Hi!

is this symbol an IEEE, or the IEC symbol? I think we decided to use IEC by default and add IEEE as variants (see e.g. wiggly resistors).

I'm asking, because this symbol looks an awfull lot like our juper symbols:
2019-02-08 09_42_48-symbol editor device_jumper

I found different styles for such symbols in IEC:
2019-02-08 09_44_47-iec 60617-1 pdf - adobe acrobat reader dc

2019-02-08 09_45_28-iec-60617 pdf - adobe acrobat reader dc

https://www.jaeger-direkt.com/onlineshop/installationstechnik/schutzschaltgeraete-sicherungen/leitungsschutzschalter/13901/ge/aeg-leitungsschutzschalter-3-polig-ep-63-c-32-a

Could you (also) provide the IEC-conforming symbol?

Best,
JAN

myfreescalewebpage:
Hi @jkriege2 That's a good point for you !

Symbol for IEEE:
ieee

Symbol for IEC:
iec

You are right it's good to do like the 4xxx family, to have both variants ("De Morgan" normal/convert).

Note: for the IEEE symbol, I suggest to add the dash line between the contacts to avoid confusion with jumpers.

Cheers,
Joel

poeschlr:
Do not use de morgan for that. Nobody will ever discover that there is a second variant. Use a separate symbol if you must.

myfreescalewebpage:
Ok @poeschlr thanks for the point !

@jremillard the conclusion is that:
- [ ] You should add a dash line between the contacts to look like the "official" symbol
- [ ] The part name should ends with "_US" for the IEEE version, so you have to rename the parts. If you wish to submit the IEC version too, the part name should end with nothing special.
- [ ] Documentation link to be corrected

Joel

poeschlr:
Parts in the device lib are not allowed to be specific to one particular part. Meaning generic name, no documentation (fill it with tilde '~') Footprint field empty but footprint filter filled sensibly.

If this is meant to represent one particular part then put it in a different library.

fauxpark:
@jremillard you've removed the underscore between `CircuitBreaker` and the pole count. I'm also not sure why you added "US symbol" to the ammeter symbol.

myfreescalewebpage:
The parts are nice now in my opinion.

@poeschlr there is no footprint filters, but I really don't know what we can indicate except `*`, it is possible to associate any connectors or any specific footprint to these symbol... What should we do ?

Thanks
Joel

PR #1499, opened 2019-02-06
Fix footprint assignemnt for SOP-4 isolators

poeschlr:
footprint pr: https://github.com/KiCad/kicad-footprints/pull/1385

http://www.vishay.com/docs/83510/tcmt1100.pdf

The second component (TLP290) is not recommended for new designs it seems: https://toshiba.semicon-storage.com/info/docget.jsp?did=12882&prodName=TLP290


poeschlr:
Second set of changes:
https://toshiba.semicon-storage.com/info/docget.jsp?did=12855&prodName=TLP290-4

evanshultz:
Is the note about NRND for the Toshiba part necessary? Surely we have this for quite a number of parts in our library. Is production status something you plan to capture in the official KiCad library?

poeschlr:
This only changes the footprint. I am a proponent of atomic pull requests. (One thing is done per PR.)

This keeps the history clean, makes reviews fast and easy, ....

I am especially unprepared to do anything that might need discussion if the change that is intended to be done in the PR does not need any discussion at all.

evanshultz:
The diff shows you added the text "not recommended for new designs, " to the description for the Toshiba part. If you did only change footprints I would have merged without any questions.

poeschlr:
Yes i added this note. This is because the datasheet has a watermark not recommended for new designs over the full page of it (on every page).

I do not suggest that we monitor every part in the lib but if we discover that a part is no longer recommended or if it is obsolete then we might really want to capture this fact in the lib.

evanshultz:
I'm not in love with this idea but you're the boss.

PR #1492, opened 2019-02-05
Add Geiger-Müller symbol

heitorPB:
I asked in the [forums about a Geiger tube symbol](https://forum.kicad.info/t/geiger-tube-symbol/15058), there was none so I decided to make one and contribute to KiCad :)

This is a generic part, without footprint.

gm

PR #1502, opened 2019-02-07
Move Geiger-Muller from Valve to Sensors

heitorPB:
I recently (#1492) added the Geiger-Müller symbol to library `Valve`, but the discussion [in the forums](https://forum.kicad.info/t/geiger-tube-symbol/15058/) said it should be in the `Sensor` lib. Here I move it from one lib to the other.

My apologies for putting it in the wrong place before.


DanSGiesbrecht:
@evanshultz @poeschlr is this considered to be a breaking change that should be saved for v6?

heitorPB:
It shouldn't be. This symbol was added yesterday to the libs.

myfreescalewebpage:
@DanSGiesbrecht yes I have merged the symbol yesterday, for me this is to be changed asap. I haven't notice it was discussed the symbol should go to the sensor lib on the forum, that's okay for me too.

PR #1508, opened 2019-02-08
Fix FP assignment TI PWQFN motor drivers

poeschlr:
Replaces old footprint with newly scripted one.

Example datasheet http://www.ti.com/lit/ds/symlink/drv8801.pdf
Footprint https://github.com/KiCad/kicad-footprints/pull/1401


evanshultz:
- The FP filter won't match because it doesn't start with an asterisk. Maybe prefix `Texas*`?
- Is it required to match the EP at the end of the FP filter since this is a specific package?

poeschlr:
done

PR #1509, opened 2019-02-08
Fix footprint for BQ25601

poeschlr:
https://github.com/pointhi/kicad-footprint-generator/pull/285

http://www.ti.com/lit/ds/symlink/bq25601.pdf


evanshultz:
Should the FP filter be prefixed with `Texas`? I can only recall that being done, as opposed to a leading asterisk, for other symbols using TI-specific footprints.

poeschlr:
Done

PR #1510, opened 2019-02-08
Improve SPICE library symbol descriptions

fauxpark:
Mostly just making them sentence case, but also reworded a couple to match the Device library descriptions (eg. DIODE and OPAMP).

---

### DIFF
--- a/Simulation_SPICE.dcm
+++ b/Simulation_SPICE.dcm
@@ -1,14 +1,14 @@
 EESchema-DOCLIB  Version 2.0
 #
 $CMP DIODE
-D Diode, for simulation only, anode on pin 1
+D Diode, anode on pin 1, for simulation only!
 K simulation
 F ~
 $ENDCMP
 #
 $CMP IAM
-D Current source, amplitude modulated (AM)
-K simulation
+D Current source, AM
+K simulation amplitude modulated
 F ~
 $ENDCMP
 #
@@ -37,8 +37,8 @@ F ~
 $ENDCMP
 #
 $CMP ISFFM
-D Current source, single-frequency modulated (FM)
-K simulation
+D Current source, single-frequency FM
+K simulation frequency modulated
 F ~
 $ENDCMP
 #
@@ -61,14 +61,14 @@ F ~
 $ENDCMP
 #
 $CMP OPAMP
-D Operational amplifier, single, +/-/OUT/V+/V-
+D Operational amplifier, single, node sequence=+/-/OUT/V+/V-
 K simulation
 F ~
 $ENDCMP
 #
 $CMP VAM
-D Voltage source, amplitude modulated (AM)
-K simulation
+D Voltage source, AM
+K simulation amplitude modulated
 F ~
 $ENDCMP
 #
@@ -97,8 +97,8 @@ F ~
 $ENDCMP
 #
 $CMP VSFFM
-D Voltage source, single-frequency modulated (FM)
-K simulation
+D Voltage source, single-frequency FM
+K simulation frequency modulated
 F ~
 $ENDCMP
 #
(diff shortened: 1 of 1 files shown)


evanshultz:
Thanks!

1. I had mostly seen all words capitalized in the description so that's why I did it like this. Either way is OK for me (actually I prefer this) but I'm not sure if this is something we care about standardizing in the library.
1. The SFFM sources would be "single-frequency frequency modulated", which sounds awkward if it's all written out together. This was the main reason why I spelled it out in parens. Is there a way to write out all letters in acronyms?
1. I really prefer more explanation about the node sequence of the opamp. That is going to trip people up if it's not clear and the slash-delimited text seems much less obvious to me.
1. Because the normal diodes in the library have pin 1 as the cathode (so bizarre...) this one has the big note which I copied from Rene's original commit after user issues. I'm fearing that if this isn't as big and clear as possible it will be more of a hindrance.

@poeschlr 
Can you take a look, especially on the last two items?

fauxpark:
(Whoops, I said title case instead of sentence case).

Since the description field is not a title per se, it should not be in title case. I think it also just looks bad, especially when there is a mix of title case, sentence case, lower case, etc. IMO this is something that *should* be in KLC because it helps make the libraries look like a more unified thing instead of a mishmash of different styles.

On the SFFM, yeah that is weird. There is actually no precedent for this in Device.lib. Perhaps AM/FM can go in the description and "amplitude/frequency modulated" can go in the keywords?

That is fair, and the generic opamp symbols don't do this. The problem I see is that the comma is already used as a delimiter in the description parts and the node sequence is all one thing. This is why I chose to use the slash. I can add back the words "node sequence" though if that helps.

For the diode, I can swap the pin 1 = anode and simulation warning and put back the exclamation mark so it makes (somewhat) more sense as a sentence.

poeschlr:
Do not remove the "simulation only" warning! It is there for a reason. (These symbols are incompatible with our footprints. Our footprints follow industry standards, simulation follows a different convention.)

fauxpark:
Yes, I was well aware of that.

fauxpark:
@evanshultz how does this look now?

evanshultz:
I'm OK with it. @poeschlr ?

Just as a side note, not to do with this PR, but there is another common node sequence for SPICE opamp subcircuits. I wonder if we should add a second one, thus providing both common options, for user convenience?

poeschlr:
We would then need to find some form of naming convention to clearly communicate to users when to use which option.

evanshultz:
Yes, but let's worry about that later. Is this OK to merge for you? If so, you can merge or let me know and I will.

poeschlr:
I did not check anything, just answered questions posted here in the comments.

evanshultz:
@poeschlr 
Would you mind checking? I'm ready to merge this but you wrote the original descriptions with the warning note and you watch the forums closely so if you have some helpful feedback it would be nice to hear it now.

evanshultz:
@fauxpark 
Can you somehow make the node sequence more clear for the opamp? Perhaps using numbers 1 through 5 would be super clear. Then I'm ready to merge.

fauxpark:
Added the numbers back. I still think it's fairly obvious, especially with the word "sequence" preceding it, that it's the pin order, but ah well.

evanshultz:
I think (and hope!) that many users may simulate circuits for the very first time with KiCad so a belt and suspenders approach to helping them is the right way to go. For example, https://fosdem.org/2019/schedule/event/ngspice/ seems to indicate that integrating circuit simulation into KiCad is a big deal for ngspice. We probably can't be too helpful.

Thanks!

fauxpark:
Fair enough.

PR #1511, opened 2019-02-08
add Pulse_H1100NL ethernet transformer

balcer:
add Pulse_H1100NL ethernet transformer

https://www.tme.eu/pl/Document/3fcf76bef7d362f3a0e2a140885dcbf0/H1102NL-Pulse-datasheet-118883.pdf

---


aewallin:
check the KLC for what to do with the NC pins. My interpretation of S4.5 and S4.6 (which might be wrong..) is that either
- draw NC pins as hidden, placed on the boundary of the symbol
OR
- include pin-count in the footprint-filter

DanSGiesbrecht:
Because the footprint contains all 16 pins, @aewallin's first suggestion is right: placing the NC pins along the border of the symbol.

poeschlr:
Yes NC pins must be included in the lib unless there are so many that the symbol would get cumbersome to use.

I have a clarification regarding NC  pin handling awaiting merging to the website repo: https://github.com/KiCad/kicad-website/pull/377

diegoherranz:
For the datasheet link, could you use one from the manufacturer's website instead of a distrubutor's? https://productfinder.pulseeng.com/doc_type/WEB301/doc_num/H1100NL/doc_part/H1100NL.pdf
The datasheet has changed quite a lot and seems to be missing the application circuit but is newer and clearer because it's just for that specific part so there are some positives too. In general I think we should use official links.

Thanks!

balcer:
Does the current symbol is acceptable?

balcer:
@DanSGiesbrecht Hej I'm planning to add 3D model to this component. But first I want to be sure that symbol is ok. When the review will end?

myfreescalewebpage:
@DanSGiesbrecht have you still time to review this pull request ? Thanks.

balcer:
What do I have to change?

evanshultz:
1. Please standardize the space after `RX` and `TX` in the description. Use it everywhere or nowhere.
1. Change `Temperature from -40 to +85` to just `-40 to 85C`.
1. Please use the HX1188 pin names for the center-tap pins.
1. Every winding needs a dot. Pins 2-3 and 7-8 don't have one.
1. Is there a reason the symbol is so tall? Can the TX and RX sections move closer to the x-axis?
1. You can get away with just 20 polylines, if the right-side center taps are a single polyline. Please take a look at that.

myfreescalewebpage:
@balcer I try to continue the review here

- [ ] Alias HX1188NL: datasheet link to be updated
- [ ] We never put the temperature in the description for transformers I think, I suggest just to remove it

Else looks good and will be able to merge it,
Joel

myfreescalewebpage:
@balcer ping

myfreescalewebpage:
No news of the author, indicate this is Abandoned

PR #1516, opened 2019-02-12
Fixed pinout of the LTC2508-32. 

PatrickBaus:
I fixed the pin layout of the LTC2508-32 (https://www.analog.com/media/en/technical-documentation/data-sheets/250832fc.pdf).

The RDLA and RDLB pins where incorrectly marked as outputs. They are in fact inputs and work like a ~CS pin on SPI devices.

I also fixed a typo in the desciption.


evanshultz:

You will find a list of rules and guidelines for the KiCad library at http://kicad-pcb.org/libraries/klc/. Please see that webpage for more information about the review items below. In fact, you may have already found it since you did such a nice job updating this symbol.

Only one thing needs to be done before this can be merged: Please remove the EP section of the footprint filter. It should end with the asterisk after the pitch. This allows other EP sizes, if they're in our library, to be used.

PatrickBaus:
@evanshultz: Thank you very much for clarifiying this. I wasn't sure whether to include the EP size and pointing my way to http://kicad-pcb.org/libraries/klc/S5.2/

evanshultz:
Thank you!

PR #1525, opened 2019-02-13
Move rf-filters to RF_Filter

aewallin:
Move recently added rf-filters to the new library RF_Filter (#605).

* LPF-B0R3
* RBP-280
* RBPF-246+   (+ indicates RoHS, maybe this is a bug and should be removed?)
* RLP-30
* SCHF-31+
* SXBP-100 (with aliases)

---


aewallin:
RBPF-246+ has a KLC error, hidden NC pin not on 100mil grid.
Not sure which is worse, the KLC-error, or placing a hidden NC-pin wholly within the symbol-box (not on the boundary) so it is on the 100mil grid.

poeschlr:
The grid warning has the weakest priority. I think we might need to adjust the KLC a bit here and maybe only make it a warning if a pin is on the 50mil grid. (especially a not connected one.)

poeschlr:
I opened an issue about that: https://github.com/KiCad/kicad-symbols/issues/1526

evanshultz:
The root issue is the pins on the component are 150mil in length. If the pins were all 100mil long, then the box is also on 100mil grids, and the NC pin lays on the outline and still on grid. Can you fix that? Then this should be OK.

aewallin:
I fixed the one symbol with KLC-issues, now has a rectangular 600x500 mil box, but not all others (for visual consistency)...
(in addition I used a 500mil square box for many other recent mini-circuits symbols)

evanshultz:
I would have asked you to use 100mil pins on all of these symbols, and that would have driven a different size of symbol body. There's no reason that they couldn't have used the recommended pin length from KLC. If you do want to change this now it would be nice, but that's too much to ask now.

Now I see what's going on with RBPF-246. I will go for consistency here, so if you want to revert that's fine but then scoot the NC pin on grid. That doesn't hurt anything and makes Travis happy.

The pins should be Passive type as this component contains only passive parts. It may not matter, but there could be a valid circuit that would trigger a false ERC error because of these pin types.

Some datasheets don't work for me. LPF-BR03, for example, should be `https://www.minicircuits.com/pdfs/LPF-B0R3+.pdf`. `www` instead of `ww2`. Can you confirm?

There is overlapping text, notably on the word "OUT" and the filter symbol. I would have wanted to get this cleared up before merging.

I just noticed the comment in your PR. Yes, RoHS does not need to be captured so you can remove the `+` from all symbols where it appears. This is the only thing I will request you to do before this PR gets merged, but I will create an issue to capture any items not done above (and applying to the pile of BFCN/HFCN/LFCN symbols) to align the entire RF_Filter library later.

aewallin:
Maybe in the future all of the passive RF-filters should be scripted? There are three different graphics (low, high, band-pass) for the box, but otherwise these just differ by the in/out/gnd pins.

evanshultz:
Thank you for moving these symbols and making the little tweaks so I can merge it!

Issue for these symbols is at https://github.com/KiCad/kicad-symbols/issues/1533. And yes, scripted would be fantastic!

PR #1527, opened 2019-02-13
Fix SANYOU SRD pin numbers

poeschlr:
Additionally: Ensure all filled polygons are closed.

This does fix https://github.com/KiCad/kicad-footprints/issues/1003

The fix was done in a way to avoid touching footprints.

screenshot from 2019-02-13 22-29-04
screenshot from 2019-02-13 22-29-26


Soldering-Stan:
This looks correct now.
I have double checked with what I had installed and pins 3 & 4 have been swapped and are correct.
As well as Sanyou, this relay also has Songle as the manufacturer.

evanshultz:
The above looks OK and I trust that Rene and folks with the relays in front of them have double-checked.

PR #1529, opened 2019-02-14
add OPA842, OPA843 and OPA846 as aliases

lorem-ipsum:
This adds three more aliases to existing symbols for the operation amplifiers [OPA842](http://www.ti.com/lit/ds/symlink/opa842.pdf), [OPA843](http://www.ti.com/lit/ds/symlink/opa843.pdf) and [OPA846](http://www.ti.com/lit/ds/symlink/opa846.pdf).


evanshultz:
Thanks! Just two description things to match the datasheet: OPA843 is missing the gain part and OPA846 is should say low noise, not low distortion.

lorem-ipsum:
made them match the datasheet description

evanshultz:
Thanks!

PR #1530, opened 2019-02-14
Texas Instruments TLV320AIC3106IZQER

jeanthom:
Hi,

Here's a symbol for Texas' TLV320AIC3106IZQER audio codec. I will try to send a footprint soon.

tlv320aic3106izqer

[Datasheet](http://www.ti.com/lit/ds/symlink/tlv320aic3106.pdf)
---


jeanthom:
A typo was fixed in the symbol, footprint should be OK

HubertD:
checked pin numbers, names, functions - looks okay to me.

would move the field reference to the symbol top left, component name to upper right

myfreescalewebpage:

Thanks @HubertD for the first step review.

- [ ] I do not agree with the symbol name, first we never include specific letter of the temperature ("I", and the package name is not correct). The correct symbol name should be `TLV320AIC3106xZQE`
- [ ] The footprint filter should be `Texas*MicroStar*Junior*BGA*5.0x5.0mm*Layout9x9*P0.5mm*` (see also travis result)
- [ ] Package name in the description is not correct, should be "BGA-80" instead of "BGA80"
- [ ] DRVDD pins should not be stacked
- [ ] All NC pins should be added to the symbol, type Not Connected, not visible
- [ ] Symbol is not centered (see travis error)

Cheers,
Joel

myfreescalewebpage:
@jeanthom are you still alive on this subject ? Thanks

evanshultz:
@HubertD 
I also thank you for taking time to review!

FYI, the ref des and value are likely auto-placed since this screenshot was taken from Eeschema. This is the default behavior when placing components.

@jeanthom 
A couple of non-required thoughts as I looked at the image above:
1. Perhaps line up the power pins so DVDD is above DVSS and so forth.
1. Since the MFP* pins are related to the serial interface, perhaps group them with the SCL and SDA pins, and perhaps even joined by RESET and SELECT. Those pins are likely to all go to a single host processor so having them together seems convenient. Shifting the GPIO pins to the left might allow this nicely.

myfreescalewebpage:
@jeanthom thanks for the first fixes, several points still to be fixed, see my previous comment at https://github.com/KiCad/kicad-symbols/pull/1530#issuecomment-468957768

myfreescalewebpage:
@jeanthom  ping!

myfreescalewebpage:

PR #1539, opened 2019-02-15
add LP3982 linear regulators

lorem-ipsum:
This adds the [LP3982](http://www.ti.com/lit/ds/symlink/lp3982.pdf) series of linear regulators from Texas Instruments as symbols. Footprint for the WSON is missing, will create it next. Most parts are exist with an extra `X` behind the package information, apparently indicating bigger package variants. Ignored those except for the `LP3982IMMX-2.82` as it does not exist without the extra X.

lp3982imm-adj
lp3982ild-1


myfreescalewebpage:

LP3982ILD-1.8
- [ ] Pin 8 (FAULT) should be type Open Collector

LP3982ILD-ADJ
- [ ] Pin 8 (FAULT) should be type Open Collector
- [ ] Pin 5 (SET) should be Passive to avoid ERC errors

LP3982IMM-1.8
- [ ] Pin 8 (FAULT) should be type Open Collector

LP3982IMM-ADJ
- [ ] Pin 8 (FAULT) should be type Open Collector
- [ ] Pin 5 (SET) should be Passive to avoid ERC errors

Cheers,
Joel

lorem-ipsum:
Hi @myfreescalewebpage,

thanks for the review, I've updated the symbols accordingly.
Is there a rule when a pin should be set to passive to avoid ERC errors, so I can try to do this correctly in the future?

Cheers,
Hendrik

evanshultz:
SET is a high-impedance pin which responds to a wide range of voltages in a linear (or at least proportional) fashion. It is an input to the IC in the sense that the pin is giving some information to the IC based on the voltage present.

Input type would also seem to be valid in similar cases (like opamp inputs) and our library uses both somewhat interchangeably. However, looking at the ERC settings you can see it's biased towards pins which have two fixed levels and catching problems that can arise when to drivers are connected together or there is a receiver but no driver. Those conditions to not apply to this SET pin.

While picking pin types on "digital/"logic" pins is usually much easier than in this case, discussion has been raised a few times about "analog" pin types without coming to a solid conclusion.

A Passive pin type will rarely trigger a DRC error, so we've often used it on "analog" pins like this. I'm not saying what's right or wrong or best or perfect, but in working with the current version of KiCad using Passive pin type here has worked well thus far.

lorem-ipsum:
Hi @evanshultz,

thanks for the detailed explanation, this makes sense to me now. I will try to use it in the future.

Cheers,
Hendrik

myfreescalewebpage:
@lorem-ipsum thanks, just need to wait the footprint before merging.

myfreescalewebpage:
Travis fails so checking by hand: footprint should be renamed here according to change on the footprint PR which has been merged, should now be "WSON-8-1EP_3x2.5mm_P0.5mm_EP1.2x1.5mm_PullBack".
Joel

myfreescalewebpage:
Thanks @poeschlr I have only restart the travis check in the travis result page, failing, but haven't think to closing/opening to solve the analysis issue.

poeschlr:
restarting via closing/open reloads the travis.yaml file which determines the python version. Restarting in the travis site only restarts the same build. We however load the scripts via git inside the script meaning it loads the new python 3 dependent scripts. (Not the best solution but with the easy workaround available i am not really motivated to invest time into implementing this in a better way.)

lorem-ipsum:
now the footprint filter should be correct. I kept the ThermalVias as they are recommended by the datasheet.

myfreescalewebpage:
@lorem-ipsum "LP3982ILD-ADJ" has still a footprint error (see travis log)

lorem-ipsum:
Hi @myfreescalewebpage, sorry for not spotting that, I have both versions currently in my footprint lib so the scripts are not complaining. Should be fixed now.

myfreescalewebpage:
Thanks, merging!

PR #1545, opened 2019-02-17
Added KX?22 series of accelerometers

SchrodingersGat:
- KX022-1020 - http://kionixfs.kionix.com/en/datasheet/KX022-1020%20Specifications%20Rev%2012.0.pdf
- KX122-1042 - http://kionixfs.kionix.com/en/datasheet/KX112-1042-Specifications-Rev-6.0.pdf
- KX222-1054 - http://kionixfs.kionix.com/en/datasheet/KX222-1054-Specifications-Rev-2.0.pdf

Single symbol with two aliases. In an existing 2x2 LGA-12 package:

Symbol copied from existing accelerometer in library.

image

**Note: The NC pin is *not* invisible because (as per the datasheet) it can be connected to various voltage levels if desired:**

image


evanshultz:
Oliver, to start with can you fix up the error caught by Travis?

SchrodingersGat:
@evanshultz please see my note in the comment above. The NC pin should be visible as the datasheet calls out multiple connection options for this pin.

evanshultz:
@SchrodingersGat 
My apologies for being so daft.

- The footprint may work, but it has no attribution. Did you confirm it's acceptable for this package? LGAs can vary a lot. In addition, -1042 is a thinner package than the others so a single footprint/3D model combo won't accurately represent the package for all 3 of these. I'll merge for now unless you want to make a dedicated footprint, but just FYI. And it also likely means our naming convention is inadequate to express all the variations in many footprints, but especially LGAs.
- There is a keyword for "spi" but not "i2c", although it doesn't bother me. Just FYI.
- Please remove the `n` from pin 10's name.
- Capitalize all letters in pin 7's name.
- Is TRIG on the right side because it usually comes from some external source and isn't related to whatever is communicating over the SPI/I2C bus?
- Could IO_VDD be closer to the SPI/I2C pins rather than VDD?
- Are the INT pins push/pull or open drain? I don't see in the datasheet.
- The symbol seems a bit lopsided, but I think that's because there's no reason to have equal pins on each side. Maybe shrink it?

SchrodingersGat:
From the [PCB layout spec sheet](http://kionixfs.kionix.com/en/document/TN008-PCB-Design-Guidelines-for-2x2-LGA-Sensors.pdf) it looks like the existing LGA-12 footprint will be acceptable.

- Added I2C keyword
- Removed n from pin 10
- Capitalized pin 7
- TRIG is on the right to separate the SPI pins from the GPIO pins. As is the case with existing accelerometer parts
- Moving IO_VDD would interfere with the SPI pin text, I don't think it's that critical
- The INT pins are push-pull (according to datasheet)
- The symbol matches the currently existing symbols. It might look a little wonky but I'd rather just keep the outline consistent with the current parts.

evanshultz:
Alrighty. Thanks!

PR #1546, opened 2019-02-18
Fix many rectangles in Device:CP

gaensli:
Device:CP (Polarized Capacitor) included 4 rectangle with either x or y size = 0. Replaced with one rectangle of same appearance.


evanshultz:
Good catch. Thanks!

Note: CP_Small looks similar but doesn't have this issue.

PR #1549, opened 2019-02-19
Added bq25895

Puchinger:
Added BQ25895
[link](http://www.ti.com/product/BQ25895)
bq25895


myfreescalewebpage:
Hi @Puchinger , thanks for contributing,

- [ ] The name of the part should include the footprint description to avoid conflict if the manufacturer propose variants, `BQ25895RTW` is great here
- [ ] Datasheet link should be `http://www.ti.com/lit/ds/symlink/bq25895.pdf`
- [ ] The package name is missing in the description, it should be `Controlled Single Cell 5A Fast Charger with MaxCharge, WQFN-24`
- [ ] Footprint filter should be `Texas*S?PWQFN?N*EP*`
- [ ] Pin 4 should be Open Collector
- [ ] Pin 10 should be Passive to avoid ERC errors
- [ ] Pin 13/14 should be Passive to avoid ERC errors
- [ ] You can center PGND
- [ ] Pin 19 should be Power Output
- [ ] Pin 23 is Power Output
- [ ] Pin 25 (Expose Pad) should be added and stacked with PGND
- [ ] The position of the pins you have chosen is very strange, something like the typical application schematic in the datasheet page 49 seems better

Cheers,
Joel

Puchinger:
@myfreescalewebpage 
Thank you for your review!
I have corrected the mistakes.
Here a new screenshot:
bq25895

myfreescalewebpage:
Hi @Puchinger thanks for the fixes !

PAD should be stacked with PGND because the datasheet indicates "Always solder PowerPAD Pad to the board, and have vias on the PowerPAD plane star-connecting to PGND and ground plane for high-current power converter."

Else sounds good, maybe you can reduce the a little the width of the device following the change above.

Cheers,
Joel

Puchinger:
@myfreescalewebpage
Know everything should be okay.
symbol

myfreescalewebpage:
@Puchinger this is great for me, merging ! Thanks.

PR #1550, opened 2019-02-19
Replace sensor/Geiger-Muller with sensor/Nuclear-Radiation_Detector

### LINKED ISSUE #1519: New Geiger-Müller symbols
We had a discussion in the forum about Geiger-Müller symbol [here](https://forum.kicad.info/t/geiger-tube-symbol/15058/). I came up with a symbol based on Google image searches. #1492 added it and #1502 moved it to the right place.

User 9V1MI in the forums found IEEE 315 that specifies how the symbol should be and paulvdh drawed these two:

symbols

We have some questions on how to proceed now:

- [x] The correct name is `Geiger-Müller`, but the letter `ü` is not allowed by the KLC. There was a comment by @evanshultz [here](https://github.com/KiCad/kicad-symbols/issues/1470#issuecomment-462957116) to replace the umlaut with `ue`: `Geiger-Mueller`. I guess it would not be a problem for people that use `u` instead of `ü`?
- [ ] How should be the names of the other 2 alternative symbols? `Geiger-Mueller_alternative` and `Geiger-Mueller_alternative2`?
- [x] How should be the names of the pins? `1` and `2`? `E1` and `E2`? `A` and `C` (anode/cathode)?
- [x] Just to be sure: they belong in the library `sensor`, right?

heitorPB:
fix #1519

Symbol on a 50 mil grid:
nuke


evanshultz:
- The description should not be a copy of the symbol name. Describe that this device does.
- I'm fine with no datasheet link (just a tilde as was done before).
- Perhaps "muller" should also be a keyword to help searching without diacritics or for users that aren't familiar with the representations of umlauts?
- I'm not sure about the pin names. Why C1 and C2? It looks like the tube has an anode and cathode, which for a diode in our library are named A and K.
- Are there any footprints existing or that you plan to add? Perhaps we can use a footprint filter?

heitorPB:
This is a generic symbol, for multiple radiation detectors. So, no footprint.

About the pin names, I renamed to 1 and 2.

As for the descriptio, `Generic radiation detector` is enough?

evanshultz:
I'd like to get more feedback on pin names. Are 1 and 2 OK, or is A and K better?

I realize this is a generic symbol, but there may be something about the footprint naming that we can key on in the footprint filter. See `LED` symbol in Device..lib I just don't know what that would be without knowing that footprints might be used. Leaving it blank is OK, too, if that's best.

The description is fine if nobody has other suggestions.

Would you please not force push commits? It's easier to see changes. Thank you. I'll squash merge so the commit history will be clean.

poeschlr:
Regarding pin names: How are they called in the datasheets? Do datasheets use numbers? If so use numbers. If not use letters.

I assume that these come only in highly specialized packages. (Unlike for example diodes) This is why i would be ok with letters.

heitorPB:
The datasheets I read about Geiger-Mueller tubes don't say anything about the pin names. Some said `ground lead` for one of the pins. But if this symbol would be more general than only Geigers, I'd be more confortable with `1` and `2`.

They come in very specialized packages and sizes.

evanshultz:
I suppose this is OK and no more comments have been added in a couple days. I'll merge this in a few hours if there is no more feedback.

evanshultz:
Thank you!

PR #1552, opened 2019-02-20
Added dual zeners to Device

matthijskooijman:
These symbols are based on the dual schottky diodes and follow the same
naming, but are changed to zener diode graphics. A footprint filter is
also added. Almost all available devices of this kind appear to use some
SOT footprint, so the footprint filter selects all SOT footprint (see e.g. [digikey](https://www.digikey.nl/products/en/discrete-semiconductor-products/diodes-zener-arrays/295?k=&pkeyword=zener&sv=0&sf=1&FV=ffe00127%2C33c0195%2C33c0196&quantity=&ColumnSort=0&page=1&pageSize=25))

These symbols have two warnings from checklib:
 - Symbol not centered on origin
   Due to the graphical nature of the symbol, the center seems a bit
   ambiguous. The centering of this symbol follows the lead of existing
   symbols.
 - Symbol possibly missing default footprint
   This seems a false positive, as this is a generic symbol, so it
   should not have a default footprint.

image


evanshultz:
Thanks!

Let's also add `SC*` as a footprint filter since those packages are also common.

matthijskooijman:
Ok, I've force-pushed a version with `SC*` added.

evanshultz:
Thanks!

In the future would you please not force push so it's easier to see the changes in each commit? I'll squash merge so commit history will still be clean. Thanks!

matthijskooijman:
> In the future would you please not force push so it's easier to see the changes in each commit? 

I'll try to remember. Thanks!

PR #1553, opened 2019-02-20
NXP SGTL5000XNLA3 audio CODEC

jeanthom:
Hi,

This is a symbol for NXP's SGTL5000 audio CODEC in QFN-20 package.

sgtl5000xnla3

[Footprint PR]() - [Datasheet](https://www.nxp.com/docs/en/data-sheet/SGTL5000.pdf)

---


jeanthom:
CI fails because the matching footprint hasn't been merged yet:

```
Checking symbol 'SGTL5000XNLA3':
  Violating S5.1
    filename
    Specified footprint does not exist
     - Footprint file Package_DFN_QFN:QFN-20-1EP_3x3mm_P0.4mm_EP1.7x1.7mm was not found
```

HubertD:
Checked pin names and numbers - looks okay to me.

Pin functions:
- Datasheet states on i2c: "The SGTL5000 is always the slave on all transactions, which means that an external master always drives CTRL_CLK" - Pin type "input" is probably okay for CTRL_CLK ✓
- Device can act as I2S master or slave device, so I2S_LRCLK and i2S_SCLK are correctly defined as bidirectional pins ✓
- Not so sure about the analog pins?
  These are all defined as "passive", though they have a clear input or output function.

Looking through audio codecs already in the library, analog pins seem to be declared as input / output as well as passive. KLC does not seem to have anything to say about analog pins?

Chosen QFN-20-1EP_3x3mm_P0.4mm_EP1.7x1.7mm looks good to me.

myfreescalewebpage:

Thanks @HubertD for the first step review.

- [ ] "sgtl5000" to be removed in the keywords
- [ ] The footprint filter never specifies the EP size, should be `QFN*1EP*3x3mm*P0.4mm*`
- [ ] Asp pointed by @HubertD I think HP* can be output, LINEIN* can be Input, LINEOUT* can be output, MIC_BIAS can be output, MIC can be input, it will not trigger errors here

Cheers,
Joel

jeanthom:
Hi,

Thank you both for your comments.

I am confused because the TLV320AIC23 uses the "Passive" type for its analog audio inputs & outputs. S4.4 isn't clear about the Passive type either... Is the TLV320AIC23 symbol incorrect ?

I will do the other changes ASAP.

HubertD:
Sorry @jeanthom,
no need to be confused by my comment.
As far as I can tell, there's nothing wrong with your pin types. 

Just wanted to state that there seem to be both usages in the library, that there is no definitive rule and I feel that it would be better to have such a rule.

thanks for your contribution, btw :-)

myfreescalewebpage:
@jeanthom do you speak of the PR https://github.com/KiCad/kicad-symbols/pull/1530 ? I don't see Passive analog I/O in this one...

The rule is very simple: use real input or output types as soon as possible. When ERC checking may fails because of specific wiring, you should use Passive type. A good example is VAG pin on this symbol.

jeanthom:
Nope, I'm talking of the TLV320AIC23* chips currently in kicad-symbols:

capture d ecran 2019-03-06 a 17 33 50

myfreescalewebpage:
Ok, this symbol need to be updated in this case, many are like that.

myfreescalewebpage:
@jeanthom any news here ? Thanks.

evanshultz:
Just a couple drive-by comments:
1. Perhaps put LINEIN* and LINEOUT* across from each other?
1. This symbol could be designed to support the extra pins in the 32-pin QFN package (even without this symbol be added right now):
- Add spots for two more CTRL* pins beneath pin 19.
- Use the extra height to place CPFILT on the left side.
- AGND can go to the left of GND so it's close to VAG (note 4 on page 59 of datasheet).

myfreescalewebpage:

PR #1556, opened 2019-02-20
TLP3123

myfreescalewebpage:
Add Toshiba TLP3123 Solid State Relay. Symbol retrieved from ASSR-1218, desription and footpritn are different.

---

https://toshiba.semicon-storage.com/info/docget.jsp?did=10047&prodName=TLP3123

capture

https://github.com/KiCad/kicad-footprints/pull/1424


evanshultz:
1. Is `1 Form A` used for these parts? I've not seen it. It applies to relays, so I see the connection, but it doesn't sound right to me. How about just getting rid of that part of the description?
1. Pin name offset should be 20mil.

myfreescalewebpage:
@evanshultz thanks for the review.

1. Agree, made a modification for this. Probably something to think more globally also as already said in https://github.com/KiCad/kicad-symbols/issues/1617
2. Oups, sorry, the symbol was retrieved from another one, just have to modify the pin numbers ! Fixed.

Joel

PR #1558, opened 2019-02-21
add AD9513, AD9514 and AD9515

lorem-ipsum:
This adds the clock distribution ICs [AD9513](https://www.analog.com/media/en/technical-documentation/data-sheets/AD9513.pdf), [AD9514](https://www.analog.com/media/en/technical-documentation/data-sheets/AD9514.pdf) and [AD9515](https://www.analog.com/media/en/technical-documentation/data-sheets/AD9515.pdf). They are very similar except their outputs. Footprint PR will follow. I placed VREF on top even though it is an output as there was space. It is used for the configuration of the S[0..10] pins. Alternatively I was thinking of splitting the part into two, moving the configuration to another part.

ad9515
ad9514
ad9513


myfreescalewebpage:

For all devices:
- [ ] Description, "Output" => "Outputs" (plural)
- [ ] The footprint filter never contain the EP size, but contain presence of EP, this should be `LFCSP*1EP*5x5mm*P0.5mm*` for this symbol
- [ ] VS pins should not be stacked, even if this enlarge the symbol a lot
- [ ] VREF should be on the right of the symbol (Power out, not power in)
- [ ] "B" in pin name is for complementary in/out. would suggest to use "~", thus `OUT0B` => `~OUT0`, `OUT1B` => `~OUT1`, `OUT1B` => `~OUT1`, `CLKB` => `~CLK`, `SYNCB` => `~SYNC`
- [ ] The datasheet indicate the EPAD sis connected to GND and is a power pin, so you should stack EPAD with GND
- [ ] Reference in top left corner, name in top right corner is great, instead of inside the symbol

Cheers,
Joel

evanshultz:
I'd prefer to keep the pin names as on the datasheet and not remove the "B" so the symbol and documentation match. KLC doesn't request this, either.

GND pins can always be stacked. Power pins should be stacked unless there's a reason not to, like the datasheet requiring a decoupling cap on each pin. The top of page 22 of the datasheet and https://www.analog.com/media/en/technical-documentation/evaluation-documentation/53447263508842AD9513_14_15_RevB_Eval_Board_Schematic.pdf both indicate to me that each VS pin needs a cap, so yes, not stacking is the right way to go here.

myfreescalewebpage:
@evanshultz 

> I'd prefer to keep the pin names as on the datasheet and not remove the "B" so the symbol and documentation match. KLC doesn't request this, either.

The datasheet indicate it is the complementary output, and we usually do that for example when a pin is "RSTn" or "CSn", they become "~RST" and "~CS".
Here do you mean it's bette to have "~OUTB" ?

Joel

evanshultz:
KLC discusses overbars for logic low signals, and not complementary/differential signals. Here the usage of "B" in pin names is mixed. I would keep the "B" on outputs but it's uncommon (to me) for logic low inputs; for consistency maybe keep all "B"? IMO the overbar adds clarity and probably wouldn't cause any confusion so what you mention works me. This isn't really covered in KLC except for the part about the datasheet winning if there's conflict. What do you like?

myfreescalewebpage:
@evanshultz thanks to point me the difference, you're right it's not really some "logic low" in/out.
I have look for other devices in the library to find examples and found very few devices, for example SAD1024, which is not with the overbar and keep the name of the datasheet. I have just removed my comment above for this reason, to keep the consistency, and we will need to remember this choice or indicate the difference in the KLC in the future.

myfreescalewebpage:
@lorem-ipsum any news here ? Thanks for the update, Joel

lorem-ipsum:
Sorry, I was occupied for some time with personal issues. Trying to slowly work off the queue of PRs now.  I've integrated the review feedback, the symbols now look like this:
AD9513
AD9514
AD9515

myfreescalewebpage:
@lorem-ipsum no worries and thanks for the update of the symbol. Just notice one remaining issue:

- [ ] On the 3 devices, RSET should be Passive to avoid ERC errors

Then only need to wait the LFCSP footprint before merging (https://github.com/KiCad/kicad-footprints/pull/1426).

Cheers,
Joel

evanshultz:
Because of all the VS pins, the symbol has a lot of unused area. Including at the bottom left. What about moving all the pins down a bit and the pushing in the sides so VS is all across the top? Here is a quick mock-up that doesn't even require making the symbol taller:
image

myfreescalewebpage:
Good proposition of @evanshultz too !

lorem-ipsum:
Good idea. I've compressed the symbol as suggested. Now only the reference and the name stick out and look a bit ugly:
AD9513

myfreescalewebpage:
That's great for me, just need to wait the footprint now.
Cheers,
Joel

myfreescalewebpage:
Footprint merged. Merging the symbols.

PR #1561, opened 2019-02-22
Added TPS2121

Puchinger:
Added TPS2121
According footprint is in pull request https://github.com/KiCad/kicad-footprints/pull/1422
http://www.ti.com/lit/ds/symlink/tps2120.pdf

symbol


myfreescalewebpage:
Hi @Puchinger , thanks for contributing,

- [ ] Name of the part should be "TPS2121RUX" to handle possible addition in the future
- [ ] Description should end with the complete package name "VQFN-HR-12"
- [ ] Datasheet field in the Fields Properties window should be empty
- [ ] GND should be Power Input
- [ ] ST Should be Open Collector
- [ ] ILIM should be Passive to avoid ERC errors
- [ ] I do not understand how you have chosen to group the pins on the left. Something like the schematic example page 25 of the datasheet seems good to follow.

Not sure about the name of the footprint, let see when the footprint will be reviewed.

Finally, when you make a PR, please look at the travis tests results because you currently have many errors you should fix yourself before the review. Thanks.

Cheers,
Joel

Puchinger:
@myfreescalewebpage
Thank you for your review!

I thought the errors of travis are because the footprint is not yet available.

I have corrected the mistakes.
Here a new screenshot:
bild

myfreescalewebpage:
Thanks for the fixes, not yet reviewed but will do asap.
All travis errors are not due to the missing footprint, you have also some matching errors (S5.1) and formatting error (S5.2) that should not appear.

myfreescalewebpage:
@Puchinger thanks, that's clearly better, but why have you got removed the footprint field itself ? This is the only acceptable error while the footprint is not merged (a single red line in the travis result). Please can you put it back ?
Thanks
Joel

Puchinger:
Okay I added the footprint again.

myfreescalewebpage:
Thanks @Puchinger for the fixes, seems ok except the footprint filter. You have the following travis result:

capture

Only the first line "Footprint file..... was not found" should be here when the footprint is pending.

To fix it:
- [ ] Footprint filter should be `Texas*VQFN*HR*2x2.5mm*P0.5mm*`

Joel

Puchinger:
Thanks for the quick response, now everything should fit

myfreescalewebpage:
Yeah, travis is better now :)
Just need to wait the integration of the footprint to merge this symbol.
Thanks for the work here
Joel

Puchinger:
@myfreescalewebpage 
I think the ST pin should be passive, because if I connect it to GND, as the datasheet says on page 4, I get an erc error.

myfreescalewebpage:
@Puchinger it's clearly an open collector output, but I assume TI has a good reason to indicate it's better to connect to GND if not used (noise for example). So in this case agree to set it Passive, you can make the modification, thanks for the report !

myfreescalewebpage:
Thanks, waiting for the footprint.

PR #1563, opened 2019-02-22
convert all analog URLs to https

lorem-ipsum:
analog supports https on all its sites, so change all datasheet links to support this.
Done by simple search and replace.


evanshultz:
I checked at least 20 and they all were good, so I'll merge. Thanks!

PR #1565, opened 2019-02-23
Connector/Conn_ARM_JTAG_SWD_10: fix pin names, types and link.

diegoherranz:
See commit message for details.

The location of the GNDDetect pin is debatable but I haven't changed it for backward compatibility.

ARM doc: http://infocenter.arm.com/help/topic/com.arm.doc.ddi0314h/DDI0314H_coresight_components_trm.pdf

arm10

Thanks!

---

- (N/A) If there are matching footprint PRs, provide link(s) as appropriate


myfreescalewebpage:
Hi @diegoherranz thanks for the fixes.

No comment to do on this. Just want to ask @evanshultz your opinion about the travis errors here: I think they should be fixed, and for me only the 1.27mm pin header should remain. What is your opinion ?

Joel

diegoherranz:
I hadn't checked the footprint or footprint filters. I saw the Travis stuff after submitting. I think we should fix it and leave just the 1.27mm one since that is the connector defined by ARM for this.

If anyone wanted to use a different type of connector for whatever reason, they're free to do so but it doesn't need to be in the footprint filter.

myfreescalewebpage:
@diegoherranz I think you can fix it like that, totally agree with you and it is still possible to use any other connector the end user wants, but ARM define one and this should be the one in the lib too.

diegoherranz:
Footprint filter fixed. Please note that it includes the pin count which being strict to KLC, shouldn't be included. However, the connector needs to be dual row so the `2x` is required. I considered `2x??` as other generic connectors, but it would all become `PinHeader?2x???P1.27mm*` which has too many question marks together and it seemed to me that `PinHeader?2x05?P1.27mm*` was clearer. Let me know otherwise.

Thanks!

evanshultz:
So to summarize, taking from the detailed initial commit log and stuff after, the changes are:
1. Documentation link
1. Footprint filter
1. Pin 1's name
1. Pin 9's type

I think all the pin types are OK, we should default to the header ARM says to use, and this symbol looks nice.

Two thoughts (I'm fine to merge it but wanted to share this):
1. The default footprint is a SMD one. Do THT headers work too? If so, perhaps there shouldn't be a default footprint and we rely on the FP filter to provide options for user selection?
1. If the connector can have pin 7 removed to act as a physical key, will our footprint filter catch that? Maybe we have a `_ARM` suffix for the header so the full 10-pin or the keyed 9-pin will be shown in CvPcb? Do you know what is specified by ARM and what is commonly used to help decide?

diegoherranz:
> The default footprint is a SMD one. Do THT headers work too? If so, perhaps there shouldn't be a default footprint and we rely on the FP filter to provide options for user selection?

Yes, THT would work too. I think it is less common on this 1.27mm pitch, but could be used perfectly fine. E.g. https://www.samtec.com/products/ftsh-105-01-f-d-k
It's probably a good idea to remove the default footprint.

> If the connector can have pin 7 removed to act as a physical key, will our footprint filter catch that? Maybe we have a `_ARM` suffix for the header so the full 10-pin or the keyed 9-pin will be shown in CvPcb? Do you know what is specified by ARM and what is commonly used to help decide?

ARM specifies that pin 7 should be removed. However, real world, I think it is more common to have that pin than not.
Having said that, even if the intention is not to have a pin 7, I think that the footprint for a keyed connector should include the 10 pads, especially knowing that it is quite common to end up using a non-keyed connector with the 10 pins (they are easir to find/source than the keyed versions).
On a similar basis, when you search for these keyed connectors on distributors like digikey, the still show as 10 pin connectors. 
In summary, I don't think I would change anything in this respect.

Thanks!

myfreescalewebpage:
Agree with @diegoherranz

myfreescalewebpage:
I have just notice there is also the 20pins version in the lib : Conn_ARM_JTAG_SWD_20. I suggest that both devices should be aligned (coherency of descriptions, keywords, footprints, footprints filter, pins names and position...)

@evanshultz also I was wondering if the Travis error should be fixed right now ? or postponed to a major release ?

Thanks
Joel

diegoherranz:
The thing with the 20 pin connector is that is not the current 20-pin trace connector but the legacy 20-pin connector which uses a different pitch (2.54mm) and different signals. The new documentation link used here doesn't include info about the legacy one for instance. I guess we could fix the footprint and/or footprint filter, though. Let me know if you want me to fix it too.

Related to this, I'm working on a symbol for the current 20-pin connector (1.27mm), which in fact, should probably be called `Conn_ARM_JTAG_SWD_20` and renaming the legacy one as `Conn_ARM_JTAG_SWD_20_LEGACY` or something similar, but that would break current designs, so maybe we can do that for v6? Other option would be calling the current one `Conn_ARM_JTAG_SWD_TRACE_20`, but it would be nice to have the current connectors with the same name, just differing on the number on pins (10 vs 20).

Thanks!

diegoherranz:
Do I fix the footprint and footprint filter on the `Conn_ARM_JTAG_SWD_20` symbol?
Regarding renaming `Conn_ARM_JTAG_SWD_20` to `Conn_ARM_JTAG_SWD_20_LEGACY`, maybe we can do as soon as 5.1 is released?

Thanks!

myfreescalewebpage:
I think you can fix the fp anf fp filter of the 20 pins symbol right now, I'm not sure about the second proposition. I don't see any good reason to have two versions of the 20pins version... Anyway, to be discussed for v6.

diegoherranz:
> I think you can fix the fp anf fp filter of the 20 pins symbol right now, I'm not sure about the second proposition.

Done. There was no default filter (OK). Footprint filter fixed: the PinHeader entry keeps the `2x10` to avoid single row headers. That's not required for the IDC, since they are always dual row.

> I'm not sure about the second proposition. I don't see any good reason to have two versions of the 20pins version... Anyway, to be discussed for v6.

There are currently at least 3 different connectors for ARM debugging:

1. 10-pin cortex, 1.27mm pitch, `Conn_ARM_JTAG_SWD_10` in the library.
2. 20-pin cortex, 1.27mm pitch. Missing in the library. 
3. 20-pin legacy, 2.54 pitch, `Conn_ARM_JTAG_SWD_20` in the library. The signals and pinout for this are completely different to the other 20-pin connector. So it makes sense to keep separate symbols.

My proposal was using `Conn_ARM_JTAG_SWD_20` for (2) and `Conn_ARM_JTAG_SWD_20_LEGACY` for (3), but let's leave that discussion for v6 when we can make breaking changes.

Thanks!

myfreescalewebpage:
Yes, can you put the previous comment in a new issue so we keep the trace of it even when merging this PR ?

What do you call the legacy version ? I have never seen it before, for me all SWD connectors are 1.27mm pitch.

myfreescalewebpage:
Current status is great for me for integration in release 5.1, more general rework to be schedule in v6 as discussed above. Merging.

diegoherranz:
> Yes, can you put the previous comment in a new issue so we keep the trace of it even when merging this PR ?

Done - https://github.com/KiCad/kicad-symbols/issues/1604

> What do you call the legacy version ? I have never seen it before, for me all SWD connectors are 1.27mm pitch.

It used to be 2.54mm until the 1.27mm debug connectors appeared. I think it is referred as "legacy" in many places, so it feels like the right term to me. Anyway, let's discuss that in due time on that issue I've just opened.

Thanks!

PR #1567, opened 2019-02-23
fix LTC6957-1: LVPECL outputs are open emitters

lorem-ipsum:
the outputs of the LVPECL outputs are open emmiters, but those of the LVDS variant are not. Remove the alias in favour of two components with the correct output type. Otherwise the symbol is not changed at all.

ltc6957xms-1


myfreescalewebpage:
Hi @lorem-ipsum thanks for this fix, no comment to do here, merging !

PR #1569, opened 2019-02-23
ISO1211 + ISO1212

myfreescalewebpage:
Create Digital Input Receiver Texas ISO1211 and ISO1212

---

http://www.ti.com/lit/ds/symlink/iso1211.pdf

iso1211

iso1212


HubertD:
checked pin names, numbers and functions. looks good to me.

evanshultz:
1. Please add the channel count to the description to better differentiate these symbols.
1. Output pins should to Tri-state type.
1. Let's make the SUB pins visible. They're not to be used in production, but they have a functional connection and could be used in testing or some other use. If they're hidden on the schematic, this symbol couldn't be used for developing a test fixture (for example) for these symbols.
1. What about putting EN at the bottom so you can align the IN and OUT pins across the symbol?
1. Flip the whole thing around so input is on the left and output on the right.

myfreescalewebpage:
@evanshultz I realize my design it ugly now ! Will do a new one, agree with your comments :)

myfreescalewebpage:
@evanshultz done! You can have a new review now :)

evanshultz:
I feel like I won the lottery!

Now they look like:
image
image

Thanks!

myfreescalewebpage:
loool thanks! Sorry to not have updated the screenshots myself :)

PR #1571, opened 2019-02-24
Fix footprint of TCA9535DBR

myfreescalewebpage:
Fix footprint and footprint filter of the TCA9535DBR. No other modifications.

---

http://www.ti.com/lit/ds/symlink/tca9535.pdf

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1432


HubertD:
Sorry @myfreescalewebpage, 
maybe I should get more sleep, but I don't think this is correct.

The SSOP-16 footprint was clearly wrong.
But I think you're mixing up two package variants:
- TCA9535DB[R|T] - SSOP-24
- TCA9535PWR - TSSOP-24

Your footprint PR tries to create SSOP-24 package (DB variant) with dimensions from the TSSOP-24 (PW) drawing?
As far as I can tell, the existing TSSOP-24 and SSOP-24 footprints would match perfectly well? 

Cheers,
Hubert

evanshultz:
There is a problem, but I believe it different than the above.

Page 36 of the datasheet:
image

Do really it should be `DBx` or just `DB`, since the `R` is about packaging, but that ship has sailed.

Page 38 shows this:
image

So the DB package dimensions are 5.3x8.2mm.

The PW dimensions are 4.4x7.8mm.

We have both footprints in the library, but should the fix be:
1. Change the symbol to `PWR` suffix?
1. Change the symbol to use a footprint with a 5.3x8.7mm body size?

When the symbol was added at https://github.com/KiCad/kicad-symbols/pull/109, it had the wrong name. The symbol was added to the old library at https://github.com/KiCad/kicad-library/pull/1563, and because many versions were added for all packages I believe it's a symbol typo and should have been `8.2` instead of `6.2`. Which means the latter choice is correct. That also avoids breaking any symbols.

myfreescalewebpage:
@evanshultz @HubertD the symbol is TCA9535DBR (I don't expect to change the name of course), this is SSOP-24 version, only the footprint and fp filter were incorrect.

@evanshultz your screenshot is showing SSOP-28 version.

According to me the modification of the symbol and the proposed footprint are correct. Dimensions are at the page 37 of the datasheet. Screenshot below.

capture

Let me known if this is correct or not but I have double checked so I expect it is :)

myfreescalewebpage:
@HubertD 

> As far as I can tell, the existing TSSOP-24 and SSOP-24 footprints would match perfectly well?

Yes that's right. But SSOP is SSOP and TSSOP is TSSOP. Height of the package is not the same and to get correct 3D rendering, we always use the right footprint, even if another match too. This also prevent confusion in the future.

Cheers,
Joel

evanshultz:
@myfreescalewebpage 
No. Check the package drawings.

PW and DB are different packages. I think that's the point that's missing. You used body dimensions for the PW package, but you're updating the DB symbol.

And the drawing I posted above if for various lead counts. I circled the correct 24 pin info.

HubertD:
I still don't get it.

Datasheet (Page 33) shows following variants:
- TCA9535DB - SSOP-24 - 5.3x8.2mm (Drawing "DB", Page 38)
- TCA9535PW - TSSOP-24 - 4.4x7.8mm (Drawing "PW", Page 37)

Both packages are in the footprint library, with the same names.
Still lacking sleep and this is making me dizzy, but I tripple checked everything and for me it seems as the very perfect footprints are already there in the library.

Now you are creating a SSOP-24 footprint with the TSSOP-24 dimensions.
As far as I can tell from the datasheet, there is no SSOP-24_4.4x7.8mm device mentioned.

The Drawing on Page 37 from your screenshot shows the TSSOP-24 variant.
Maybe that's the fault here, and you think there is a SSOP-28 and a TSSOP-24 variant?
That's not the case. 
The SSOP drawing on Page 38 shows 28 pins, but it's just a generic figure and in this case defines the SSOP-24 device.

Anyway, I will take my nap now. Maybe everything is clearer tomorrow :-)

myfreescalewebpage:
My bad, I was obviously too much tired ! My confusion comes from that the ordering information table starts with DB package, but the first described is the PW version. You are right, sorry, I apologize for this. Will make a new commit with the correct footprint now.

myfreescalewebpage:
Also fixed a travis error (pin offset - no impact on the symbol itself)

evanshultz:
Great. Sorry that I missed your updated commit.

myfreescalewebpage:
No worries :) Thanks for the review!

PR #1578, opened 2019-02-25
Switch dual fix

lorem-ipsum:
Fix the pins of the existing `SW_Push_Dual_x2`. It had the pins set to be shared between both components, resulting in both parts having numbers 3 and 4. The first commit does nothing, but saves it with Kicad 5.0.1 to get a clean commit afterwards. It could be dropped theoretically.


myfreescalewebpage:
@lorem-ipsum good catch ! No comment, merging, thanks for this fix.

PR #1579, opened 2019-02-25
Add IRM-03, -05, and -10 symbols

evanshultz:
Finishes the items at https://github.com/KiCad/kicad-footprints/issues/1320.

Here are the missing symbols for the existing footprints in our library. Hopefully by having the symbols already, if I haven't messed anything up, users won't need to create their own symbols and can avoid any trouble. Hopefully.

Pin numbers are not given in the datasheets so I selected them to match the existing footprints.

https://www.meanwell.com/Upload/PDF/IRM-03/IRM-03-SPEC.PDF
https://www.meanwell.com/Upload/PDF/IRM-05/IRM-05-SPEC.PDF
https://www.meanwell.com/Upload/PDF/IRM-10/IRM-10-SPEC.PDF

image

I also corrected other IRM datasheets to the official ones. I think the URL scheme is obvious.

---


evanshultz:
There are two optional things I can do also:
1. Fix pin names to match the datasheet. For example, "-V" instead of "-Vout".
1. Force the footprint filter to only match SMD or THT footprints since the symbols are specific to the package mounting type.

Let me know if you'd like to do that here.

myfreescalewebpage:
@evanshultz I agree with the two propositions above to get more accurate devices

antoniovazquezblanco:
Sorry to stick my nose here.

In order to be coherent with the IRM-02 series I would go the following way:
- For non-SMD versions I would stick to the 1 to 4 numbering.
- For SMD I would go for the numbering shown in the table of each device.
- I would try to force the THT/SMD in fp filtering that can easily cause trouble if users do not pay full attention.
- Maybe those could be aliases? I am away and cannot check at the moment but there may be good candidates. Can check later.

Thank you very very much!

evanshultz:
@antoniovazquezblanco 
Not at all. I'm glad you popped in.

I totally agree with you about THT pin numbering. However, the footprints already existed and so users may have created their own symbols. I kept the footprint pin numbering for these THT footprints, even if it seemed weird, so that user-created symbols wouldn't break and our existing footprints didn't need to be touched. Thinking that this may get into 5.1, I felt continuity was most important.

Similarly, IRM-05 and IRM-10 could be ALIASes. But since there are unique footprints I kept separate symbols. The option now that I see would be to make a combo footprint and then make one symbol with many ALIASes use that combo footprint. But keeping them split, since unique IRM-05 and IRM-10 footprints already existed, didn't seem too bad and maintains existing bits in our library.

No SMD symbols are in this PR and the two that already exist match the datasheet.

Just updated FP filters to be more specific.

evanshultz:
Regarding pin names the current "+/-Vout" doesn't show up anywhere in the datasheets. But take a look at https://www.meanwell.com/Upload/PDF/IRM-03/IRM-03-SPEC.PDF and you'll see:
image

Should they be "+/-V" or "+/-Vo"? It shouldn't be a problem to change just the pin names, but what do we call them? I personally like just "+/-V", and for the datasheets that only contain THT packages (IRM-05 and up) this is the terminology used. Sound OK?

antoniovazquezblanco:
Whatever you prefer for naming. I am okey either way. I think it is explicit enough to know the function.

Thanks!

evanshultz:
OK. Renamed.

Once this is merged I'll create an issue to capture what things we could do for these symbols and footprints after 5.1 when we can make more invasive changes.

antoniovazquezblanco:
Thanks!

PR #1580, opened 2019-02-26
Add/correct FP filters for jumper symbols

evanshultz:
In Jumper.lib some of the symbols didn't have a properly formatted FP filter so they didn't pick up any footprints in CvPcb. And the jumper symbols in Device.lib didn't have any FP filter. So... here is the fix.

I also found one FP filter that was missing the leading "Solder".

Checked in KiCad and jumpers appear in CvPcb now. Yay!

---


poeschlr:
thanks.

PR #1124, opened 2018-11-07
Add LMP8640

cejpmart:
DS: http://www.ti.com/lit/ds/symlink/lmp8640.pdf

The symbol is based on INA138, although IMO it may mislead somebody into thinking it is a standard op amp.

Imgur

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1588, opened 2019-02-28
adding Texas Instruments INA253 current sensor, advice needed

rs-development:
I would like to add the INA253 current sensor to the library.

[Datasheet](http://www.ti.com/lit/ds/symlink/ina253.pdf)

image

But since the IC is kinda different i would like to ask how to approach the symbol design. Are there any similar devices in the library for orientation?

Should the IS, IN and SH Pins go to the left? 
Or should i use the classic amplifier symbol and try modifying it?

image

Thanks for your tips :+1:


myfreescalewebpage:
Hi @rs-development 
I think something similar to the op amp is great, I have recently merged this curretn sensor too: https://github.com/KiCad/kicad-symbols/pull/1124
Joel

myfreescalewebpage:
@rs-development I see you are pushing some commits, please indicate when the review can be done on this symbol, thanks !

rs-development:
Pin 15 is NC but should be connected to ground. That's why i made it passive instead of NC.

@myfreescalewebpage have a look

myfreescalewebpage:
OK, doing the review from this point.

myfreescalewebpage:
Several comment I have on your contribution:

- [ ] A comma is to be added before the package name in the description, and the package name should be fixed: `High Voltage, Bidirectional, Zero-Drift, Current-ShuntMonitor With Integrated, 2-mΩ, Precision, Low Inductive Shunt Resistor, TSSOP-20`
- [ ] NC pins should not be stacked
- [ ] Pin 15 should be renamed GND and stacked with pin 6
- [ ] REF1 and REF2 should be Passive to avoid any ERC errors

Else, design is ok.

Joel

myfreescalewebpage:
Thanks @rs-development no more comment, merging.

daviid5:
Hey!
I have at the moment the following problem using the symbol:
grafik

The Symbol does not really make sense, without knowing the internal structure of the INA253 you cannot understand that this component is "only" a shunt with an integrated amplifier ...

I would suggest to change the symbol in the following way:
- IS- on the left side of the symbol
- IS+ mirrored to IS- on the right side
I just created the following mockup:
grafik

And in addition it would be really nice to show the internal structure in the symbol
See the datasheet:
grafik
This would make it much eassier to understand the functionality of the component when you don't have the datasheet in mind ...


What do you think?

Background information:
grafik

I have this circuit, when you look at it, the functionality is really easy to understand.
And you can understand directly the "flow" from left to right through the H-Bridge.
Except of the part with the INA253, which looks really weird and confusing.

evanshultz:
Another option would be a multi-unit symbol. This would have the advantage of cleanly separating the (also-separate) functional units, while making it clear that there is an opamp symbol inside. This separation is shown in section 8.2 of the datasheet on page 12.

Something like this:
image
```
# INA253
#
DEF INA253 U 0 5 Y Y 2 L N
F0 "U" 250 200 50 H V C CNN
F1 "INA253" 350 -150 50 H V C CNN
F2 "Package_SO:TSSOP-20_4.4x6.5mm_P0.65mm" 0 -900 50 H I C CNN
F3 "" 0 -1000 50 H I C CNN
$FPLIST
 TSSOP*4.4x6.5mm*P0.65mm*
$ENDFPLIST
DRAW
P 4 1 1 10 -300 300 300 0 -300 -300 -300 300 f
P 2 2 1 0 0 -90 0 -100 N
P 2 2 1 0 0 90 0 100 N
P 5 2 1 0 0 -30 40 -45 0 -60 -40 -75 0 -90 N
P 5 2 1 0 0 30 40 15 0 0 -40 -15 0 -30 N
P 5 2 1 0 0 90 40 75 0 60 -40 45 0 30 N
X REF2 10 100 -300 200 U 50 50 1 1 P
X REF1 11 0 -300 150 U 50 50 1 1 P
X NC 12 0 -150 100 U 50 50 1 1 N N
X ~ 13 400 0 100 L 50 50 1 1 O
X DNC2 14 -100 150 100 D 50 50 1 1 N N
X GND 15 -100 -300 100 U 50 50 1 1 P N
X IN+ 16 -400 100 100 R 50 50 1 1 P
X IN- 5 -400 -100 100 R 50 50 1 1 P
X GND 6 -100 -300 100 U 50 50 1 1 W
X DNC1 7 0 150 100 D 50 50 1 1 N N
X NC 8 -100 -150 100 U 50 50 1 1 N N
X VS 9 -100 300 100 D 50 50 1 1 W
X ~ 1 0 200 100 D 50 50 2 1 P
X ~ 17 100 100 100 L 50 50 2 1 P
X ~ 18 0 -200 100 U 50 50 2 1 P N
X ~ 19 0 -200 100 U 50 50 2 1 P N
X ~ 2 0 200 100 D 50 50 2 1 P N
X ~ 20 0 -200 100 U 50 50 2 1 P
X ~ 3 0 200 100 D 50 50 2 1 P N
X ~ 4 100 -100 100 L 50 50 2 1 P
ENDDRAW
ENDDEF
```

daviid5:
I like your idea, but I don't see any application where it would make sense to use the shunt and the anmplifier seperated from eacht other.
Beside this, the datasheet says in 8.3.1 "this resistor is not intended to be used as a stand-alone component"

So my suggestion would be something like this:
grafik
It's a really qick&dirty mockup :D

myfreescalewebpage:
@rs-development for me IS+ should go next to SH+/IN+ and IS- should go next to SH-/IN-. The design of @evanshultz is not bad. @daviid5 your design is a bit away of what we are used to see in the library.
@rs-development if you want to submit a fix pull request, it's ok, but I think the dire ction of @evanshultz is better.
Joel

evanshultz:
@daviid5 
In section 6 for pins 4 and 17 and then again in section 9.1.1 an external filter is discussed. This is why I made them separate units; easy to connect together as you've shown but straightforward to add the filter.

daviid5:
@evanshultz 
Well, that's correct.
But I think in that use case (using only the amplifier, or only the shunt stand alone) is such uncommon that we should not really care about that case.

I think it's much more confusing to have a splitted package. Imagine to you place the Shunt-Symbol a bit away from the amplifier.
Noboby will get the fact that the shunt and the amplifier are together in one package ... (it's such a special part ...), the only you to find that out is to look at the Designator.
For me it would make much more sense to bring it into one symbol, it leaves much less room for misinterpretations ;)

PR #1589, opened 2019-03-01
Fix pin numbering for Omron G6K relays.

marcsaegesser:
The pin labled 12 should really be 8 for these parts.

For reference, the G6K datasheet is available at
https://omronfs.omron.com/en_US/ecb/products/pdf/en-g6k.pdf

---


evanshultz:
@marcsaegesser 
Thanks for reporting this and submitting a PR! I apologize for the library having a problem but at least with your help and initiative it can be fixed.

First, here are the existing symbols:
image

You are correct that pin 12 should be 8. Oops.

I checked all footprints and they use pin numbers 1-8, so this won't break anything.

Thanks again!

PR #1357, opened 2018-12-31
Fix pin 9, 10, 11, 12, 13, 14 of 74LS157 and 74157

pushpull13:
Fix pin 9, 10, 11, 12, 13, 14 of 74LS157 in 74xx.lib and 74157 in 74xx_IEEE.lib

- http://www.ti.com/lit/ds/symlink/sn74ls157.pdf

PR #1592, opened 2019-03-03
fixed pins of 74157

pushpull13:
Fix pin 9, 10, 11, 12, 13, 14 of 74LS157 in 74xx.lib

http://www.ti.com/lit/ds/symlink/sn74ls157.pdf


pushpull13:
Minor bug in pins of 74157

poeschlr:
screenshot of old vs new would help. Is the datasheet you use the one that was originally used? (Maybe TI uses slightly different pinout to some other manufacturers.)

diegoherranz:
Thanks for your contribution, @pushpull13.

It looks like the pin names currently in the library match this Motorola part: https://www.uni-kl.de/elektronik-lager/417724

Actually, because both muxes (C and D) are interchangeable, it doesn't make much difference.

If we want to follow the Texas numbering/arranging, we should change the pin names too to match their datasheet. E.g. Zd -> 4Y. Would you change that?

A separate thing is that this symbol could use multiple units, but that would break the backward compatibilty so we can't do it until we start working on v6 stuff.

Thanks!

myfreescalewebpage:
@pushpull13 seems it is the same PR than https://github.com/KiCad/kicad-symbols/pull/1592, do you confirm ? You should not create a new PR each time a branch conflict occurs, instead, simply update your local branch and solve conflicts.

pushpull13:
Actually, this is first time I am contributing. So, a bit less idea what to
do

myfreescalewebpage:
Well, you should ask or google the info. This PR is breaking existing schematics design and cannot be merged right now. It will be possible only when v6 will be released, and new branch conflict will occur again before it appends! Even for you it's too much job to create a new PR each time a conflict appears :)

myfreescalewebpage:
I have just read again comment of @evanshultz at https://github.com/KiCad/kicad-symbols/pull/1357#issuecomment-468826194, probably it's good to merge right now.

Thus, accepting this PR, rejecting the old one, but please next time @pushpull13, follow the normal process, thanks!

pushpull13:
Thank you

PR #1593, opened 2019-03-03
add generic SW_Push_DPDT

lorem-ipsum:
This adds a generic symbol for DPDT push bottons, such as the one in [this footprint PR](https://github.com/KiCad/kicad-footprints/pull/1454). The symbol was created from the existing `SW_Push_SPDT` symbol. I just continued the numbering scheme.

sw_push_dpdt


poeschlr:
We have a new naming convention for switch symbols and footprints (only to be used for new additions)
See: https://github.com/KiCad/kicad-symbols/issues/580#issuecomment-450632899

lorem-ipsum:
Ah, cool. Thanks for pointing it out.

Looking at the new rules the name of this one should already be ok. If I understand it correctly it should get at footprint filter for `SW_Push_DPDT*` (or `SW*Push*DPDT*`?).

myfreescalewebpage:
@poeschlr if you have the answer for @lorem-ipsum ? Joel

myfreescalewebpage:
@lorem-ipsum just made a review of this pull request. I think the issue pointed by Rene is still in discussion and because your symbol is consistent with all the other SW_Push devices now, I merge it right now.

antoniovazquezblanco:
For new additions we should stick to the latest naming convention. See https://github.com/KiCad/kicad-symbols/issues/580#issuecomment-450632899

Thanks!

myfreescalewebpage:
@antoniovazquezblanco I have understood from this thread that the status is not clear as indicated above. I have asked. No response. The pull request is consistent with all the other SW, probably this is something to plan for v6.

poeschlr:
Why do you think i pointed to that issue if i did not intent it to be followed? The last post in it also clearly states that its suggestion should be used for all new additions. The reason why only new additions is because we do not want to disrupt existing installations.

myfreescalewebpage:
Sorry @poeschlr but the author was waiting for an answer how to rename it's part. I have personally not understood how to name it too. I have judge that the proposition of the author was not bad based on existing devices. Submitter and reviewer clearly need a detailed description of the rules in KLC with some examples, else, this is very complicated to understand for such kind of topics.
I suggest that the symbol should be reworked all at the same time for v6. Meanwhile, every v5.x will be consistent so it's not too bad.

evanshultz:
The old paradigm will stick around in this case to avoid breaking existing schematics.

The new paradigm is used on new PRs since it works better and captures what we have learned from using the old style (and the problems it has).

You're right that documenting this in KLC so new PRs can follow, and for general clarity, sounds like a good approach. Then we can say KLC is correct and there are legacy symbols to avoid breaking schematics. An issue, targeted at the future release, can capture the intent to make updates to existing symbols when the time is right.

PR #1600, opened 2019-03-03
Create LT3748

myfreescalewebpage:
Creation of LT3748, MSOP-16 package

---

https://www.analog.com/media/en/technical-documentation/data-sheets/3748fb.pdf

capture

https://github.com/KiCad/kicad-footprints/pull/1457

No error except pending footprint


evanshultz:
1. Let's put RREF on the left side with other Passive pins.
1. In just this lib we have INTVCC pins Power Output, Passive, Power Input, and Output types. It is a complicated pin to classify. I guess this is fine.

myfreescalewebpage:
Done :)

evanshultz:
So here is what we have now, next to a standard FET symbol:
image

What would you think about moving SENSE down relative to GATE, so the wire can connect straight across, and then moving both pins towards the bottom to give more space around RFB?

myfreescalewebpage:
Not a bad idea :)
Done !

PR #1601, opened 2019-03-04
Add NCN4555MNR2G

jneiva08:
Added NCN4555MNR2G

Datasheet: https://www.onsemi.com/pub/Collateral/NCN4555-D.PDF
Footprint: https://github.com/KiCad/kicad-footprints/pull/1458

Screenshot:
ncn455

---


myfreescalewebpage:
Hi @jneiva08 , thanks for contributing,

- [ ] I disagree with the symbol name, the datasheet indicate the QFN version is "MN suffix" so the symbol should be "NCN4555MN"
- [ ] No comma in the keywords
- [ ] Pin offset should be 20mil (KLC error S3.6, see travis result)
- [ ] Expose pad is missing (pin 17) and should be stacked with GND pin 10
- [ ] You can reduce a little bit the size of the symbol as shown below if you want, let me known:

capture

Else sounds good, just need to wait the footprint.

Cheers,
Joel

jneiva08:
@myfreescalewebpage , have made alterations you recommended. the symbol don't look exactly to yours
ncn4555

myfreescalewebpage:
Your proposition is great too :) Just realize we can switch pins 13 and 14 or 9 and 11 so that the CLK is in front of SIM_CLK and RST in front of SIM_RST. Can you address it ? Thanks.

myfreescalewebpage:
@jneiva08 thanks for the modifications, no more comment here, just need to wait the integration of the footprint to merge this.

jneiva08:
@myfreescalewebpage yesterday have received some NCN4555MNR2G from Farnell.

Have tried to verify if GND (pin 10) have continuity to EP, but it doesn't have.
Probably is better to remove EP from GND.

In MPU-9250 the EP isn't connected to GND.

myfreescalewebpage:
@jneiva08 that's not because you have continuity that the pad should not be connected to ground. I have just read a more complete description in the datasheet page 9:

> The exposed pad of the QFN−16 package will be connected to the ground as well as the unconnected pins (NC). A relatively large ground plane is recommended.

So, I think we should (you should also in your design) stack the EP and all the NC pins too (we have keep NC pins not connected at this moment but I'm no more sure we have to).

Let me known,
Joel

myfreescalewebpage:
PS : MPU-9250 do not respect KLC at all (pins are too long, graphic style of pin is not correct, too much lost space, Power Input on the right etc.)

myfreescalewebpage:
@jneiva08 merging following integration of the footprint.

jneiva08:
Thanks

PR #1608, opened 2019-03-05
WE 750343373

myfreescalewebpage:
WE Flyback Transformer 750343373
Created from another transformer, simply renumber pins and add 4 not connected pins

---

https://katalog.we-online.com/ctm/datasheet/750343373.pdf

sym

https://github.com/KiCad/kicad-footprints/pull/1465

Component outline thickness from the original transformer (S3.3) and pending footprint (S5.1)


evanshultz:
1. We really need to sort out a description scheme for transformers...
1. I don't really think NC pins are needed here since the footprint doesn't have the pin count in it. There is only one possible footprint and it will always be matched. But for consistency, maybe they should stay?
1. While I'm not a big stickler on this and I think we go with what works best for users, IEEE and ANSI show `T` for a transformer ref des. Our library has a majority of `TR`, a number of `T`, and a single `U`. This ought to be straightened out.

myfreescalewebpage:
@evanshultz thanks for the review

1. I agree, I propose this should be part of https://github.com/KiCad/kicad-symbols/issues/1617
2. I agree too. We ask them for several reasons, don't know really if this can cause a problem here if they are not present... Anyway the presence of all NC pins is not bad according to mend is a best practice to have. I vote to keep them.
3. So `T` is probably the best to be kept for us. I can change here and maybe I can open a new pull request for that if you have time to review it after ?

Joel

evanshultz:
3. Since TR is obvious, and they are the majority, I'm fine sticking with it. If you're suggesting to stick with TR and update all others to that, I'm fine doing the U one now and waiting on the T ones until closer to v6. @poeschlr ?

myfreescalewebpage:
@evanshultz 

> until closer to v6

Hummm I think we will have many work for the v6 so starting right now is not bad and I'm ready to do it for the transformers. That's also why I decided to participate solving some pending issues we have like the Valve library in an other PR.

However, you indicated above T was the best option according to IEEE and ANSI, why keeping TR ?

Joel

evanshultz:
I don't want to start breaking people's designs that use this library now when the v5 branch will be the release one for likely a couple years. More severe changes can be accepted closer to a new major release of KiCad, with stability in the interim where possible. That's why my feeling is to defer work to later (but as you've said it should be captured and then planned for) which will affect users.

Fair point about TR vs T. Since most of them are TR I defaulted to that. I don't have evidence one way or the other, but with ref des especially we have gone with what's common rather than strictly adhere to any spec. Do any other part types use T? Do we prefer a single letter versus two? I don't know but it might help answer this. Personally I prefer T to have a smaller ref des but I'm only one users' voice.

evanshultz:
Shall I merge this and move the discussion to an issue?

myfreescalewebpage:
> I don't want to start breaking people's designs that use this library now when the v5 branch will be the release one for likely a couple years. More severe changes can be accepted closer to a new major release of KiCad, with stability in the interim where possible. That's why my feeling is to defer work to later (but as you've said it should be captured and then planned for) which will affect users.

User that don't want to have breaks in their design should use the library delivered in the KiCad install package, and should not follow the every day change of the github repos :) A roadmap can be made but I m not sure it is easy to draw it now...

> Fair point about TR vs T. Since most of them are TR I defaulted to that. I don't have evidence one way or the other, but with ref des especially we have gone with what's common rather than strictly adhere to any spec. Do any other part types use T? Do we prefer a single letter versus two? I don't know but it might help answer this. Personally I prefer T to have a smaller ref des but I'm only one users' voice.

Agree with you and I have not personal preference too. I was just pointing the content of your previous message. Maybe we can speak of that in a dedicated issue we can opened, with v6 milestone, that probably the best option to get librarians an eye on that.

evanshultz:
This method of only intentionally breaking things, bugs excluded, near major releases is Rene's preference. He should be central to any discussion that is in regards to changing this method.

poeschlr:
The devs gave us two options. Avoid braking existing designs or accept that the users do not get library updates shiped with minor releases.
(Or an implicit option: make a separate branch that only gets bugfixes that is then shiped with kicad for minor releases. I however doubt we have enough mapower for this.)


So my verdict is as follows: No changes that break designs just to fix a KLC error. If there is a true bug in a symbol then go ahead and fix it. If fixing that bug already breaks designs then you are free to also fix KLC errors. Otherwise see rule number one. (It is better to have a few KLC errors in the lib than to have broken designs by users.)

poeschlr:
We can do it similarly than what we did with buttons and switches. Think about better naming schemes and use it for new contributions. (Yes this will result in a mixed state of such libs. Yes i am aware this is ugly. No i do not think the benefits of reorganizing this right away outway the downsides for users.)

evanshultz:
@myfreescalewebpage 
Can you please check the Travis issues. Isn't this the Python 2/3 thing and it can be ignored?

myfreescalewebpage:
That a scripts issue I have logged here https://github.com/KiCad/kicad-library-utils/issues/292, should be solved by closing/opening according to Rene.

evanshultz:
I already did that and it cleared the missing footprint issue but didn't make Travis green. Hmmm...

myfreescalewebpage:
I'm currently fixing the footprint issue, I have to rename the manufacturer here too.

myfreescalewebpage:
Done.

evanshultz:
Thanks!

(These errors are bogus and caused by having the galvanic isolation be a filled box.)

myfreescalewebpage:
Yes I know :) Thanks for the review !!

PR #1610, opened 2019-03-05
Add Texas TPS568215

myfreescalewebpage:
Add Texas TPS568215

---

http://www.ti.com/lit/ds/symlink/tps568215.pdf

symbol

https://github.com/KiCad/kicad-footprints/pull/1466


evanshultz:
1. I'm thinking to put the noun at the end of the description, align with the datasheet title, make it clear the switching frequency isn't continuously adjustable, and simplify the footprint a bit. How about `4.5V-17V Input, 8A Synchronous Step-Down SWIFT Converter, Adjustable Output, 400kHz/800kHz/1.2MHz Switching Frequency, Texas VQFN-18`?
1. Can VIN be on the top?
1. Put PGOOD on the right side and 17-18 on the left side.
1. AGND should be a discrete pin for layout. Several parts need to be grounded specifically to AGND, and without separate nets it's harder to ensure just a single point on the board bridges PGND and AGND. Keeping them separate here and using a net tie will allow this.

myfreescalewebpage:
@evanshultz thanks for the review.

1. I agree, will do.
2. It is a step down regulator....
3. Agree too.
4. Yes that's right, making the modification

Joel

evanshultz:
2.Can you clarify what you mean? Our "standard" linear regulator symbol, with just 3 pins, is small and tight with VIN on the left side. But here I'm not understanding why VIN can't be on top. Let me know what you're thinking?

myfreescalewebpage:
@evanshultz you are thinking to the "Regulator_Linear" library. This one is in "Regulator_Switching" and is like many other. Just have a look to all symbols in this one.

evanshultz:
Yes, I know. I was giving an example of a library where we specifically decided to put a power pin on the side in violation of KLC.

Here, why can't VIN be on the top to conform with KLC? No downside that I can see. Other legacy symbols in this lib don't but we can't change them so I say to leave them alone but recognize they're not a good reference for the current library guidelines.

myfreescalewebpage:
Well, I have accepted many pull request in regulator switching library with input voltage on the left. Compared to the regulator linear library, I don't see any functional differences, this is a regulator with current flowing from the input to the output as shown on the above schema.

Capture

VIN on the top will look like a little bit strange I think.

However I agree this not apply to regulator that drive a FET with a transformer like the LT3748 I have proposed in https://github.com/KiCad/kicad-symbols/pull/1600.

Joel

myfreescalewebpage:
Just seen pull request of @SchrodingersGat at https://github.com/KiCad/kicad-symbols/pull/1710 that have VIN to the top. Not bad anyway.

However, this is not the traditional way to draw this kind of regulator, even the datasheet has:

Capture

evanshultz:
The library is designed to KLC, and http://kicad-pcb.org/libraries/klc/S4.2/ says power pins should be on top. There are a no exceptions listed. Yet when Jan reworked the linear reg lib it was decided to make simple 3-terminal linear regs have VIN on the left. We both know KLC doesn't cover everything, by design, but sadly I haven't been proactive in keeping it updated where new info would clearly be helpful. And while each vendor may have their own style for making symbols that doesn't mean it matches KLC. As you said there are many switching regulator symbols with VIN on the side today, but we can stick closer to KLC starting now to reduce the number of abnormalities. Or we can decide to make more exceptions or abolish this clause of KLC. My point is that I can't see a downside to doing what KLC says in this example.

I will merge this and I raised the VIN pin position on switching regulator symbols at https://github.com/KiCad/kicad-symbols/issues/1711. There are many other undecided topics floating around so it's just another item on the list.

myfreescalewebpage:
I agree with you, we have to decide globally if this is an exception for all regulators, or not and in this case a work will need to be done across the library before v6 :)

SchrodingersGat:
I think that logically for regulators it makes sense to have the VIN pin(s) on the top left of the symbol. This is currently what we do for all the linear regs.

myfreescalewebpage:
> I think that logically for regulators it makes sense to have the VIN pin(s) on the top left of the symbol. This is currently what we do for all the linear regs.

@SchrodingersGat take care, this is not top, this is side. Precisely on the top of the left side :)

SchrodingersGat:
Sorry you're right, I didn't express that very well.

```... on the top (of the left side) of the symbol ```

myfreescalewebpage:
@SchrodingersGat I know you know :)

evanshultz:
Footprint is merged and CI is green. So... merge this?

myfreescalewebpage:
@evanshultz I cannot do it I'm the author :)

evanshultz:
Right. I'm making sure you're ready for me to merge it. And it sounds like you are.

myfreescalewebpage:
I am :)

PR #1237, opened 2018-12-08
Transistor_FET: Add Diodes Inc. N-Channel SOT-23 MOSFETs

diegoherranz:
Semi-scripted

By filtering https://www.diodes.com/products/discrete/mosfets/mosfet-master-table/export/mosfet-master-table.xlsx and processing it with a quick and dirty script. All of them added as aliases.

Selected:
 - N-channel
 - SOT23 (I manually checked that they're all same pinout)
 - Without ESD gate protection so that they can be aliases of BSS138
 - Automotive qualified variants (with a Q suffix) excluded (even if there's a separate datasheet). I guess this is somewhat similar to temperature variants which we don't alias?
- Excluded parts already in the library (obviously)

After filtering the spreadsheet (`diodes_nfets_sot23.ods`), datasheet column replaced by a plain text URL (was a hyperlink) and saved as CSV.

The CSV gets processed by `fet_aliases_gen.py` which prints the CMPs to be added to the `.dcm` file and the list of part numbers to add to the `.lib`.

I pasted the aliases on the `.lib` and reordered manually (alphabetically).

I pasted the CMPs at the end of the .dcm and opened and saved from within KiCad so that they get ordered properly.

Note: Some existant components were not in the right order and have been reordered too.

Datasheets and script used: [fets_spreadsheets_script.zip](https://github.com/KiCad/kicad-symbols/files/2659589/fets_spreadsheets_script.zip)

Thanks!

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful (Not applicable)
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required

PR #1615, opened 2019-03-08
Added 2N7002K

0xdec:
Added 2N7002K as an alias of BSS138
Datasheet: https://www.diodes.com/assets/Datasheets/ds30896.pdf

---


evanshultz:
@diegoherranz 
Are you going to take this? While `2N7002K` (and other parts not sole-sourced) has varying parameters depending on which datasheet you read, this is probably fine so this PR looks good to me.

myfreescalewebpage:
I would just say it's better with no extra space before values and units. I would ask to remove the spaces before "A" and "V" in the description. @0xdec

evanshultz:
No space is my preference too, but because of the inconsistency everywhere with descriptions that go back before I started helping out here, and no guidelines in KLC, I just merge if I have no other comments during a review. But that's me. Maybe I'm wrong?

myfreescalewebpage:
I'm not against merging @evanshultz, I know this is something to review and fix in all lib.

diegoherranz:
Yes, I was planning on taking this.

The main comment I wanted to make was that this device does include gate protection diodes.
When I worked on https://github.com/KiCad/kicad-symbols/pull/1237, I didn't include those with gate protection diodes since I thought that they would need a different symbol. Assuming that is the case, this would need a new symbol too. Although, obviously, it "works" with the symbol as is.

Any opinions on this?

Thanks!

evanshultz:
IMO it's not worth it. This symbol still captures the device and the diodes don't add much. A note in the description wouldn't hurt, but I don't even think is necessary to merge.

Reviewing and polishing this library would be nice in the future, but I personally am in favor of merging as-is.

0xdec:
@myfreescalewebpage I also prefer no spaces in front of "A" and "V", but I just copied and modified the description from the existing 2N7002 devices. I'm happy to make any changes and rebase/force push if you think it's necessary.

0xdec:
@myfreescalewebpage fixed the extraneous spaces issue, and also went through and fixed them for every other component in Transistor_FET in #1619

diegoherranz:
> IMO it's not worth it. This symbol still captures the device and the diodes don't add much. A note in the description wouldn't hurt, but I don't even think is necessary to merge.

OK. I'll merge.

Thanks, @0xdec, for your contribution.

PR #1120, opened 2018-11-04
Added atomic part and dcm data for PIC18(L)F25/26K83 in the narrow DI…

fragraider:
Commit message:
"Added atomic part and dcm data for PIC18(L)F25/26K83 in the narrow DIP-28 package to MCU_Microchip_PIC18.lib and MCU_Microchip_PIC18.dcm via KiCad's library editor."

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/40001943A.pdf

image

image

image

image

image

Commits for other packages would follow after feedback for this one (since a lot of copy paste would be involved).
I feel overwhelmed by the "Check the output of the Travis automated check scripts" since I could not find much information online and seems to require login to https://travis-ci.org/ ?

Thanks for all your hard work on KiCad!

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #1624, opened 2019-03-11
Removed placeholder x form PIC18(L)F25/26K83x-x_SP. It is now PIC18(L…

fragraider:
…)F25/26K83-x_SP. The Dip package (SP) has no option at this possition, making the x misleading.

This error was discovered after merging:
https://github.com/KiCad/kicad-symbols/pull/1120
Details at the end of that thread. 

This pull request has therefore this single letter removed form the part numbers.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/40001943A.pdf

billede

---


myfreescalewebpage:
Thanks @fragraider for this, good catch !

For me, regarding to product identification system page 840 of the datasheet, the part should be `PIC18F25K83-xSP`, `PIC18F26K83-xSP`, `PIC18LF25K83-xSP` and `PIC18LF26K83-xSP` (no "_").

Let me known what do you think.

Joel

fragraider:
Hello Joel

I must admit that I do not really have an opinion:
- The "_" marks the spot where the footprint would have an /, but I find it visually disturbing, even ugly.
- No "_" would maybe leave the reader confuse as to where the temperature indication ends and the package indication starts. (If the reader is looking for such a hint).

I am certainly more than happy to remove the underscore, especially the more I look at your comment. The small x really differentiates the itself form the capital SP (And would so in the other footprints too).

Hannes

P.S.: All honor goes to @diegoherranz, he was really patient and helpful and found the issue right after he merged https://github.com/KiCad/kicad-symbols/pull/1120

myfreescalewebpage:
Thanks for the answer, I have based my comment also on existing devices in the library. Most of them do not have "_", just look at the PIC18F family. Some few of t he PIC have "_" but I suspect this is not really conformed to the datasheet too. I also realize many do no discard the temperature letter, but they should.

myfreescalewebpage:
Thanks, merging.

fragraider:
I thought to just remove the underscore. Should it be decided to go back, I get a chance to learn how to return to a previous state in git :)

@myfreescalewebpage 
> I also realize many do no discard the temperature letter, but they should.

From the KLC I understood that every aspect of a part number should be preserved?
http://kicad-pcb.org/libraries/klc/S2.2/

Here is a screenshot of what was comited:
billede

fragraider:
Thanks!

myfreescalewebpage:
@fragraider next release will be 6.0, so it is permitted to break symbols, particularly to enhance them according to the KLC, if you wish, submit a PR we will review with pleasure :)

fragraider:
@myfreescalewebpage
I think the symbol is currently compliant! I will soon get working on the versions for the other footprints :)

fragraider:
@myfreescalewebpage
Doh, sorry, it took some time to get your comment... Yeah, would actually be a nice effort :)

poeschlr:
@fragraider sadly this was wrong information. The next release will be 5.1.1 and after that 5.1.2, .... (v6 is at least two years away.)

This means for the next year or so we will need to be very careful of what to include. This particular pull request is on the brink to be honest and should have probably been delayed till we truly start with v6.

But it is low impact enough and fixes something that can be considered a bug if you squint hard enough. Meaning i will keep it in for 5.1.1. (Just be aware of that for the future.)

fragraider:
@poeschlr I kinda started to make kopies for this component for each footprint available. Should that wait too (I.e. I would NOT make a pull request when I would be ready with the those)?

The currently, there is only the DIP package included.

poeschlr:
Adding new things never is a problem. Changing existing assets in a way that breaks existing installations is. (This is mainly because kicad right now has no proper caching for symbols. The cache lib/rescue lib stuff simply has too many places where it can go wrong. It is unintuitive for users which leads to many projects being broken and sadly also breaks way to often even if the user does not really do something wrong. Not even wayne fully trusts that feature this is why we are advised to limit the impact of our changes.)

fragraider:
@poeschlr Understood! Then I think we where good, but will not modify anything in the current library, thanks for the info!

myfreescalewebpage:
Thanks @poeschlr for the info !

Issue #1523, opened 2019-02-13
Ground connection of 3 pins crystal symbols are not align with 50mil grid


hildogjr:
Ground connection of 3 pins crystal symbols are not align with 50mil grid, instead these ground pins are aligned with a 25mil grid.

evanshultz:
You are referring to the crystal symbols that end in `_Small`, like this one:
![image](https://user-images.githubusercontent.com/1936989/52725067-55d90880-2f65-11e9-9f6b-5958cad6c7ca.png)

There are four of them in the library.

http://kicad-pcb.org/libraries/klc/S4.1/ doesn't allow any pins off a 100mil grid, and almost all of the other `_Small` symbols still use a 100mil grid. Only the coupled inductors (like `L_Core_Iron_Coupled_Small`) have pins on a 50mil grid except for these crystals.

KLC allows shorter pin lengths for small discrete components, which apply to the `_Small` symbols, but there is no allowance for pin grid placement. So these symbols are in violation.

Because this makes it hard to connect wires to the pins if they're off grid, I think KLC is correct and these symbols are a problem. I propose changing the coupled inductor and these grounded crystal symbols to put all pins on a 100mil grid. Possible fixes:
1. The coupled inductor symbols can just be shifted up by 50mil.
1. Crystal ground pins can shrink to 25mil in length and be placed at a Y location of +/-100mil to be on grid. The pin number then smacks into the symbol but if the pin number size is changed to 30mil it just clears the symbol geometry. Leaving the pin number text size alone isn't that bad, only a bit ugly.

@poeschlr 
Sound OK to you? Should this be done right away and treated as a bug, or wait until v6?

@hildogjr 
Thanks for catching this issue and reporting it!

hildogjr:
I think as a small fix (not change any folder/name) should be interesting already be avaible at v5.1.
Yes, the ground pin could be shrink 25mil or yes better placing it down 25mil and use a line segment?

evanshultz:
@hildogjr 
Indeed, I would like to see this fixed in v5.1 as well.

Moving the crystal ground pin down by 25mil puts it at [0, -150], which is not on a 100mil grid. So that's not a fix to me. We want all symbol pins to be on a 100mil grid for wiring.

hildogjr:
Moving it down at [0, -200] mil, i think, will make the symbol bigger. Shrink it appear to be the best option, I just concerned about the pin name text.

PR #1625, opened 2019-03-11
Fix small gnd crystal ground pin location

evanshultz:
Fixes https://github.com/KiCad/kicad-symbols/issues/1523.

The above case noted that grounded crystal symbols are not on a 100mil grid as required by http://kicad-pcb.org/libraries/klc/S4.1/ clause 1. Clause 2 allows short pins for simple devices, which I think we have here, so I went with that scheme. The pin number text was shrunk to 30mil to be readable.

Here is an example symbol to give you the idea:
image

The whole symbol could be shrunk slightly, or just vertically to give more room around the pin numbers. But I went simple here and solicit your feedback if this is OK to merge as-is.

---


myfreescalewebpage:
@evanshultz thanks for the fix, no comment to do, merging.

poeschlr:
This one is of similar impact to the triangle symbol "fix" and one can argue that it was a bug so i allow it to stay in 5.1.1.

PR #1630, opened 2019-03-13
Change MAX31855KASA SO pin to tri-state

VictorLamoine:
This only changes the type of the SO pin of the MAX31855KASA symbol to tri-state.
The current output type leads to DRC errors if multiple MAX31855KASA symbols are added to a SPI bus.

Helps with #232


evanshultz:
@VictorLamoine 
Thanks! I see this is your first contribution so thanks for taking the time to find this issue, search to find an existing conversation about it, and then submit this fix.

Do you know the SO pin is actually tri-state? I imagine it is but I don't see this in the datasheet. If it's not, then the DRC error is valid. I'm not familiar with this particular IC so I don't know, and I'd like to get a confirmation before merging.

VictorLamoine:
`SO` pin corresponds to the `MISO` line of the SPI bus. If this pin was not tri-state we could not have multiple peripherals on the bus. (note: there is no `MOSI` line for this component because it is read only).

So it is logical that the pin is tri-state and is confirmed by the fact that I have already used multiple MAX31855 chips on the same SPI bus.

evanshultz:
@VictorLamoine 
I understand the SPI bus but wanted confirmation on this particular IC. Thank you.

PR #1637, opened 2019-03-15
added LMZ14201

rs-development:
http://www.ti.com/lit/ds/symlink/lmz14201.pdf

image

I think there is no matching footprint right now, is this correct?


rs-development:
I put the VIN Pin to the left, since similar devices do the same.

myfreescalewebpage:

- [ ] Sizes in the description are useless, you can simply write `1A, 42V Input, 0.8V - 6V Output, Step-Down DC/DC Power Module, TO-PMOD`
- [ ] Footprint and footprint filter should be defined
- [ ] Expose Pad should be stacked with GND
- [ ] RON and SS should be Passive to avoid ERC errors
- [ ] I will personally prefer RON above EN looking at the schematic first page of the datasheet

Cheers,
Joel

myfreescalewebpage:
Hi @rs-development thanks for the fixes. The footprint will be required, why not setting this ? If it is not yet available in the library, you should submit a PR on the footprint repository and indicate the link here to enable following this. It will not be possible to merge without footprint.
Regarding to stacking of EP, you should rename pin 8 to `GND`(see also travis error).
Joel

myfreescalewebpage:
@rs-development any news here ? Thanks :)

myfreescalewebpage:

PR #1639, opened 2019-03-16
Logic_LevelTranslator: Add TI SN74AVC8T245PW

zonque:
This part is a classic 8-Bit dual-supply bus transceiver.

The datasheet is here:

  https://www.ti.com/lit/ds/symlink/sn74avc8t245.pdf


zonque:
image

myfreescalewebpage:

- [ ] The name of the part is not enough, you should specify which part you are drawing. If you want to add VQFN, TSSOP and TVSOP, you should submit three symbols (in the same pull request it's great)
- [ ] Footprint is missing, footprint filter should also be filled, according to previous comment
- [ ] Keywords are not correctly spelled
- [ ] Description should end with the package name, for example `, TSSOP-24`if you choose the TSSOP version, etc.
- [ ] Pins length should be 100mil (symbol < 100pins)
- [ ] GND pins should be stacked (meaning all at the same place, first pin Power Input visible, other pins Passive invisible)
- [ ] You should finally center the symbol (see also travis results which is very nice to get something more consistent with the library)

Cheers,
Joel

zonque:
Thanks a lot for this review, much appreciated!

I addressed all of your points, with the exception of the footprint. For the PWR variant, `Package_SO:TSSOP-24_4.4x7.8mm_P0.65mm` fits very well.

The symbol now looks like this:

image

myfreescalewebpage:
Thanks @zonque for the job here. Some new comments:

- [ ] For the TSSOP version, the name should be `SN74AVC8T245PW` to cover all TSSOP-24 packages
- [ ] You are vioalting KLC S5.2 (see travis log), footprint filter should be `TSSOP*4.4x7.8mm*P0.65mm*` for this symbol

Else, seems great!

Joel

zonque:
Thanks again. Both issues have been addressed now.

myfreescalewebpage:
@zonque only the SN74AVC8T245PW device is expected, you should remove SN74AVC8T245PWR, see git diff at https://github.com/KiCad/kicad-symbols/pull/1639/files

zonque:
That's strange, I merely renamed the symbol. Thanks for noting, removed now.

myfreescalewebpage:
No worries, thanks for the update. Merging.

PR #1640, opened 2019-03-16
Timer_PLL: add Cirrus CS2000-CP

zonque:
Components such as the CS2000-CP don't seem to have a good place to live
yet. Hence, create a new library called 'Clock' for this.


myfreescalewebpage:
Hello @zonque thanks for this contribution. I think library Timer is okay for this part.
Two additional requirements we have: dcm file should be submitted two for this PR, and please can you complete the description according to the template, with screenshot, link to datasheet etc. That's really help us when performing a review and speed up the process.
Thanks,
Joel

zonque:
Thanks! Right, the Timer library seems okay for it, moved it there. Here's a screenshot of the symbol in the editor:

image

zonque:
`nCS` was changed into `~CS` now:

image

zonque:
`nCS` was changed to `~CS` now, and the `GND` is now of type "Power Input":

image

myfreescalewebpage:
Thanks, for what I see now, the symbol is not bad but to big, should be reduced, you have plenty of blank space everywhere. Also, we usually put input on the left and output on the right.
You can have a look to the MCP7940N-xMS for an example.

zonque:
Alright, thanks again. Made the symbol a bit more compact now, but I still have one input pin on the right side. It feels wrong to rip the logical groups apart. What do you think?

image

zonque:
Moved the symbol to `Timer_PLL` now, as this is where other N-fractional PLL devices live.

myfreescalewebpage:

- [ ] Description should ends with the package name, add `, MSOP-10` at the end
- [ ] The footprint filter should be `MSOP*3x3mm*P0.5mm*`
- [ ] XTO type is Output
- [ ] SCL/CCLK type is Input
- [ ] SDA/CDIN type is Bidirectional
- [ ] I suggest the position of the pins should follow the example page 6 of the datasheet.

Cheers,
Joel

zonque:
Right, that's much nicer. This is how it looks like now:

image

There's more space now inside the symbol, but I can't see how that could be improved unless the power pins are not put on the 0-axis.

myfreescalewebpage:
@zonque that's ok for me like that, thanks, merging!

PR #1642, opened 2019-03-16
- added TI SM72295 Full Bridge Mosfet Driver

daviid5:
grafik
http://www.ti.com/lit/gpn/sm72295

---


daviid5:
@myfreescalewebpage Hey, is there any further action needed?

myfreescalewebpage:
@daviid5 I have only added pending review labels, but the review has not been made at all. The backlog is plenty of pull request to review, but don't worry your turn will come :)

rs-development:
I think there are some errors in the symbol.

- OVS should be in input
- LIB should be inüput
- AIN and BIN should be output

myfreescalewebpage:
Hi @daviid5 , thanks for contributing,

First thanks to @rs-development to start the review.

A single additional comment I have during my review:

- [ ] PGOOD type should be Open Collector

I have no idea if the position of the pins is nice to make the schematic, but I trust in you about this point, this seems not bad from my point of view. if you have a sample schematic image you can share to render it, that's very nice :)

Cheers,
Joel

evanshultz:
1. PGOOD should stay Output type since there's an internal pullup. Same with OVP.
1. IIN and IOUT might best be Passive type since just a resistor connects there (with an internal buffer why do anything else to that sensitive pin?).
1. Perhaps PGOOD and OVP should be on the left side since they may be likely to go to the same part of group of circuits that generates the PWM drive signals.
1. I'd put pins 1/2 and 13/14 below the FET drives on each side because the current sense is done at the very front and end of the power circuitry. OVS could be above pins 13/14 on the right side. From the top of the right side, HOA, HBA, HSA, and LOA as a group, perhaps with room to fit a cap between HBA and HSA, and then the B side pins in a similar order. As Joel said above, showing a schematic should make it more clear how to arrange pins in the least-tangled way.

myfreescalewebpage:
@evanshultz following your comments:
1- not agree with you, the ERC error will fails but it should not if you have several open drain PGOOD outputs in parallel, which is possible to do here with multiple regulators going to the same MCU control input for example. The presence of the pull up is not relevant.
2- just notice IIN is output. IOUT is output too. Why Passive ? the resistor+capacitor on the schematic is intended to make a filter, but the connection is then going to something else, such as an analog input of an MCU to measure the current.
3- for many regulators we claim PGOOD on the right and we don't consider if EN/other pin are connected to the same group or not.
4- that's a the question asked to the author.
Cheers,
Joel

evanshultz:
1. Oh, right. Good point.
2. As mentioned above, there is an internal buffer so I can't fathom why this signal would be taken directly to another part. This high-Z signal can stay local and the buffered signal goes elsewhere. I don't know why the datasheet shows an arrow. Either type will work but, as you mentioned above, since it's a current output one could parallel multiple pins to develop more voltage across an external resistor (not sure why you'd need to do that, but it's possible).
3. Just thinking about signal routing.

myfreescalewebpage:
About point 2: agree with you, the signal to be used is the buffered output, not this one immediately. Modified my previous review to add this.

daviid5:
Thanks for your feedback

regarding the pin-layout, what you think about this one?
grafik

I also corrected the pin types (see picture above)

regarding the Layout, at the moment I've got a project with this layout:
grafik
(symbol is already changed to ne "version" from the screenshot above, while the traces around have not been corrected yet

myfreescalewebpage:
Not bad, maybe SIA and SOA can be on the right too so all "power" pins (current measure and FET drivers) are grouped on the same side.

evanshultz:
As can be seen in the screenshot above, PGOOD and OVP go back towards the PWM input signals (here possibly a header but they are all likely to go to/from the same place. I mentioned putting them on the left side above and that still seems reasonable. We have routinely prioritized nice pin placement over inputs on left and outputs on right.

daviid5:
Well I would suggest to place the driver outputs symetrical around the X-Axis
Move PGOOD and OVP to top left, and SIA SOA to top right

myfreescalewebpage:
Why not, can you do it and update your schematic sample ? Thanks @daviid5

myfreescalewebpage:
Hi @daviid5 there is a branch conflict here to solve and also some changes pending (see my previous comments).
Joel

myfreescalewebpage:
@daviid5 the merged failed, see https://github.com/KiCad/kicad-symbols/pull/1642/files and following screenshot displayed when opening KiCad:

Capture

Joel

evanshultz:
Not sure when or where this happened, but looking at the file I can see the text "DRAW" appears 155 times and "ENDDRAW" appears 77 times. For the closing draw statements there should only be 154 opening statements. A quick looks shows the bottom of symbol `TC4422` is cut off for some reason. You may need to manually fix this since master looks fine.

myfreescalewebpage:
Thanks @daviid5 
Fix at https://github.com/KiCad/kicad-symbols/pull/1642#issuecomment-479566321 not implmented, else, this is nice.
Joel

myfreescalewebpage:
Thanks @daviid5 
Last comment I have forgotten in the previous review: the name of the device should be "SM72295MA" to avoid conflict if the manufacturer add package options
Joel

myfreescalewebpage:
@daviid5 the change has duplicated the device, see at https://github.com/KiCad/kicad-symbols/pull/1642/files. There is also now a branch conflict to solve.
Thanks,
Joel

myfreescalewebpage:
Thanks for the fix, great job here ! Merging.

Issue #153, opened 2018-01-09
Missing symbols for Vigortronix VTX-214-010 family


antoniovazquezblanco:
The footprint exists (`Converter_ACDC_Vigortronix_VTX-214-010-xxx_THT.kicad_mod`) but no symbols for the component.

myfreescalewebpage:
@antoniovazquezblanco added VTX-214-010 with https://github.com/KiCad/kicad-symbols/pull/1648

antoniovazquezblanco:
I will review it. Need some time but I will find it ;)

antoniovazquezblanco:
Thank you @evanshultz! Faster than me ;)

PR #1648, opened 2019-03-18
Create VTX-214-010 family

myfreescalewebpage:
Fixes https://github.com/KiCad/kicad-symbols/issues/153

Create VTX-214-010 family following issue at https://github.com/KiCad/kicad-symbols/issues/153.
The symbol is based on VTX-214-015 already in the library

---

http://www.vigortronix.com/10WattACDCPCBPowerModule.aspx

Capture

GND pin at the right, warning from travis check, followed convention for this kind of symbol.


evanshultz:
1. Datasheet page 2 shows pin names of `L`, `N`, DC+` and `DC-`. Should we adhere to that? It seems more clear than either set of pin names shown on page 3. This goes for the existing VTX symbol too.
1. The -107 version is 7.5V output.
1. Can the `xxx*` be removed from the FP filter? Possibly swap it to `THT*`?

myfreescalewebpage:
@evanshultz thanks for the answer.

1- Yes and page 3 shows other names too. I have made some screenshot below.



So I just decided to keep the same name than VTX-214-015 family. Let me known what is the best, the datasheet has no clear table of the pins names...
2- fixed.
3- Keep consistency of the format with VTX-214-015 family too.

Joel

evanshultz:
1. The table on page 2 was the only place that was clear about the two AC pins, so that's one reason why I favored it. Both VTX symbols could use this naming?
3. Right now, this part of the FP filter does nothing. If there was an SMD later then possibly adding "THT" could help. But likely there will only be the THT version so the footprint and the symbol don't need to capture that. We don't specify SMD or THT for most footprints if there's only one version.

Neither are blocking to merge this, but thinking that having v5.1 released gives us more latitude to change I looked at it with fresh eyes. :)

myfreescalewebpage:
1- Yes they can, I will fix both symbols right now
2- My intention was to help the issue #153 which is only about the missing symbol but if it is more consistent we can also review the footprints names, let me know

myfreescalewebpage:
Changing the output names also solve one travis warning :)

myfreescalewebpage:
@evanshultz any new comments ? Thanks.

evanshultz:
Nope. Looks OK.

antoniovazquezblanco:
Thank you both!

PR #1652, opened 2019-03-19
SN65HVD1050

SchrodingersGat:
Added 5V CAN transceiver in SOIC-8 package
http://www.ti.com/lit/ds/symlink/sn65hvd1050.pdf

image

image

Symbol copied from existing SN65xxx series of CAN chips for compatibility.


myfreescalewebpage:
Hi @SchrodingersGat , thanks for contributing,

- [ ] Package in the description should be completed: "SOIC-8"
- [ ] VREF should use capital letters
- [ ] According to typical application schematic page 19, I think VREF should be moved between CANH and CANL pins
- [ ] "transceiver" in keywords badly spelled

Cheers,
Joel

evanshultz:
@SchrodingersGat Ping!

SchrodingersGat:
All covered, I think. Thanks for the review

myfreescalewebpage:
I have trouble with the design of the right part of the symbol, because it is not vertically centered.

Your symbol:

Capture

An other symbol I think better in the library (LTC2875-DD), except it is too large and that for this device PAD should be stacked with GND:

ltc

SchrodingersGat:
Should we consolidate all the CAN symbols then? There are a few different sizes and shapes for them.

myfreescalewebpage:
Good question, at least KLC errors can be improved but something more consistent between the symbol is good to have IMO. @evanshultz yours ?

evanshultz:
I fully support the idea, but to follow the devs and Rene we can't until close to v6. But for now we can at least make future contributions attractive.

SchrodingersGat:
image

Updated

myfreescalewebpage:
@SchrodingersGat no more comment here, merging. Thanks for this job!

PR #1653, opened 2019-03-19
Added MCP2542FD

SchrodingersGat:
- MCP2542WFD alias
- DFN-8 3x3mm package (matches existing footprint)

image

Datasheet - http://www.microchip.com/mymicrochip/filehandler.aspx?ddocname=en585412


myfreescalewebpage:
Hi @SchrodingersGat , thanks for contributing,

- [ ] Can you explain from where the name of the device comes from ? I do not found any information about "FD" of "WFD" you have set in the name of the symbols
- [ ] Chapter 1.7.10 in the datasheet recommends to stack PAD pin with VSS
- [ ] I disagree with EP size, I think the footprint should be `Package_DFN_QFN:DFN-8-1EP_3x3mm_P0.65mm_EP1.47x2mm`

Cheers,
Joel

SchrodingersGat:
Hi @myfreescalewebpage thanks for the review;

From the first page of the datasheet:

image

image

The other points can easily be addressed. I'll await your feedback from the other PR

myfreescalewebpage:
@SchrodingersGat thanks to point me this. What I do not understand is that there is no difference between the package version.... Have you got more details about that ? Else, you will need to rename the symbol to "MCP2542FD_DFN8" for example

SchrodingersGat:
There is a difference on Pin 5 - one package has VIO and one package has NC

myfreescalewebpage:
> There is a difference on Pin 5 - one package has VIO and one package has NC

Sure. However I was speaking about the package name. How the difference is made between DFN and SOIC for example . I do not found the information...

SchrodingersGat:
Oh, sorry you are correct, I see what you mean now.

The ordering guide at the end of the datasheet has this:

So a suffix of ```MF``` would be appropriate here.

image

myfreescalewebpage:
Thanks, so your devices should be `MCP2542FDxMF` with alias `MCP2542WFDxMF` (added to previous checklist).
Joel

SchrodingersGat:
I have updated the part numbering as suggested.

I think the footprint selection is correct based on the exposed-pad sizes given in the datasheet:

image

SchrodingersGat:
Also fixed the datasheet links which were incorrect

myfreescalewebpage:
Thanks! Made a complete review, no more comment now. Merging.

PR #1490, opened 2019-02-04
Remove part name from description in power ports

JacobEFO:
Remove part name from description

---
Description: Several power ports contained part name in the description violating http://kicad-pcb.org/libraries/klc/S6.3/

Fixed that issue on several power ports. The filtering function remains unchanged.

Output from Travis check:

Checking symbol '+10V' - No errors
Checking symbol '+12C' - No errors
Checking symbol '+12L' - No errors
Checking symbol '+12LF' - No errors
Checking symbol '+12P' - No errors
Checking symbol '+12V' - No errors
Checking symbol '+12VA' - No errors
Checking symbol '+15V' - No errors
Checking symbol '+1V0' - No errors
Checking symbol '+1V1' - No errors
Checking symbol '+1V2' - No errors
Checking symbol '+1V35' - No errors
Checking symbol '+1V5' - No errors
Checking symbol '+1V8' - No errors
Checking symbol '+24V' - No errors
Checking symbol '+28V' - No errors
Checking symbol '+2V5' - No errors
Checking symbol '+2V8' - No errors
Checking symbol '+3.3VA' - No errors
Checking symbol '+3.3VADC' - No errors
Checking symbol '+3.3VDAC' - No errors
Checking symbol '+3.3VP' - No errors
Checking symbol '+36V' - No errors
Checking symbol '+3V3' - No errors
Checking symbol '+3V8' - No errors
Checking symbol '+48V' - No errors
Checking symbol '+4V' - No errors
Checking symbol '+5C' - No errors
Checking symbol '+5F' - No errors
Checking symbol '+5P' - No errors
Checking symbol '+5V' - No errors
Checking symbol '+5VA' - No errors
Checking symbol '+5VD' - No errors
Checking symbol '+5VL' - No errors
Checking symbol '+5VP' - No errors
Checking symbol '+6V' - No errors
Checking symbol '+7.5V' - No errors
Checking symbol '+8V' - No errors
Checking symbol '+9V' - No errors
Checking symbol '+9VA' - No errors
Checking symbol '+BATT' - No errors
Checking symbol '+VDC' - No errors
Checking symbol '+VSW' - No errors
Checking symbol '-10V' - No errors
Checking symbol '-12V' - No errors
Checking symbol '-12VA' - No errors
Checking symbol '-15V' - No errors
Checking symbol '-24V' - No errors
Checking symbol '-2V5' - No errors
Checking symbol '-36V' - No errors
Checking symbol '-3V3' - No errors
[...]

PR #1654, opened 2019-03-19
Add Symbol for E73-2G4M08S1C BLE5 module (nRF52840)

GitWellBack:
Schematic symbol for EByte E73-2G4M08S1C BLE5 module (nRF52840 SoC with ceramic antenna)

Datasheet: http://www.ebyte.com/en/downpdf.aspx?id=445

Screenshot:
E73-2G4M08S1C-symbol

Footprint PR: #1490

Credit: Tried to keep this symbol matching the layout of the other E73 symbol already in the library (created by @isundaylee)

KLC violations: No errors from script-testing.

---


myfreescalewebpage:
Hi @GitWellBack , thanks for contributing,

- [ ] I suggest to create a new library "RF_Bluetooth" for this kind of device. Can you do it for the symbol and the footprint ?
- [ ] VBUS should be moved to the top (Power Input pin)
- [ ] GND pins must be stacked
- [ ] I suggest to put P1.xx on the left to reduce a bit the size of the symbol. Suggestion below.

Capture

Cheers,
Joel

GitWellBack:
>     * [ ]  I suggest to create a new library "RF_Bluetooth" for this kind of device. Can you do it for the symbol and the footprint ?
> 
>     * [ ]  VBUS should be moved to the top (Power Input pin)
> 
>     * [ ]  GND pins must be stacked
> 
>     * [ ]  I suggest to put P1.xx on the left to reduce a bit the size of the symbol. Suggestion below.

I'll get on these and make commits to both PRs tomorrow - thanks for the suggestions.

evanshultz:
@myfreescalewebpage 
Note that sym-lib-table isn't updated in-place if the user just grabs new libs from GH or the repo on the KiCad website. So in that case, limiting the creation of libraries will help this symbol get into everybody's library without as much hassle. Prefer to use an existing library if possible.

myfreescalewebpage:
@evanshultz I know. I'm not sure there is Bluetooth device symbols at the moment in the lib so I asked here to create the new library right now.

GitWellBack:
Changes made - including pushing relevant parts to RF_Bluetooth (symbol, footprint and 3dshape). @myfreescalewebpage and @evanshultz - both valid points, I've gone with the move to RF_Bluetooth, purely because on studying the contents of RF_Module, I think it might be more appropriate to have the Bluetooth modules separate from the more esoteric RF modules. On top of which there's already an RF_Bluetooth folder for symbols in the repo (possibly not in sym-lib-table), just not for footprints or 3dshapes. Basically KiCAD modules still needs a sort-out ;-) I realise this will also slow uptake due to fp-lib-table changes, but I guess that's life.

myfreescalewebpage:
Looks great. Just one comment here: can you order the positive power supply pins so that VBUS will be on the left of VDD ? I guess this will be easier to wire the supply from the USB connector/cable.

GitWellBack:
> 
> 
> Looks great. Just one comment here: can you order the positive power supply pins so that VBUS will be on the left of VDD ? I guess this will be easier to wire the supply from the USB connector/cable.

And this ladies and gentleman is why I should stick to my rule of testing components in my designs before submitting/committing! :-)
Thanks for catching that - definitely should have been the other way around - if only to keep things straight as you're looking (I tend to label things off to a connector/comms sheet, but I do like clustered functionality.)

myfreescalewebpage:
loooool, thanks, the symbol looks great now :)
Just need to wait the footprint before merging

myfreescalewebpage:

PR #1655, opened 2019-03-19
add AD9912 DDS

aewallin:
AD9912 DDS in 64-pin LFCSP

https://www.analog.com/media/en/technical-documentation/data-sheets/ad9912.pdf

AD9912

Questions/ideas:
- all DVSS pins can be stacked?
- all AVSS pins can be stacked?
- EPAD should be connected to GND, but datasheet doesn't say DVSS or AVSS - so we leave it as a separate pin
- I would rather not guess how many de-coupling caps the designer wants to use - so maybe leave all the supply pins as separate pins?
- should AD-naming of footprint be used "CP-64-7" or the more generic LFCSP-64-1EP_9x9mm_P0.5mm_EP5.21x5.21mm ?

https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/lfcspcp/cp_64_7.pdf

Footprint PR is now: https://github.com/KiCad/kicad-footprints/pull/1492

---


aewallin:
I don't understand why the symbol editor has touched AD9951 also in this commit? only change is addition of AD9912.

evanshultz:
Different versions of KiCad have a nasty habit of re-ordering the lines in library files. It's likely that this library was last saved with a different version of KiCad that you used.

Answers:
1. If the "Pin No" column on pages 9-10 have more than one number, they can be stacked unless their are other requirements. You've been around so you know this. :) Ground pins should always be stacked.
1. Page 10 says the EPAD is connected to Analog ground, so that sounds like it can stack with AVSS.
1. Page 26 says "With each group, use bypass capacitors of 1 μF in parallel with a 10 μF." That matches the eval board at https://www.analog.com/media/en/evaluation-boards-kits/evaluation-board/AD9912_Schematic.pdf, so I'm good with grouping the power pins in this way.
1. Generic footprint name.

Review notes:
1. Once pins are stacked the symbol can be squeezed down.
1. You can intersperse the NC pins with other pins so the symbol doesn't grow just for NC pins.
1. The S* pins can he Hi-Z, so let's make the tri-state type.
1. Pins 31 and 48 can be Passive type.
1. Is there an arrangement of pins 34/35 and 40/41 that won't require crossing wires in a normal usage?
1. I would be inclined to put the S* pins at the bottom of the left side. What do you think?
1. Perhaps pin 31 should be at the top right so be close to the AVDD pins, assuming the external filter is normally used.
1. Set the pin name offset to 20mil.
1. The footprint filter should match the footprint, but without the EP section.

aewallin:
here is revision 2.
- power supply pins grouped according to p26/40
- GND pins stacked
- pin types as suggested
- fp-filter set
- S-pins bottom left
- loop_filter pin top right
- smaller symbol, might be hard to squeeze it further 

AD9912

aewallin:
I don't understand the travis-error about the DVSS pin-stack.
DVSS pin2 is visible a and 4,6,8,56,57 invisible.

aewallin:
and now I understand the travis-error even less. it messes up the pin-coordinates for the DVSS stack??

  Violating S4.3
    Rules for pin stacking
    Pinstack @ (-100,1000) must have exactly one (1) visible pin
     - Pin DVSS (56) is INVISIBLE
     - Pin DVSS (57) is INVISIBLE
  Violating S4.4
     - Pin DVSS (56) @ (-100,-1000) is of type PASSIVE
     - Pin DVSS (57) @ (-100,-1000) is of type PASSIVE

evanshultz:
Can you please take a look at #5 above? All other review items look good. Thank you!

I'm not sure about the DVSS pin stack. Looks good to me from checking out the file. Let's restart CI.

evanshultz:
Let's put digital supplies on the left and analog on the right, irrespective of the order in the datasheet.

The symbol can be squeezed more. All the power pins can pack together. Pin 31 can slide under them and all pins on the right side can move down.

You can see just before `P N` on pins 56 and 57 they are `0 0` and not `1 1` like the other DVSS pins. This is because those pins are common to all units and body styles. If you fix that it should pass CI. Then all that will be left is the footprint.

aewallin:
rev 3:
- DVDD on left, AVDD on right
- further pack pins to squeeze down size
- power-input stacks have one visible/power-input and the rest invisible/passive
- I don't see a big issue with out/outb and connection to feedback/feedbackb - there is a low-pass filter or transformer in-between usually anyway. perhaps example-schematics would help to see if there is a better arrangement. output-pins now follow datasheet fig39 roughy.
AD9912r3

evanshultz:
I did like the left-side pins spaced out like it was before.

Please put the ref des on top left and value on top right instead of in the middle.

aewallin:
AD9912r4

evanshultz:
Looks good to me. Thanks for your time to submit the symbol and footprint!

aewallin:
great, thanks for the review and suggestions!

PR #1657, opened 2019-03-20
Adds +3V0 to the power flags in the power library

gvandewaetere:
Adds +3V0 to the power flags in the power library.


evanshultz:
Thank you.

This is your first contribution so thanks for taking to time to add to the official library and hope to see you again!

diegoherranz:
Quick question. Why is this 3V0 and not simply 3V?
For example there are 12V and 5V symbols, not 12V0 or 5V0.
Thanks!

evanshultz:
@diegoherranz 
When I saw both `+3.3V` and `3V3` symbols I figured we weren't picky about formatting and would accept whatever users submitted. Maybe I was too quick to merge?

gvandewaetere:
I actually prefer the 3V3, 2V8, ... notation for a simple reason. Some parts of the world (Belgium for example) use a comma to separate the integer and fraction digits. To avoid ambiguity in international situations, I always the letter for the units where the decimal point should be. Writing the 0 or not is just a case of personal preference I guess. Also I like the way that if you list typical voltages +1V8, +2V5, +2V8, +3V0, +3V3, +5V0, +12V,+24V,... they all have the same amount of characters. The only one where it 'feels' a little bit weird (for no logic reason) for me is the +5V0 one (here I prefer +5V).

poeschlr:
We might want to look into this in more detail once the feature set of the new file format is clear. For now i would suggest not to be too picky but don't let this lib grow too much.

PR #1663, opened 2019-03-20
Addition - DMG2301L

gcormier:
Just another MOSFET, added an alias. **Please** double check my alias selection for pins/GDS.

image

image

https://www.diodes.com/assets/Datasheets/DMG2301L.pdf


myfreescalewebpage:

- [ ] Package name in the description should be simply "SOT-23", not "SOT-23-3"

Your alias is correct.

Cheers,
Joel

gcormier:
51aaa93 should fix that up!

myfreescalewebpage:
@gcormier good catch for the TP0610T package name, but your new symbol DMG2301L has still wrong package format :)

gcormier:
Ahh gotcha. Okay, I updated all of them to SOT-23 which is the standard. Might as well clean them up.

myfreescalewebpage:
Thanks! Merging

PR #1665, opened 2019-03-21
Add 74LS07

b1ackmai1er:

---


myfreescalewebpage:
@b1ackmai1er thanks for this contribution. Several question and comments:
- Why the git diff is showing a so large difference ? This is not normal. Maybe you have got some issues with lien ending.
- I think the device can be an alias of 74LCX07, isn't it ?
- Finally, we expect from you to complete your pull request message with several information such as datasheet link and screenshot of the symbol, please update your previous message, this clearly simply the review of contribution and it is very important for us.
Thanks,
Joel

b1ackmai1er:
Hi Joel,
Line endings fixed - I think I understand what is causing this now (my text editor).
I saw the 74LXC07 part but the symbol did not match the data sheet symbol (center symbol).
Would you prefer and alias or a the new symbol which matched the data sheet? I am not sure what the center symbol signifies.
image
Apologies for not completing information - I rushed this.

Regards Phillip

evanshultz:
That symbol means the pin is open collector. It's in IEEE 91-1984.

I'm not aware of any rules/guidelines for this in the library now so perhaps it's time to decide. Librarians? My feeling is that this helps to catch mistakes with visual checking, but a more robust solution is to have automated checking (like for missing pullup resistors, in this case) and that's a better way to go. My experience is that these open collector symbols are wildly inconsistent in the industry so it's not obvious there is a "correct" solution and we can decide what goes into KLC on this point.

b1ackmai1er:
Oops forgot data sheet again: www.ti.com/lit/ds/symlink/sn74ls07.pdf

myfreescalewebpage:
@b1ackmai1er I do not know about the symbol. @evanshultz probably this is something to think more generally. Anyway, the 74LCX07 has Open Collector output too, so your symbol can be an alias of this one:

Capture

Joel

evanshultz:
Yes. I say move ahead here and pin symbols can be a separate discussion. We already require separate symbols if pin types are different, so this is only a visual thing and won't affect our library otherwise.

b1ackmai1er:
Hi gents, Changed to aliases as requested.

myfreescalewebpage:
Thanks @b1ackmai1er for the modification.
However, the new footprint filter for DIP-14 should be `DIP*W7.62mm*`.
Joel

b1ackmai1er:
Hi I have update this and other DIP?14* entries to DIP*W7.62mm*
In the process I have actually rause a PUSH/Patch request by mistake. Any idea how to cancel that?
Travis reporting huge amount of errors which I do not understand.

myfreescalewebpage:
The travis warnings and errors are due to the fix of the footprint filters, this is normal.
No more comment to do, I think we can merge.

PR #1667, opened 2019-03-21
Add ltm8063

eeyrjmr:
Adding LT8063 uModule switching regulator.

New footprint required. Analog_BGA-28_4.00x6.25mm created and pull request to kicad-footprints to be requested

Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/8063fa.pdf
Symbol outline from: https://www.analog.com/media/en/dsp-documentation/evaluation-kit-manuals/DC2494AF.PDF


eeyrjmr:
All Travis checks pass. This PR is ready for review

Untitled

myfreescalewebpage:
Hi @eeyrjmr , thanks for contributing,

- [ ] Description should ends with package name: `40VIN, 2A Silent Switcher µModule Regulator, BGA-28`
- [ ] You have a strange dot at the end of keywords, to be removed
- [ ] The datasheet link contain some space that break it, please review
- [ ] Pin name position offset should be 20mil (see also travis warning)
- [ ] Footprint is not set, should be `Package_BGA:Analog_BGA-28_4x6.25mm` this will fail in travis but it is normal while the footprint is pending
- [ ] Footprint filter will be `Analog*BGA*4x6.25mm*`
- [ ] Pin name should use capital letters only
- [ ] GND type should be Power Input
- [ ] PG should be on the right of the device (block diagram page 11 of the datasheet)
- [ ] SYNC should be on the left of the device (block diagram page 11 of the datasheet)
- [ ] VOUT pins should be stacked

Cheers,
Joel

eeyrjmr:
@myfreescalewebpage thank you for reviewing.

Untitled

- [x] Description should ends with package name: 40VIN, 2A Silent Switcher µModule Regulator, BGA-28
Done.
 - [x] You have a strange dot at the end of keywords, to be removed
This did not appear in the UI but did appear in the datafile. Action 1 rewrite the raw file and this was corrected
 - [x] The datasheet link contain some space that break it, please review
This did not appear in the UI but did appear in the datafile. Action 1 rewrote the raw file and this was corrected
- [x]  Pin name position offset should be 20mil (see also travis warning)
Done
- [x]  Footprint is not set, should be Package_BGA:Analog_BGA-28_4x6.25mm this will fail in travis but it is normal while the footprint is pending
Done and expected Travis failure
- [x]  Footprint filter will be Analog*BGA*4x6.25mm*
Done
- [x]  Pin name should use capital letters only
Done
- [x]  GND type should be Power Input
Done
- [x]  PG should be on the right of the device (block diagram page 11 of the datasheet)
Done
- [x]  SYNC should be on the left of the device (block diagram page 11 of the datasheet)
Done 
- [x]  GND pins should be stacked
Done 
- [x]  VOUT pins should be stacked
Done. Resultant Travis failure

**Travis failures:** 
 **Violating S4.4**. As per special cases in S4.4 ( http://kicad-pcb.org/libraries/klc/S4.3/ ) the additional VOUT have had their type set to PASSIVE to ensure no DRC failure

**Violating S5.1** Footprint does not exist at this moment. Request KiCad/kicad-footprints#1494 covers this need

myfreescalewebpage:
Thanks @eeyrjmr great job, this is clearly better. Several comments:
- [ ] Missing a comma before package name in description
- [ ] VOUT stacking is incorrect, all pins should be Power Output. All stacking rules at http://kicad-pcb.org/libraries/klc/S4.3/
- [ ] "LTM8063" should not be bold and to be moved in the top right corner of the symbol, "U?" should go in the top left.
Joel

eeyrjmr:
@myfreescalewebpage recommendations incorporated

- [X] Missing a comma before package name in description
- [X]  VOUT stacking is incorrect, all pins should be Power Output. All stacking rules at http://kicad-pcb.org/libraries/klc/S4.3/
- [X]  "LTM8063" should not be bold and to be moved in the top right corner of the symbol, "U?" should go in the top left.

**Travis failures:
Violating S4.3**    

>  - Pin VOUT (F2) : POWER OUTPUT should be visible in a power-in/power-out/output pin stack
>  - Pin VOUT (F2) : POWER OUTPUT should be an invisible PASSIVE pin power-in/power-out/output pin stack

Commit 63b5e40 undid 3715f6d which aimed to meet S4.3 but 2nd review comment is indicating the stacked VOUT should all be power not passive 

**Violating S5.1** Specific part does not exist. Ticket open 

Untitled

myfreescalewebpage:
For VOUT, all should be Power Output, only one should be visible. This should makes travis happy.

eeyrjmr:
ok that's odd because all VOUT are "Power output" and only one is visible (F1).  Ill look at example parts and see if there is something I am missing 

Untitled

myfreescalewebpage:
Just checked, One should be visible, type Power Output, all the other one should be Passive, invisible

eeyrjmr:
Ok I have essentially reverted 63b5e40 bringing the pins closer to commit  3715f6d  Travis now has the pins back as a warning and the footprint as an error (known)

myfreescalewebpage:
Thanks, great job and responsive <3
Just need to wait for the footprint before merging the symbol.

evanshultz:
I would suggest swapping pins A3 and B2 so that if SYNC gets an external source the two pins below it doesn't have wires that cross. TR/SS and RT just have passive parts going to ground.

Footprint is merged.

eeyrjmr:
Valid point. RT and TR/SS  are always to 0V and while sync can be it doesn't have to be. 

Swapped positions of PG and FB for a similar reason. FB always to local GND while PG is open-collector

myfreescalewebpage:
@eeyrjmr ok for SYNC pin but I don't think moving PG is good here because VOUT and FB should be clearly grouped ! please revert this one.
Joel

evanshultz:
@myfreescalewebpage 
This one is a bit weird. FB is the bottom resistor in the divider, and the top resistor is inside the package. So FB does in fact have just a single resistor going to ground from it. I can imagine other esoteric ideas, but this is the intended use and probably the case in the vast majority of designs. So anyway, it probably does make sense since PG will go to other circuitry if it's used, but FB is unlikely to go elsewhere. What do you think?

eeyrjmr:
@evanshultz  exactly. for some switches the feedback resistor goes between SW and the FB pin and thus such pins being next door to each other makes sense.
This chip (and others in the series) has the resistor going to GND, as it completes the divider. 

@myfreescalewebpage If you wish I can swap them back but I believe this will cause capture compromises if an end-user wants to use the PG pin 

below is an example sketch for a 3v3 configuration

Untitled

myfreescalewebpage:
@evanshultz @eeyrjmr thanks to point me that, I haven't realized before only a single resistor was required. In this case, agree with the current design of course :) Merging!

PR #1671, opened 2019-03-23
Clearer descriptions for power symbols

poeschlr:
All power symbols where reffered to as power flags in their description. I made it clear that they create global labels and added a special descritpion to the PWR_FLAG symbol to tell users that this one is to be used to tell ERC where power comes from.


myfreescalewebpage:
Hi @poeschlr no comment to do here, great enhancement, merging.

PR #1681, opened 2019-03-25
Driver_LED: Add MPQ2483

zonque:
This adds the MPQ2483 from Monolithic Power Systems. The datasheet is here:

  https://www.monolithicpower.com/pub/media/document/MPQ2483_r1.05.pdf

One note on the pin arrangement: the FB pin lives on the right side of the
symbol even though it's an input pin, because it is alway fed from the
switching loop that's fed from the BST pin.

This symbol needs a new footprint which was submitted here:

  https://github.com/KiCad/kicad-footprints/pull/1498

image


zonque:
The footprint is now merged. Can the travis job be restared?

myfreescalewebpage:
Restarting travis.

evanshultz:
1. Remove pin count and EP from footprint filter leaving `DFN*1EP_3x3mm*P0.5mm*`.
1. What about putting OVP on the right side since it senses the output voltage?
1. How about INGND to the left of GND and spaced apart a bit to fit the resistor to ground?
1. Can BST and SW spread apart so a cap symbol can fit there? And FB and SW to fit the resistor?

zonque:
I'm uncertain about the pin arrangement. How about the version I just pushed? This resembles the schematics in the datasheet as well.

image

evanshultz:
Hmm... yeah. What about OVP just above or below FB, COMP on the left side, and smash the symbol a bit? Put INGND on the bottom or side, whichever looks best to you. Then let's see what we have. Or if you've got other ideas bring them up.

zonque:
Nah, I figure OVP needs to be the topmost on the left, as that matches all typical schematics examples.

How about this (just pushed)?

image

evanshultz:
Looking better. Thank you.

I'd rather see OVP near the bottom left. I find the schematics in the datasheet to be poorly drawn because of this reason. I can't think of a good reason to put OVP on top as shown.

Perhaps shrink a bit? If INGND goes to the right it can get a bit shorter.

zonque:
> I'd rather see OVP near the bottom left. I find the schematics in the datasheet to be poorly drawn because of this reason. I can't think of a good reason to put OVP on top as shown.

OVP on the left side, really? OVP measures the voltage of the generated output, so I really think it should be on the right. Otherwise you need to go all around the symbol to feed the signal back in. FB is fed from the return path, so I think it makes more sense to have it underneath OVP.

INGND is the reference for the RSET input, so they should be close to each other (ie, INGND on the bottom left).

What about this then?

image

evanshultz:
My mistake. I meant OVP on the right but typed left. My fault for being confusing. Sorry.

I like the image above. My only comment left is to space apart OVP and FB from SW (move BST and SW up as well?) so the top OVP resistor can more easily fit after the inductor between SW and OVP. Right now the empty space on the right side of the symbol isn't being used so why not use it?

evanshultz:
And also please change the underscore in the footprint filter to an asterisk.

zonque:
Like so?

image

evanshultz:
Sure. Let's do it. Thank you!

zonque:
Yeah, thanks a lot for all the review. Updated my schematics now, and the new arrangement in fact pays out :) One small issue showed up when I did though, sending a fix right now.

PR #1682, opened 2019-03-26
NLSV2T244

SchrodingersGat:
- Added three footprint variants of the same IC
- Dual non-inverting level translator
- Datasheet: http://www.onsemi.com/pub/Collateral/NLSV2T244-D.PDF
- One variant requires the footprint here - https://github.com/KiCad/kicad-footprints/pull/1502
- Symbol copied from existing logic level translators
- Only KLC issue is that the footprint isn't merged yet.

image


evanshultz:
1. Remove `R2G` from the symbol names since that's non-functional.
1. You've got pins 4 and 5 swapped.

SchrodingersGat:
Good catch on the pins. Changes made.

evanshultz:
Just update this to use the OnSemi footprint name, since it was changed.

SchrodingersGat:
Done!

evanshultz:
The footprint looks good. Please update the filter and we're ready to merge.

SchrodingersGat:
Oops, sorry for missing that one

evanshultz:
I think we haven't been doing leading asterisks. What about smacking `OnSemi` up front?

SchrodingersGat:
Ok, done

PR #1683, opened 2019-03-26
TXB0304RUT and TXBN0304RUT

SchrodingersGat:
Added two new symbols:
- TXB0304RUT as alias of TXB0104RUT
- TXBN0304RUT as new symbol (difference is that OE is active-low)
- Datasheet: http://www.ti.com/lit/ds/symlink/txb0304.pdf
- Footprint already exists in KiCad libs

image


evanshultz:
1. Both symbols have two rectangles. Only one is needed (the one with background fill).
1. Shouldn't A* and B* pins on both symbols be Tri-state type?

SchrodingersGat:
Not sure how that weird extra square appeared. Anyhow, fixed now.

PR #1684, opened 2019-03-26
Added ZXMP4A16GTA P-Channel FET

SchrodingersGat:
- Datasheet: https://www.diodes.com/assets/Datasheets/ZXMP4A16G.pdf
- Symbol copied from existing FETs (added extra pin for tab)

image


myfreescalewebpage:
@SchrodingersGat have you checked if an existing symbol can be used as an alias ?

SchrodingersGat:
I could not find one that was in a SOT-223 package with this pinout.

evanshultz:
Agreed. There's only one other SOT-223 FET and it's an N-channel.

1. Description says 4.6A but should be 6.4A.
1. `TA` in symbol name is for packaging so get rid of it.
1. Use the `SOT-223-3_TabPin2` footprint so it doesn't need the extra pin.

SchrodingersGat:
Done!

PR #1697, opened 2019-03-30
Added  PIC32MK1024GPD064-I/PT

ttshaw1:
Added PIC32MK1024GPD064-I/PT,  TQFN-64, to PIC32 symbol library. Datasheet for IC is here: http://ww1.microchip.com/downloads/en/DeviceDoc/PIC32MK-General-Purpose-and-Motor-Control-%20(GPMC)-Family-Datasheet-60001402E.pdf
image

Error checking scripts showed no errors. Pins 47 and 48 are Inputs rather than the typical Bidirectional in accordance with the datasheet. Pin 42 (Vbat) is specified as nonfunctional and should be connected to Vdd, so I set it as a power input and put it near Vdd. 

I followed the convention of the other symbols in the library for the symbol properties. But I did notice some issues with them, such as the description not being comma separated and using KB for kilobytes (kB). I think the other symbols in the library also use the wrong pin name offset. I'm not sure how best to reconcile those differences. 

---


ttshaw1:
Forgot the .dcm file. Will try again.

PR #1700, opened 2019-03-30
ADXL343 symbol

nonninz:
I've added a symbol for the ADXL343 accelerometer.

Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/ADXL343.pdf

Footprint is already in library: Package_LGA:LGA-14_3x5mm_P0.8mm_LayoutBorder1x6y

Screenshot:
Screen Shot 2019-03-30 at 21 57 40


evanshultz:
Thanks for this contribution and welcome!

1. Unfortunately, that footprint doesn't quite match. And we haven't added fine-grained enough info to the footprint names to rule it out. The differences are small, but compare this datasheet with https://www.st.com/resource/en/datasheet/lsm303dlhc.pdf (from where the footprint was made). Both 0.49mm dimensions are 0.5mm in the footprint source's datasheet and the pad width is 0.813mm instead of just 0.8mm. @poeschlr Do you want to let this slide or make a new footprint? If the latter, how to name it since the existing footprint name is so high-level it fits both packages?
1. Please remove `-14` from the footprint filter.
1. You can add the `/` in pin 1's name.
1. Pins 3 and 11 should be visible and near the pin they can optionally be connected to (GND or VS).
1. Please write out pin 13's name as shown. (Pin 12 should also be the same but it's a really long name. Would you see if there's a nice way to fit it? I reckon all pins could be on the same side if that looks nice.)

nonninz:
Hi Evan, thanks!

1. Actually, now that I check on page 34 of [the datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/ADXL343.pdf) there is a recommended land pattern that I missed before, which is larger than the footprint I selected. I can create another one with the reccommended land pattern, but I'm not sure if then it should be a specific footprint for this component?
2. Ok
3. I'm not sure what do you mean. Where in `Vdd_IO` should I add a slash?
4. Ok
5. That may be tricky. I'll try to get some inspiration from the other packages :)

evanshultz:
1. The problem remains that our general footprint naming convention doesn't have enough information to distinguish between these two footprints. The option we could accept is to make the footprint non-general and put `Analog_` in front of it (resulting in `Analog_LGA-14_3x5mm_P0.8mm_LayoutBorder1x6y`) which then clearly distinguishes them.
3. You can call it `VDD_I/O`.

Let us know if you have more questions. Thanks again for taking time to enhance the KiCad library!

nonninz:
New Preview:
Screen Shot 2019-04-07 at 16 20 17

Footprint should be ready soon

nonninz:
Footprint updated and created PR: https://github.com/KiCad/kicad-footprints/pull/1533

nonninz:
@evanshultz the issues should be resolved now.
Again this is the PR for the footprint: https://github.com/KiCad/kicad-footprints/pull/1533

Unfortunately the cla tool seems not to be working for me in that repo - I tried signing it several times but it keeps the status to "pending".

evanshultz:
@nonninz 
Sorry for the delay. Let's revert the last footprint stuff, using the existing footprint, and merge it. OK?

evanshultz:
@nonninz 
Do you agree to the above? If so, we can make the edits I mentioned above and get this PR merged.

nonninz:
@evanshultz Sorry for the delay. I've reverted the package to the generic version.

evanshultz:
@nonninz 
Great! Thanks!

PR #1701, opened 2019-03-30
Add Si446x

lorenz:
Adds the Si446x series of Sub-GHz transceivers from Silicon Labs [Datasheet](https://www.silabs.com/documents/public/data-sheets/Si4464-63-61-60.pdf)
Si446x

The GPIOs are broken up because GPIO0/1 have different characteristics and are mostly used as an input (for direct RF modulation or CTS) whereas GPIO2/3 are usually used for output control.

---


lorenz:
`Footprint filter 'QFN-20*4x4mm*P0.5mm*' seems to contain pin-number, but should not!`: Yes, it should since the symbol is not complete (it has electrically non-connected pins).

evanshultz:
1. Are the GPIO connections you mention above the way you use this part? I'm not familiar with this IC myself, but it appears the GPIO assignment is totally flexible. Is this in an app note? I don't see any fixed use of GPIO in the dev kits, so if the GPIO are truly flexible they can be placed together.
1. NC pin needs to be added to the symbol. See S4.5 to S5.2 at http://kicad-pcb.org/libraries/klc/.
1. Then change the footprint filter to `QFN*1EP*4x4mm*P0.5mm*`.
1. Use the "p" and "n" nomenclature for pins 2 and 3 to follow the datasheet.
1. Pin 17 can be Input type.
1. It would be nice to space apart pins 17 and 18 to fit a crystal symbol there if it doesn't enlarge the symbol. I'm not seeing this possibility now, but mentioning it just in case.
1. Please stack pins 18 and 21.
1. I think this would be best if the symbols were separate instead of using a wildcard. Can you please make ALIASes for all parts or else just choose one? The descriptions should note the bandwidth and output power to differentiate the parts.
1. Can you smash down the symbol? It looks a bit large.

evanshultz:
1. You took this from page 24 of the datasheet, I see. As I said, I'm not familiar with this part so I'll go with you. Can you link to some implementation of this IC or modules using this IC?
2. Perhaps the wording can be improved. We only leave off pins if there are as many or more NC pins than functional pins. It's happened a couple times. For this part, with only 1 NC pin, please add it.

lorenz:
Si446x
I did some redesigning, @evanshultz what do you think of this? I did some layout experiments and it's actually not much worse for the layout if the GPIOs are all grouped and it allows for a much smaller symbol and also allows for some space to put in a crystal.

evanshultz:
Looks good to me. Thanks! If you're happy please push this and we can merge.

A couple notes:
- Please put the end of the NC pin along the symbol rectangle edge so it's unlikely to get connected accidentally.
- Make sure the stacked pins have the same name.

lorenz:
I've pushed the changes. Aliases are still an open question, I've put them in now, but the base is still Si446x. Should I change this to the 4460 or the 4464?

evanshultz:
Looks good to me. Thank you!

Regarding the ALIASes, remove the generic one and name the symbol 4660 with 3 ALIASes. Using a generic versus specific PNs is mutually-exclusive.

Lastly, remove the space after `MHz` in the description and I think this is ready.

lorenz:
Ok, I've made the Si4460 the base symbol and removed it from the aliases and removed the space in between the number and MHz. Should be all good now.

evanshultz:
Please also remove the space _after_ `MHz`. Then let's merge. :)

lorenz:
Oh, didn't even notice that space. My font renderer left maybe 4px of whitespace there. I've readded the other one since it is more consistent with other symbols.

evanshultz:
Great. Thank you!

PR #1707, opened 2019-04-02
Added the NXP PCF8525T I2C Real Time Clock

johnfwhitmore:
Sorry been a while and I've lost track of things. This is the addition of the 'T' device.

I have to check the footprint for the other two, as a cursory glance didn't find them. Creating pull request to check Travis.

Screenshot from 2019-04-02 19-36-06


swedishhat:
Hi @johnfwhitmore, I cloned your repo and ran the SchLib helper script on your modified library. It threw the following error: 
image

Your pin position offset is set to 40mil (1.016mm), which is still technically compliant to the KLC since it's less than the 50mil max:
image

However, I'd still probably recommend changing it to 20mil (0.508mm) to avoid this collision between the VSS pin and the INT1/CLK0 pin -- changing the pin position offset fixes it:
image

Also, please rename pin 7 from "INT1/CLKO" to "~INT1~/CLKO". This will place a bar over the "INT1" part showing that it is an active low signal.
image

Otherwise, looks good! :+1:

johnfwhitmore:
Thanks for your comments. It's still a bit cramped down there but might be better then increasing the size of the symbol.

Screenshot from 2019-04-06 15-42-13

johnfwhitmore:
Thanks for your comments. It's still a bit cramped down there but might be better then increasing the size of the symbol.

swedishhat:
Looks good to me. I'm new to this and don't have merge rights or anything but hopefully we can get an admin to weigh in and move it along

myfreescalewebpage:
Hi @johnfwhitmore , thanks for contributing,

First thanks to @swedishhat for the first step review :)

- [ ] Name of pin 7 should be `~INT1~/CLKOUT`
- [ ] The design of the symbol is a little bit different than the other RTC in the library. Maybe you can arrange the pin looking at DS1307Z+ example. Proposition below.

Capture

Cheers,
Joel

swedishhat:
Hi @myfreescalewebpage, not a problem. It looks like "Name of pin 7 should be ~INT1~/CLKOUT" is already fulfilled since it has the bar over the top. His commit also says "X ~INT1~/CLKO 7 400 -200 100 L 50 50 1 1 O"

johnfwhitmore:
Can I respond from email.

Going to look at that this evening so if there are comments I'll address
them later, but I think the negative logic pin is sorted.

myfreescalewebpage:
I'm not pointing the negative logic but the name itself, CLKOUT vs CLKO. The datasheet indicate CLKOUT, so you should follow this name for better understanding.

johnfwhitmore:
Oops, sorry don't know where I pulled that from. I'll change it.

myfreescalewebpage:
No worries :)

evanshultz:
Note that clock output and interrupt pins on this IC are all open drain.

myfreescalewebpage:
Good catch @evanshultz I have forgotten this point ! @johnfwhitmore can you fix INT/CLKOUT pin type too ?
Thanks,
Joel

johnfwhitmore:
Screenshot from 2019-04-10 06-47-01

johnfwhitmore:
Sorry for the delay over the last few days. Normal laptop had to go in for repair so fighting with an old 32bit device.

myfreescalewebpage:
@johnfwhitmore that's great except the symbol is not vertically centered (see travis warning). Can you fix it ? Then ready to be merged :)

johnfwhitmore:
I actually looked for that when I did it but didn't see the axis in the symbol editor. Must not have been awake.


Screenshot from 2019-04-10 23-28-37

myfreescalewebpage:
No worries, thanks this looks great! Merging :)

johnfwhitmore:
Thanks a million for your help!

myfreescalewebpage:
You're welcome :)

PR #1709, opened 2019-04-03
Added mikroBUS socket symbol

ppielorz:
Added mikroBUS socket symbol according to datasheet (http://download.mikroe.com/documents/standards/mikrobus/mikrobus-standard-specification-v200.pdf).

I didn't put GND pins to the bottom of the symbol because of original shape.
I left all of the pins as Passive because this is just a connector, not a full device.

Footprint pull request: https://github.com/KiCad/kicad-footprints/pull/1519.

image

---


myfreescalewebpage:
Hello @ppielorz thanks for contributing,

I have not made a complete review, but the following points should be fixed for the moment:
- [ ] Pin should be grouped by function, not according to their physical position
- [ ] +3.3V at the top, GND stacked at the bottom
- [ ] Pin type should be specified

Cheers,
Joel

ppielorz:
Hi Joel,
Should I group these pins even tough it's a connector?

Paweł

myfreescalewebpage:
Yes exactly, that's a KiCad policy.

myfreescalewebpage:
@ppielorz thanks, doing a full review right now:

- [ ] I think the symbol can be called simply `mikroBUS` isn't ?
- [ ] Graphic style of SCL and SCK should be Line
- [ ] You should group TX with RX to simplify the schematics, this break the rule of input on the left and output on the right but it's not a problem. All pins should be group by fucntion. For example serial port is a function, spi port is a function etc.
- [ ] RST in bottom will be better to wire it
- [ ] You should stack GND pins
- [ ] You can have the symbol more symmetrical. Example below.

Capture

Joel

ppielorz:
@myfreescalewebpage 
Referring to your remarks:
> * [ ]  I think the symbol can be called simply `mikroBUS` isn't ?
Well, in mikroBus datasheet there is a clear division between socket and addon board
> * [ ]  Graphic style of SCL and SCK should be Line
Done
> * [ ]  You should group TX with RX to simplify the schematics, this break the rule of input on the left and output on the right but it's not a problem. All pins should be group by fucntion. For example serial port is a function, spi port is a function etc.
Done. I changed a bit your proposition to get more linear pin number order but they are grouped.
> * [ ]  RST in bottom will be better to wire it
Done.
> * [ ]  You should stack GND pins
Done
> * [ ]  You can have the symbol more symmetrical. Example below.

New screenshot below:
image

myfreescalewebpage:
@ppielorz thanks for the fixes.

> Well, in mikroBus datasheet there is a clear division between socket and addon board.

Ok, in this case keep like that. Can you simply change "socket" to "Socket" for consistency with the footprint name

Also, I just realized the footprint filter is not correct, you should correct to "mikroBUS*"

Cheers,
Joel

ppielorz:
Done.

myfreescalewebpage:
@ppielorz thanks, just made a final review, notice a final bug here reported in travis: GND pins are not properly stacked. The rule is: first pin Power Input visible, other pin Passive invisible. To solve the issue on your symbol, can you set pin 16 Passive ? Thanks!

myfreescalewebpage:
Thanks, just need to wait for the footprint now :)

myfreescalewebpage:

myfreescalewebpage:
No news of the author on the footprint, indicate this is Abandoned

PR #1710, opened 2019-04-04
Added LTC7138

SchrodingersGat:
* 400mA 140V buck converter
* Symbol copied from LTC3638
* Slight pinout differences (LTC7138 has a "ANODE" pin)
* https://www.analog.com/en/products/ltc7138.html#product-overview
* https://www.analog.com/media/en/technical-documentation/data-sheets/7138f.pdf

image

image


evanshultz:
Just one comment: If there was a bit more room between SW and ANODE would a diode symbol fit better?

SchrodingersGat:
To accomplish this I think the SW pin should be moved up - should I do this for the LTC3638 symbol too?

evanshultz:
No, just this one. That would break existing designs using LTC3638 and we want to keep that stable through the v5 cycle.

SchrodingersGat:
Ok, I have moved the SW pin upwards a little

evanshultz:
Thanks!

PR #1712, opened 2019-04-05
Fix LEM LV25 doc link

hildogjr:
Just fix the datasheet link


evanshultz:
Good catch. Thank you!

PR #1713, opened 2019-04-06
Driver_LED: MPQ2483DQ: Fix thermal pad assignment

zonque:
Fix a small issue with the [recently added](https://github.com/KiCad/kicad-symbols/pull/1681) MPQ2483DQ: the thermal pad has pin
number 11, not 13.


myfreescalewebpage:
@zonque thanks for the fix, no comment to do, merging.

PR #1714, opened 2019-04-06
Added NCP1030 and NCP1031 PWM controllers

cafee:
Hello,

I've added NCP1030/NCP1031 PWM controllers symbols to switching regulators library.

Datasheet: https://www.onsemi.com/pub/Collateral/NCP1030-D.PDF

Screenshot of NCP1031DR - others symbols are similliar
ncp1030


myfreescalewebpage:
@cafee thanks for this contribution and welcome ! Before we can review, can you sign the CLA please ? That's a mandatory condition to get your contribution accepted. Thanks, Joel

evanshultz:
Welcome and thanks for making a contribution to the official KiCad library!

Here are my review notes:
1. The description doesn't differentiate the parts. At the end of the existing description, add `, 1A` or `, 2A` for the switch current depending on the part number.
1. Then after that, add the package like `, SOIC-8` or `, DFN-8`.
1. Change CT and COMP to Passive type. Let's also do this for VDRAIN also since no other pin type will really help to catch issues as it's not a logic pin.
1. The DR symbol has the wrong footprint assigned. Should be 3.9x4.9mm package.
1. On the MN symbol, pin 9 must be visible and of Passive type. I don't see that it has an internal connection in the datasheet so it should be a visible pin named EP. Correct me if I'm wrong. (I also notice that it optionally can have pullback leads which is terribly annoying.)

If you have questions or need help, please review http://kicad-pcb.org/libraries/klc/ first and then ask here.

cafee:
Thanks for feedback.
I had addressed all notes from you and prepared next commit. 
To description i had added the targeted power range (3W, 6W respectively) and footprint.

evanshultz:
Thanks!

1. Power as the differentiator is good with me. Remove ` Up to` and ` package` from the description, and also add a hyphen before the pin count for the MN version.
2. The correct DR footprint is 'SOIC' and not just 'SO'. I'll look into why these footprints are in the library.
3. The MN version probably most commonly connects the EP to ground, so please that pin at the bottom and please post a new screenshot for easy review.

myfreescalewebpage:
@cafee are you still working on this subject ?

PR #1715, opened 2019-04-06
Regulator_SwitchedCapacitor: add TPS60151DRV

zonque:
Add TPS60151, a 5V, 140mA charge pump in a WSON-6 package.

The datasheet is here:

  https://www.ti.com/lit/ds/symlink/tps60151.pdf

Some space is left between the pins on the right side so a cap can fit in
the schematics.

image


myfreescalewebpage:

- [ ] Name of the symbol should be `TPS60151DRV`to avoid issues if the manufacturer add new packages options
- [ ] VIN pin should be on left side, in front of VOUT, not on the top for this kind of symbol
- [ ] Footprint filter should be `WSON*1EP*2x2mm*P0.65mm*`
- [ ] Pin 7 should be added to the bottom of the symbol, called "EP", type Passive, NOT stacked with GND (it is not expected according to the datasheet)

Cheers,
Joel

zonque:
Thanks! All done now.

image

myfreescalewebpage:
@zonque thanks for the fixes. TPS60151 (without DRV) is still in the pull request. Can you remove it ? Then able to merge this. Joel

zonque:
You mean in the title of the PR or in the commit log message?

myfreescalewebpage:
In the git diff, look at https://github.com/KiCad/kicad-symbols/pull/1715/files

zonque:
Ah, okay. The editor does strange things sometimes when symbols are renamed. Haven't really found a pattern to report that though. Anyway, removed now with the latest commit.

myfreescalewebpage:
Thanks, yes it's sometimes strange :) Looks good now, merging.

PR #801, opened 2018-08-06
Added TUSB4020BI

jstjst:
Added symbol for TUSB4020BI USB Hub
http://www.ti.com/lit/ds/symlink/tusb4020bi.pdf
Used the TUSB4041I symbol as a template
grafik

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #1719, opened 2019-04-07
Added TUSB4020BI

jstjst:
Something with my git went wrong, so here is PR #801 resubmitted in a separated branch.
grafik

http://www.ti.com/lit/ds/symlink/tusb4020bi.pdf


myfreescalewebpage:

- [ ] The name of the part should be `TUSB4020BIPHP` to avoid issues if the manufacturer add packages options
- [ ] The footprint should be `Package_QFP:HTQFP-48-1EP_7x7mm_P0.5mm_EP3.62x3.62mm_Mask3.62x3.52mm_ThermalVias` (error on EP size)
- [ ] The description should ends with package name at the end : `Two-Ports USB 2.0 Hub, HTQFP-48`
- [ ] Name of pin 49 should be GND
- [ ] The symbol is very large for no reason, you can reduce a bit.
- [ ] RSVD pins must be moved on the outline box of the symbol
- [ ] I'm not sure of the position of the pin on the left, is it easy to wire pins 36/21/35 like that ? Have you got a sample schematic ?

Cheers,
Joel

jstjst:
- [ ] The name of the part should be TUSB4020BIPHP to avoid issues if the manufacturer add packages options
- [ ] The footprint should be `Package_QFP:HTQFP-48-1EP_7x7mm_P0.5mm_EP3.62x3.62mm_Mask3.62x3.62mm_ThermalVias` (error on EP size)
- [ ] The description should ends with package name at the end : Two-Ports USB 2.0 Hub, HTQFP-48
- [ ] Name of pin 49 should be GND
- [ ] The symbol is very large for no reason, you can reduce a bit.
- [ ] RSVD pins must be moved on the outline box of the symbol
- [ ] I'm not sure of the position of the pin on the left, is it easy to wire pins 36/21/35 like that ? Have you got a sample schematic ?

myfreescalewebpage:
Thanks @jstjst just need to wait for the footprint before merging the symbol. Joel

jstjst:
> Thanks @jstjst just need to wait for the footprint before merging the symbol. Joel
Working on that, there is a footprint for HTQFP 64, but in some dimension it seems to differs from the TI recommendation. Is there a standard source for footprint drawings used in kicad?

myfreescalewebpage:
For Texas footprints I'm not sure but standard footprints are scripted using https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO.

jstjst:
I'll take a look at this

evanshultz:
RSVD should be visible. They have some connection and the current symbol would not be able to support a test fixture or similar application. They are not intended for connection by the end user but they do in fact have a connection (so they're aren't Not Connected).

Why is the mask size specified above? If it's the same as the thermal pad it doesn't need to be given separately in the footprint name. Only if the mask and EP size differ is the mask size provided.

myfreescalewebpage:
Thanks @evanshultz for the additional comment, yes fogotten to indicate that reserved pins should be visible, type Passive, to allow some testing connections for advanced users. @jstjst can you handle this ?
I'm not a footprint expert but the comment of @evanshultz about that is good too.
Joel

jstjst:
I agree, that the RVSD could have a different electrical type.
But when I look at the TUSB4041I symbol, the RVSD pins are also invisible.
I also think, that there is really no use to have them on the schematic for a normal user.

@evanshultz Your are right with the comment about the mask size. I just got started with the footprint rules. I still didn't figured out the footprint generator and I didn't found any documentation for the scripts. So with that point there is help wanted.

myfreescalewebpage:
> I agree, that the RVSD could have a different electrical type.
But when I look at the TUSB4041I symbol, the RVSD pins are also invisible.
I also think, that there is really no use to have them on the schematic for a normal user.

This is a general KiCad library guideline to follow.

Joel

jstjst:
I hope the pin placement is fine
grafik

myfreescalewebpage:
Thanks, looks good !

myfreescalewebpage:
@jstjst any news about the footprint here ?
Thanks,
Joel

antoniovazquezblanco:
Summarizing, only footprints described at http://www.ti.com/lit/ds/symlink/tusb4020bi.pdf#page=42 are missing.

Naming should be:
`Texas_PHP_S-PQFP-G48_EP3.62x3.62mm`
`Texas_PHP_S-PQFP-G48_EP3.62x3.62mm_ThermalVias`

myfreescalewebpage:

myfreescalewebpage:

PR #1721, opened 2019-04-07
Added pololu Step-Down Voltage Regulators

jstjst:
D24V5Fx family
D24V10Fx family
D24V22Fx family
Overview at pololu: https://www.pololu.com/category/131/step-down-voltage-regulators
grafik


myfreescalewebpage:
@jstjst thanks for this contribution,

I have not made a full review, but I see:
- [ ] Description should not contain the symbol name and should finish the package name
- [ ] No footprint and footprint filter are defined

Cheers,
Joel

jstjst:
>  Description should not contain the symbol name and should finish the package name

I dont think their exists a package name that could be added
> No footprint and footprint filter are defined

Would suggest to only provide a footprint filter and leave the footprint field empty (because their are multiple footprint options). Or is it necessary to define a footprint (than I would choose the horizontal variant)?

myfreescalewebpage:
> Would suggest to only provide a footprint filter and leave the footprint field empty (because their are multiple footprint options). Or is it necessary to define a footprint (than I would choose the horizontal variant)?

The symbol is not useful without a footprint, it is required and you should submit it. Else, how is it possible to use the symbol ?

myfreescalewebpage:
Note: it is also possible to have two footprints (horizontal and vertical mounting) for some small modules.

myfreescalewebpage:
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1535

myfreescalewebpage:
Making a full review now.

D24V10F3:
- [ ] Footprint filter should be `Converter*DCDC*Pololu*D24V10Fx*`
- [ ] Reference should be `U` for this symbol

D24V22F3:
- [ ] Footprint filter should be `Converter*DCDC*Pololu*D24V22Fx*`
- [ ] Reference should be `U` for this symbol

D24V5F3:
- [ ] Footprint filter should be `Converter*DCDC*Pololu*D24V5Fx*`
- [ ] Reference should be `U` for this symbol

Need to wait for the footprint to merge this pull request.

Cheers,
Joel

jstjst:
I thought, that the reference should be `A` because it is a sub-assembly?
Or how is a sub-assembly defined?

myfreescalewebpage:
@jstjst I have already seen `U` in the library for some assembled modules, but never `A`... Have you got example where `A` is used in the library ?

jstjst:
@myfreescalewebpage I don't have an example, just interpreted the KLC wrong.
I have changed all the things mentioned in the review

myfreescalewebpage:
No news of the author on the fp, indicate this is Abandoned

PR #1723, opened 2019-04-08
UEXT connector

nerdyscout:
**_outdated description, please see below!_**

Hi,

basically this is a generic 2x5pin connector with specific signals.

https://en.wikipedia.org/wiki/UEXT

image

---


myfreescalewebpage:
@nerdyscout thanks for this contribution,

I have not made a full review, but you should follow some KiCad convention even for this kind of symbol: pins grouped by function, 3V3 at the top, GND at the bottom.

Have you got a better link for the datasheet ? A PDF file is better usually.

Name of pin 10 is `~SSEL`.

Cheers,
Joel

nerdyscout:
well this is not a actual part, it is much more a 2x5 pin connector with fix predefined signals. The AVR-ISP is an equivalent to it. The AVR-ISP does have the power signals on top/buttom therefore I agree to change it to keep it similar.
On the other hand I would like to keep it as it is so it does look like a connector. If I apply rule S4.2 all other signal pins have to be moved as well. But depending on if this part is used for a host or slave device the signal direction will change from input pins to output and vice versa.
Therefore I think it is easier to keep it as it is.  

The naming of SSEL seems to be not consistent. The datasheet below and the picture in wikipedia say SSEL, the table shows ~SSEL.  As this depends on the actual chip used behind this connector it is impossible to define now.

datasheet:
https://www.olimex.com/Products/Modules/UEXT/resources/UEXT_rev_B.pdf

myfreescalewebpage:
The actual design is not conform with KLC guidelines so if you want to integrate it you should conform to. Your PDf datasheet is clearly a better reference than wikipedia.

> depending on if this part is used for a host or slave device the signal direction will change from input pins to output and vice versa.

In this case two symbol should be defined and signal type should be clearly indicated. You can have a look to the mini-pcie socket recently merged here for reference: https://github.com/KiCad/kicad-symbols/pull/1196.

Joel

nerdyscout:
# Host
image
# Slave
image

chschlue:
Pins 5 and 6 as well as 7, 8, 9, 10 belong together logically.
Is there a particular reason you mixed I²C and SPI in your pin groups?

chschlue:
> The naming of SSEL seems to be not consistent. The datasheet below and the picture in wikipedia say SSEL, the table shows ~SSEL. As this depends on the actual chip used behind this connector it is impossible to define now.

SPI slave select is usually (universally?) active low. I'd go with ~SS.

nerdyscout:
sry my fault. I accidently swapped the both clock pins

myfreescalewebpage:
Back from vacation, just made a complete review.
- [ ] You can remove the final dot in the descriptions
- [ ] Pin offset should be 20mil (see travis log)
- [ ] Footprint filter should be "Connector*:*2x05*" (no underscore allowed, see travis log)
Then, should be ready to merge.
Joel

myfreescalewebpage:
Thanks, merging.
Joel

PR #1728, opened 2019-04-10
BQ25570: Nano power boost charger and buck converter

ObKo:
http://www.ti.com/lit/ds/symlink/bq25570.pdf

bq25570
bq25570_2

https://github.com/pointhi/kicad-footprint-generator/pull/339

https://github.com/KiCad/kicad-footprints/pull/1543


jneiva08:
hi,

The NC pins shoud be invisible and in the edge of square
The VSS should be stacked

evanshultz:

1. Page 4 of the datasheet says the two NC pins should be connected to ground. So I'd support stacking them on the VSS pins.
1. Let's use the datasheet description of `Nano Power Boost Charger and Buck Converter
for Energy Harvester Powered Applications`.
1. Then after that add on `, QFN-20` to indicate the package.
1. The footprint selected does not match the package for this part. We don't have any 3.5mm square QFN-20 footprints, but you can create it using the simple script at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON.
1. VIN_DC should be on the top.
1. If pin 4 is at the bottom of the left side the cap will have a clean path to ground. That pin can also change to Passive type.
1. VOUT should be Input type.
1. Change pins 16 and 20 to Passive type.
1. Once all the pin movement is settled down see if you can shrink the symbol. It looks overly large above and stacking the pins should help even more.

@jneiva08 
Thanks for taking a look.

ObKo:
> VOUT should be Input type.

`Power input` or `Input`? And why, it's output of buck converter provided power, or do I misunderstand `Power output` pin type?

> Once all the pin movement is settled down see if you can shrink the symbol. It looks overly large above and stacking the pins should help even more.

The main problem are resistor divider inputs on the right side (pin 7, 10-12) - all dividers must be driven by pin 8 (see example schematic). So, they should be together and near pin 8. The best I can do:

bq25570_n

> The footprint selected does not match the package for this part.

Oh, that was surprisingly. It indeed narrower that common QFN-20, I will open new pull request for footprint.

ObKo:
https://github.com/pointhi/kicad-footprint-generator/pull/339

https://github.com/KiCad/kicad-footprints/pull/1543

ObKo:
bq25570_2
IMO better keep VIN_DC on the left side, on same level as VOUT for clear VIN->VOUT path.

evanshultz:
5. The wire going to VSTOR does make it look ugly, but that connection depends on the application. Our library requirements are power pin at the top and I'd like to stick with that.
7. The purpose of VOUT, as shown in the block diagram on page 14 of the datasheet, is to provide input to the PWM controller. So it's an input. This one is unusual because the VOUT_SET pin means this isn't a dedicated feedback pin that has a resistor (or other network) from the voltage output. The same net will need a PWR_FLAG symbol to indicate it can source power. Unless another librarian has another suggestion.
9. Looking at the original symbol, my main suggestion was to shrink vertically once VSS is stacked and the NC pins are gone.

ObKo:
BQ25570

evanshultz:
Thanks! Just a couple things left in addition to the footprint:
1. Remove the space at the front of the keywords
2. Change the footprint filter to `QFN*1EP*3.5x3.5mm*P0.5mm*`.

evanshultz:
Looks good now. Thanks!

PR #1731, opened 2019-04-11
Rename existing MCP1754S models and add SOT23 variants

antoniovazquezblanco:
http://ww1.microchip.com/downloads/en/DeviceDoc/20002276C.pdf

Screenshot from 2019-04-11 16-07-37
---


evanshultz:
Let me summarize what I think is happening:
1. The SOT-89 version is being renamed to adhere to the proper convention (since only Microchip makes this part it doesn't need the package type as a suffix).
1. SOT-23 version is added.

Is that correct? If so, perhaps we should delay this or only include the SOT-23 symbols for now. 5.1.1 is coming very soon and 5.1.2 seems likely. There's little reason to break designs using the SOT-89 symbols now, I think?

And before merging anything, the `E` in the symbol name should be changed to a `x` to wildcard temp.

antoniovazquezblanco:
> Is that correct?

Yes

I will split the PR. Thanks!

PR #1732, opened 2019-04-11
PCF8523TK

johnfwhitmore:
The same symbol as for the PCF8523T but changed footprint

Screenshot from 2019-04-11 19-26-28


johnfwhitmore:
I might have gotten the HVSON*8* wrong?

evanshultz:
I'm going to assign Joel to this for obvious reasons.

But in the interim, two things:
1. Change it to `HVSON*1EP*4x4mm*P0.8mm*`. Because of the footprint filter in CvPcb we will remove the pin count and removing the EP size allows the user to choose other variations if they want to deviate from the 'standard' footprint.
1. Add a hyphen in the description before the pin count, so it's `HVSON-8`.

johnfwhitmore:
Not directly related to this Pull Request but the two files for "MCU_Microchip_SAMD" keep showing up as changed and not even git checkout clears those changes. I assume kicad has auto changed the files but can't understand why checkout doesn't just clear the changes.

Not going to commit them but hope someone does.

myfreescalewebpage:
@johnfwhitmore thanks for this contribution,

Thanks @evanshultz for the first step review :)

A single remaining comment on this symbol:
- [ ] You should add pin 9, stacked with VSS. (This is the expose pad)

Cheers,
Joel

johnfwhitmore:
Oops completely missed that -

Screenshot from 2019-04-15 18-12-54

myfreescalewebpage:

PR #1737, opened 2019-04-12
added TLV733PDBV

jstjst:
added TLV733PDBV LDO regulator
http://www.ti.com/lit/ds/symlink/tlv733p.pdf
grafik


myfreescalewebpage:

- [ ] Description should ends with package name, add `, SOT-23-5` at the end
- [ ] Description and keywords can be enhanced, see other LDO format
- [ ] It seems that the device can be an alias of "LD39015M08R" already in the library, isn't it ?

Cheers,
Joel

myfreescalewebpage:
Thanks, seems very nice, merging.

jstjst:
Thanks to you and the other librarians for the great :thumbsup:
It must be much more time consuming to check all the PR, as I thought at my first look

myfreescalewebpage:
Thanks @jstjst :)

PR #1740, opened 2019-04-13
PIC18F87K22-I/PT Addition of symbol

johnfwhitmore:
I've copied this from an existing symbol and the only significant difference is that I've stacked the multiple power pins on top of each other. I've made a number Passive and invisible but one of each visible. I think that's how I've been asked to do it before but Travis might say different.

Datasheet - http://ww1.microchip.com/downloads/en/DeviceDoc/39960d.pdf

Screenshot from 2019-04-13 19-24-41


myfreescalewebpage:
@johnfwhitmore thanks for this symbol.

I have not made a full review, but from the screenshot:
- [ ] The name of the pins should be the main function only, a second one if it is very important, but we do not add all functions now
- [ ] Positive power supply pins should not be stacked for this symbol

Cheers,
Joel

myfreescalewebpage:
@johnfwhitmore thanks, making a full review from this point:

- [ ] The name of the device should be `PIC18F87K22-xPT` (no temperature information in KiCad symbols)
- [ ] The description is a little bit long and not with the correct format. Based on other devices I propose `PIC18F nanoWatt XLP Technology MCU, 64MHz, 128KB Flash, 4KB RAM, 1KB EEPROM, 1.8-5.5V, 69 GPIO, TQFP-80`
- [ ] Keywords `Flash Based 8-Bit Microcontroller` are enough and more consistent with other devices in the library
- [ ] The symbol is very large for no reason, you should compact a bit
- [ ] Name of pin 9 should be `RG5/~MCLR`
- [ ] Pins RA6 and RA7 should be grouped with RA pins and should be Bidirectionnal
- [ ] Maybe VDDCORE/VCAP pin will be better below RJ port to simplify the connection of the capacitor
- [ ] ENVREG is Input

Cheers,
Joel

johnfwhitmore:
Screenshot from 2019-04-23 20-08-30

myfreescalewebpage:
Thanks @johnfwhitmore for all those fixes. Some few comments to terminate this:

- [ ] You can get VDD pins a little more to the left, to get them centered. AVDD can be moved to the left too. Same comment for VSS and AVSS if you increase a little bit the height of the symbol.
- [ ] RJ aligned to RE, RH aligned to RD, RG aligned to RC and RF aligned to RB can be very nice

Joel

johnfwhitmore:
Screenshot from 2019-04-25 16-34-34

johnfwhitmore:
Screenshot from 2019-04-25 17-26-24

myfreescalewebpage:
Thanks, looks better.
RJ7 is not correctly placed. At the same tiem can you move the RF pins to have RF1 in front of RB1, ..., RF7 in front of RB7 ?

johnfwhitmore:
Screenshot from 2019-04-30 17-30-12

myfreescalewebpage:

PR #1749, opened 2019-04-14
AD8015 transimpedance amplifier, in SOIC-8

aewallin:
https://www.analog.com/media/en/technical-documentation/data-sheets/AD8015.pdf

this is a transimpedance amplifier, not a pure op-amp. not sure if there is a better library where this can be placed?

AD8015

---


myfreescalewebpage:

- [ ] Your symbol is very tall, triangle symbols are usually a little bit smaller and with pin offset of 5, not 10. Moreover, I wonder if Vbyp should be on the left or on the bottom? An example to follow can be symbol THAT2181 in the library. Let me known about Vbyp position.

Maybe @evanshultz can also give his opinion on this design, he is used to review the triangle symbols :)

Cheers,
Joel

evanshultz:
I would also support having the input on the middle left side an Vbyp on the top or bottom.

myfreescalewebpage:
Yes @evanshultz probably it's better :)

aewallin:
updated symbol:
AD8015

evanshultz:
Here is an example with differential outputs in the library. It would be nice to commonalize.
image

Also, calling the power pins `+V` and `-V` has been done on all triangle symbols. It may be slightly different than the datasheet but it's still understandable and helps out with this small symbol size.

myfreescalewebpage:
@aewallin your proposition is not yet aligned with the above exampel shown by @evanshultz (which is great I think)
Take care at the outputs dimensions and positions, and Vbyp should ne just on the right of -V
Joel

aewallin:
the example-triangle mentioned above, THAT2181, had corners on a 200-mil grid, but that is too small for the pin-placements shown in LTC6362 - so now the triangle size is larger again following LTC6362.
AD8015

myfreescalewebpage:
@aewallin thanks, according to me it's great now. @evanshultz your opinion is important for me here, what do you think of the above proposition ?

evanshultz:
Looks good to me as well. Thanks all!

myfreescalewebpage:
Thanks @aewallin for this symbol and @evanshultz for your confirmation, merging :)

PR #1769, opened 2019-04-20
ADP71x2 series

jstjst:
I put some work in the symbols for the ADP7142 and ADP7182.
You can see the details in the commit messages.
https://www.analog.com/media/en/technical-documentation/data-sheets/ADP7142.pdf
https://www.analog.com/media/en/technical-documentation/data-sheets/ADP7182.pdf

grafik
grafik
grafik

We have to check if it is the right time to merge all these changes, because some of them might break existing designs (pins got moved, renaming because of typo).


jstjst:
Changed EP size, because this footprint https://github.com/KiCad/kicad-footprints/pull/1364 was created for the analog `RD-8-1` package

myfreescalewebpage:

ADP7142ARDZ:
- [ ] Footprint filter should be `SOIC*1EP*3.9x4.9mm*P1.27mm*`
- [ ] Name of pin 3 is `SENSE/ADJ`
- [ ] Pin 9 (the expose pad) should be added and stacked with GND

ADP7142AUJZ:
- [ ] Name of pin 4 is `SENSE/ADJ`

ADP7182AUJZ:
- [ ] Name of pin 4 is `ADJ`
- [ ] Can you move pin 4 100mil to the bottom so it's symmetrical with the ADP7142AUJZ ?

Aliases of ADP7182AUJZ:
- [ ] Should be a separated symbol with pin 4 name `NC`, Not Connected, invisible, moved on the border outline of the symbol.

Cheers,
Joel

PR #1744, opened 2019-04-13
modifed the Raspberry Pi symbol and added the Run header

jstjst:
modifed the Raspberry Pi symbol and added the Run header

Also the symbol name of the Raspberry Pi models may should be changed to `Raspberry_Pi` or `Raspberry_Pi_40-Pin`, because today all models have the 40 Pin header (there is a new version of the version 1, that had original 26 pins and now has also 40 pins). But I think just `Raspberry_Pi` is the way to go, since there is no symbol in the library with 26 pins

grafik

PR #1771, opened 2019-04-21
Discussion about template for RPi

jstjst:
Previous discussion see here: https://github.com/KiCad/kicad-symbols/pull/1744#issuecomment-485093620
In this PR, I want to check the template for the final symbol variants.
grafik


jstjst:
Things to do:
- [ ] Decision about designator `A` or `U`
- [ ] Check description and keywords
- [ ] Check footprint naming and filter

myfreescalewebpage:
@jstjst we do not allow "templates" in the library. I have opened an issue with the list of all RPi at #1772.
Two other points: all GND should be stacked, else this is not valid, and we were agree to not stack positive power pins, I think it's better too to allow various configurations.

jstjst:
Here is a footprint PR about the name: 
In the end the decided for `Raspberry_Pi_Zero_Socketed_THT_FaceDown_MountingHoles`

@myfreescalewebpage Forgot to mention it: It is not my intention, that this PR is merged, just wanted a Travis output and also a new topic fr discussions.

myfreescalewebpage:
Ok, added to the issue at the moment, you can update it after if required.

jstjst:
The remaining GND Pins will be stacket, but they are not needed for all symbols, so I left them aside for now.
I will revert the power pin stacking

myfreescalewebpage:
Ok, can you update it in #1772 ? For now, I will close this PR since it's not to be merged.

PR #1773, opened 2019-04-21
Add missing hysteresis symbol on 74LS14 (and 74HC14 alias)

hzeller:
Add missing hysteresis symbol on 74LS14 (and 74HC14 alias)
---

- [ x] Provide a URL to a datasheet for the symbol(s) you are contributing
(the symbol is already existing, I just added the missing hysteresis symbol)
  https://www.ti.com/lit/ds/symlink/sn74hc14.pdf

- [ x] An example screenshot image is very helpful
before:
74ls14-before

after:
74ls14-after

- [x ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
(footprint not changed)


myfreescalewebpage:
@hzeller thanks for contributing.

I have made a review of the modification. Looks great except that this symbol has multiple aliases. There are 74HCT04 (the symbol itself), 74HC14, 74HC04 and 74LS14.

I agree to have hysteresis symbol on the 74HC14 and 74LS14 but not on the two other symbols.

I suggest you to duplicate the 74HCT04, keeping the existing symbol like that, and add the hysteresis symbol on the new symbol only (74HC14 and alias 74LS14). To avoid breaking existing design, please do not make any other modifications to the symbol, just duplicate and add hysteresis.

Cheers,
Joel

hzeller:
Done as suggested. 74HCT04 (orig) and 74HC04 (alias) are now one symbol, 74HC14 (orig) and 74LS14 (alias) is the other symbol.

Side-by-side:
7404vs7414

hzeller:
Given that these are two submits, but only in conjunction we arrive at the proper state, I suggest to Squash-merge these.

myfreescalewebpage:
Thanks, no comment, just merging. Do not worry about the commit history, it's very common to have multiple commits during the reviews :)

Issue #756, opened 2018-07-14
Hidden power pins in 74xGxx library (KLC 4.7.ii)


RileyStarlight:
I have recently updated the schematics libraries and found that the new symbols for the 74xGxx family have hidden power pins. This makes it absolutely incompatible with mixed voltage environments, and from what I've been seeing, it also goes against KLC.

I'm using the latest libraries for kicad 4 repository, but the ones in this repo are identical.

thomasonw:
+1 on this issue.  Embed and per-defined and hidden forcing of power and ground to Vcc and GND is IMHO very poor form.  Precludes distributed power / ground designed and I would offer this really needs to be addressed before doing Ver 5 release.

jcyrax:
I've started work on this issue (PR:https://github.com/KiCad/kicad-symbols/pull/371), but due to time constraint its unfinished. Symbol part is basically done, but my PR needs review and maybe some further changes based on that. I have missing footprints created, but need to make PR for them.

evanshultz:
@jcyrax 
Have you pushed the symbols? And did you script them (symgen or other)?

If the footprints are already broken now, adding power pins would still be an upgrade with no drawback and potentially able to merge right away IMO.

jcyrax:
Yes, the latest changes are in my PR (https://github.com/KiCad/kicad-symbols/pull/371). They are scripted with symgen.

henryhallam:
Any movement on this? Personally I don't care whether the power pins are a separate unit symbol in a box, separate unit symbol as "floating pins", or on the first unit symbol - but the present situation with them being hidden pins is highly unsatisfactory.

kwendenarmo:
I think that each unit has its own power pins is most appropriate. And so it is in for example op-amps, both multipart and single.

poeschlr:
My vote is on having power pins as a separate unit as it will offer the flexibility to support all usecases.
I am not really sure if we can include such a change in 5.1 or if we need to wait till v6. We might want to ask the devs about that.

Regarding the pin count filter: I am not so sure that the filter is broken by having pins on all units. (They should not be counted multiple times as long as the pins are set as common to all units.)

There even is a ERC check now that reduces the chance of connecting multple nets to different duplicates of the same pin. (However there are also some bug reports indicating that it does not quite work right now.)

poeschlr:
[...]

PR #1774, opened 2019-04-21
Fix mirrored hysteresis symbol, datasheet link for 74LVC1G14, 74LVC3G14, 74LVC3G14

hzeller:
Fix mirrored hysteresis symbol, datasheet link for 74LVC1G14, 74LVC3G14, 74LVC3G14

The Schmitt-Trigger NOT gates had a mirrored hysteresis symbol, different from common way of displaying hysteresis. Also, for the single inverter, the hysteresis symbol was a little off-center.

Also fixed links to datasheets as the original links were links that resulted in 404.

(There is more wrong with these symbols (not yellow; has hidden power pins #756), but _not_ changed in this pull request for the sake of not changing things that might break existing designs.)

---

Note, this contains two commits (one for the hysteresis change, one of the datasheet update). Feel free to squash-merge.

- [ x] Provide a URL to a datasheet for the symbol(s) you are contributing
  Fixed datasheet links:
  https://www.ti.com/lit/ds/symlink/sn74lvc1g14.pdf
  https://www.ti.com/lit/ds/symlink/sn74lvc2g14.pdf
  https://www.ti.com/lit/ds/symlink/sn74lvc3g14.pdf

- [ x] An example screenshot image is very helpful

**Before:**
74g14-before

**After:**
74g14-after

(not changed footprint)


hzeller:
The Travis build is complaining, but it only complains about things that are already broken and that should probably not be changed at this point (e.g. the hidden power pins that were postponed to v6 according to the discussion in #756).

So this should be still good to go.

hzeller:
I see there is also pull request #371 which somehow got stuck in review a year ago.
Probably that should first be finished.

myfreescalewebpage:
@hzeller thanks for this contribution,

No comment about this pull request.
Thanks to point the other errors and PR, yes this is existing issues we have on the 74xx family, your contribution help to go in the right direction so I will merge right now.

Cheers,
Joel

PR #1779, opened 2019-04-22
fixing 40374, and add 74AHC/AHCT244 alias

JonathSpirit:
for the 40374 :
pin '1' is not a reset pin '~R' but a enable pin '~E'
and the 40374 is a tri-state output buffer, not a simple output buffer.
(and try to fix Travis error)

https://www.cmos4000.com/cmos/40374.html
https://www.digchip.com/datasheets/download_datasheet.php?id=369790&part-number=HEF40374BDB

(before change)

40374_b

(after change)
40374_a

For the 74HC244 :

adding 74AHC244/74AHCT244 alias.

the keyword for the 74HC244 say 'TTL' but it's 'HCMOS'.
same for the 74HCT244 the keyword say 'TTL' but it's 'HCTMOS'.

---


JonathSpirit:
A small question,
is that ok to create an CMOS alias '74HCxxx' on a TTL '74LSxxx' ? Or it must be 2 separate component ?

evanshultz:
As long as the symbols are compatible you can use an ALIAS. I think it's OK here.

Two things to do, both on 40374:
1. Make the power pin invisible. We know this is bad practice but that's the way things are for now.
2. There are 4 polylines now instead of just two. Can you revert this?

JonathSpirit:
I redo the polylines and set the power pin invisible.

evanshultz:
Thank you. I'd really prefer to see the diff only include the pin 1 name change and pin name offset change on 40374. We don't want users with this symbol in their schematics already to run into problems. Can you revert any other changes to the symbol, which might be easiest to do by hand?

JonathSpirit:
All change is reverted and i only add pin change (pin 1 '~R' to '~E' and output pin to tri-state).

evanshultz:
Thank you! The logic symbols are all bad but breaking schematics is worse.

PR #1782, opened 2019-04-23
Add LMR33630 switching regulator family in 8-pin HSOIC package

aurel32:
The ADDA, BDDA and CDDA regulators only differ by they switching
frequency.

http://www.ti.com/lit/ds/symlink/lmr33630.pdf


aurel32:
Gentle ping about that one.

The CI build fails, but I don't see how to fix that. On that chip VCC is an output for a decoupling capacitor and not a power input or output. Here is the corresponding datasheet description:

`Internal 5-V LDO output. Used as supply to internal control circuits. Do not connect to external loads. Can be used as logic supply for power-good flag. Connect a high-quality 1-μF capacitor from this pin to GND.`

chschlue:
CI errors are false positives in this case.
However, the default FP does not match TI's recommendation (mask cutout is to small).

aurel32:
@chschlue Thanks for the quick review. It happens we already have the DDA0008J footprint in the library, so I have just updated it to the correct one.

chschlue:
You're welcome, but this is far from a full review.
I was just wondering why this has an SMD EP at all and happened to notice the wrong dimensions.

aurel32:
I had to rebase my PR as the LMR33640 symbol has just been merged in PR #2528. I have updated the commit so that the LMR33630 is now an alias of the LMR33640. The LMR33640 is just an upgraded version of the LMR33630 released recently.

I have also added a commit to fix the default footprint, as it lacks thermal via like in the datasheet.

@cpresser could you please have a look at this PR as you merged the LMR33640 device recently? It has been almost 11 months.

cpresser:
@aurel32 Sorry, I was not aware that there is a conflict. I just picked #2528 from the list of open PRs since it seemed easy to finish (I am looking for low-hanging-fruits^^).

This one does not seem much different; I will assign myself.


They seem to be pin compatible:
image
image

- [x] I see you already made this PR only contain aliases which is good!
- [ ] The description should also contain the package-type as last item ` ,HSOP-8`
- [ ] I missed the above point on #2528 Would you please fix their description as well
- [ ] Why did you exchange the default footprint?
- [ ] Please delete the screenshot in the first post since its outdated and misleading

aurel32:
> @aurel32 Sorry, I was not aware that there is a conflict. I just picked #2528 from the list of open PRs since it seemed easy to finish (I am looking for low-hanging-fruits^^).
> 
> This one does not seem much different; I will assign myself.

Thanks.

> They seem to be pin compatible:
> image
> image

Yep, the LMR33640 (the latest chip released) datasheet explicity says so:
> Pin compatible with:
> –  LMR33610, LMR33620, and LMR33630 (36 V, 1 A, 2 A, or 3 A)
> –  LMR36510 and LMR36520 (65 V, 1 A, or 2 A)

>     * [x]  I see you already made this PR only contain aliases which is good!
> 
>     * [x]  The description should also contain the package-type as last item ` ,HSOP-8`

Ok,  I have just done that.
 
>     * [x]  I missed the above point on #2528 Would you please fix their description as well

Ditto.

>     * [ ]  Why did you exchange the default footprint?

I have done that in the second commit. I updated the footprint to add the thermal vias. That way it matches the recommended footprint from the datasheet. I haven't change the footprint filter so that people can deliberately not use thermal vias.
 
>     * [x]  Please delete the screenshot in the first post since its outdated and misleading

aurel32:
@cpresser That was fast, thanks a lot!

PR #1784, opened 2019-04-24
Driver_LED: Add TLC59108IPWR

zonque:
This adds a symbol for TLC59108IPWR, a 8-Bit Fm+ I2C-BusConstant-CurrentLED
Sink Driver in a TSSOP-20 package.

The datasheet is here:

  https://www.ti.com/lit/ds/symlink/tlc59108.pdf

image


myfreescalewebpage:

- [ ] Name of the device should be `TLC59108xPW`
- [ ] I don't see thermal pad pin 21 on the TSSOP version, maybe I'm wrong ?
- [ ] REXT should be Passive to avoid ERC errors
- [ ] Would say OUT pins should be Open Collector

Cheers,
Joel

zonque:
Thanks for the review. All issues addressed now.

myfreescalewebpage:
Thanks, merging.

PR #1788, opened 2019-04-26
Add MCP41HVX1

suzizecat:
Add High voltage MCP41HVx1 digital potentiometers, based upon MCP45HV stuff (I2C version)
Provides both TSSOP and QFN versions.

DS : http://ww1.microchip.com/downloads/en/DeviceDoc/20005207B.pdf

image

---


suzizecat:
I see that all violating pins are the NC ones. They are not of  "Non connected" type because they actually can be connected without harm to the chip, but are simply not linked to anything internally. (doesn't fall under KLC S4.5) 

How should I handle that ? I may have misunderstood KLC about "No Connect" type which, for me, represent a pin which should not be connected under any circumstances. Could you clear up this point for me ?

Thanks,
Julien

evanshultz:
In this case, since the symbols are rather different it would be nice to show both. Here is the QFN one:
image

- Can you please put a "x" before the "MQ" or "ST" to  indicate temp as a wildcard? This is how we do other Microchip parts.
- MCP41HV31-ST has the wrong type of serial interface in the description.
- Without looking into the details myself, I see that the QFN symbol has two rectangles and 4 polylines while the other one has one rectangle and 5 polylines. Can you clean up the graphics?
- The correct QFN footprint has 3.25mm square thermal pad. Please create a correct footprint if it doesn't exist.
- The QFN footprint filter needs a `1EP` in it to ensure the footprint selected has a thermal pad.
- Could you put the lowest number (*31) at the symbol and the higher number (*51) as the ALIAS?
- ST: Pins 8 and 9 are flipped.
- MQ: Pin 21 can be Passive type.
- MQ: Because of the many NC pins the symbol is rather wide. Would it be possible to move the pot pins up and wrap the NC pins around the bottom left corner to shrink the symbol any?

suzizecat:
I'll update this when I find time. However : 
 - could you please explain me what do you mean by _Pins 8 and 9 are flipped_ ?
- Would it be fine to put some of the NC pins of the QFN version at the top ? This way, TSSOP and QFN are interchangeable on most IOs (haven't checked is those are already interchangeable or not.)

Moreover, I will probably update the MCP45HVx stuff which most probably suffers from the same issues since those are basically the same symbols. Can I add this in this PR ?

evanshultz:
Answers:
- Look at the ST symbol in your original post and page 21 of the datasheet. Pin 8 is DNGD and pin 9 is NC. You have them flipped.
- Putting NC pins whereever they fit is OK with me, but since the datasheet says connect them to DGND or VL let's make it easy and keep them at the bottom.

Because changing existing symbols would affect schematics using those symbols, let's not change them now. Closer to v6's release would be acceptable.

myfreescalewebpage:
@suzizecat are you still working on this ? Can you sign the CLA please. Thanks

suzizecat:
@myfreescalewebpage Hi, I don't know why GitHub don't recognize this commit as coming from me (since I've already agreed to the CLA).
However, since I'm not working on it anymore, we might as well drop it. 
If I come again on this component, I will resubmit a new pull request.

myfreescalewebpage:
@suzizecat thanks for your answer, I indicate the PR is Abandoned and if you or somebody else want to continue on this it is still possible.
Joel

myfreescalewebpage:
PS : for the CLA, it is linked to your email address, maybe you have change it on your GitHub account or on your local fork of the repository ? We also sometime face this kind of issues with CLA.

PR #1789, opened 2019-04-26
Fix 8259 pinout: RD and WR were swapped

ab0tj:
The pinout for the 8259 chip had RD and WR swapped. See the datasheet here: https://pdos.csail.mit.edu/6.828/2009/readings/hardware/8259A.pdf - which is the same one referenced in the library itself.


evanshultz:
Oops. Good catch. Thanks!

PR #1590, opened 2019-03-01
Add PIC16F1786-xML

juanjorojash:
**Addition of PIC16F76-IML** to the MCU_Microchip_PIC16 library. The _PIC16F73-IML_ was used base symbol and the footprint is: _Package_DFN_QFN:QFN-28-1EP_6x6mm_P0.65mm_EP4.25x4.25mm_

**Datasheet:** http://ww1.microchip.com/downloads/en/DeviceDoc/40001637C.pdf
image

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

PR #1792, opened 2019-04-27
addPIC16(L)F1823 SOIC PDIP TSSOP

roger225:
add PIC16(L)F1823 PDIP SOIC TSSOP
- PIC16(L)F1823-I/P
- PIC16(L)F1823-I/SL
- PIC16(L)F1823-I/ST

PIC16(L)F1823-I/P
image

PIC16(L)F1823-I/SL
image

PIC16(L)F1823-I/ST
image

http://ww1.microchip.com/downloads/en/DeviceDoc/40001413E.pdf


diegoherranz:
Thanks, @roger225, for your contribution. Sorry for the delay reviewing this. I haven't had much time during the holidays period.

A few comments:
- [ ] Pin name offset should be 20mil (0.508mm). See [KLC S3.6](http://kicad-pcb.org/libraries/klc/S3.6/).
- [ ] `RA3/MCLR/VPP` should be `RA3/~MCLR~/VPP`. See [KLC S4.7](http://kicad-pcb.org/libraries/klc/S4.7/).
- [ ] `RA3/~MCLR~/VPP` is input only. See page 12 of the datasheet.
- [ ] The temperature rating should use a wildcard. E.g. `PIC16F1823-xP`. See [KLC S2.2](http://kicad-pcb.org/libraries/klc/S2.2/). Can you change them all? I know that many of the PIC symbols in the library don't follow this at the moment and that is confusing when creating new symbols.
- [ ] Can you follow the comment on https://github.com/KiCad/kicad-symbols/pull/1590#issuecomment-471752125 for the description and the keywords?
- [ ] According to [KLC S4.1](http://kicad-pcb.org/libraries/klc/S4.1/), pin length should be 100 mil (2.54mm).

> "The number of characters in the pin number determine a pin’s length. When there are two characters in the pin number then 100mil (2.54mm) long pins should be used"

- [ ] I don't think this is enforced but my personal preference is having the reference designator top left and the component name top right. See PIC18F25K83-xSP as an example. This is optional however, so I'm OK if you don't change it.
- [ ] Maybe it could be a bit smaller horizontally?

Comments which don't require any action:
- I think it was agreed that the alternative functions of microcontroller pins wouldn't be included in the library, but given how small these devices are, I think having a few of the alternative functions as you've done is very useful and doesn't make the symbol overly crowded. So I would leave it as is. In fact, this is in line with what was done in https://github.com/KiCad/kicad-symbols/pull/1590.
- Horizontal pitch for the PDIP in the datasheet is 0.310 inches (7.874mm), but the footprint chosen is W7.62mm. I guess it's fine because every other DIP footprint used in the PIC libraries is 7.62mm. It's been a long time since I used a DIP part :)
- Datasheet says that both SL and SL indicate SOIC. However, only SL is available on their website to buy. SL is the one you used, so I think it's OK.

Thanks!

diegoherranz:
Hi, @roger225 
This repo is getting locked down. See https://github.com/KiCad/kicad-symbols/blob/master/README.md
Would you be able to finish this PR before the repo moves to gitlab?
Thanks!

myfreescalewebpage:
@diegoherranz I propose the PR should be tag Abandoned :)

diegoherranz:
> @diegoherranz I propose the PR should be tag Abandoned :)

I agree. I've just seen you've tagged it already.

PR #1793, opened 2019-04-27
Corrections

dxrpl:
I've changed type (electrical properties) of 6 pins:

1. Pin 3 (FB) is output of the error aplifier (not input)
2. Pin 14 (REF) is power output of internal voltage reference regulator (not power input)
3. Pins 8, 9, 10 and 11 are not exactly open collector/emitter because it depends on configuration. Read explanation on the bottom.

I know it's minor change but it causes DRC errors potentially hard to debug for new users.

Datasheet: http://www.ti.com/lit/ds/symlink/tl494.pdf

Explanation of point 3:

The problem is that given pin is open collector/emitter depending on configuration. As in datasheet (first page):

```
The uncommitted output transistors provide either common-emitter or emitter-follower output capability.
```

This causes DRC errors no matter witch configuration you choose, because of connection of output pin straight to GND/VCC. Because there is no such thing like changing pin type depending on configuration, we have 3 options:

1. Change pin types to undefined - this solve problem with DRC errors but potentially can lead to other issues (in my opinion preferred option).
2. Create separate footprints for all 4 possible configurations (kinda sucks).
3. Leave those pins as is and hope that user is going to figure it out in reasonable time (kinda sucks but not so much as point 2.).


evanshultz:

On the first two points I agree and that's good. Thank you.

On the second, the message above is misleading. Perhaps just wording or perspective. The emitter and collector pins are truly "open" and will be configured as open collectors or open emitters, as the user chooses. Open emitter seems to be the only configuration giving a problem as my check test in 5.1.2 showed (the error is from the lower C1/C2 pins):
image

This is what you're talking about? If so, then yes, this PR is OK to merge. But note that you changed the pins to Passive type and NOT Undefined.

dxrpl:
Yes, you are right. Wording and perspective has ruined my explanation :) the thing is that both "open collector" and "open emitter" are considered signal outputs so when you connect collector lead to power output (VCC, +12V etc.) it's not "open collector output" even though it literally is one of the transistor leads (it is open collector but no longer signal output but actually power input). So when you connect signal output to power output the DRC raise an error.

To be honest I think the configuration with emitter connected straight to GND should also raise an error. Maybe I don't understand algorithm behind DRC.

And my move with passive was lame I admit :)
So changed passive to undefined.

dxrpl:
Oh no, I've just realised that if you make pin type unspecified, then DRC raise errors too. This is where the idea of passive comes from. I think the best solution is to leave pin 8, 9, 10 and 11 as is.

kicad

evanshultz:
Yes, those four pins should be Passive type. I was noting that your message said 'Undefined' type but the commit (and the correct type) were not Undefined. If you change them back to Passive I'll merge.

poeschlr:
Unspecified is kind of reserved for symbols that are not meant to be used without specializing them inside the project. I think the idea would be to use them where one really can not tell at symbol design time and wants to ensure the user thinks about it. I however fear that this might be a feature that will not really be understood by the typical user of kicad. Which is why we normally do not use that pin type.

dxrpl:
OK, so finally pins 8, 9, 10 and 11 are passive type. Many thanks for discussion. I'm going to contribute more on-the-fly every time I notice some problems.

evanshultz:
Thanks! Hopefully there aren't any more problems but surely there are some lurking.

PR #1794, opened 2019-04-28
add IR26-21C_L110_TR8

roger225:
add IR26-21C_L110_TR8 IR LED, 940 NM, 20 deg, SMD 1206 

http://www.everlight.com/file/ProductFile/IR26-21C-L110-TR8.pdf


evanshultz:
Thanks for contributing!

- Remove " package" from the end of the description.
- Please base this symbol off of the Device:LED symbol as it's the current new style. The old symbols in this library are a poor reference.
image

evanshultz:
Looks good. Thanks!

BTW, please don't force push changes since that makes it harder to review. We squash merge so it's all good.

PR #1795, opened 2019-04-29
TC4421/2 added

dxrpl:
Hi,
I added TC4421 and TC4422 ICs. Very popular mosfet/igbt drivers.

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/20001420F.pdf

kicad


evanshultz:
Thanks for contributing!

Would you please post updated screenshots? When there are significant visual changes, like there are here, this makes review much easier.

Also please stack pins, use the pin naming from the datasheet, and compress the symbol if possible.

evanshultz:
Please use more specific commit messages. "some corrections" doesn't tell us exactly what was changed and will make it harder to review. Thank you.

dxrpl:
Hi,
What do you mean by "stack pins"? The Travis CI build failed when I made pins on the sides. It said something like "VDD pins has to be on top and GND on the bottom".
I've changed pin names according to datasheet but I think compressing them is not possible. Remember that it has to be on 100 mil grid so slight shrinkage will cause Travis CI build fail.

kicad1

kicad2

evanshultz:
Stacking pins can result in something like this (I didn't run the check script but I think the pin types are right for stacking):
image

```
# TC4422
#
DEF TC4422 U 0 20 Y Y 1 F N
F0 "U" -250 250 50 H V C CNN
F1 "TC4422" 200 250 50 H V C CNN
F2 "Package_DIP:DIP-8-N6_W7.62mm" 0 -800 50 H I C CNN
F3 "" 0 0 50 H I C CNN
$FPLIST
 DIP*W7.62mm*
$ENDFPLIST
DRAW
S -300 200 300 -200 0 1 10 f
X VDD 1 0 300 100 D 50 50 1 1 W
X INPUT 2 -400 0 100 R 50 50 1 1 I
X NC 3 -300 -100 100 R 50 50 1 1 N N
X GND 4 0 -300 100 U 50 50 1 1 W
X GND 5 0 -300 100 U 50 50 1 1 P N
X OUTPUT 6 400 0 100 L 50 50 1 1 O
X OUTPUT 7 400 0 100 L 50 50 1 1 P N
X VDD 8 0 300 100 D 50 50 1 1 P N
ENDDRAW
ENDDEF
```

All pins should be Line type and not bubbles. Use an overbar for negative polarity.

evanshultz:
See the CI log:
- Change pin name offset
- Fix output pin stacking; see http://kicad-pcb.org/libraries/klc/S4.3/

Also, you can compress the symbol as I showed above.

Thanks!

evanshultz:
See the CI log:
- Change pin name offset
- Fix output pin stacking; see http://kicad-pcb.org/libraries/klc/S4.3/

Also, you can compress the symbol as I showed above.

Thanks!

dxrpl:
OK I've fixed it. Further shrinkage is going to make symbol ugly.
According to KLC I've also changed invisible output and power input pins type to passive.

Screenshot_2019-05-08_18-02-50

evanshultz:
While I prefer the smaller symbol, let's do it. Thanks for your contribution!

PR #1797, opened 2019-04-30
TMC5130-TA added

serkanxselcuk:
TMC5130-TA step motor driver added.

Datasheet: https://www.trinamic.com/fileadmin/assets/Products/ICs_Documents/TMC5130_datasheet_Rev1.15.pdf

image

Footprint: Package_QFP:TQFP-48-1EP_7x7mm_P0.5mm_EP5x5mm_ThermalVias

---


serkanxselcuk:
Exception to KLC S4.5
NC pins are to connect with GND (see Datasheet page 10)

pin | number | function
--- | --------- | -------
N.C. | 6, 31, 36 | Unused pins;connect to GND for compatibility to future versions

Should I stack NC pins?

image

poeschlr:
There seem to be two GNDs. So with which one should the NC pins be connected? With which one should the EP be connected. If there is a clear answer to that then yes you can stack them. If they are stacked then pin type not connected is wrong. 

But if there are legitimate reasons not to connect them then i would suggest to keep them separate with pin type NC as you have it now. (We need to offer symbols in the most generic way as we do not know the use case of our users. This means some specialties can not really be represented in the symbols we provide. This might just be one of these cases)

serkanxselcuk:
Hi @poeschlr , nice to write with you again.

GNDP (power GND), NC and EP should connected to normal GND of circuit (NC not because of technical needs, only for compatibility with future chip designs of manufacturer). I don't know if any user would like to ignore this, or if we can allow to ignore. Personally I would care manufacturers suggestion.

I would stack GNDP (as shown power input), NC (as passive) and EP (as passive) pins, if it is OK for you?

GNDA (analog GND) can be connected direct as well, but some of users can decide to build filter or use different level for analog pins, so it must stay as separate pin.

serkanxselcuk:
image

Stacking NC and EP pins under GNDP accured CI errors:

Checking symbol 'TMC5130-TA':
  Violating S4.3
    Rules for pin stacking
    Pinstack @ (-100,1200) have different names
-Pin GNDP (13)
-Pin NC (31)
-Pin NC (36)
-Pin GNDP (48)
-Pin EP (49)
-Pin NC (6)

  Violating S4.6
    Unused pins should be set as NOT CONNECTED and should be INVISIBLE
    NC pins are not correct pin-type:
-Pin NC (31) @ (-100,-1200) should be of type NOT CONNECTED, but is of type PASSIVE
-Pin NC (36) @ (-100,-1200) should be of type NOT CONNECTED, but is of type PASSIVE
-Pin NC (6) @ (-100,-1200) should be of type NOT CONNECTED, but is of type PASSIVE

Will we ignore these errors, if we intentionally want to stack NC pins under GND pins?

evanshultz:
EP should be a separate pin; page 12 clearly says "a GND plane" so the user can choose. Stack 6, 31, and 36 and leave them as a separate pin so the user can choose which ground they go to (since there is a ground pad it will be easy to connect them all to the same potential as the EP and thus I believe they can be stacked). You will get an error from this since it's atypical, but we can ignore this CI failure.

- I think the description could have `Standalone driver for two-phase bipolar stepper motor` changed to `Power Driver For Stepper Motors`.
- Flip the position of pins 1 and 2 (1 is always tied to ground for users, I presume).
- Do we know the type that pin 1 should be? If not, use Passive.
- Please use underscores in pin names as the datasheet does, for example pin 4.
- Put signal power pins on the left and output power ones on the right; pins 10 and 13/48 are good examples.
- Pins 23-25 should be on the left side.
- Put ref des at top left and value at top right.

Please post a new screenshot when all these changes are made. Thanks!

serkanxselcuk:
TODO List:
* [ ] EP should be a separate pin; page 12 clearly says "a GND plane" so the user can choose. Stack 6, 31, and 36 and leave them as a separate pin so the user can choose which ground they go to.
* [ ] I think the description could have `Standalone driver for two-phase bipolar stepper motor` changed to `Power Driver For Stepper Motors`.
* [ ] Flip the position of pins 1 and 2 (1 is always tied to ground for users, I presume).
* [ ] Do we know the type that pin 1 should be? If not, use Passive.
* [ ] Please use underscores in pin names as the datasheet does, for example pin 4.
* [ ] Put signal power pins on the left and output power ones on the right; pins 10 and 13/48 are good examples.
* [ ] Pins 23-25 should be on the left side.
* [ ] Put ref des at top left and value at top right.
* [ ] Please post a new screenshot when all these changes are made. Thanks!

evanshultz:
Thanks!!

My suggestion for pin 1 being Passive is that the datasheet doesn't go into detail on the pin. I assume it's a digital interface. Maybe it's really bidirectional? Passive is a safe choice to avoid ERC if this symbol is used by someone who really knows the interface and would connect to it (like a test fixture).

I would think the VS pins can be stacked and they should go to the right side of the symbol, along with the GNDP pin.

Also, the symbol looks like it can be a bit shorter but I didn't try this myself so please take a look.

serkanxselcuk:
Hi @evanshultz 
> My suggestion for pin 1 being Passive is that the datasheet doesn't go into detail on the pin. I assume it's a digital interface. Maybe it's really bidirectional? Passive is a safe choice to avoid ERC if this symbol is used by someone who really knows the interface and would connect to it (like a test fixture).

* [ ] Okay, I will change pin1 to passive.

> I would think the VS pins can be stacked and they should go to the right side of the symbol, along with the GNDP pin.

KLC S4.3.2 would not match with it, because VS and VSA pins connected to +Vm power supply signal but with decoupling capacitors.
image
(Figure 1.1, 1.2, 1.3 in datasheet)

> Also, the symbol looks like it can be a bit shorter but I didn't try this myself so please take a look.

* [ ] I try to get symbol smaller.

image

evanshultz:
- OK, no stacking of VS pins.
- The overbar on the pins names to replace the `N` was lost and that was good.
- VS should stay on top and GNDP should stay on bottom.

serkanxselcuk:
@evanshultz  thank you for review.

> * The overbar on the pins names to replace the `N` was lost and that was good.

I couldn't understand clear; do you mean it was better with over bar (simplicity) or better with `N` (represents datasheet)

* [ ] VS should stay on top and GNDP should stay on bottom.

image

evanshultz:
Better with the overbar. This is captured at http://kicad-pcb.org/libraries/klc/S4.7/.

@poeschlr 
Any problem adding this file extension at https://github.com/KiCad/kicad-symbols/pull/1797/commits/5626eb5299d66e7d3d700f56101cd404276ff6b5?

serkanxselcuk:
hi @evanshultz , 
I just wanted to remember, that changes are made.

serkanxselcuk:
@myfreescalewebpage 
Hi Joel, would you have time to review this PR? @evanshultz seems not to have time or working for more important things.

myfreescalewebpage:
@serkanxselcuk I made a small review, looks good but:

- I think this should be better to follow existing examples of TMC devices in the lib to choose the position of the pins. 5VOUT on the top, VCP, CPO, CPI on the rigth etc. Check TMC2130-TA for example.
- You are violating S5.1 (travis error), the footprint should be "Package_QFP:TQFP-48-1EP_7x7mm_P0.5mm_EP5x5mm_ThermalVias" to fix this.

Joel

serkanxselcuk:
@myfreescalewebpage 
Hi Joel,
TMC2130A-TA was not merged as I started to work on this PR. I think design compability with TMC2130A-TA is now provided. Travis error with footprint is not occurred after last changes.

Thanks for your review and useful comments.

serkanxselcuk:
TMC5130A-TA
image

TMC2130-TA (to compare)
image

PS: At pin 3, 4, 5, 7, 24, 25, 29, 30 on TMC2130-TA used underscore "_" as separator. We might change it to slash "/". Should I make quickly this change in this PR?

evanshultz:
Sorry about the delay and thanks for your patience.
1. The three NC pins should probably not be stacked because the datasheet intimates they may have functionality later and one presumes they will not be the same net. Unfortunately, 2130 was already merged and it is the same in this regard.
1. Your symbol appears wider than 2130. Can it be smaller?
1. Your symbol is also taller than 2130. Can you move just the top down to match? Maybe move pins 11/12/1 up to align the bottom?
1. The visible right-side pins are not aligned with 2130.

Pin 49 should not be stacked, I believe. The datasheet says 'Connect the exposed die pad to _a_ GND plane' but not which one. I think this should be application-specific and up to the user. And this one is a discrete pin which is great. But TMC2130 is not. Would you want to make it's thermal pad pin a discrete pin here too?

The datasheet uses underscore as a separator so we'll do that for the pin names in this symbol too.

serkanxselcuk:
image
TMCM5130A-TA

image
TMC2130-TA

image
TMC2130-LA

serkanxselcuk:
Hi @evanshultz ,

I have another PR to add TMC5160, which has 20 pins for motor side (right side). I thing it is better to move all other pins to left side for pin compability. So control signals side is not needed to change on a design, if user wants to replace motor driver. For this reason I moved VCP, CPO and CPI pins to left side, because 20 pins would not pass to right side of TMC5160.

So, TMC2130-LA, TMC2130-TA and TMC5130A-TA symbols are now pin compatible. And TMC5160 will be pin compatible as well. Casually I separated NC and EP pins on TMC2130-TA and TMC2130-LA.

If you find left side pins too close to each other, symbol should be longer in Y direction.

serkanxselcuk:
image
TMC5130A-TA

image
TMC2130-TA

image
TMC2130-LA

myfreescalewebpage:
Hello @serkanxselcuk why modification of gitignore file ? Joel

serkanxselcuk:
Hi Joel,

Git (or meld) added an .orig file in library folder, which is actually not included in libraries. I thought we can add .orig to gitignore, so git doesn't import changes on this file, even if we use "git add *" accidentally.

If there is a reason not to add it in gitignore or you don't like this, I can change it back.

myfreescalewebpage:
Get don't use orig file I think, probably you done this with an extra tool, meld probably.
We cannot add all extensions of all extra tools people are using, so my point of view is to limit gitignore to KiCad files. orig extension has no place here.

serkanxselcuk:
@myfreescalewebpage 
Hi Joel,
I deleted ".orig" from gitignore back.

serkanxselcuk:
> Hi @evanshultz ,
> 
> I have another PR to #1811 add TMC5160, which has 20 pins for motor side (right side). I thing it is better to move all other pins to left side for pin compability. So control signals side is not needed to change on a design, if user wants to replace motor driver. For this reason I moved VCP, CPO and CPI pins to left side, because 20 pins would not pass to right side of TMC5160.
> 
> So, TMC2130-LA, TMC2130-TA and TMC5130A-TA symbols are now pin compatible. And TMC5160 will be pin compatible as well. Casually I separated NC and EP pins on TMC2130-TA and TMC2130-LA.
> 
> If you find left side pins too close to each other, symbol should be longer in Y direction.

Hello @evanshultz ,
would you have time to review or give answer, if my opinion about pin compatibility with TMCM5160  is correct?

serkanxselcuk:
> Hi @evanshultz ,
> 
> I have another PR to #1811 add TMC5160, which has 20 pins for motor side (right side). I thing it is better to move all other pins to left side for pin compability. So control signals side is not needed to change on a design, if user wants to replace motor driver. For this reason I moved VCP, CPO and CPI pins to left side, because 20 pins would not pass to right side of TMC5160.
> 
> So, TMC2130-LA, TMC2130-TA and TMC5130A-TA symbols are now pin compatible. And TMC5160 will be pin compatible as well. Casually I separated NC and EP pins on TMC2130-TA and TMC2130-LA.
> 
> If you find left side pins too close to each other, symbol should be longer in Y direction.

Hi @evanshultz and @myfreescalewebpage ,
can you give a feedback about my opinion with TMC5160 pin compability? Do you find this design good, or would you like me to revert back to older pin positions of TMC2130 and TMC5130?

myfreescalewebpage:
@serkanxselcuk just realized we haven't answer you. I think pin compatibility is always good to have.

evanshultz:
@serkanxselcuk 
I like the idea of having control pins on the left and motor driver pins on the right. Go ahead and do that if you're still willing. I would also like to see the control pins spaced out a bit.

myfreescalewebpage:
@serkanxselcuk are you still alive on this subject ?

serkanxselcuk:
Hi @myfreescalewebpage , jes, I'm here.
@evanshultz, I still want all control pins on left and only motor pins on right. I will categorize left side pins and push soon.

Thank you for your feedback.

serkanxselcuk:
image

image

image

evanshultz:
The EP pin is the digital ground, so Power Input type is correct.

If 5130 is just an older version of 5130A, as it seems to me, please get rid of it.

I checked all pin types on all 3 symbols and they look good. And it appears the pin locations line up perfectly. And the assigned footprints look good. So just those two things above and this is ready. Thanks!

evanshultz:
Since the 'A' is the newest one, please keep it and not the suffix-less symbol name.

evanshultz:
Great! Thank you!

PR #1799, opened 2019-04-30
[WIP] New auto-generated symbols for Xilinx Spartan/Artix/Kintex-7 and Zynq-7000 FPGAs

ObKo:
# Intro

This pull request is part of reworking Xilinx FPGA symbols and footprints:

https://github.com/KiCad/kicad-footprints/issues/1560
https://github.com/KiCad/kicad-library-utils/pull/293
https://github.com/pointhi/kicad-footprint-generator/pull/345

Xilinx has CSV files describing FPGA pinout (https://www.xilinx.com/support/package-pinout-files.html). This PR is result of script generating symbols from those files, similar to current STM32 generator.

Some background:
* Xilinx FPGA pins are divided into **banks**
* Each bank can be powered separately (from 1.2V to 3.3V)
* Most of pins are general purpose and bidirectional
* There are some dedicated pins (like reset, JTAG, config clk, etc). All of them are placed into **bank 0**
* Each FPGA has at least 3 power rails - VCCINT, VCCAUX, VCCBRAM each comes from several pins.
* There is always huge amount of GND pins.
* Total number of I/O pins >100

Scripts generates multi-unit kicad symbol, each **KiCAD unit = FPGA Bank**

## Example
Here you can see generated symbol for **Spartan-7** FPGA:

Spartan-7

Bank 0 is on top and contains all dedicated pins and common power rails. Other two units - Bank 14 and Bank 34 with 50 I/O and power.

Part of **Artix-7** symbol:

Artix-7

Here you can see bank with GTP transceivers

**Zynq-7000** symbol with memory bank:

Zynq-7000

## Current progress:
- [x] CSV parsing and pin extraction
- [x] Pin grouping by banks and function
- [x] Symbol generation and pin layout
- [x] Electrical types
- [x] VCC* and GND pin stacking
- [ ] Footprint generation - _should be synced to current PRs in KiCAD libraries: https://github.com/KiCad/kicad-footprints/issues/1560_
- [x] Spartan-7
- [x] Artix-7
- [x] Zynq-7000
- [x] Kintex-7
- [x] Virtex-7

Library checking scripts generates S3.1 errors for some symbols because **script checks symbol origin using pins regardless of unit it belongs to**

---


evanshultz:
@ObKo 
I believe all Xilinx 7 series footprints are now merged. Would you mind updating since almost all of the CI errors are due to missing footprints? Then I think we'll be in a good position to review your work here.

Thank you!

ObKo:
I've just regenerated libraries with new footprints - script successfully found correct footprints for all parts.

Only violation is in S3.1 now, I'll try to fix checklib script for multi-unit symbols. 

This and https://github.com/KiCad/kicad-library-utils/pull/293 can be reviewed and merged now.

herostrat:
It will probably take a while to sanity check this huge contribution, and not only because of the size but because it also replaces a lot of existing footprints.
Until the S3.1 update gets checked I will start with first checks and suggestions.

---

First comparison with existing footprint, XC7A15T-CPG236
grafik

Thoughts:
 - I like the stacking of the power pins
 - The splitting in banks and functional group is sensible and better than before, except on the supply bank. Imho it would make sense to have a separate one for this. Reason are:
    - There is no logical connection between the power bank and configuration pins
    - Decoupling condensators are easier to place and organize if they are horizontal and not vertical (for this case). See:
grafik
Compared to:
grafik
I know in some cases the designer chooses to place the C separately (i.e. not directly conneted via a wire) on the page, but I think this is not as clear and nice.
 - I could not find parsing errors are wrong pin definitions
 - The footprint assignment is correct on all instances I checked (mainly the problem childs, i.e. with multiple versions in one name e.g. Xilinx_CLG484_CLG485
 - The Travis warnings are mainly centering problems. I don't know how hard this would be to fix and how important it is (I had no problem with clicking and catching symbols given their size) but I guess it would be cleaner of they are centered.

---

Please note that I am not qualified to merge this, all this above are my thoughts and could be wrong.
So before you start modifying, let's wait until someone more qualified chimes in.

I already said it but let me say it again: this is impressive and a invaluable contribution, thank you very muh   
 - 
 - Obviously the symbols are not compatible with the old ones. I don't know what the

ObKo:
> The splitting in banks and functional group is sensible and better than before, except on the supply bank. Imho it would make sense to have a separate one for this. Reason are:

I agree that power pins looks better if they are at sides in this case. But with pin stacking, there is usually only 5-20 power pins, so separating them to dedicated bank isn't necessary, IMO. As a compromise, may be we can place power pins at right side of config bank, like this:

spartan-7

same-functional parts, current one on the left, new on the right. New looks much more cleaner for me, however, it violates KLC S4.2

Biggest part (maximum number of power pins):

Virtex-7

What do you think?

> The Travis warnings are mainly centering problems. I don't know how hard this would be to fix and how important it is (I had no problem with clicking and catching symbols given their size) but I guess it would be cleaner of they are centered.

https://github.com/KiCad/kicad-library-utils/pull/299

ObKo:
Modified script for testing, here what it generates (xc7a200t-ffg1156):

xc7a200t

ObKo:
Any progress on this? 

Should I move power pins to the sides or keep them on top?

herostrat:
Sorry that there is no review for this PR.
Just to let you know I added it to the milestone for 6.

Imho we can not merge this as it would regress older used xilinx symbols.
Once we target v6 we do no longer need to be downwards compatible.

myfreescalewebpage:
@herostrat v6 is coming, do you ant to continue the review of this PR ? Thanks for the answer

herostrat:
@myfreescalewebpage I would actually like to, but I am not sure I am qualified enough for the review myself.
Would it be ok if you look over my findings and give a second opinion on them?

ObKo:
Hello.
I'm ready to improve this PR. 
What do you think about power pins?
IMO, placing it on top looks bad for big parts. 
I see two options: move them to right side of configuration bank (as on screens above) or make separate power bank and place pins on left side.

PR #1805, opened 2019-05-01
Added NUF2030XV6

SchrodingersGat:
- Alias NUF2042XV6

https://www.onsemi.com/pub/Collateral/NUF2030XV6-D.PDF

image


evanshultz:
- Since the series resistor is the same, why include it in the description?
- Please put the package in the description.
- Can the data lines indicate their usage and better match the datasheet pin names? like D1I and D2O or something? I also realize you've tried to make a compact and attractive symbol, but since the pins aren't directly connected together I would like for them to have different names.

evanshultz:
@SchrodingersGat 
Oliver, are you still around?

myfreescalewebpage:
@evanshultz I think we can tag Abandoned now

PR #1806, opened 2019-05-01
MPL115A1: Change footprint

jneiva08:
Update default footprint and filter.

Footprint: https://github.com/KiCad/kicad-footprints/pull/1570

---


myfreescalewebpage:
Hi @jneiva08 thanks for this contribution,

Just one comment, footprint should be `NXP*LGA*3x5mm*P1.25mm*` here.

Joel

jneiva08:
Hi @myfreescalewebpage , have done the change

evanshultz:
Since height is specific for this part and different from other related NXP sensors, perhaps height should be in the footprint filter also?

Note that this is potentially capricious to do during 5.1 since it affects and existing symbol but I'll take the responsibility, Joel.

myfreescalewebpage:
@evanshultz I have clearly wondered if it was to be included, but removing the height permit more user possibilities as we do for the EP size.
Footprint and symbol specific to this part only and changed at the same time, I don't think this is a problem.
Joel

evanshultz:
I don't agree with you.

We leave off the EP size on QFNs because the footprint fits the package perfectly and also because the EP size doesn't affect the 3D model (varying EP sizes or via/no-via footprints can share a 3D model).

In this case, the height of the package is known and this footprint filter allows footprints with a 3D model of an incorrect height to be selected. The footprint filter (because of the 3D model association) allows more than just footprints with varying copper and a correct 3D model.

Does that make sense? Have I overlooked something?

Now, is this a big issue? No. But I'm sharing what should be thought through before deciding interchangeability and from there what possible footprints (and 3D models) CvPcb should show to the users when the footprint filter, err, filter is enabled.

myfreescalewebpage:
I have forgotten about the 3D model, you are right, height is probably better in this case, sorry. @jneiva08 if you wish to fix the EP filter back with height, can you make a PR please ? Sorry for the annoyance.

PR #1812, opened 2019-05-03
add LT3042 voltage regulator, MSOP-10

aewallin:
https://www.analog.com/media/en/technical-documentation/data-sheets/3042fb.pdf

LT3042xMSE

slightly reduced font-size on right hand lower pins, to make everything fit.

---


evanshultz:
- Footprint filter should be `MSOP*1EP*3x3mm*P0.5mm*` to avoid considering the pin count (CvPcb can filter that) and ensure an EP footprint is used.
- Pin 3 should be Input type.
- Put pin 4 on the right side.
- Pins 5 and 7 should be on the sides (typically only power and ground on top and bottom).
- You will have to grow the symbol a bit to make things fit.

aewallin:
The datasheet has many different pin-configs for the example circuits, so it's not really clear what pins should go left and right...
LT3042xMSE

evanshultz:
I think this will work well. Thank you!

PR #1819, opened 2019-05-05
add IRM-H238T_TR2

roger225:
add IRM-H238T_TR2

image

http://www.everlight.com/file/ProductFile/IRM-H238T-TR2.pdf

footprints PR: https://github.com/KiCad/kicad-footprints/pull/1564
---


evanshultz:
- Description should have `kHz` and not `KHz`.
- You can also remove `, SMD type` from the description.
- Pin name offset should be 20mil.
- Footprint filter should be `Everlight*IRM*H238T*TR2*` (don't include hyphens and add trailing asterisk).
- Stack another ground pin, pin 4, on the existing one. See my comment in the footprint PR.
- Can power pins be on the top and bottom? I don't see why not and that's the default guideline for the library.
- The drawing is rather complex. Is there a way to simplify it?

evanshultz:
@roger225 
Are you still around and ready to finish this PR in the next few days? If so, I left a comment on the footprint PR and I have review comments above for the symbol. Thank you.

myfreescalewebpage:
@evanshultz I think we can put Abandoned tag now and see if the author come back to this later, but I'm not sure he will...
(I try actually to have a real status of Abandoned PR before the migration to GitLab and to see how many remaining alive PR we have)

evanshultz:
@myfreescalewebpage 
Sure. I also tagged the footprint with Abandoned.

Issue #1772, opened 2019-04-21
Raspberry Pi board


myfreescalewebpage:
A work a currently in progress to include symbol/footprints for Raspberry Pi board. A list with proposed naming is given below. Checked will be added when new symbols are defined.

- 26 GPIO
  - Raspberry Pi 1 A => `RaspberryPi-1A`
  - Raspberry Pi 1 B => `RaspberryPi-1B`
- 26 GPIO with `RUN, GND`
  - Raspberry Pi 1 A => `RaspberryPi-1A_Run`
  - Raspberry Pi 1 B => `RaspberryPi-1B_Run`
- 40 GPIO
  - Raspberry Pi 1 A+ => `RaspberryPi-1A+`
  - Raspberry Pi 1 B+ => `RaspberryPi-1B+`
  - Raspberry Pi 2 B => `RaspberryPi-2B`
  - Raspberry Pi 3 B => `RaspberryPi-3B`
  - Raspberry Pi 3 A+ => `RaspberryPi-3A+`
  - Raspberry Pi 3 B+ => `RaspberryPi-3B+`
  - Raspberry Pi Zero => `RaspberryPi-Zero`
  - Raspberry Pi Zero W => `RaspberryPi-ZeroW`
- 40 GPIO with `RUN, GND`
  - Raspberry Pi 1 A+ => `RaspberryPi-1A+_Run`
  - Raspberry Pi 1 B+ => `RaspberryPi-1B+_Run`
  - Raspberry Pi 2 B => `RaspberryPi-2B_Run`
  - Raspberry Pi 3 B => `RaspberryPi-3B_Run`
  - Raspberry Pi Zero => `RaspberryPi-Zero_Run`
  - Raspberry Pi Zero W => `RaspberryPi-ZeroW_Run`
- 40 GPIO with `PEN, RUN`
  - Raspberry Pi 3 A+ => `RaspberryPi-3A+_Pen_Run`
  - Raspberry Pi 3 B+ => `RaspberryPi-3B+_Pen_Run`
- 40 GPIO with `PEN, RUN` and POE
  - Raspberry Pi 3 B+ => `RaspberryPi-3B+_Pen_Run_PoE`
- 40 GPIO with `RUN, GND` and `TV, GND`
  - Raspberry Pi Zero => `RaspberryPi-Zero_Run_Tv`
  - Raspberry Pi Zero W => `RaspberryPi-ZeroW_Run_Tv`

[...]

PR #1820, opened 2019-05-05
Raspberry Pi: 40 pin GPIO with Run

jstjst:
First set of RPi symbols, related issue: #1772 
grafik


myfreescalewebpage:
@jstjst thanks for this first raspberry pi contribution following our discussion, a single comment below for all the symbols:

- [ ] The datasheet should not be a schematic, we prefer a 'real' datasheet, i.e. a documentation of the product.

Also, Have you got proposed some footprint for all of them ? I havent' seem then at the moment. You should provide a link to the footprint PR here. Thanks.

Cheers,
Joel

jstjst:
@myfreescalewebpage Sadly there is no documentation document for the Raspberry Pis available. A document like that form the Compute Modules don't exits. I dug through the documentation page at the website and found the linked schematic and a general page about the GPIOs. But I think, that the schematic is more useful than this: https://www.raspberrypi.org/documentation/usage/gpio/README.md

There is no footprint PR yet and I plan to do the symbols first. I just wanted to submit the first set, so I can see what I have to change before creating more of them.

myfreescalewebpage:
@jstjst just checked and me too haven't found better datasheet as already exist for Compute Modules. Probably it's better to keep the schematic in this case.

No other comment to do on the symbol, you can proceed with the footprints :)

myfreescalewebpage:
@jstjst any news of the footprints for the RPi symbols ? Thanks, Joel

myfreescalewebpage:
@jstjst ping! see my previous message about footprints. Thanks

PR #1822, opened 2019-05-06
Added RCLAMP3328P

SchrodingersGat:
8-way ESD clamping diode in UDFN-9 package

Datasheet: https://media.digikey.com/pdf/Data%20Sheets/Semtech%20PDFs/RCLAMP3328P.pdf

image

image

Footprint: https://github.com/KiCad/kicad-footprints/pull/1572


SchrodingersGat:
Added some other aliases based on compatible parts available on DigiKey

SchrodingersGat:
Note the only KLC error is because the footprint has not yet been merged

evanshultz:
- Remove "package" from the end of the description.
- Should the upper pins be named K1, K2, etc.?

SchrodingersGat:
Changes made as requested

evanshultz:
Thanks!

`RCLAMP3328P` lost the package name in addition to the word `package`.

evanshultz:
Looks good. Thanks!

PR #1823, opened 2019-05-07
Add Texas UCC246xx sync rect controllers

evanshultz:
http://www.ti.com/lit/ds/symlink/ucc24610.pdf
image
image

http://www.ti.com/lit/ds/symlink/ucc24612.pdf
image

http://www.ti.com/lit/ds/symlink/ucc24630.pdf
http://www.ti.com/lit/ds/symlink/ucc24636.pdf (this is an ALIAS of the above)
image

Needs footprint from https://github.com/KiCad/kicad-footprints/pull/1581.

---


myfreescalewebpage:
Hi @evanshultz 

UCC24610DRB:
- [ ] For me the package at the end of the description should be "Texas DRB0008A" instead of DFN-8
- [ ] I think the footprint filter should be `Texas*DRB0008A*` (I have seen several time the complete package name, not only the DRB letters)

Else no comment on the other symbol, just need to wait the Texas footprint before merging.

Cheers,
Joel

evanshultz:
@myfreescalewebpage 
Yeah, this isn't somewhere that's very consistent. Let me write down what I did and why, then you let me know what you think. That way we can both be on the same page.
- If the user is looking at both options in Eeschema, they may not know the suffix 'D' vs 'DRB' at the end of the symbol name. SOIC in the description will be clear, and I thought DFN-8 was much more clear than duplicating DRB. Those are rather different packaging technologies so I thought having a salient way to differentiate the symbols was nice. I don't know if this is common across the library but I do recall using simple names like this in some descriptions.
- The three letter name is the base name that TI uses. See page 32 of the datasheet. The '0008A' is something like a four-digit version and a one-letter revision. Or something like that. I can see over time these last five characters change as a drawing is revised, but the three letters captures the package. You can see this at http://www.ti.com/packaging/docs/searchtipackages.tsp?packageName=SON if you search for "drb" on the page. In the past, naming TI custom packages was a mess but you recall I talked with them and shared that the three-letter name is the unique designation from TI so I believe we should start using that for new contributions. If there is an updated document then the footprint could be updated as well and just using 'DRB' in the FP filter means the symbols are still good to go. (That being said, I named the footprint as 'DRB0008A' since it was built to that document version but perhaps I should be more generic there as well??)

evanshultz:
@poeschlr 
I'm not sure of the best approach to either item above. Would you be able to share your thoughts?

myfreescalewebpage:
@evanshultz forgotten to answer you previous comment, but :

- I was not pointing the name of the symbol, it's very nice for me. It's just about the description, and for me it's not so evident to indicate DFN when the package is DRB0008...

- The questions you have about footprints remind me my issue at https://github.com/KiCad/kicad-footprints/issues/1536 : Texas, two devices, same footprint name, but two different footprints (first one is 3x3mm second one is 4x4mm) So definitively using the name given by Texas is not enough...

If @poeschlr can also give his opinion it's very nice too !

Joel

evanshultz:
@myfreescalewebpage 
No problem. Lately I'm falling behind in reviewing PRs too...
- So am I. I'm saying 'DFN-8' when picking the symbol is probably much more obvious than 'DRB' for most users. And the user can still see the symbol name which includes the 'DRB' text as well. Here is a screenshot of the symbol chooser to show how these symbols look. Check out the first two and it seems like having 'DFN-8' vs 'SOIC-8' is most clear. But I will give in if that's what y'all want. What are your thoughts?
image
- I'm not seeing the issue here. 'DRB' is the name, and '0008A' is some revision. So if a newer revision is released, just including 'DRB' in the footprint filter means the footprint can be updated without having to touch the symbol. I think I explained this above so I'm not sure what else I should write. Am I confusing?

myfreescalewebpage:
Well, the package look like DFN but it's not DFN, so i'm not sure.... Texas have some very strange footprints for some devices which are not totally DFN, not totally QFN....

As pointed by my footprint issue, DRB00008A is not enough for me. Nothing indicate that the same name is not used for another device, but with different dimensions.

In https://github.com/KiCad/kicad-footprints/issues/1536:
- TPS63060 : package "S-PWSON-N10"
- TPS54541 : package "S-PWSON-N10"
but the two package are not the same....

evanshultz:
Joel, read the comments I already wrote in the issue you linked above. We've been over this already. The three letter designation, `DRB` in this case, **is the most unique identifier!**

myfreescalewebpage:
I trust in you, that Texas which made me confused :)
What about the fixes in my previous comment about the fp filter ? I wonder if it needs to include the revision ?

evanshultz:
Regarding the FP filter, let's say Texas updates their drawing to DRB0008B or DRB0009 (I don't know their revisioning scheme). If our footprint is then updated to match, which should be safe since it's the same package, the symbol then doesn't need to be touched. Perhaps then the footprint should just be called Texas_DRB? We then lose track of to which revision the footprint is designed. Not sure the best way to handle this.

The package is kinda wacky so DFN or QFN in the description I'm not sure about. I personally don't find 'DRB' to be helpful but I agree if there's a better way to do our library it can be gray area. You prefer just sticking with 'DRB', I believe?

myfreescalewebpage:
I don't think the footprint should be DRB only, because you have DRB00016 too (16 pins packages). I had a look to other Txes devices in the lib, ok to merge like that.
Joel

evanshultz:
Thanks for merging.

So you're saying the number ('0008' in this case) means the number of pins? http://www.ti.com/packaging/docs/searchproductbypackage.tsp?orderablePartNumber=&packageDesignator=DRB&pinCount=&results=results#resultspage shows that DRB only comes in 8 pins, as can also be seen at http://www.ti.com/packaging/docs/searchtipackages.tsp?packageName=SON. Maybe it is. I couldn't find any results Googling 'DRB00016' as shown above. Where did you see that package?

myfreescalewebpage:
I'm not saying DRB00016 is existing but in my understanding, the number was the number of pins... Maybe I'm wrong, maybe TI has another logic !

PR #1831, opened 2019-05-08
add BUF602 buffer amplifier, SOIC-8

aewallin:
- I am guessing the "I" in the part-name is temperature range, so ending is "xD"
- position of VREF output pin can be discussed, as it is usually used for an AC-coupled input pin.

http://www.ti.com/lit/ds/symlink/buf602.pdf

BUF602xD

---


evanshultz:
Add `3.9x4.9mm*P1.27mm*` to the end of the footprint filter.

I'm fine with VREF where it is and it makes sense to use it as a DC bias for circuitry on the input. But I think it will work OK for that application. Do you have other suggestions? Can the text size be increased?

aewallin:
50mil text size for VREF fits in the symbol - just.
BUF602xD

evanshultz:
Thanks!

PR #1833, opened 2019-05-09
Add CTRIM_Small to Devices

qu1ck:
This is generic small trimmer cap symbol based on C_Small and with same visuals as CTRIM just scaled down.
It goes well with C_Small and R_Small if your schematic already uses those symbols instead of full scale versions.

Image for comparison

Imgur

* [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
* [x] An example screenshot image is very helpful
* [ ] Ensure that the associated footprints match the official footprint library
* [ ] A new fitting footprint must be submitted if the library does not yet contain one.
* [ ] If there are matching footprint PRs, provide link(s) as appropriate
* [x]  Give a reason behind any intentional library convention rule violation.


myfreescalewebpage:
@qu1ck thanks for this contribution !
No comment to do, merging.
Joel

PR #1834, opened 2019-05-09
Added LM43603 voltage regulator

SchrodingersGat:
LM43603 Voltage Regulator

Datasheet: http://www.ti.com/lit/ds/symlink/lm43603.pdf

image

Footprint PR is outstanding - https://github.com/KiCad/kicad-footprints/pull/1585


myfreescalewebpage:
Hi @AUTHOR, thanks for contributing,

- [ ] Footprint filter should be `HTSSOP*1EP*4.4x5mm*P0.65mm*`
- [ ] Pins length should be 100mil
- [ ] Pin 1 should be Power Output
- [ ] Pin 17 should be stacked with PGND according to the datasheet

Cheers,
Joel

evanshultz:
Pin 1 goes directly to an inductor in the normal application of this IC. We have used Passive type in the past since this pin is not intended to directly supply power.

myfreescalewebpage:
@evanshultz on this part (not all I agree), the power clearly goes out from SW pin. There is an internal mosfet between VIN and SW.

SchrodingersGat:
Fixes made except for shortening the pins - I think at 100mil the pins are a bit too cramped...

myfreescalewebpage:
@SchrodingersGat pin length is a convention used every where in the library. Two digits in pins number => pin length 100mil.
Joel

SchrodingersGat:
Sure - updated as follows:

image

myfreescalewebpage:
Thanks.
Footprint filter still to be fixed, else just need to wait the footprint.
Joel

SchrodingersGat:
Sorry I missed that earlier comment regarding the leading `*`. Is fixed now.

myfreescalewebpage:
Indicate Abandoned because no news of the author (on this PR and others)

evanshultz:
@myfreescalewebpage 
Yes, I understand electrically the current goes out of the SW pin. But the purpose of pin types is for ERC. And there must be an inductor between the SW pin and the voltage output of the converter. If SW is Power Output, the SW will will directly function as a source and not create any errors if it is connected to Power Input pins. This is not a correct usage of this IC. Thus, making SW Passive forces the user to add a Power_Flag symbol to mark the output of the inductor as a power source if ERC errors are to be eliminated. In this way, the choice of pin type helps the user to avoid misonfiguration.

myfreescalewebpage:
@evanshultz I know.
One question maybe to have a better understanding: why the user should add the power flag (I personally never use it) ? It he SW pin is Power Output, isn't it equivalent to Passive+Powerflag ??

chschlue:
The inductor is in series with the load. A switch pin set to power output only drives the net between SW and L.
The user still gets an ERC error about the net to the right of the inductor not being driven but on the flip side doesn't get an error if it's in fact wired wrong (because a power input is directly connected to the switch pin, for example.)
Does that make sense?

Out of curiosity: If you never use power flags, what _do_ you do instead?

myfreescalewebpage:
I should try with a pratical example probably.

I never use power flags and I never used ERC tool in fact for my personal project. I have just finish a personal project on which everything is working correctly but I have still 122 ERC errors/warnings... Mainly because missing of power and no connection flags...

Is it something you always do in your projects ?

Joel

evanshultz:
Oh yeah. ERC is there to help you. It's definitely on the checklist of things to do before a peer review. (Other schematic capture tools have a much more powerful ERC engine than KiCad, though.) I wouldn't move onto layout without getting down to 0 ERC errors, or very few and an explanation for why they are there.

myfreescalewebpage:
Yes I need to improve on this, I will do, after thinking all the night probably it's better to avoid mistakes.
So for Buck regulator like this one, SW should be Passive. Will remember.

PR #1835, opened 2019-05-09
Fix 'Inductor' description.

Matir:
Fix reference to 'Capacitor' in 'Inductor' description.


evanshultz:
Welcome and thank you! Good catch on this typo.

PR #1838, opened 2019-05-10
Add Murata MGJ3 symbol

evanshultz:
https://power.murata.com/datasheet?/data/power/ncl/kdc_mgj3.pdf

image

I used with a single symbol trying to catch all variants and then struggled a bit to name the pins since the outputs can vary and the pin names on page 11 of the datasheet doesn't apply to all versions. Thoughts?

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1591

---


myfreescalewebpage:
Hello @evanshultz several comments:

- [ ] The footprint filter should be `Converter*DCDC*Murata*MGJ3*`
- [ ] How have you chosen the name of the output pins ? I refer to the table page 11 of the datasheet.

Cheers,
Joel

evanshultz:
- Fixed. Good catch and thank you.
- Please see the comments in my original post. I specifically pointed out that the pin names on page 11 don't apply for all models. Maybe it would be best to split up the symbols and follow the pin names given on page 7? If you have a better convention please share it.

myfreescalewebpage:
Well, I understand the position but I have doubt about the current pin names.... it is not very convenient according to me...

Maybe the best is to create several symbols, with the right pin names for each of them. I also notice that for all other DCDC converter we have created an alias for each voltage output, which is nice to use.

Joel

evanshultz:
@myfreescalewebpage 
I thought about re-jiggering the 5V part into all three versions with unique pinouts for each configuration, but that seemed like overkill to me. There would be nine symbols: three for each configuration and there are three input voltages. It also results in not having the exact MPN as the symbol name which is what I see in this library.

So I chose to use text to indicate the pinout in the three configurations so there will just be three symbols.

What you think of this approach?
image

I'm a little unsure if Power Output is the right pin type in this case, but I believe that is the best choice. Do you have thoughts on that?

myfreescalewebpage:
Well, this is good if a single device can be configured, but I assumed it's not the case ? In this case several symbols are required, at the moment the symbol is not clear...

evanshultz:
Each symbol is for one MPN, and it has various outputs. By using different connections the user can derive various outputs. The outputs are selected for three use cases: IGBT, SiC, or MOSFET. So each pin lists the voltages for that application and there is one NC pin for each use case which floats.

So the user will pick their use case and then connect to three of the four right-side pins.

Is that unclear? Do you have another suggestion? This still seems the simplest method to me.

I only made one symbol as an example above. Two other symbols would be required for all three MPNs and I didn't make them yet so we can discuss the above proposal.

myfreescalewebpage:
Just have a better analysis of the datasheet:
- There is three devices: `MGJ3T05150505MC`, `MGJ3T12150505MC` and `MGJ3T24150505MC`. They can be aliases, only the description will change.
- The usage MOSFET/IGBT/SIC is depending of the wiring as shown on the table page 6 of the datasheet.

=> Your proposition is good
=> You should just change the part name according to the three above order code I have given, to avoid conflict because it seems other packages can exist
=> You should add the two aliases for 12V and 24V versions
=> You have an error for description of pin 10, it is "IGBT: +15V SIC: +20V MOSFET: +15V"

evanshultz:
Excellent. Sounds like we're in agreement. Thanks for taking a look.

I fixed the mistake you noted above and added the ALIASes.

myfreescalewebpage:
:) No more comment to do on this PR, merging, thanks @evanshultz

PR #1841, opened 2019-05-10
Add Abracon ABx8x5 RTCs

evanshultz:
Designed to have interchangeable symbols.

https://abracon.com/Precisiontiming/AB08X5-RTC.PDF
image
image

https://abracon.com/Precisiontiming/AB18X5-RTC.pdf
image
image

The package drawings are identical and at https://abracon.com/Support/PackageDrawing/Precisiontiming/AB08XX.PDF and https://abracon.com/Support/PackageDrawing/Precisiontiming/AB18XX.PDF (no idea why they need to be separate...). I chose a nearly-compatible footprint submitted at https://github.com/KiCad/kicad-footprints/pull/1199. The lead width and length is slightly different, but our footprint names don't capture that info so how could we tell? In addition, the footprint used is a VQFN but the height of that part and these RTCs are all in the same range so no real difference to the 3D model.

---


myfreescalewebpage:
Hi @evanshultz, now reviewing this one :)

For the 4 symbols:

- [ ] Footprint filter: I suggest `VQFN*1EP*3x3mm*P0.5mm*` (as done for TQFP, HTSSOP...)
- [ ] XO should be Output

Cheers,
Joel

evanshultz:
Changed XO pin type. For the FP filter, I felt leaving off the `V` gave the most choices if the user doesn't keep the default footprint since the differences between QFN and VQFN and TQFN are often arbitrary. I alluded to this in the original post above, but maybe this is silly. What do you think?

myfreescalewebpage:
Well actually in the library we have both: some device have complete package name `VQFN...` and some have `*QFN...`. I have a preference for the first one which is the best to not leave the place to questions during the reviews. I think the majority of users do not play with the package and keep the default anyway...

evanshultz:
Understood. Since we're using the generator it's easy to make a footprint with any name, but some vendors call a package VQFN which is identical to a TQFN by another vendor. Quite messy. I changed to VQFN but this might be something to keep an eye on and see if we run into it frequently.

Thanks for reviewing!

myfreescalewebpage:
Yes I agree, it's messy.
I think you made an unwanted modification on AB0805, EXTI has been renamed to EXT which is wrong.
Joel

evanshultz:
Why yes, I did mess that up. Thanks for noticing. Fixed now.

myfreescalewebpage:
Thanks, merging.

PR #1842, opened 2019-05-10
Add UCC39002

evanshultz:
http://www.ti.com/lit/ds/slus495i/slus495i.pdf

image

Since there is not a hard requirement to split symbols by package in this library, I left off a default footprint and added FP filters to catch the possible footprints.

---


myfreescalewebpage:
@evanshultz why not creating a symbol for each footprint ?
Joel

evanshultz:
See the note above. Because it's not required and this suited my need while only requiring one symbol. Last I read anything, Rene allowed either style although the library was moving toward a 1:1 relationship between symbols and footprints.

myfreescalewebpage:
That's a trouble for me, sorry for that but it's not because it suit your need it is to be integrated in the library... I'm asking people to make atomic part every day and the rule also applies to the reviewers... Else, this will be difficult to ask contributors something we don't do.

evanshultz:
Both full specified and not are OK as far as I knew. I have merged similar symbols recently and perhaps I shouldn't have. Making this PR fully specified is OK for me, but let's ask since I'm not sure anymore.

@poeschlr 
https://github.com/KiCad/kicad-website/pull/407, to me, clearly says that the workflow where the footprint is chosen in CvPcb is accepted. I think you've written about this other places but I didn't find them just now. Would you like all new contributions to be fully specified or do you also allow this style (no default footprint with relevant FP filters) to be merged at this time?

poeschlr:
Yes the workflow for choosing footprints with cvpcb is an accepted one. but we limit such generic symbols to a limited subset of libraries (For a long time now. It must have been the idea of either you, jan or oliver back when i started working on the lib. I seem to remember it was a combination of you three that came to that conclusion in the first KLC discussions for KLC 2.)

The power management lib is not one of these generic libs.

Meaning make 3 symbols. One per package. (with either the part number parameters to identify them or use the clear text footprint name suffix if it is produced by many manufacturers.)

evanshultz:
@myfreescalewebpage 
This now covers just the SOIC package.

@poeschlr 
It must have been Jan and/or Oliver. I don't recall this guideline and will stick to that in the future. IMO the KLC update I referenced above does NOT make this concept clear and perhaps could use a little more polish.

myfreescalewebpage:
@evanshultz just made a review:

- [ ] Name of pin 3 is `VDD`, not VCC
- [ ] I have a doubt about the position of the CA- and CS+ pins, I'm not sure if it is not better to have them on the same side, I refer to application page 12 for exemple....

Cheers,
Joel

evanshultz:
I fixed the pin 3 name. Good catch!

This part can sense across a low-side or high-side sense element. In one case, CS+ might be best on top and in one case it might be best on the bottom, if they're on the same side. By putting them on opposite sides, either routing becomes easier with less wire crossovers. At least, that was my thinking. Compare the figure on page 12 you mentioned to the one on page 1 or page 11, and consider the CS+/CS- wiring. Maybe you see something better that I didn't?

myfreescalewebpage:
> Maybe you see something better that I didn't?

Not really I just wanted to get your opinion on this :)

No more comment t o do, merging.

PR #1846, opened 2019-05-13
System automation library

serkanxselcuk:
Library for automation schematics. I'd like to know, what do you think about this type of libraries.

These libraries are not for PCB, useful for machine design or electrical cabinets. Will we start to add automation components for this purpose?

**Please don't merge yet. First of all, I wanted to get your opinion and serve some examples, which are not ready now.**

image
image
image

---


evanshultz:
@poeschlr

myfreescalewebpage:
In my opinion it's good to have

poeschlr:
I am not yet fully sure how we should go on with these. In general kicad is not really capable of doing a good job at the detailed usecase. On the other hand these symbols might be useful to show offboard components. (KiCad as system design tool)

---

**Why is kicad not really suited to electrical cabinet design?**
Its focus on hierarchical design runs kind of counter to it. (schematics for electrical cabinets are designed in flat designs as they are intended to be printed.) Flat design really is not supported well in kicad (see https://forum.kicad.info/t/hierarchical-or-flat-schematic-design-what-is-best-for-me/16477) Making a schematic with >20 pages might become quite hard to do. There is also no easy way to define a page order (or insert a page in the middle later on) as the page number does not matter at all in a proper hierarchical design (that one is not intended to be printed anyways as it is basically unreadable on paper)

Another major missing feature are proper cross references. Not just of nets but also different symbol units (it is quite common to have the coil of a relay on a different page to its contacts. There is no way to cross reference that easily. At least not in both directions)

---

**The offboard usecase is missing features as well**
There is not really a good way to mark symbols as "do not include in pcb" which would be a requirement here. The only option we have is starting the reference with "#" but that will also remove them from the BOM. (We definitely need file format support for this to be implemented properly.) I will ask over on the mailing list for advice (sometime this evening)

The system design tool usecase might also be expanded to wanting to have multiple pcbs from a single system design drawing. That is definitely something that will not be possible anytime soon (unless i did miss something major with the new file format proposal)

serkanxselcuk:
@poeschlr thanks for detailed explanation. I see, it will be difficult to implement clear and good solution for a short time. Some users can make tricks for themselves to make limited features possible, but we can't support fully features of electrical cabinet or system design only by adding libraries.

BOM would be the most interesting tool. On system design we would need separate BOM lists for each PCB and a separate list for external parts. Alternative make a BOM list of full project as well...

How will we go on?

1. Shall we wait for more comments of other librarians/contributors?
2. Shall we leave PRs for these purpose open for future? So if SW-boys make some changes on eeschema or release new SW for, we will be ready to serve libraries.
3. Shall we close this PRs now and stop thinking of it? If this proposes are ready on eeschema or new SW anytime, we will start to develop libraries after that?

evanshultz:
@poeschlr is the head librarian so you may not need anybody else to chime in.

IMO KiCad isn't (yet?) the right tool for this and I support either the second choice if you've submitted a request for what you'd like to the devs, and the third option if you haven't.

myfreescalewebpage:
Not really agree with you guys, we have spice symbol, they are not intended for making a PCB, but only there for simulation purpose. Drawing a pure cabinet schema (without mixing with a PCB) can be a nice possibilities for some users.

evanshultz:
The KiCad dev team intentionally added circuit simulation support. As Rene detailed above, support for cabinets is not really present in the current release. Robust cabinet support is possible, however, and the originator was encouraged to request this feature of the KiCad dev team above.

myfreescalewebpage:
ok

PR #1848, opened 2019-05-13
Add AMC1100/AMC1300

evanshultz:
http://www.ti.com/lit/ds/symlink/amc1100.pdf
http://www.ti.com/lit/ds/symlink/amc1300.pdf

image

I used the pin names from the AMC1100 datasheet.

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1595

---


myfreescalewebpage:
@evanshultz nice device!

ADC1100DWV
- [ ] There is an extra space in front of the description, to be removed

AMC1300DWV
- [ ] Datasheet link should be `http://www.ti.com/lit/ds/symlink/amc1300.pdf`
- [ ] This should be a separated device because pin have not the same name...

Cheers,
Joel

evanshultz:
Fixed the extra space and datasheet link. Regarding the pin names, as these symbols are compatible I wanted to keep them the same. It's unclear to me why the datasheets use different names but they're interchangeable.

myfreescalewebpage:
Same comment at https://github.com/KiCad/kicad-symbols/pull/1842#issuecomment-491969301

evanshultz:
Would anybody prefer a triangle symbol here? I could shrink pin number text to avoid collisions.
image

@myfreescalewebpage 
I don't think the comments at that PR apply here. Maybe I'm missing something?

@poeschlr 
What is your opinion on this? Both parts have the same function and are interchangeable. We do pin name substitution on other opamp symbols, like for power and input pins. However, I do like having pin names match the datasheet for clarity.

myfreescalewebpage:
@evanshultz not bad the triangle here, I like it :) The comment was that the part should be atomic here, like on the other PR. You can have the same design so the parts remain interchangeable.

evanshultz:
The maintenance considerations of having two otherwise identical symbols is my concern, and why I made them an ALIAS. This has nothing to do with the footprint assignment and they're using a generic footprint so the use of the word 'atomic' isn't following with the most recent use since KiCon. Perhaps you can clarify?

myfreescalewebpage:
Well, we claim to the contributor for atomic part very often, even if the design of the symbol are the same.... And in this specific case, the symbol is not totally the same because of the pin names. I understand your position but I think this will be better to have to separated symbol here.

evanshultz:
@poeschlr 
Can you provide input on pin names and symbol commonality as discussed above?

evanshultz:
@poeschlr ?

myfreescalewebpage:
@evanshultz @poeschlr ping

evanshultz:
@chschlue 
You're the new big dog. Do you mind taking a minute to review the above and share your thoughts on pin naming here?

chschlue:
I'm fine with the triangle symbol and simplified pin names.

evanshultz:
@myfreescalewebpage 
Updated with a triangle symbol:
image

It turns out the one above wasn't quite even (back corners at -200 and front corner at 300). So the one in this comment doesn't quite match the one I sent earlier. It looks a little funny to me; not too bad, just a little funny. If you have some other ideas please let me know.

PR #1850, opened 2019-05-13
Rpi 40pin run pen

jstjst:
Next set of RPi symbols, related issue: #1772
grafik

PR #1851, opened 2019-05-13
Rpi 40pin 2

jstjst:
40 pin symbols (2) created from #1850 
grafik


myfreescalewebpage:
@jstjst no comment here, just need to wait the footprint.
Cheers,
Joel

PR #1852, opened 2019-05-13
Rpi 40pin 1

jstjst:
40 pin symbols (1) created from #1820
grafik


jstjst:
Reminder: be careful when creating the footprints, this one already exists: https://github.com/KiCad/kicad-footprints/pull/1066

myfreescalewebpage:
@jstjst no comment here, just need to wait the footprint, except for RaspberryPi-Zero as you indicated the footprint already exist.
Cheers,
Joel

PR #1857, opened 2019-05-16
Added 4009, 4010 Hex Buffers with separated VCC, VDD and high sink cu…

xennonflash:
Added 4009, 4010 Hex Buffers with separated VCC, VDD and high sink current capability.
4009 (inverter) and 4010 can be used to convert cmos level (Vcmos=VDD) to ttl level (Vttl=Vcc)
and invert (4009) or not (4010) the logic.
4009 and 4010 have high current sink capability.

I generated them from the 4000 lib based on 4049 and 4050 which are pin compatible but lack the VCC and VDD split feature. So I edited the file with an editor and checked with KICAD library editor if my edits were correct.
Also I checked the pinning according to the data sheets.

Cheers xennonflash

---


myfreescalewebpage:
Hi @xennonflash , thanks for contributing,

- [ ] For both symbols, pin 13 (NC) is missing, should be added to the symbol, Passive, Invisible, on the border outline of the symbol
- [ ] For both symbols, the keywords are too much, just write "cmos" like the other 4xxx is enought

Cheers,
Joel

evanshultz:
@xennonflash 
Please take a look at the auto-generated notes in the first post above. One of the requirements is to post an image of the symbols. Could you please do that?

Having hidden power pins is a problem across this library and should be addressed for the next major KiCad release. I would like to maintain the style in this library even if it's bad at this time.

myfreescalewebpage:
@evanshultz do you suggest I should merge right now without the NC pin ?

evanshultz:
I'm only suggesting that we stick with the style in the library already for consistency, even if the current style is bad. I'm not sure how this particular symbol could work with hidden power pins, since there are two positive supplies for I and O, but in general I wouldn't deviate from the current style since v6 is going to take a while.

@xennonflash 
In addition to an image of the symbols, you're requested to provide the datasheet link.

Thanks for contributing and welcome!

myfreescalewebpage:
Ok, merging now :) The library should be unified later.

PR #1863, opened 2019-05-16
Small fix in ublox SAM

jneiva08:
Removed "_N" from pin name, already have "~" in the beginning.

Probably should renaming but probably only for V6

---


evanshultz:
This is fine to do now. It won't affect designs because no pins or symbol size are changing.

Thanks!

jneiva08:
Thanks.

When i was saying is to change from ublox_SAM-M8Q to SAM-M8Q, to be identical to the other GPS from ublox

PR #1616, opened 2019-03-08
Added Si7141DP

0xdec:
Datasheet: https://www.vishay.com/docs/65596/si7141dp.pdf

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #1893, opened 2019-05-29
added NHD display

rs-development:
[datasheet](https://www.mouser.de/datasheet/2/291/NHD-C0216AZ-FN-GBW-18835.pdf)

Pin 2 is described as power input but is connected to a potentiometer in the reference circuit, so i made it pin type passiv.

image

not done yet, work in progress


rs-development:
waiting for footprint [#1616](https://github.com/KiCad/kicad-footprints/pull/1616)

serkanxselcuk:
Hi @rs-development ,

thank you for contributing. I am not official reviewer by KiCad library team, i can only give some suggestions.

Pin connection points must be outside of symbol on [KLC S3.5](http://kicad-pcb.org/libraries/klc/S3.5/). Do you have a special reason to put right side pins (2, 4, 5, 6) in symbol rectangle?

I would prefer datasheet link on manufacturer's official webpage in datasheet field (http://www.newhavendisplay.com/specs/NHD-C0216AZ-FN-GBW.pdf)

myfreescalewebpage:
Hello @rs-development thanks for this contribution,

Thanks also to @serkanxselcuk for the first steps review, your comments are right and @rs-development you can implement them.

The control pins on the right are a little bit strange. Can you make something like the `NHD-0420H1Z` symbol which is already in the library, to have both devices consistent (single difference, your device has no back-light so just forget A, K and VO, but it is nice to have the same design!)

Cheers,
Joel

rs-development:
i changed the footprint to match the existing, and i also changed the existing display footprint.

In my opinion VO should be passiv since it is recommended to connect it with a voltage divider to vdd, so it should be passive, otherwise there are erc errors

myfreescalewebpage:
@rs-development thanks for the fixes. Some few comments:

- [ ] The footprint and footprint filter should be set
- [ ] Name of pins 7-14 are D0-D7, not DB0-DB7
- [ ] VO position to be fixed, see below

Capture

Cheers,
Joel

myfreescalewebpage:

PR #1894, opened 2019-05-30
Is31fl3733

fauxpark:
An I2C 12x16 LED matrix driver, with several individual breathing modes.

sym

Datasheet: http://www.issi.com/WW/pdf/IS31FL3733.pdf

---


myfreescalewebpage:
Hi @fauxpark , thanks for contributing,

- [ ] Devices should be atomic, this mean you should create `IS31FL3733-QFLS4` and `IS31FL3733-TQLS4`, and set footprint for each of them, and fix description
- [ ] Positive power pins should NOT be stacked
- [ ] INTB type should be Open Collector
- [ ] RSET type should be Passive to avoid ERC errors
- [ ] Pin 49 (Expose Pad) should be added and stacked with GND
- [ ] I have doubts about the position of the SW pins, something like the figure 1 in the datasheet seems better to wire the LEDs...

Cheers,
Joel

fauxpark:
> Devices should be atomic

Done, but I left off the `LS4` since it is common to both.

> Positive power pins should NOT be stacked
  INTB type should be Open Collector
  RSET type should be Passive to avoid ERC errors

Done.

> Pin 49 (Expose Pad) should be added and stacked with GND

Whoops, done.

> I have doubts about the position of the SW pins, something like the figure 1 in the datasheet seems better to wire the LEDs...

Done, there is a bit more wasted space on the left side now but is probably much better for wiring, yeah.

sym

myfreescalewebpage:
Thanks, no comment to do, merging.

PR #1895, opened 2019-05-30
MCU_Microchip_SAMD: Fix line endings to Unix-style

diegoherranz:
Not much more to say.
Dealing with git is a bit painful with these 2 files with the wrong line endings.

Thanks.


evanshultz:
A quick review looks like it fits the description above. Thanks for catching this.

PR #1900, opened 2019-06-02
Switch: SW_DPST_x2: make every pin passive

diegoherranz:
Unit B pins were defined as "input" rather than "passive".
Not much more to say.

Thanks!


myfreescalewebpage:
@diegoherranz no comment to do on this PR, thanks, merging.
Joel

PR #1907, opened 2019-06-06
Added diode MBRA340 as aliases to the existing B120-E3 symbol

Trubochkin:
- Added diode MBRA340 as aliases to the existing B120-E3 symbol

- Datasheet: https://www.onsemi.com/pub/Collateral/MBRA340T3-D.PDF

- Fotprint: https://github.com/KiCad/kicad-footprints/blob/master/Diode_SMD.pretty/D_SMA.kicad_mod

- Screenshot: 
SharedScreenshot

---


evanshultz:
We don't accept automotive-qualified parts, so can you please remove the NRV* prefix symbol? What is unique about the SBRA part?

I believe whatever diodes you end up adding can be an ALIAS to the existing `B120-E3` symbol. Please confirm and make that change if you agree.

poeschlr:
Clarification of:
> We don't accept automotive-qualified parts, so can you please remove the NRV* prefix symbol?

It is not that we do not accept symbols for such parts. It is that we do not include this qualifier in the part name. (It is a non functional qualifier. To avoid ballooning the lib such qualifiers are left out if they are at the end of the part number or replaced with wildcard "x" if in the middle.)

Trubochkin:
> We don't accept automotive-qualified parts, so can you please remove the NRV* prefix symbol? What is unique about the SBRA part?
> 
> I believe whatever diodes you end up adding can be an ALIAS to the existing `B120-E3` symbol. Please confirm and make that change if you agree.

Done

evanshultz:
This is good now. Thanks!

@poeschlr 
Appreciate the extra clarity where I didn't do a great job of explaining.

PR #1909, opened 2019-06-07
Add symbol for 4021 shift register

dogtopus:
Datasheet (TI): https://www.ti.com/lit/ds/symlink/cd4021b.pdf
Datasheet (Nexperia): https://assets.nexperia.com/documents/data-sheet/HEF4021B.pdf

Somehow this is missing. Add it as somebody else and I might need it.

---


myfreescalewebpage:
Hi @dogtopus , thanks for contributing,

- [ ] Footprint filter should be only `SOIC*3.9x9.9mm*P1.27mm*` and `TSSOP*4.4x5mm*P0.65mm*` except if you can provide some other datasheet indicating the other package versions exist for this device
- [ ] Pin 10 graphic style should be "Line"

Cheers,
Joel

dogtopus:
> Footprint filter should be only SOIC*3.9x9.9mm*P1.27mm* and TSSOP*4.4x5mm*P0.65mm*

That's true for Nexperia. However TI provides them in `DIP*W7.62mm*` and `SOIC*5.3x10.2mm*P1.27mm*` form factor as well. (See CD4021BE and CD4021BN respectively)

However I believe that `SOIC*7.5x10.3mm*P1.27mm*` was indeed added by mistake and I will remove it soon.

Also do I need to include multiple datasheets in case that different manufactures produces the same part of different form factors? If so, how?

> Pin 10 graphic style should be "Line"

This is the clock pin of the shift register so AFAIK it should have the clock graphic style. However I checked the other 4xxx symbols with the clock pin. Some of them had the clock pin set as clock, some don't. So my question is: is there any rule for when a clock pin does not need to be set as clock? Or it is just simply a historical thing and all new 4xxx symbols should not have the clock pins set as clock?

myfreescalewebpage:
> That's true for Nexperia. However TI provides them in DIP*W7.62mm* and SOIC*5.3x10.2mm*P1.27mm* form factor as well. (See CD4021BE and CD4021BN respectively)

Great!

> However I believe that SOIC*7.5x10.3mm*P1.27mm* was indeed added by mistake and I will remove it soon.

Yes please, thanks!

> Also do I need to include multiple datasheets in case that different manufactures produces the same part of different form factors? If so, how?

It's not possible, just keep the current one.

> This is the clock pin of the shift register so AFAIK it should have the clock graphic style.

The KLC now indicate we do no more use other styles, we use only Line.

Joel

dogtopus:
Resolved

myfreescalewebpage:
Thanks, merging.

PR #1911, opened 2019-06-09
Added symbol for STM8S001J3.

arnmulligan:
I reused the Package_SO:SOIC-8_3.9x4.9mm_P1.27mm footprint since it seems to be equivalent to the SO8N as specified in the datasheet. Only difference is the length of the pads is a fraction more generous than the datasheet recommendation.
---

- [x ] Provide a URL to a datasheet for the symbol(s) you are contributing
https://www.st.com/resource/en/datasheet/stm8s001j3.pdf
- [x ] An example screenshot image is very helpful
STM8S001J3_symbol
- [x ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
Existing footprint reused.
- [ x] If there are matching footprint PRs, provide link(s) as appropriate
No Pull Requests that I could find.
?
- [ x] Give a reason behind any intentional library convention rule violation.
Nothing intentional.


myfreescalewebpage:
Hi @arnmulligan , thanks for contributing,

- [ ] Name of the device should be `STM8S001J3M` top avoid conflict if the manufacturer adds packages options
- [ ] Package format at the end of the description is wrong, should be `16MHz, 8K Flash, 1K RAM, 128 EEPROM, SO-8`
- [ ] SO is SO and SOIC is SOIC. We should not make some confusion so you should use SO package. If it does not exist, you should submit it with the SO generator at https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO
- [ ] Pin length should 100mil (symbol < 100pins)
- [ ] The size of the symbol can be reduced. I wonder if the first pin name is not enough for this symbol...

Cheers,
Joel

arnmulligan:
Review comments hopefully correctly addressed.
I updated my version of KiCad and found a more appropriate SO footprint (pretty sure it wasn't there before). Again, only a slightly longer pad length than the datasheet recommendation.
Pins 5&6 grouped to the left of the symbol, since their primary function is I2C.

STM8S001J3_symbol

myfreescalewebpage:
@arnmulligan , thanks for the update, great job!

- [ ] I still not agree with the package, the datasheet indicates it is "4.9x6mm", it is not possible to use the selected footprint.

Joel

arnmulligan:
@myfreescalewebpage :

>     * [ ]  Also, pin offset should be 20mil (see travis warning)

I thought I already fixed the Travis error? Currently I only see "The Travis CI build passed ".

>     * [ ]  I still not agree with the package, the datasheet indicates it is "4.9x6mm", it is not possible to use the selected footprint

I am not infallible, but this is my reasoning:
Let's take a look at some other KiCAD symbols and footprints to compare, D & E are the overall package dimensions including pins.

################
STM8S003K3T
Datasheet:
10.1 LQFP32 package information
Figure 44. LQFP32 - 32-pin, 7 x 7 mm low-profile quad flat package outline
D (typ) = 9.000mm
E (typ) = 9.000mm
KiCAD:
Package_QFP:LQFP-32_7x7mm_P0.8mm

STM8L101F2P
Datasheet:
10.5 TSSOP20 package information
Figure 49. TSSOP20 - 20-lead thin shrink small package outline
D (typ) = 6.500mm
E (typ) = 6.400mm
KiCAD:
Package_SO:TSSOP-20_4.4x6.5mm_P0.65mm

STM32L151CCTx
Datasheet:
7.3 STM32L151xC STM32L152xC LQFP48, 7 x 7 mm, 48-pin low-profile quad flat package information
Figure 38. LQFP48, 7 x 7 mm, 48-pin low-profile quad flat package outline
D (typ) = 9.000mm
E (typ) = 9.000mm
KiCAD:
Package_QFP:LQFP-48_7x7mm_P0.5mm

And now my suggestion: 
STM8S001J3M
Datasheet:
10.1 SO8N package information
Figure 37. SO8N – 8-lead, 4.9 x 6 mm, plastic small outline, 150 mils body width, package outline
D (typ) = 4.900mm
E (typ) = 6.000mm
KiCAD:
Package_SO:SO-8_3.9x4.9mm_P1.27mm
################

So you can see that when the ST datasheet specifies the package size it sometimes/often/usually uses the size of the plastic body excluding the pins. KiCAD seems to use the size of the plastic body excluding the pins (correct?).

In the case of the STM8S001J3M, the datasheet specifies the package size as the overall size INCLUDING pins. I feel, to maintain consistency, the KiCAD size should still be just the size of the plastic body.

So how do I proceed?

myfreescalewebpage:
> KiCAD seems to use the size of the plastic body excluding the pins

Correct.

The datasheet also includes the size of the plastic body only:

Capture

They are D and E1.

Joel

arnmulligan:
>     * [ ]  Also, pin offset should be 20mil (see travis warning)

Fixed. No more Travis warnings.

> > KiCAD seems to use the size of the plastic body excluding the pins
> 
> Correct.
> 
> The datasheet also includes the size of the plastic body only:
> They are D and E1.

Yes, D=4.9 E1=3.9, which is what I am already using in the footprint name, so it's correct?

myfreescalewebpage:
> Yes, D=4.9 E1=3.9, which is what I am already using in the footprint name, so it's correct?

Sorry, yes it is ! I'm back from vacation and now no more tired, double checked and yes you are right. The datasheet is confusing, it indicates 4.9*6mm on page 77, but it is 3.9*4.9 !

Merging.
Joel

PR #1912, opened 2019-06-10
Add IS31FL3218

fauxpark:
Another LED matrix driver by ISSI.

sym

Datsheet: http://www.issi.com/WW/pdf/31FL3218.pdf
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1640
Script PR: https://github.com/pointhi/kicad-footprint-generator/pull/374

---


evanshultz:
Should the body size and pinout be compatible with the existing 3216 symbol? I've not used either part so I don't know if that's something to consider or not. If not, can the symbol be a bit smaller? It looks like empty space at the bottom. If not, then everything looks good and I'll merge.

FYI, review is a bit easier if all symbols are in screenshots and not just one. :)

evanshultz:
Can you capitalize words in the description? That seems to be the convention.

fauxpark:
> Should the body size and pinout be compatible with the existing 3216 symbol? I've not used either part so I don't know if that's something to consider or not.

The pinouts of all these ISSI drivers seem wildly different between them. This part has two extra OUT pins, does not have the GPIO capability on OUT9-OUT16, nor does it have the audio input pin, so I'd say it is not at all compatible.

> If not, can the symbol be a bit smaller? It looks like empty space at the bottom. If not, then everything looks good and I'll merge.

Done.

> FYI, review is a bit easier if all symbols are in screenshots and not just one. :)

I figured it would imply the only difference between them is the extra pin on the QFN symbol for the EP, but fair enough. Sorry.

evanshultz:
All good. Thanks!

PR #1913, opened 2019-06-11
Add PCA9635

fauxpark:
16-bit Fm+ I2C-bus LED driver, TSSOP-28.

sym

Datasheet: https://www.nxp.com/docs/en/data-sheet/PCA9635.pdf

---


evanshultz:
1. Can you capitalize words in the description? That seems to be the convention.
1. Can you shrink it a bit? There's room at the bottom, and the off-center rule is weak and intended to catch gross errors; slight off-centering if the symbol looks better is OK.

fauxpark:
>     1. Can you capitalize words in the description? That seems to be the convention.

Not sure what you're referring to here. The description being in sentence case seems reasonable to me. If you meant the keywords, I don't see how that matters as the search should be case insensitive.

>     2. Can you shrink it a bit? There's room at the bottom, and the off-center rule is weak and intended to catch gross errors; slight off-centering if the symbol looks better is OK.

Done.

evanshultz:
While this library is fairly mixed, most DCM files have all words in the description capitalized. Not sure when or why this was done. Not required so I'll merge now.

PR #1914, opened 2019-06-11
Improve IS31FL3216/A

fauxpark:
Fixing up the IS31FL3216 LED controller.

Before:
sym-1
After:
sym-2

 - Stacked GND pins, changed `1EP` pin to `GND`
 - Flipped `OUT9` to `OUT16` pin names so `OUTx` is along the body
 - Tweaked description and keywords
 - Shrunk and centered symbol
 - Corrected pin name offset

The `CLK` pin is used for synchronising multiple controllers, so I left it as bidirectional instead of input as Travis suggests: http://www.issi.com/WW/pdf/IS31FL3216.pdf#page=2
I also did not swap `SCL` and `SDA` as I figured that would be too disruptive.

---


evanshultz:
The symbol shrinking/moving will break existing schematics, and we don't want to do that during v5. Can you revert that part of it?

Agreed that CLK should be bidirectional.

fauxpark:
Done.

evanshultz:
Thank you!

With that done and the diff much less noisy, I took a look at the changes. This is mergeable now, but let me propose one thing: Remove temp from the description. KLC would wildcard this so it can be considered extraneous in this context.

fauxpark:
Yeah, I did think that shouldn't have been there but left it alone just in case. Removed.

evanshultz:
Great. Thanks!

PR #1921, opened 2019-06-12
74HC(T)164

mpkopec:
Added 8-bit serial-in parallel-out shift register.

image

---


mpkopec:
The symbol is off-center, since its asymmetry.

rjj100:
Thank you! Would be nice to have a 74HC165 which is similar and not in the library.

evanshultz:
@mpkopec 
A couple of quick things:
1. Please always include a datasheet. For consistency of pin names, please select from the most common vendor in the library already (as best you can).
1. I can't read the pin types above, but I'm sure they shouldn't all be the same for left and right pins.
1. Pin 8 shouldn't use the clock pin style (all pins in the lib are normal style).

mpkopec:
@evanshultz 
1. There is a datasheet. Checked other 2 or 3 symbols for shift registers and I can't find any major difference from my symbol.
2. In Polish "Wejście" is "Input" and "Wyjście" is "Output", so everything is ok, however, for a foreigner. it might be confusing. :)
3. Done.

Thanks a lot for the input and sorry for the delay.

evanshultz:
Thanks!

I meant please include the datasheet in the PR body, but somehow I didn't write that. Sorry.

PR #1924, opened 2019-06-14
Add AK7742EQ

evanshultz:
https://www.akm.com/akm/en/file/datasheet/AK7742EQ.pdf
image

The analog input and output pins were somewhat obvious in placement, and since I put the digital power/ground on the right side that meant the clock and I2C config pins should be on the right side also so it can get to the digital power or ground.

I then continued with the I2C interface and reset on the right side, thinking they would come from a one part. With nothing on the left side, I packed in the I2S pins which likely go to another part. And then I filled in the oscillator pins where there was the most room.

Maybe this isn't the best pin arrangement so let me know if you see something better.

---


diegoherranz:
A few comments:
- [ ] I think that the GND pins should be stacked
- [ ] The analog outputs should be outputs instead of inputs
- [ ] I think `LFLT` should better be output as defined in the datasheet.  Same for `AVDRV` and `VCOM`. Any reason to need passive?
- [ ] I think that `TEST1` should better be input as in the datasheet to avoid leaving it open without the DRC complaining.
- [ ] There is an exposed pad so please add a pin 49. Do not stack this pin since "The exposed pad must be open or connected to the ground."
- [ ] The footprint (and filter) will need to have an exposed pad.  `LQFP-48-1EP_7x7mm_P0.5mm_EP3.6x3.6mm` may be good enough? What do you think? Otherwise, a new footprint is needed.
- Because of [S4.7](http://kicad-pcb.org/libraries/klc/S4.7/), `IRESETN` could/should be `~IRESET`

Regarding the symbol arrangement, I like the logical approach you've followed. However, for the bottom half, most of the pins on the left are outputs and most of the pins on the left are outputs. So I think that the opposite would feel more natural to me. Let me explain my reasoning and let's see if you agree with any of it :)

- Reset feels natural on the left
- I2C is mostly communication into this AK7742E so again feels better on the left
- The I2C address, clock selection pins and similar pin are clearly inputs so again left.
- If you want, the analog and digital power/gnd pins can be reversed.

But hey, this is all very personal preference. Let me see what you think. See screenshot mock-up for reference.

akm

Thanks!

evanshultz:
Thanks for the review!

- We have typically used Passive for pins which go to passive parts like pins 6, 31, and 42. This is because it lessens the chance of a false DRC error compared to other pin types. If they can source current it's a different story, but these pins don't.

- TEST1 has some function, but it's not expected to be used by the customer. If this symbol is used for a test fixture, Passive will work. But we don't know what's inside the IC so that avoids DRC errors in a normal use case and allows connection for somebody who does know the IC in case TEST1 is an input or output or both.

Regarding the two items above, maybe there's a better way. I wrote what has been done but if there is a gotcha please mention it so we can handle these kinds of pins better.

- My symbol is for the LQFP package, which doesn't have an exposed pad. Note the EQ suffix. EN is the suffix for the QFN package with an exposed pad.

I moved the pins like you mentioned above and also thinned the symbol. Pin names are a little tight near the top but similar to what I see elsewhere. Current symbol is below. Looks OK?
image

diegoherranz:
Many thanks for making the changes in 2 commits. It makes the review much easier. Appreciated.

Regarding the passive pins, I get your point and I know that we've used passive for similar pins in the past. I guess it works. For TEST1 I hadn't thought about someone making a test fixture. In that case, yes, passive works better. For the other 3 pins passive works as well, so I think we can leave it as is.

> My symbol is for the LQFP package, which doesn't have an exposed pad. Note the EQ suffix. EN is the suffix for the QFN package with an exposed pad.

Aggh, got confused with that. Sorry.

I think it's all good to go. Thanks!

PR #1925, opened 2019-06-15
Commodore 6502 Microprocessor

StormTrooper:
*Commodore Semiconductor Group 6502 Microprocessor*

---

Datasheet http://archive.6502.org/datasheets/mos_6500_mpu_nov_1985.pdf

Screenshot from 2019-06-25 06-34-47


fauxpark:
Hi, just a couple things I can see:

 - Please reduce the pin length to 100 mil.
 - The component name should be horizontal, and either centred at (0,0), or placed along the top or bottom right of the rectangle like the refdes. Probably the latter as the symbol is not really big enough to justify placing the refdes/value inside the box.
 - Please move the footprint field closer to the symbol. I like to centre them at (0,0).
 - The `Vss` pins can probably be stacked.

obsolete/MCU_Commodore.dcm
```diff
@@ -0,0 +1,9 @@
+EESchema-DOCLIB  Version 2.0
+#
+$CMP 6502
+D NMOS Microprocessor, 65Kb memory, DIP-40
```
fauxpark:
```suggestion
D NMOS microprocessor, 65kB memory, DIP-40
```

fauxpark:
This library should probably also be renamed to `CPU_Commodore` and added to the `sym_lib_table` file.

evanshultz:
@fauxpark 
Thanks for the great review!!

@StormTrooper 
Thanks for taking the time to contribute! This is a nice symbol and we appreciate you adding to the existing library.

1. Since we won't need a manufacturer-specific lib, let's go with `obsolete/CPU.lib` for this symbol. Thoughts?
1. `IN` and `OUT` in pin names can be all capital letters.
1. Pin 37 is for phase zero (not one).
1. Does it make sense to have the clock pins be 37, 3, and then 39, top to bottom?
1. Is pin 2 only an Input?
1. Is pin 38 an Input?
1. How about putting a bit of space below the data pins?
1. Please change pin 40's name to `~RES`.
1. What about pins 34 and 40 on the left side? I truly don't know what's best, so just asking.
1. Even if the symbol is a bit off-center, you can still slide up the bottom. Having the square fit around the pins more attractively is more important than a slight off-centering.

Lastly, please update the screenshot in the first post or add a new one during the review process. This makes things much easier. Here is the current symbol:
image

StormTrooper:
Thanks for the review and picking up on the errors. have corrected the pins as requested.

evanshultz:
Thanks!

Not sure if you noticed items 9 and 10 above. Maybe item 9 is best as it is? Don't know. But if pin placement stays as it is, the symbol can be a little less tall (item 10). You can move around the NC pins and make things bit a little nicer so there's aren't as many big gaps between pins.

StormTrooper:
I've moved around the pins to address item 10.

evanshultz:
Thanks again!

Are pins 34 and 40 good to have together? They can both stay on the lower right but shift around the NC pins, if so.

Two other things:
1. The pin at the bottom should be all capital letters.
1. NC should be oriented so the pin line is inside the symbol body.

StormTrooper:
Made your suggested corrections. Should hopefully all be ok now,

evanshultz:
Thanks again.

Both pins 1 and 21 need the same name, all capitalized.

And since you're touching the symbol again anyway, it might be nice to push the three clock pins on the left down and move the NC pin above them so the clocks have a pin space above?

StormTrooper:
Pins 1 & 21 changed to both have the same name, and I've moved the pins down as suggested.

evanshultz:
Thanks for all the work. Looks great. I know this was a bit of a long process.

Since 5.1.3 is planned to be released soon you should see your work in KiCad then.

StormTrooper:
Lovely stuff. Thanks for all the help. Some good pointers I can use when submitting the next symbol :)

PR #1928, opened 2019-06-20
add AD7298 8 channel ADC

lorem-ipsum:
This adds the [AD7982](https://www.analog.com/media/en/technical-documentation/data-sheets/AD7298.pdf) 8 channel ADC. Matching footprint PR: https://github.com/KiCad/kicad-footprints/pull/1653

I've set the VREF pin to passive as it can be either used as power reference output or input.

AD7298

---


evanshultz:
1. FP filter should be `LFCSP*4x4mm*1EP*P0.5mm*` (missing the body size).
1. Even if it makes the symbol a bit off-center, you can slide up pin 7 and all right-side pins to shrink the symbol. Centering is really just meant to catch a wildly off-center symbol and not small issues which are to the detriment of the symbol's aesthetics.
1. It's unclear to me if the thermal pad should always be connected to pin 9. I didn't see the datasheet mention any internal connection or required external connection. How did you know to stack them? Interestingly, https://www.analog.com/media/en/technical-documentation/user-guides/UG_254.pdf shows GND and GND1 connected together and the thermal pad isn't discretely shown.

lorem-ipsum:
Hi @evanshultz,

thanks for your review. Maybe at some point I will be able to remember all parts of the footprint filter -.-

I also shrunk the symbol as suggested. In the pin listings there is a statement that the pad should be soldered to the PCB ground. I interpreted this as it should also be soldered to the electrical GND. But maybe I am mistaken.

evanshultz:
I read that too. But GND1 and GND are both ground but specifically mention different what pins should decouple to them, so I know they should be separate pins. And you got that. But for the thermal pad it only says "ground" so it's unclear to me if one of the GND pins is best or it doesn't matter. There is talk about the thermal sensor and the ground pad but that doesn't tell me either. I would suppose GND and pin 21 are shorted, like you have, and GND1 is connected at just a single point for best noise performance. But I just don't see that clearly mentioned. Do you see details I missed?

lorem-ipsum:
No, I don't think there is anything clearly mentioning where to connect it to. As it is unclear, should I separate GND and EPAD?

evanshultz:
Yes. That's probably most safe. Thank you.

Would you then post a new screenshot?

lorem-ipsum:
here we go:
AD7298

evanshultz:
Thanks!

PR #1929, opened 2019-06-20
add AD7949 family of 8 channel ADC

lorem-ipsum:
This adds the [AD7949](https://www.analog.com/media/en/technical-documentation/data-sheets/AD7949.pdf), [AD7689](https://www.analog.com/media/en/technical-documentation/data-sheets/AD7682_7689.pdf) and [AD7699](https://www.analog.com/media/en/technical-documentation/data-sheets/AD7699.pdf) 8 channel ADC. They share the same package and symbol, only the resolution and speed differ.

The reference pins are set to passive as they can either be used to supply a reference or need bypass capacitors attached.

AD7949

---


evanshultz:
1. Thanks for shrinking.
1. Let's also separate pin 21 here since it's not connected internally.
1. 7689 and 7699 come in two packages, so they need to be named like `AD7689xCP` to note the symbol only applies to the LFCSP package. The last one can be named this way but doesn't have to be.

Would you please post a new screenshot?

lorem-ipsum:
AD7949

evanshultz:
The suffix for the package should be `xCP`, not `xxPZ`. The `Z` is for RoHS, which we don't capture, and there is only one character for the A or B grade. `CP` is the package's short name.

Actually, 7949 and 7699 only come in B grade so you can use `AD7949BCP` and `AD7699BCP` if you like. 7689 comes in both grades so we need either both or a wildcard.

lorem-ipsum:
changed them, and only kept the wildcard for the 7689

evanshultz:
Thanks!

PR #1932, opened 2019-06-21
HT12D

G-Pereira:
A Serial Decoder with a comparator.
image

https://www.farnell.com/datasheets/1525377.pdf


myfreescalewebpage:
Hi @G-Pereira , thanks for contributing,

- [ ] The footprint should be SOP, not SOIC
- [ ] The description should ends with the package name. I propose: `2^12 serial decoder, SOP-20`
- [ ] The datasheet should be from the manufacturer website, do you have one ? Else OK.
- [ ] The Dx pin should be Output
- [ ] The symbol is very thin and can be arranged to make easy connection in a schematic. I propose based on the datasheet:

Capture

Cheers,
Joel

G-Pereira:
Which footprint specifically? I have confirmed all the dimensions and while searching on the internet I found out that SOIC is the same. I have found this weird but in fact I have used this footprint and it fits perfectly.

G-Pereira:
There is no SOP-20 in the libraries

G-Pereira:
This force push was to revert the last commit

G-Pereira:
Ok, I got confused with Travis output, sorry the mess. It should be ok now, the force push changes nothing because I've used it to go back to the "Set Dx pins to output" commit and then I've pushed again the "Center Symbol" commit

myfreescalewebpage:
Thanks for the fixes.
We always ask to create the exact footprint, even if the dimensions are the same. This is to avoid future confusion by the users of your symbol. You can create the footprint with the generator (https://github.com/pointhi/kicad-footprint-generator) and submit it.

G-Pereira:
What about creating a copy of the SOIC footprint? It is exactly the same.

myfreescalewebpage:
The generator should be updated anyway. But yes it's simple to do it because then you just have to use a compare tool to check both generated footprints are the same :)

myfreescalewebpage:
Thanks. Will need to wait the integration of the footprint before continue here.

chschlue:
FP merged

G-Pereira:
There was an issue with the footprint dimensions. This is not ready to merge

G-Pereira:
Check https://github.com/KiCad/kicad-footprints/pull/2440

G-Pereira:
Ready to merge!

myfreescalewebpage:

PR #1933, opened 2019-06-21
Add Adafruit Feather M0 symbol

matthijskooijman:
This adds a symbol for the Adafruit Feather M0 module, see https://learn.adafruit.com/adafruit-feather-m0-basic-proto for docs on the basic version. [Schematic here](https://learn.adafruit.com/assets/28801). Matching footprint is in [another PR](https://github.com/KiCad/kicad-footprints/pull/1657).

image

The symbol is specific to the Feather M0, but the footprint is
intentionally generic to all Feather boards (since AFAIK all Feather
boards are intended to use the same footprint, though pin assignments
can vary between different MCU's). Note that "Feather M0" is still
somewhat generic, since there are a few different versions (e.g. one
with an SD card onboard, etc.), but all of these different versions have
the same pinout (differences are only on the board itself and affect
pins internally routed on the Feather board).

Pin naming is based on the silkscreen and schematic (except power pins,
which I named after the ones from the Arduino Nano). Pin numbering is
fairly arbitrary (counterclockwise with the USB connector down, like the
similarly sized Arduino Nano), since the part does not have its own pin
numbering (in the Adafruit schematic, both pin headers are numbered separately).

The symbol is based on the Arduino Uno symbol, so it can be used as a
drop-in replacement (most pins will still align, and some special pins like
reset, aref and the analog pins are in the same place). Other special pins,
like SPI and I²C are not in the same place, but (at least SPI) are not
specially marked or placed in the Arduino Uno pinout anyway. Instead, I
just used the logical pin numbering (e.g. the Dx pin numbers used in the
software) to order the pins on the symbol. An alternative would be to
use the physical pin ordering instead (which is slightly different here
and there), which could make it easier to make drop-in replacement parts
for other Feather variants (which use the same physical footprint but
not necessarily the same logical pin functions). The downside of that is
that the analog pins would no longer be separate (but halfway the list
of pins), breaking compatibility with the Arduino Uno.

On the symbol, I'm not sure how to mark the power pins. 3V3 is conncted
to a regulator, so certainly output, but I believe both the BAT and
5V pins could be used as output (when a battery or USB is connected),
but also as in input (to connect and external battery or 5V supply).
However, Adafruit does [not recommend][1] using them as inputs, so I've
marked them as output (except for GND, which I think would cause
problems connecting multiple GNDs together, which is not typically an
actual problem).

[1]: https://learn.adafruit.com/adafruit-feather-m0-adalogger/power-management#alternative-power-options-4-19


myfreescalewebpage:
Hi @matthijskooijman , thanks for contributing,

- [ ] Name of pin 10 is `VBUS`
- [ ] Name of pin 13 is `~RESET`

Cheers,
Joel

matthijskooijman:
Thanks for you review, I'll update the symbol. As for the VBUS pin, I guess that's the name of the pin (or at least the name of the net it is connected to) in the schematic, but would not make more sense to stick to the name on the silkscreen instead (which is `USB`)? The silkscreen names (or slightly expanded versions) are also used in the [pinout documentation](https://learn.adafruit.com/adafruit-feather-m0-basic-proto/pinouts).

Looking at the other pin names, in the proposed symbol, I think they all match the silkscreen now (though sometimes a bit more expanded than the silkscreen, e.g. `RESET` vs `RST` or `+3V3` vs `3V`), except for `+5V` vs `USB`.

So, I would name the pin `USB` rather than `VBUS`, would you agree?

myfreescalewebpage:
> So, I would name the pin USB rather than VBUS, would you agree?

Ok for this one because the doc speak about "USB" too (not because oft he silkscreen which is something I don't care about - the most important is to be correct with the docs).

myfreescalewebpage:
@matthijskooijman ping

matthijskooijman:
@myfreescalewebpage Apologies for the late reply, I was busy :-)

I've fixed the two remaining pins (see updated screenshot in first post) and rebased on top of master.

Thanks again for reviewing!

myfreescalewebpage:
Thanks for the update. The ping was to have news from you I'm sorry :)
Just need to wait the footprint now.
Joel

nickoe:
@myfreescalewebpage  I think the footprint is in https://github.com/KiCad/kicad-footprints/pull/1657

nickoe:
@matthijskooijman I guess the docs say:

> USB - this is the positive voltage to/from the micro USB jack if connected

So it is probably better to just call it USB, although it would have been nicer if it was named VBUS from adafruit.

matthijskooijman:
> @myfreescalewebpage I think the footprint is in KiCad/kicad-footprints#1657

I had interpreted @myfreescalewebpage's remark as "wait for the footprint to be reviewed and ready to be merged", not "for a footprint to be submitted", but maybe I was wrong.

> So it is probably better to just call it USB, although it would have been nicer if it was named VBUS from adafruit.


Yeah, agreed.

I also just realized that the pin was still called `+5V` so still needed to be changed (on my previous push, I though that no change to `VBUS` was needed, but I missed that a change to `USB` was needed). So I added a commit to change the pin name, should be good now I think (this time I added a new commit rather than amending the previous commit, just remembered that that was preferred for this repository IIRC).

fauxpark:
I think this should be renamed to `Adafruit_Feather_M0_Proto` to distinguish it from the variants of this board with a BLE module in place of the protoboard - see #2473.

There is also a 32U4 version as with the Bluefruit LE: https://www.adafruit.com/product/2771.

matthijskooijman:
> I think this should be renamed to Adafruit_Feather_Proto_M0 to distinguish it from the variants of this board with a BLE module in place of the protoboard - see #2473.

I was under the impression that the external pinout is identical for all M0 feathers, just the internal connections are different, so this symbol could be used generically. However, I just noticed that at least the [Feather M0 Lora](https://learn.adafruit.com/adafruit-feather-m0-radio-with-lora-radio-module/pinouts) has indeed different pins (it replaces GND with an Lora IO pin) and even has some extra pins.

So yeah, renaming this is a good plan, I'll do that (but later, bit busy now).

However, I do not think this needs to be renamed to distinguish it from the Bluefruit module you linked, since that has a 32U4 rather than an M0 MCU, so that is already distinguished by the current naming.

Also, thanks for linking to your Bluefruit symbol and footprint. Would it be worthwile to make them a bit more compatible where possible? I think the symbols could be modified to have pins which are (logically) the same in the same place. From a quick look:
 1. the `USB`, VBAT` and `EN` pins are common but in different places, so could be moved to match. I would prefer my version, since it is compatible with the Arduino Uno and Arduino Leonardo symbols.
 2. you're using `VBUS` to match the schematic, while I've used `USB` to match the documentation and silkscreen (see discussion above).
 3. you've labeled the `MISO`/`MOSI`/`SCK` pins without their `Dxx` equivalent and sorted them in physical pin order with some separation, I've added `Dxx` and sorted them in logical pin number order. Given the logical pin number ordering is different between both boards, sorting them in physical pin order is probably the only way to make these symbols compatible, and then some separation between the SPI pins and the other digital pins seems appropriate (since the sorting changes). The same argument could be made for the I²C pins (and also `RX`/`TX`, since those are at logical pin 0/1 in both boards, but e.g. the [Feather Huzzah](https://learn.adafruit.com/adafruit-feather-huzzah-esp8266/pinouts) has other pin numbers again.


As for the footprint, I think those *are* intended to be identical across all feather boards (with some exceptions, such as the Lora board I mentioned which has some extra pins). Looking at our both footprints, they look identical except for pin numbering and orientation (but that's pretty arbitrary, at least in mine). I think it would be good to just have a single footprint for both symbols, would you agree? I would then suggest using my version as-is, since it is already named generically and has a version with and without mounting holes.

fauxpark:
> I was under the impression that the external pinout is identical for all M0 feathers, just the internal connections are different, so this symbol could be used generically.

The Proto board has a `GND` pin in place of the `DFU` pin that goes to the BLE module on the Bluefruit.

> However, I do not think this needs to be renamed to distinguish it from the Bluefruit module you linked, since that has a 32U4 rather than an M0 MCU, so that is already distinguished by the current naming.

There is a 32U4 and M0 version of the Bluefruit LE and proto boards. I've added the M0 one as an alias, as the pinouts are the same between them, at least. You've only added the M0 here.

> the `USB`, `VBAT` and `EN` pins are common but in different places, so could be moved to match. I would prefer my version, since it is compatible with the Arduino Uno and Arduino Leonardo symbols.

Fair enough, I've moved the power pins to match and switched `AREF` and `EN`. However, I think the `~DFU` pin should be near the `~RESET` pin, as it serves a similar purpose, but for the BLE module.

> you've labeled the MISO/MOSI/SCK pins without their Dxx equivalent and sorted them in physical pin order with some separation, I've added Dxx and sorted them in logical pin number order.

I used the Arduino pin numbers - however, at first I couldn't find the numbers for the SPI pins, on the Leonardo they are only connected to the ISP header. [Turns out they are on digital pins 14-17](https://forum.arduino.cc/index.php?topic=466897.0). I don't know where Adafruit got 20-24 from. Maybe that's the pinout for the M0, in which case an alias won't work here.

RE footprint: yours is probably more detailed, yes.

matthijskooijman:
> There is a 32U4 and M0 version of the Bluefruit LE and proto boards. I've added the M0 one as an alias, as the pinouts are the same between them, at least. You've only added the M0 here.

Ah, good point, I had assumed they would be more different. It does seem the logical pin numbers of the SPI and I²C pins are different, so having them as an alias prevents adding the logical pin numbers to those pin names. I think this would make the pin names and the symbol easier to use, so I would be inclined to duplicate both symbols for 32u4 and M0, and including the Dxx in the pin names (I think Kicad 6 will support pin renaming in aliases, so later some of this duplication can be reduced again).

> Fair enough, I've moved the power pins to match and switched AREF and EN. However, I think the ~DFU pin should be near the ~RESET pin, as it serves a similar purpose, but for the BLE module.

Hm, but both symbols (as well as the Arduino symbols) have an `AREF` pin, so would be good to also keep that in a consistent location if possible, but I see your point about grouping `~DFU` as well. How about something like this:

image

This puts the `EN` down with the `AREF` pin, so both can match among all feathers, and the `AREF` pin can match with the Arduinos as well.

Note that your footprint is still a little mismatched with the Arduino Uno/Leonardo, it's body is slightly lower and the `AREF` and analog pins are shifted by one:

image

(note that this is the old version of your symbol, seems you haven't pushed the power pin and `AREF`/`EN` swapping yet)

> I used the Arduino pin numbers - however, at first I couldn't find the numbers for the SPI pins, on the Leonardo they are only connected to the ISP header. Turns out they are on digital pins 14-17. I don't know where Adafruit got 20-24 from. Maybe that's the pinout for the M0, in which case an alias won't work here.

Did you mean 14-16 there? Logical / Arduino pin 17 does not seem related to SPI? Also, where does 20-24 come from? I can't find those pin numbers for SPI in either the M0 or 32u4 pinouts?

So, what to do with the SPI/I2C/UART pins? I think we can either:
 - Leave them as-is, sorted by logical/Arduino pin number, and accept that the symbols are not complete drop-in replacements if any of these special pins are involved. Note that even though it seems at first glance that the SPI pins are matched, the logical pin numbers are not in the same order for M0 and 32u4, so even those will not match.
 - Try to give these special pins a specific place, which I think could look something like this:
   image
   Note that I moved the I²C pins to the right, to again match the Arduino symbols (which seems like a nice bonus), and also freeing up some room at the left to keep some space between the digital, UART and SPI pins. It also accidentally matches the SPI pins with the Arduino Uno (not the Leonardo), but that does not seem really relevant here (since the Arduino does not label them as such on D11/D12/D13).



@myfreescalewebpage @nickoe, maybe you have any input on this?

fauxpark:
Seems GitHub had a bit of a hiccup, I pushed the changes right after I made that comment, but the PR didn't update for a bit. It also showed my closed footprint PR in the open list 😄

Arduino pin 17 is connected to the SPI slave select pin (PB0), but it's NC on the Feather 32U4 boards: https://cdn-learn.adafruit.com/assets/assets/000/028/668/original/adafruit_products_schem.png?1447886222
Instead, PB4 is used as the chip select pin for the BLE module, for some reason, and is also not broken out.

myfreescalewebpage:
Not read everything but yes I'm waiting the footprint to be reviewed before the symbol can be merged.
About renaming the symbol, maybe the LoRa can just be suffixed with "_LoRa" ?
Joel

myfreescalewebpage:
About pin order I prefer to have the pin as done today for consistency across the library.

matthijskooijman:
> About renaming the symbol, maybe the LoRa can just be suffixed with "_LoRa" ?

Yeah, that should certainly happen, but it turns out there's more Feather M0 variants that have subtly different pin assignments (so need a different symbol), so my earlier assumption that this symbol could be used generically across all M0 feathers is not really true and the generic `Adafruit_Feather_M0` name for this symbol no longer seems warranted.

Looking at the [full list of feathers](https://www.adafruit.com/feather), the only feathers that can use this symbol are the basic proto, adalogger and circuit express version.The Bluefruit BLE version, the Wifi version, the LoRa version and the RFM69 radio version need a different symbol. Also in the interest of having symbols that describe unique parts, I've gone ahead and renamed this symbol to `Adafruit_Feather_M0_Basic_Proto` and added aliases `Adafruit_Feather_M0_Adalogger` and `Adafruit_Feather_M0_Circuit_Express`.

> Seems GitHub had a bit of a hiccup, I pushed the changes right after I made that comment

@fauxpark Great, with our combined changes, seems our symbols are pretty much compatible now. I have some more suggestions, but I'll comment them on your PR rather than here.

> About pin order I prefer to have the pin as done today for consistency across the library.

@myfreescalewebpage, what do you mean by consistentency here exactly? Consistent with what? I am suggesting changes to make the Feather symbols more compatible and consistent with each other?

Also, I think that I/O pins have no clearly defined ordering rules, except that [grouping by function is recommended](https://kicad-pcb.org/libraries/klc/S4.2/).

Currently, we have the following pin ordering for the digital pins (I included the Bluefruit feather and Arduino Uno to illustrate pin compatibility, note that the two feathers do *not* have matching SPI pins due to the logical pin number sorting):

image

Now, my suggestion would be to pick out the I²C, SPI and UART pins and group them separately and sort them by function rather than logical pin number (Dxx). Since all different Feather boards have these same pins, this makes it easier to switch from one feather board to another, reducing the amount of connections that must be redrawn in the schematic. Since all feathers even put these functions in the same physical pins, a lot of the board layout can even be preserved.

With this grouping, it could look like this:

image

Note that compared to my earlier suggestion, I removed the space between UART and SPI pins, to leave a bit more room for additional GPIO pins, and swapped SPI and UART pins to match the physical layout.

Thinking on this a bit more, I actually think that this proposal is really just implementing the function-based grouping recommended by [KLC S4.2](https://kicad-pcb.org/libraries/klc/S4.2/). In the current situation above, it looks like the pins are also grouped by function (UART, I²C and SPI pins are side by side), but that is just an accidental side effect of their logical pin numbering. Even more, the Feather ESP32 has [SPI pins on logical pins 5, 18 and 19](https://learn.adafruit.com/assets/41624), so with the current pin ordering scheme, those would not end up together.

@fauxpark, @myfreescalewebpage, would you agree that this pin ordering is indeed an improvement? Or am I totally overthinking this?

matthijskooijman:
I've updated pin numbering to match the rotated footprint (see https://github.com/KiCad/kicad-symbols/pull/2473#issuecomment-659327413).

myfreescalewebpage:
@matthijskooijman I was referring to pin order. I know it's good to group by functions, but for this kind of modules, as for the MCUs, we usually group by pin name. I.E. DO then D1 then D2 etc, without grouping by functions.
However, if all feather modules are designed with the same rules, we can accept the way you want. Maybe you can submit some other symbols right now in this PR to have a more general view ? Lora version for example.

matthijskooijman:
>  Maybe you can submit some other symbols right now in this PR to have a more general view ? Lora version for example.

Yeah, I think I will (though I'm away for vacation the next 2 weeks, so not before that).

myfreescalewebpage:
@matthijskooijman ping

matthijskooijman:
> However, if all feather modules are designed with the same rules, we can accept the way you want.

Yup, seems they even wrote down most of these rules here: https://learn.adafruit.com/adafruit-feather/feather-specification

>  Maybe you can submit some other symbols right now in this PR to have a more general view ? Lora version for example.

I now added the ESP-based feathers (since those have very different GPIO-numbering) and all the M0 variants (including the LoRa version). See below for some specific discussions of some of the problems or peculiarities I found and how I handled them.

The Lora/RFM board was indeed the most problematic, since it adds extra pins that are hard to squeeze in. I'm not entirely sure what the best solution is there, so opinions are welcome. See below for details.

## Newly added symbols

### ESP32
The pin labeling of the ESP32 is a bit weird. Rather than `Dx` logical "Arduino" pin numbers, they're labeled with logical `GPIOx` numbers as defined by the ESP chip, so this numbering is all over the place. These pin numbers are shown as `IOx` on the schematic, which is consistent with other ESP-based board (except some input-only pins, which are labeled as `Ix`).

Also, all but one GPIO pin are also analog pins, so those pins also have a (more consistently numbered) `Ax` number. Grouping all analog pins on the right won't work because there are too many, but just putting the first 6 on the right (like other feathers) and then putting all other GPIO on the left and ordering them all using their `Ax` number gives a consistent picture which is also symbol and footprint compatible with the other feathers. Note that the special-function GPIO pins (i.e. SPI/I2C/UART) are *not* analog pins, so have not `Ax` number. They do have a `IO` number, so that's what I'm showing in addition to the special function (this matches the silkscreen, the schematic sometimes omits either the IO number of the special function, I've suggested adafruit to fix this is the schematic).

### ESP8266
This has a lot less pins (just one analog pin, with 5 NC pins where the analog pins are on most other feathers).

The pins are labeled `GPIOx` in the schematic, but I've used `IOx` in the example, to make everything fit better, and for consistency with the ESP32 version. Other ESP8266 boards often also use `IOx` labeling, so that seems acceptable to me. The `RXD` and `TXD` pins do not have any associated `GPIO` number in the schematic, but I've added them here based on the datasheet for the ESP12 module used (and suggested adafruit to add them in the schematic as well).

The IO pins have no really meaningful "logical" pin order (there are the `IOx` numbers, but those are nonconsecutive and random. To keep footprint/layout compatibilitity with the other feathers, I've opted to order these pins by their physical pin numbers (since all other feathers have matching logical and pin numbers, otherwise this would be the only exception).

Also, the SPI pins are duplicated in the place where other feather pins have normal GPIO pins (just exactly the same pins, internally connected). I just put these on the symbol twice, with the second copy among the normal GPIO pins (based on physical pin numbers). I considered doing pin stacking for these, but then KiCad insists connecting each pair of pins together in the layout, which is not really something you'd typically want for these signal pins, I think.

This board also has an extra `CH_PD` pin (chip powerdown) that is unique to the ESP8266. Since this is similar in functionality to the reset and enable pins, I've put it up there.

### M0 Wifi
Identical to the M0 basic proto, except the spare GND pin is replaced by wifi chip WAKE pin.

### M0 RFM69HCW Packet Radio and M0 RFM9x LoRa Radio
These are actually exactly the same board, just with a different radio soldered on. They even share a single schematic. I've created a single symbol with an alias for them.

These are also nearly identical to the M0 basic proto, except the spare GND is replaced by a radio IO pin, but there are also 4 additional pins (so different footprint): 3 more radio IO pins and a pin for the antenna.

This one is actually problematic: These extra pins need some squeezing to get in. Ideally, the four radio IO pins (labeled `DIOx`) would be grouped together, but that requires moving either the GPIO pins at the top left (breaking compatibility with the Uno) or moving the special-purpose pins at the bottom left (which looks a bit less clean, but is probably ok. This should happen to *all* feathers, to keep them compatible). One other alternative is to split the radio IO pins, rather than keeping them grouped. I tried the last two options, see below:
image

I think I would prefer the left version, any other thoughts? We could make all feather symbols bigger, breaking compatibility with the Arduino Uno (and other Arduinos based on that), I think I'd rather try and keep that.

The 32u4 RFM versions are in a slightly better position, since those only break out 3 radio IO pins rather than four (so also have one less pin, yet another footprint).

### Generic symbol
I also added a generic symbol, that has all the special pins labeled, and for the analog and digital pins just assumes some `Ax` and `Dx` labels. The "Spare" I/O pin (that varies between each board) is labeled as such. This symbol can probably be used in a lot of cases where no specific symbol is available for a Feather board.

## Other changes
I also noticed that pin 4 is actually the standard GND pin, and pin 16 is usually something different, just happens to be GND on the M0 proto. So I swapped pin 4 and 16 there, so 4 is now visible and 16 hidden.

Compared to my earlier proposal, I've also swapped the special function (e.g. `MISO` and `Dx` names on the 32u4 and M0 feathers) so that the name closest to the pin reflects the name that was used to position it (e.g. `Dx` or `Ax` for the generic pins, and the function for the special function pins).

I also made the symbol descriptions and keywords a bit more detailed and consistent.

##  Symbols not added yet
Most of the others are very similar, usually only replacing the "spare" (GND on M0) pin with something else.

The WICED board is a bit weird in its assignments of analog pins (weird numbering and one of them is also labeled "WAKE"). It also has 1 extra pin besides the normal headers for SWD, so that would probably need its own footprint.


# Symbols so far
Here's an overview of the current state of all symbols in this PR. It includes  both options for the RFM versions, if we go for the left version, than all other symbols will have to be updated to also move their SPI and UART pins one position down as well.

image


# All feather versions
This is a (hopefully complete) list of all feathers that are currently being sold. Most of them are listed in the [intro to feathers](https://learn.adafruit.com/adafruit-feather/feathers) tutorial, but some of them I found only in the learning system and/or shop.

### M0/SAMD21-based
 - https://learn.adafruit.com/adafruit-feather-m0-basic-proto/ 
 - https://learn.adafruit.com/adafruit-feather-m0-adalogger/
 - https://learn.adafruit.com/adafruit-feather-m0-express-designed-for-circuit-python-circuitpython
 - https://learn.adafruit.com/adafruit-feather-m0-bluefruit-le/
 - https://learn.adafruit.com/adafruit-feather-m0-wifi-atwinc1500/
 - https://learn.adafruit.com/adafruit-feather-m0-radio-with-rfm69-packet-radio
 - https://learn.adafruit.com/adafruit-feather-m0-radio-with-lora-radio-module

### AVR 32u4-based
 - https://learn.adafruit.com/adafruit-feather-32u4-basic-proto
 - https://learn.adafruit.com/adafruit-feather-32u4-adalogger
 - https://learn.adafruit.com/adafruit-feather-32u4-bluefruit-le
 - https://learn.adafruit.com/adafruit-feather-32u4-radio-with-rfm69hcw-module
 - https://learn.adafruit.com/adafruit-feather-32u4-radio-with-lora-radio-module/
 - https://learn.adafruit.com/adafruit-feather-32u4-fona (cellular)

### NRF-based
 - https://learn.adafruit.com/introducing-the-adafruit-nrf52840-feather (nRF52840 Express)
 - https://learn.adafruit.com/adafruit-feather-sense (nRF52840)
 - https://learn.adafruit.com/bluefruit-nrf52-feather-learning-guide (nRF52832 Bluefruit)
 - https://learn.adafruit.com/adafruit-nrf52-pro-feather (nRF52832)

### Single feathers (only one version with this particular MCU available)
 - https://learn.adafruit.com/adafruit-feather-huzzah-esp8266
 - https://learn.adafruit.com/adafruit-huzzah32-esp32-feather
 - https://learn.adafruit.com/introducing-the-adafruit-wiced-feather-wifi
 - https://learn.adafruit.com/adafruit-feather-m4-express-atsamd51
 - https://learn.adafruit.com/adafruit-stm32f405-feather-express
 - https://www.adafruit.com/product/3200 (Teensy adapter, schematic only)
 - https://learn.adafruit.com/adafruit-feather-328p-atmega328-atmega328p

myfreescalewebpage:
@matthijskooijman new steps of the review now:

- [ ] Except Adafruit_Feather_M0_RFM69HCW_Packet_Radio I suggest all the footprints should be "Module:Adafruit_Feather" (currently some are using the "with holes" version).
- [ ] Except "Adafruit_Feather_32U4_BluefruitLE", on all the other symbol, VBAT, 3V3 and USB should be Power Input
- [ ] There is an unwanted "Adafruit_Feather_M0_RFM69HCW_Packet_Radio_copy"

Adafruit_Feather_M0_RFM69HCW_Packet_Radio
- [ ] Have you made the footprint PR ?

Joel

matthijskooijman:
>  Except Adafruit_Feather_M0_RFM69HCW_Packet_Radio I suggest all the footprints should be "Module:Adafruit_Feather" (currently some are using the "with holes" version).

Will fix.

>  Except "Adafruit_Feather_32U4_BluefruitLE", on all the other symbol, VBAT, 3V3 and USB should be Power Input

Yeah, I guess. These pins typically provide power, but can also be used to input power, so then power input makes more sense, so I'll change that. I think I based this on the Uno symbol, which also uses power output, but that should be separately fixed I guess.

>  There is an unwanted "Adafruit_Feather_M0_RFM69HCW_Packet_Radio_copy"

We still have to figure out how to layout this symbol (and others) exactly, this shows one of the two options I've considered. Any thoughts on that? Once we figure that out, I'll remove the other one.

>  Have you made the footprint PR ?

Nope, but now that the generic footprint is merged, I'll do this one as well.

myfreescalewebpage:
Thanks, pin me for the next step of the review

matthijskooijman:
@myfreescalewebpage I'll ping you once I make the changes. However, before I can continue, we still need to decide how to handle the extra pins on the RFM versions. I'm really curious to hear your ideas on the options I've outlined. For clarity, below is the relevant section from my earlier comment (copied verbatim), which shows the challenge and options I'm considering. Maybe you could respond to that? @fauxpark, maybe you have some insights as well?


### M0 RFM69HCW Packet Radio and M0 RFM9x LoRa Radio
 - https://learn.adafruit.com/adafruit-feather-m0-radio-with-rfm69-packet-radio
 - https://learn.adafruit.com/adafruit-feather-m0-radio-with-lora-radio-module

These are actually exactly the same board, just with a different radio soldered on. They even share a single schematic. I've created a single symbol with an alias for them.

These are also nearly identical to the M0 basic proto, except the spare GND is replaced by a radio IO pin, but there are also 4 additional pins (so different footprint): 3 more radio IO pins and a pin for the antenna.

This one is actually problematic: These extra pins need some squeezing to get in. Ideally, the four radio IO pins (labeled `DIOx`) would be grouped together, but that requires moving either the GPIO pins at the top left (breaking compatibility with the Uno) or moving the special-purpose pins at the bottom left (which looks a bit less clean, but is probably ok. This should happen to *all* feathers, to keep them compatible). One other alternative is to split the radio IO pins, rather than keeping them grouped. I tried the last two options, see below:
image

I think I would prefer the left version, any other thoughts? We could make all feather symbols bigger, breaking compatibility with the Arduino Uno (and other Arduinos based on that), I think I'd rather try and keep that.

The 32u4 RFM versions are in a slightly better position, since those only break out 3 radio IO pins rather than four (so also have one less pin, yet another footprint).

myfreescalewebpage:
> I think I would prefer the left version, any other thoughts? We could make all feather symbols bigger, breaking compatibility with the Arduino Uno (and other Arduinos based on that), I think I'd rather try and keep that.

I think this is the best option, I do not care of Feather / UNO compatibility. If we go in this direction why not creating a symbol with raspberry pi pin compatibility ? It has no sense. Feather boards are feather boards. UNO is something different.
I prefer to have all nice Feather symbols compatible each other.

matthijskooijman:
> If we go in this direction why not creating a symbol with raspberry pi pin compatibility ?

Well, the Arduino series and Feather series are in roughly the same category of easy-to-use microcontroller boards, while the rpi (and a lot of othe dev boards) are really a different category. Starting a project out with an Arduino Uno and later deciding to a Feather seems like a reasonable usecase to me, though maybe not super common. Maybe the Arduino nano is more similar to a feather, but looking at the current footprints there, I just noticed that the Nano and Uno already have a slightly different layout (most of the pins are in the same place, but the Nano is slightly lower than the Uno), so that does mean the compatibility a bit more arbitrary.

So I guess that just making the symbol bigger is a good plan indeed. I'll probably stick to the same pin layout (got to pick something, right), but make it a few lines higher, so it will still be a drop-in-replacement for the Uno in most areas, except for the lower pins (including the GND pins).

Thanks!

myfreescalewebpage:
I think this is a good choice.
You can still upgrade UNO symbol later (not right now, we can first try to focus on this symbol), but you will be able to propose it on GitLab soon! Migration on 2020-10-01 :)

matthijskooijman:
Ok, I tried making the symbol a bit bigger. I did the LoRa/RFM version first, which has the most pins. I now have this:

image

How is that?
I now moved all the fixed-function pins (SPI, UART, I²C) to the left, leaving room at the right for all the radio pins (on most other feathers the area on the bottom right will be empty).

Any remarks about the pin layout?

I also tried the WICED Wifi version, since it has the weirdest pin assignments (the fixed-function ones are still the same, but the pin numbering is a bit of a mess because it uses the CPU's port/pin assignments). I just placed them based on the physical pin numbers (so overall, all feather symbols will have all GPIO pins in a fixed place based on their physical pin number, except when such a pin has a clearly different function, such as the `SWDIO` below).

image

If I would do just this WICED symbol, I might choose a slightly different layout, but this has consistency and does not seem too bad to me?

One more question: Is there a policy for placement of the Reference and Value fields? I couldn't find this in the KLC. I see a lot of symbols with the ref in the top left and the value in the lower right, which seems like a nice way to do it (I think this is also what @fauxpark did in the bluefruit versions I imported).

myfreescalewebpage:
The design you proposed is great :) 

> One more question: Is there a policy for placement of the Reference and Value fields? I couldn't find this in the KLC. I see a lot of symbols with the ref in the top left and the value in the lower right, which seems like a nice way to do it (I think this is also what @fauxpark did in the bluefruit versions I imported).

There is no rule because the ref and name are moved automatically when a symbol is placed on a schematic. A common way to do it is ref on top left and name on top right or bottom right. My opinion is : as soon as it is clearly indicated and do not overlap with the symbol it's great. Same comment apply for the footprint name. I'm used to put it below the symbol when I contribute to the lib.

fauxpark:
> There is no rule because the ref and name are moved automatically when a symbol is placed on a schematic.

This is not always true as you can turn that off. Personally I dislike it so I agree there should be something in the KLC regarding placement of refdes and symbol name.

myfreescalewebpage:
Good remark of @fauxpark . I agree with you but there is no rule today.

matthijskooijman:
Ah, just also noticed the autoplace behavior. But I've made the placement consistent now (ref at top left, value at bottom right) anway. I've also fixed the issues @myfreescalewebpage mentioned, except for the missing footprint which I'll do next. I've now changed all symbols to use the same, bigger layout as I suggested above, so with that, I think I have no more changes to be made to this PR (other than adding the footprints).

Here's the overview again:

image

There's still a few other Feathers missing, but since those are fairly obscure and I don't want to spend too much more time on this, I'll leave them for now. Also because at a later point, it might be good to do scripted generation of these symbols, and I think the new symbol library format could also greatly simplify a series like this (by allowing omitting and renaming pins from a base symbol, I think).

fauxpark:
@matthijskooijman your last commit is incorrect; Atmel (well, Microchip now) uses `ATmega32U4` with a capital U: https://www.microchip.com/wwwproducts/en/ATMEGA32U4

myfreescalewebpage:
@fauxpark I agree with you but the Feather board is spelled "32u4". The MCU on the board itself does not matter for me.

myfreescalewebpage:
@matthijskooijman just one comment on Adafruit_Feather_M0_RFM69HCW_Packet_Radio:
- [ ] Pin number should be in anticlockwise order, you have to change pin number for pins 29-32 to 32-29

Else just need the missing footprint.
Joel

PS : the library is migrated to GitLab in few hours now so if you want to work on this right now to not suffer of the migration on this PR, do it quickly (only 4 pads to be added)

myfreescalewebpage:
Oh just notice also the travis errors:
- [ ] Symbols are not centered, you have to move everything down
Joel

matthijskooijman:
Re naming, I checked the Adafruit feather product name and said something about the MCU name from memory. But indeed, the board name seems more relevant here than the MCU name here indeed.

I also did one more doublecheck of all pin names against the Adafruit documentation, looks all good.

I added the missing footprints here: https://github.com/KiCad/kicad-footprints/pull/2507

I fixed the pin ordering, also in the footprint.

I fixed the symbol centering, should have checked that before pushing.

Found one symbol where the ref/value fields were not consistent yet, so fixed that too.

myfreescalewebpage:
Need to wait footprints now to continue

myfreescalewebpage:
@matthijskooijman 

Adafruit_Feather_M0_Basic_Proto
- [ ] Fp should be `Module:Adafruit_Feather`

Joel

matthijskooijman:
W00ps, must have missed that one before. Fixed now.

myfreescalewebpage:
No problem; Just need to check footprint now before merging here.

myfreescalewebpage:

myfreescalewebpage:
And now travis is happy with the fp. Merging the symbols

myfreescalewebpage:
Done before moving to GitLab, hooray!

matthijskooijman:
w00tw00t! Thanks for you many reviews and advice, much appreciated :-D

PR #1934, opened 2019-06-21
Fix checklib.py errors in MCU_Module

matthijskooijman:
This:
 - Sets some pin name offsets to 20mil
 - Removes some symbol names from the description fields

This leaves the CHIP symbols with errors, since those will be reworked
in #1240.

This also leaves missing default footprints for some SODIMM modules.
Currently, there are no SODIMM footprints in the library yet, but even
when there will be, there can very well be different sockets that are
all SODIMM-compatible but with different footprints (effectively, these
modules do not have a footprint of their own, but plug into a socket
that has a footprint).

Note that some SODIMM sockets are being added in
https://github.com/KiCad/kicad-footprints/pull/1187


myfreescalewebpage:
@matthijskooijman I'm the original author of the CHIP devices and I have reworked the them since, and I'm waiting for the review at #1240. Can you revert your modifications for the CHIP devices, so I will not have a large conflict to solve ! Thanks.
Joel

matthijskooijman:
@myfreescalewebpage, thanks for pointing that out. I've backed out those changes.

myfreescalewebpage:
@matthijskooijman thanks. I have made the review of the other modifications, no comment to do. Thanks for this fix, merging.

antoniovazquezblanco:
@myfreescalewebpage please use the squash feature of GH to keep history clean. Thanks!

PR #1897, opened 2019-05-31
Added AP65111

DecadoPL:
https://www.tme.eu/Document/6fc4ea95d789e43795384f666d828c26/AP65111.pdf

image

It fits Package_TO_SOT_SMD:TSOT-23-6

PR #1938, opened 2019-06-27
Added AP65111

DecadoPL:
I've been working on https://github.com/KiCad/kicad-symbols/pull/1897 . I'm learning Git and Kicad so I've deleted old repo and did it from the very begining. And now I have no idea how to pull it into the same topic, so I create new one. BTW even if i signed CLAassistant it is still pending. Hope it will work this time.

https://www.diodes.com/assets/Datasheets/AP65111A.pdf

image

Instead of naming it AP65111WU i named it AP65111AWU, this name exist in manufacturer database and shops database. Correct me if I'm wrong


DecadoPL:
I signed CLA license multiple times and it is still shows that it has been not signed 

image

evanshultz:
@myfreescalewebpage 
Since you reviewed the original PR I'm assigning you.

myfreescalewebpage:
Hi @DecadoPL thanks to continue this work. Some comments here:

- [ ] SW should be Power Output
- [ ] Just notice "KHZ" should be "KHz" in the description

Else, no other comment to do, great job !
Joel

evanshultz:
Should SW be Power Output? It does not supply power directly to other ICs because there is a series element (inductor) between the SW pin and the actual converter output. For that reason, maybe Output or Passive is better? I recall this being discussed before but a quick search didn't turn it up. And it may not make any real difference anyway.

myfreescalewebpage:
Well @evanshultz I agree the classical usage of this is to have an inductor on SW output, but that's only a typical application. In the real life, it's an output and it is able to drive power so... I don't remember of this discussion except for another device but it was a different converter topology.

DecadoPL:
@myfreescalewebpage 
So what is the final decision? How should i modify it?

evanshultz:
@DecadoPL 
It definitely should not be Power Input. Assuming the inductor is there it probably won't matter since you'll need the PWR_FLAG symbol after the inductor. And the library isn't consistent about this already. You can change it as requested above.

myfreescalewebpage:
@DecadoPL let say you can keep Output only. That something will need to be more consistent in the lib, one day!

DecadoPL:
@myfreescalewebpage Done!

myfreescalewebpage:

PR #1941, opened 2019-06-27
Split IRS2092 symbol by package

evanshultz:
As can be seen at https://www.infineon.com/cms/en/product/power/class-d-audio-amplifier-solutions/?redirId=104912&intc=0560030, the DIP16 package of IRS2092 is no longer available. The SOIC16 package is still good to go.

So this PR takes the original symbol and makes it just for the DIP16 package. This is the part numbers scheme anyway, without the "S". This means no symbols are removed (but a default footprint is added and the SOIC footprint filter is removed.

I copied the symbol, tacked on the "S", and then set the default footprint and only the matching FP filter.

Since no symbol is removed and the valid symbol added, I believe this should be OK to merge now. Let me know if this has to wait for v6.

Images of both symbols to show nothing changed here:
image

That being said, this symbol is rather awkward to use. If changing just IRS2092S is OK or if a significant usability improvement can be considered for both of these specialized symbols I can detail my proposed change.

---


myfreescalewebpage:
@evanshultz no comment on this PR.
The modification do not break existing schematics, merging now.
Joel

PR #1943, opened 2019-06-28
Add KSZ9563

evanshultz:
http://ww1.microchip.com/downloads/en/DeviceDoc/KSZ9563R-Data-Sheet-DS00002419C.pdf
image

Copied from `KSZ9893RNX`, but pins 39 and 40 changed from NC and placed at the lower right. It's a little packed but I felt that symbol compatibility was nice.

From http://ww1.microchip.com/downloads/en/DeviceDoc/EVB-KSZ9563-Evaluation-Board-Users-Guide-50002726B.pdf it's clear each power pin should be discrete, like it already was.

---


evanshultz:
Fixed some pin types. Pin 44 is a bit tricky because it can be active high or active low, and in either case a pull up/down resistor is needed. I guess Output is a suitable pin type but it's a tricky one.

Should I do the same to `KSZ9893RNX`, from which I copied this symbol? It has the same problems with pin types and this is a non-breaking change.

myfreescalewebpage:
@diegoherranz any progress on the review of this ? I'm also interested by this symbol.

diegoherranz:
I haven't had time to help with libraries lately but I should have some time during the Christmas holidays.

diegoherranz:
Sorry for the delay in reviewing.

Things checked which look OK:
- Pinout names and numbering correct
- Exposed pad connected to GND - OK
- The [HW design checklist](http://ww1.microchip.com/downloads/en/DeviceDoc/KSZ9563RNX-Hardware-Design-Checklist-00002770A.pdf) recommends a decoupling cap for each power pin, so it is OK not to stack the power pins.
- Footprint matches Microchip recommendation: including center pad size and 3D model height
- Pin types look OK except LEDs (see below). Note: A few pins are defined as I/O on the datasheet but I think that they are only inputs when used as strap pins on boot. For that, one would use pull-ups or pull-downs so Output type should work OK. 

Things which need fixing:
- The LED pins can work as GPIO so I think that they should be defined as bidirectional 

Thanks!

evanshultz:
Are you talking about table 3-3 at http://ww1.microchip.com/downloads/en/DeviceDoc/KSZ9563R-Data-Sheet-DS00002419D.pdf#page=16? I didn't catch this before, so thank you, but that table doesn't sound like GPIO to me so I want to be sure I understand.

diegoherranz:
No, I mean "LED Override Register" and "LED Output Register" on page 71.

evanshultz:
I see. They can be GPOs. Changed type to Output. Thanks for noticing that.

diegoherranz:
Looks OK now. Merging.
Thanks!

PR #1949, opened 2019-07-03
Change Power Integrations S pin to Power Input type

evanshultz:
You're going to need some background to sort this out. So let me proceed and please ask questions if you have them.

First, here is are representative datasheets (just one from each family):
- LinkSwitch: https://ac-dc.power.com/sites/default/files/product-docs/linkswitch-3_family_datasheet.pdf
- TinySwitch: https://ac-dc.power.com/sites/default/files/product-docs/tinyswitch-4_family_datasheet.pdf
- TOPSwitch: https://ac-dc.power.com/sites/default/files/product-docs/topswitch-jx_family_datasheet.pdf

Here is the TinySwitch block diagram:
image

The S (source pin) is the drain of the main FET. I assume because of this, the pin type was Open Emitter. This is a bit strange already because this pin can't be pulled up to apply voltage/current to a load between this pin and ground, as would be expected of an open emitter. Hmmm...

Furthermore, you can also see that the control circuitry is grounded to the S pin. So there is no choice but to put this pin at the circuit reference which will normally be primary ground.

If a single power supply is built with this IC there's likely no problem. And so this pin type wouldn't be a problem.

On alternative system uses this IC as an auxiliary controller with a bigger power supply. In that case there's likely another power supply controller IC and wouldn't be surprising to have an opamp, comparator, or some other ancillary circuit in the system. Here, multiple ICs will be on a common ground and they will sink their quiescent and active current to the primary ground net. And their pins on primary ground will be Power Input type, requiring the primary ground net to have a PWR_FLAG symbol on it to avoid ERC errors.

Now let's see what happens when we connect the PWR_FLAG symbol (with a Power Output pin) to the primary ground net, also connected to the Power Integrations' IC's Open Emitter pin and run ERC:
image

We get an error. Because typically an open emitter will pull up and that would short out a power supply. So yeah, the pin types on these ICs are wrong.

Here I've correct that problem by making the visible pin Power Input type and any stacked pins to Passive. No screenshot because if the above explanation is clear the file diff is all you need to approve.

---


evanshultz:
Any chance to get this into 5.1.5? In a normal schematic using these symbols currently produces an ERC error so I'd almost lean towards this being a bug.

evanshultz:
Ping! Is anybody able to take a look at this?

A similar issues exists on the LM319 symbol.

myfreescalewebpage:
@evanshultz reviewed ok. I trust in you for the ERC test, this seems logical also to me event if it is not so evident. Merging.

PR #1950, opened 2019-07-03
Add STGAP2S symbols

evanshultz:
https://www.st.com/resource/en/datasheet/stgap2s.pdf
image
image

I question the pin type for pins 6 and 7, but I'm not sure it really matters so I left them as Output like all the other gate drivers.

---


myfreescalewebpage:
@evanshultz OK for Output on pin 6 and 7 when looking at typical applications.
No comment to do on the devices, merging.
Joel

PR #1955, opened 2019-07-06
Fix timing capacitor pin naming of 4538

davefiddes:
HEF4538B-UnitA
HEF4538B-UnitB
HEF4538B-UnitC
HEF4538B-usage

This fixes the pin names used for the timing capacitors used
by the monostable units (A&B) to be Cx as per the datasheet.
These are connected internally to VSS but they do not need
to be connected externally to anything other than the
timing capacitor.

The symbol outline and center has been changed to match
style guidelines (fixing S3.1 and S3.6 validation failures).

Validation passes apart from S5.1 which is common across
all symbols in this library.

The outline of the power supply Unit C has been increased to match Unit A and Unit B to fix S3.1 validation error. I'm not sure if this is correct or not.


myfreescalewebpage:
@davefiddes thanks for this contribution.
No comment to do except that it breaks the existing designs and need to wait KiCad V6 before merging.
We keep this open and it will be merged when the V6 release will be soon.
Joel

davefiddes:
That's great. Thanks for the review.

evanshultz:
Also, please note two things:
1. Pin name corrections can be made at any time. And if something is a true error, we will make breaking changes. But centering is just a convenience thing to update these older symbols as Joel mentioned.
1. A total overhaul of the logic libraries is hopefully coming for v6. So while this PR staying open hopefully helps to ensure this work gets done, the PR may ultimately not be merged in favor of a global, scripted approach to logic symbol generation.

PR #1880, opened 2019-05-24
Bugfix swapped RX and TX Pins of NXP Transceivers

Deadeye5589:
Bug where the RX and TX pin of the NXP Transceivers TJA1021 and TJA1029 where swapped and dit not match the datasheet. 
Fix related to issue no. #1875

**Datasheet Links:**
[https://www.nxp.com/docs/en/data-sheet/TJA1021.pdf](url)
[https://www.nxp.com/docs/en/data-sheet/TJA1029.pdf](url)

**Example Picutures:**
Fixed_TJA1021
Fixed_TJA1029

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

PR #1965, opened 2019-07-11
added TJA1057T as an alias to TJA1051T

sebastianbergt:
datasheet: https://www.nxp.com/docs/en/data-sheet/TJA1057.pdf
screenshot: its just an alias
footprint: its just an alias

---


fenugrec:
I notice this contains a lot of changes to other CAN transceivers too ? Maybe they should be in a separate commit

sebastianbergt:
Hi, I just made the whole library compliant to the checker scripts. What would be the value of making a separate commit with only a part of the library compliant?

evanshultz:
@sebastianbergt 
Thank you for contributing!

Let me mention two things to the point made by @fenugrec :
1. The subject of the PR doesn't mention anything but the ALIASes, so the user expectation would be that nothing else changed. Changes should be grouped in unique comments. It's much easier to review. So even if the PR contains multiple things, use a separate commit for each change in the PR.
1. Even if there are multiple changes that are related, separate PRs aren't a bad idea. Especially if it helps to segregate additions from changes/fixes.

Here is one reason why: Adding symbols won't affect anything else. But changing pin locations or symbol size on existing symbols, which could be placed into current schematics, is not allowed just anytime because we don't want a working schematic one day to be updated and break tomorrow.

Pin name offset won't break existing schematics so the changes are OK to do any time. I'll review everything together and we can merge this all at once. The above is FYI for the future.

Review:
1. You can remove the `DFN-8-1EP/` from `TJA1029TK`'s description. Thanks for catching that.
1. For `TJA1057T`, change `silent mode` in the description to `Silent Mode Input` since this is an option.
1. For `TJA1021` and `TJA1029` the pin name and type changes are not accepted. Even though NXP seems backwards we will name pins per the datasheet. This was corrected recently at https://github.com/KiCad/kicad-symbols/pull/1880.
1. NC pin placement changes on lines 1282-1310 need to be reverted. The symbol is +/-400mil wide so the adjusted placement puts the pins outside the symbol body where they could be accidentally connected. The NC pins should be on the symbol body outline, where they were, if possible.

Thanks again for adding to the library and making what's there even better!

myfreescalewebpage:
@sebastianbergt can you fix the above comments of @evanshultz please ?

sebastianbergt:
Hi Joel,

I am sorry, I am no longer using KiCAD these days. Shall I delete my PRs?

Best regards,
Sebastian

--

myfreescalewebpage:
No we just put tag "Abandoned" so if another user want to continue working on this it is still possible.
Thanks for your message

PR #1967, opened 2019-07-11
added MB85RS256TY as an alias to MB85RS16 in Memory_NVRAM

sebastianbergt:
datasheet: https://www.fujitsu.com/us/Images/MB85RS256TY-DS501-00046-6v0-E.pdf
screenshot: just an alias

---


evanshultz:
Is there a non-automotive version of this IC? If so, please use that one. Otherwise it looks OK.

sebastianbergt:
The others are non automotive. 
However I misread the datasheet. I thought those would run at 5.0V instead of 3.3V. They have the same voltage characteristics and the same footprint as the ones that are already in the library. So I guess, it is not worth adding it to the library now.

Issue #951, opened 2018-09-18
 Transistor_BJT: S5.1 and S5.2 errors


antoniovazquezblanco:
KLC Rules:
http://kicad-pcb.org/libraries/klc/S5.1/
http://kicad-pcb.org/libraries/klc/S5.2/

Splitting issue #28 I have found the following violations:
```
Checking symbol 'BC846BS':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'SC?70*' does not match the footprint 'Package_TO_SOT_SMD:SOT-363_SC-70-6' set for this symbol.
     - could not match 'SOT-363_SC-70-6' against filter 'SC?70*'
     - could not match 'Package_TO_SOT_SMD:SOT-363_SC-70-6' against filter 'SC?70*'
    Footprint filter 'SC?88*' does not match the footprint 'Package_TO_SOT_SMD:SOT-363_SC-70-6' set for this symbol.
     - could not match 'SOT-363_SC-70-6' against filter 'SC?88*'
     - could not match 'Package_TO_SOT_SMD:SOT-363_SC-70-6' against filter 'SC?88*'
    Symbol has a footprint defined in the footprint field, but several (3) footprint filters set. If the symbol is for a single default footprint, remove the surplus filters. If the symbol is meant for multiple different footprints, empty the footprint field.
checking ruleS5.2
checking ruleS5.1
Checking symbol 'BC856BS':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'SC?70*' does not match the footprint 'Package_TO_SOT_SMD:SOT-363_SC-70-6' set for this symbol.
     - could not match 'SOT-363_SC-70-6' against filter 'SC?70*'
     - could not match 'Package_TO_SOT_SMD:SOT-363_SC-70-6' against filter 'SC?70*'
    Footprint filter 'SC?88*' does not match the footprint 'Package_TO_SOT_SMD:SOT-363_SC-70-6' set for this symbol.
     - could not match 'SOT-363_SC-70-6' against filter 'SC?88*'
     - could not match 'Package_TO_SOT_SMD:SOT-363_SC-70-6' against filter 'SC?88*'
    Symbol has a footprint defined in the footprint field, but several (3) footprint filters set. If the symbol is for a single default footprint, remove the surplus filters. If the symbol is meant for multiple different footprints, empty the footprint field.
```

PR #1970, opened 2019-07-12
Change BC8[45]6BS footprint filters

chschlue:
Fixes issue #951

---


evanshultz:
@chschlue 
Thanks for taking on a bug fix with your first contribution here!

We want to go about this a bit differently. Since these are fully-specified parts with a defined package in the datasheet, we will keep the existing default footprint. But the footprint filters can remove all but `SOT?363*`. That should work.

chschlue:
@evanshultz like so?

evanshultz:
Indeed. Thank you!

PR #1981, opened 2019-07-14
Add XC6206

fenugrec:
Add XC6206PxxxMR LDO Positive voltage regulator (fixed), SOT-23

Some comments:

- Part available in 0.1V increments so it made more sense to me to wildcard the MPN
- It is an LDO, but Torex doesn't call it exactly that.
- tons of second-source mfg with different prefix than XC
- checklib.py claims it's violating S3.1 (not centered) - ok, but all other 3-terminal regs are identical.
- checklib claims it's violatin S3.6, not clear why ? Drawing is a re-numbered copy of LM3480. I notice checklib reports S3.6 violations for a lot of other parts, so maybe something's wrong with the script ?

https://www.torexsemi.com/file/xc6206/XC6206.pdf
xc6206

---


diegoherranz:
Thanks for your contribution. We hope to see more of these :)

Could this one be created as an alias of a similar LDO? If possible, it's always easier to maintain.

Thanks!

fenugrec:
@diegoherranz , I could check again for aliases, but I don't remember seeing similar LDOs with the same pin numbering ?

diegoherranz:
I've had a look and I think that `APE8865N-12-HF-3` may be a good candidate where to add an alias. Could you double check and if you agree turn the XC6206 into an alias?

Thanks!

fenugrec:
Thanks, symbol-wise I agree they're equivalent. I just ran a test with xc6206 as an alias of that APE8865 part you mentioned. Is it expected that in the symbol editor, the text field for "component name" shows the full APE8865* number ? Dropping the component in a schematic seems to work as intended, i.e. the component is correctly named XC6206*.

diegoherranz:
Yes, I think it is expected. See that when you double click on the XC6206 alias on the libraries menu on the left, it will automatically take you to the APE8865* symbol. That is why you see the APE8865* number for "component name".

Thanks!

fenugrec:
Done

diegoherranz:
Thanks for your changes.

Could you add Torex to the keywords so that it shows up if someone searches that term?

Also, I'm wondering what to do with the -G suffix. Options:
1. Creating 2 aliases: MR and MR-G. I think this would go against KLC S2.2.
2. Adding another wildcard: MRx
3. Adding another wildcard variant b: MR-x
4. Leaving it as is: MR

What do you think? I think I would go for 2 or 3. Maybe a slight preference for 3.

Thanks!

fenugrec:
> Could you add Torex to the keywords
yep

> -G suffix

that's the for halogen / rohs stuff; regular PN doesn't include the trailing "-" of course. So "MRx" I think would be meaningless since the G variant isn't called "MRG", but "MR-G".

I do think this is just adding noise to the lib; I would just go with #4 (especially given the number of second-source parts, which may have different suffix conventions). But I do realise the KLC specifically mentions the rohs case as an example, so I guess I should add another wildcard for MR-G in the .lib, and add a $CMP block in the dcm.

diegoherranz:
> so I guess I should add another wildcard for MR-G in the .lib, and add a $CMP block in the dcm

[KLC S2.2](http://kicad-pcb.org/libraries/klc/S2.2/) explicitly discourages creating aliases for cases like this:

_To capture every possible combination of part variation requires a large number of symbol aliases - this is to be avoided._

So I think that option 1 is not possible and it's got to be 2, 3 or 4.

I understand that in this case, because of the extra dash, it's a bit weird. Maybe option 2? Any help @KiCad/librarians? What has been done in the past for similar cases? If we agre something, I can update S2.2 so it is clear for similar cases in the future.

Thanks!

fenugrec:
I still prefer 4, but how about changing the PN to "MRxx" It would match MR and MR-G.

diegoherranz:
> I still prefer 4, but how about changing the PN to "MRxx" It would match MR and MR-G.

Good idea. I think it's the option I prefer. Could you rename it?
Thanks!

fenugrec:
Done. I'm not sure why it doesn't show up here yet, usually github detects the forcepush pretty fast...

diegoherranz:
It doesn't show up yet.
I can see it on your fork, though. Weird...
Let's wait a bit and see.

Thanks!

evanshultz:
We have usually just left off trailing wildcards, which would be option 4. I can't come up with an example right now, though...

fenugrec:
Ah ! I had just re-pushed with the trailing wildcards, but I'm more than happy to go back to what I had initially, i.e. XC6206PxxxMR

fenugrec:
So, do I go back to no trailing wildcards ?

diegoherranz:
Sorry, I had forgotten about this one.

If we usually left off trailing wildcards, I'm OK for you to do the same here.

Thanks!

diegoherranz:
Also, I've just noticed that depending on the voltage, the output current can go from 60 to 250 mA. Can you change the description to say 60-250mA? And can you remove the current from the keywords? I guess it doesn't make much sense there given it is in the description too.

Thanks!

fenugrec:
No worries. Good point about the variable current rating, I hadn't noticed.

- updated description to 60-250mA
- removed spurious keyword
- removed trailing wildcards

diegoherranz:
Thanks for the changes. I think it looks good so I'll merge.

Just a minor comment regarding force-pushing commits. Every time you force-push, the reviewer has to review everything again because you lose the history of what's changed. In this case, given it is a small patch, it's not hard to review again, but especially on big submissions, it is desirable to just review the newest commit which will be a very small diff.

Note: Travis reports Violating S3.1 (Origin is centered on the middle of the symbol) and S3.6 (Pin name position offset) but that's for the existing symbol so in case that had to be fixed (which may be OK for this symbol), it would be a different story altogether.

diegoherranz:
Forgot to say, we always squash the PR before merging, which means it becomes a single commit in the repo, so don't be scared of having many commits in a PR, it won't pollute the history.

Thanks again for your contribution and I hope to see more submissions :)

fenugrec:
Thanks. I wanted to spare you guys the trouble of rebase/squashing, but I'll leave you that priviledge next time !

diegoherranz:
It's really easy to do, it is a single click operation in GitHub so don't worry about it.

PR #1985, opened 2019-07-16
Add ICE40HX4K-BG121 alias

twam:
Only added an alias to existing component.

Although no differences to be expected as this is a pin-compatible 'lower' version of the same device I checked all pins individually.

---


twam:
Any chance to get this merged?

cpresser:
Hi,
I did find some pins that are not identical
- J7
- H4
- H5
- D6

I did this comparison by taking the two .xlsx files, sorting them by the BGA-121 row and putting them next to each other.

twam:
You're right! I only checked that all pins on the lower version equal the higher one, but didn't expect that there are more pins on the higher version.

cpresser:
you can do this by doing a copy of the existing symbol first.
then change just the pins that differ.
if there are two separate commits for that its easy to review for me.

twam:
Something is fishy there. The excel file misses those pins but in the data sheet they have exactly the same number:
Screenshot 2019-10-17 at 16 42 17

cpresser:
That is indeed weird.
Are there other files you can check? BDSL? Or perhaps from within ICEcube2?

twam:
I wrote an eMail to the technical support from Lattice. Let's see if I get an answer.

twam:
> I wrote an eMail to the technical support from Lattice. Let's see if I get an answer.

Didn't get any reply so far. Back then the Technical Support form didn't work. Tried again to today and was able to file a report. ID is C154546-010620.

myfreescalewebpage:
@twam @cpresser what is the status here ?

twam:
Still no reply from Lattice.

I'm pretty sure, that this is only an error in the excel file, but cannot guarantee that. If you want absolute certainty I suggest to discard this.

cpresser:
Unsure. We have conflicting information.

I did some more googling and found this which indicates they are compatible:
https://github.com/YosysHQ/icestorm/pull/132/files

I compared the two arrays `8k-bg121` and `8k-bg121:4k` and they are the same.
That is one more data-point to show they are identical.

The pinout .xls files have not been updated since 2017
image


There are OrCad symbols for download, But I don't have the software to open them for inspection.

cpresser:
The BG121 parts are not in the OrCad file. Yay.
But there is another document called MigrationSheet. There the 4k and 8k are identical. 
Also people on IRC confirmed that they have the same die, so why would they be bonded differently.

PR #1986, opened 2019-07-16
added MAX7456

rs-development:
Not sure about the library, but i could not find similar devices.

https://datasheets.maximintegrated.com/en/ds/MAX7456.pdf

image


rs-development:
I compressed the symbol, because the last ones where no big :), we will see what you think.

rs-development:
VIN is VIDEO IN but triggers an error because the script thinks it should be power input.

cpresser:
Hi,

The `VIN` and `VOUT` are false positives by travis CI.
But the other items should be fixed.
- [ ] Centering 
- [ ] Don't stack NC pins
- [ ] The EP pin needs to be renamed to AGND (see KLC S4.3)
- [ ] Pin name position offset
- [ ] The symbol has a default footprint, so please add it to the description (KLC S6.3)

I have not checked the Pins yet (position, type, numbering), but will do so soon.

myfreescalewebpage:
@rs-development are you still working on this symbol ?

PR #1988, opened 2019-07-16
Added Texas Instruments TPS61252 switching regulator (tiny adjustable…

g-gg:
… boost)

The repo is already equipped with plenty of buck regulators, but not so many boost.

The TPS61252 has
- adjustable current limit (up to 1.5A)
- adjustable output voltage (up to 6V)
- input voltage range suitable for battery operation
- tiny 2x2mm WSON-8 package

Datasheet: http://www.ti.com/product/TPS61252

---


evanshultz:
Thanks for this contribution. Sorry for the first PR didn't work out but we appreciate your persistence.

Please post an image of the symbol as requested above, ideally showing the pin types too. I've done it here so you can see what we're looking for.
image

1. Let's condense the description a bit. Start with the datasheet description and then add other details after it. How about ` Tiny 1.5-A Boost Converter With Adjustable Input Current Limit, 2.3-6.0V Vin, 3.0-6.5V Vout, WSON-8`.
1. Pin name offset should be 20mil.
1. Footprint filter should be `WSON*1EP*2x2mm*P0.5mm*`.
1. Stack pins 1 and 9 as the datasheet says they must be connected on page 3.
1. Pin 5 type should be Open Collector (KiCad doesn't have Open Drain).
1. Pin 7 should perhaps be Passive type since it doesn't match what the ERC engine thinks is Power Input.
1. Move pin 8 to the top, and then you could shrink the symbol vertically.

g-gg:
Wow... thank you for your professional and kind feedback. This is my first symbol I contribute to the kicad library, I've learned a lot. And I really didn't recon with a response that fast.
I hope now everything is ok. Here is the new screenshot:
Screen Shot 2019-07-17 at 20 44 08

evanshultz:
:) And thank you!

Just two things:
1. Now with a pin at the top please put the ref des at top left and value at top right.
1. The stacked power pin (pin 9) must be Passive type. See the last clause at http://kicad-pcb.org/libraries/klc/S4.3/. (Now that I read it again that paragraph could really use a re-write for clarity.)

g-gg:
Thanks for your guidance. I even read that page, but somehow didn't understand it the first time.

Screen Shot 2019-07-17 at 21 01 44

evanshultz:
Looks great! Thank you!

We appreciate your first contribution here and hope to see you back soon!

PR #1990, opened 2019-07-17
add HMC1099PM5E and HMP8500PM5E power amplifiers

lorem-ipsum:
This adds the [HMC1099PM5E](https://www.analog.com/media/en/technical-documentation/data-sheets/HMC1099PM5E.pdf) and almost identical [HMC8500PM5E](https://www.analog.com/media/en/technical-documentation/data-sheets/HMC8500PM5E.pdf) RF power amplifiers. There is also the (now EOL) [HMC1099](https://www.analog.com/media/en/technical-documentation/data-sheets/HMC1099.pdf), but it uses a slightly different footprint.

It is probably the biggest pin stack I have every created... The NIC pins should also be connected to GND, so I've added them next to it. Footprint has been created using the generator, PR will be there in a minute.

Screenshot_20190717_221631

---


lorem-ipsum:
So about the footprint PR...

I've just realised that I posted a footprint with that name already in a different PR for a different part. However the two are not completely identical. The package height is different, but also some tolerances differ.
- [Analog cp-32-2](https://www.analog.com/media/en/package-pcb-resources/package/414143737956480539664569cp_32_2.pdf) (added in previous PR)
- [Analog cg-32-2](https://www.analog.com/media/en/package-pcb-resources/package/pkg_pdf/lfcsp-cav/CG-32-2.pdf) (this package)

How should I proceed with this? Should I create them as manufacturer specific versions? And if so, where should the package name go?

myfreescalewebpage:

HMC1099PM5E
- [ ] Footprint filter should be `LFCSP*1EP*5x5mm*P0.5mm*`
- [ ] NIC pins should not be stacked and should be on the border outline of the symbol or inside the symbol. You can reduce the length of the NIC pin to have all of them inside. We can also accept they will not be on the 100mil grid to manage this.

Cheers,
Joel

lorem-ipsum:
Hi @myfreescalewebpage,

thanks for the review. I fixed the footprint filter. I would like to keep the NIC pins accessible as the datasheet states that all characteristics are measured with these pins tied to ground. So while they are not connected internally they are required to match the specifications. Hiding them in the body will make this impossible to implement.

Cheers,
Hendrik

myfreescalewebpage:
In this case I suggest to stack the NIC pins with GND. What do you think ? Joel

evanshultz:
Is there any situation where the user wouldn't connect all NIC pins to the same potential? Meaning do we need to consider a case where the NIC pins can't all be stacked?

For the footprints, they are different. Let's go with `Analog_LFCSP_CAV-32-1EP_5x5mm_P0.5mm_EP3.1x3.1mm.kicad_mod` and `Analog_LFCSP_VQ-32-1EP_5x5mm_P0.5mm_EP3.1x3.1mm.kicad_mod`. The part of a FP filter `...LFCSP*1EP...` will catch either one but I think this will work. Does that sound OK?

myfreescalewebpage:
To be very accurate the datasheet indicates "No Internal Connection. These pins are not connected internally. However, all data was measured with these pins connected to RF and dc ground externally."

For some pcb routing maybe all NIC will not be connected but because it's an RF device, I presume it's not very recommended to do some strange thing with the pins.

@lorem-ipsum your advice ?

lorem-ipsum:
Its just not clean what to do based on the datasheet. In my application I tied them all to GND, just to be sure.

I chose the original solution of having them both next to each other as it leaves the decision to the end user. Tying them together is a matter of seconds and it is very clear what happens. If they are stacked with GND or burried inside it is not clear from the symbol what will happen with them, requiring users to check the symbol in detail and possibly edit it to their desired behaviour.

myfreescalewebpage:
What trouble me in this case is that the NIC pins are connected all together, and there is not the possibility to not connect them together. The design of Analog is probably something to follow, I suggest to stack the NIC pins on GND.

lorem-ipsum:
I stacked them on top of the GND pins and rebased against master. Now looks like this:
HMC1099PM5E

myfreescalewebpage:

PR #1992, opened 2019-07-17
add DAC1220E 20-bit DAC

lorem-ipsum:
This adds the [DAC1220E](https://www.ti.com/lit/ds/symlink/dac1220.pdf) from TI.

I am not really sure about the correct type for the VOUT pin. The script would like it to be a power output due to the name. Is this correct?

Screenshot_20190717_222834
---


cpresser:
Hi, a few suggestions
- [ ] space XIN and XOUT so a `Crystal_Small` fits between the pins.
- [ ] suggested pin order on the right side: `VREF` `C1` `VOUT` `C2`. And space them so that C_Small fits between them.
- [ ] `VREF`should be power input, so DRC complains if there is nothing driving the net
- [x] footprint and filter are good
- [x] descr. and keywords are good

lorem-ipsum:
thanks for the review. All fixed, now looks like this:
Screenshot_20191225_154636

cpresser:
Looks good. thanks.
One more thing: I took a look at other DAC symbols regarding the `Vref` pin. For example `MCP4901-EMC` or `MAX5215`. They have `Vref` at the top, pin-type is power-input. This also conforms to KLC.
Moving Vref to the top is also possible here. With C1 on the left-top, wiring up C1 is also not an issue.

chschlue:
> * [x]  `VREF`should be power input, so DRC complains if there is nothing driving the net

Really?
Apart from the fact that Vref inputs aren't supposed to draw current (and by extension little to no power), there are lots of cases where the pin isn't going to be connected to an actual power output (but to a shunt reference, series reference, run-of-the-mill zener diode, basic voltage divider, whatever), so DRC warnings are the problem here, not the solution.

cpresser:
> > * [x]   `VREF`should be power input, so DRC complains if there is nothing driving the net
> 
> Really?
> Apart from the fact that Vref inputs aren't supposed to draw current (and by extension little to no power), there are lots of cases where the pin isn't going to be connected to an actual power output (but to a shunt reference, series reference, run-of-the-mill zener diode, basic voltage divider, whatever), so DRC warnings are the problem here, not the solution.

No. Not really. Thanks for pointing that out.
Dunnow what was going on in my head when I wrote this. Perhaps I was locked into thinking that a voltage-reference-chip needs to be connected there.

Pin-type `Input` will throw a DRC error it its unconnected, but not if its connected to a resistor divider or similar passives. So that seems like the way to go.
@chschlue Any Ideas about the pin-position? If we make it an input it should go to the left. And to make wiring easy C1 should be close by. One option would be to switch places of XIN/XOUT with Vref and C1.

evanshultz:
A floating `Input` pin will not create any warning or error by the ERC engine at the default settings. Check the pin matrix. That seems like a fine type to me, though.

Putting it on the right side makes sense here. To me, at least. It's a voltage for the analog side of the IC, and that's the right side. IMO it might be better to flip VREF/C1 and VOUT/C2 pins for easier wiring and access to ground, but I think the right side works well. Putting inputs all on the left is a rather weak rule and often overridden if there are wiring or aesthetic concerns.

cpresser:
As far as I understand Input<->NC will generate an error. Those are the defaults in my KiCad installation.
image

I would rather have `Vref` on the top of the right side for access to positive supplys. I don't see why one would need GND more than positive voltages from the top. What am I overlooking?
My bike-shed has the pins as shown in the recent screenshot from @lorem-ipsum :smile:

chschlue:
I'm with @evanshultz on the type but I don't think putting it on the right side is a good choice.
This isn't a feedback input. Why would you want your reference voltage circuitry to be on the right in the schematics?
Also, why would you want it to be on top?
Generally speaking: Vref not being your supply voltage is exactly the point of having a separate pin.
And in this particular case:
AVdd is 0-6V while Vref is 2-3V _and_ it's a 20bit DAC. I would want a precise reference.
(C1 should go wherever Vref ends up, BTW)

evanshultz:
Oh yeah. I'm stupid. `Input` does generate an error when unconnected.

@cpresser 
If KLC doesn't control it and there isn't an informal guideline about it (this library lacks consistency in many places), then you can decide pin placement before you merge unless you want to block this PR based on resolution of discussion on a topic. The logic I wrote above has been used on a number of symbols throughout the library but it's not a rule. Putting myself in the shoes of a contributor, I'd rather see my contribution merged if the sticking point wasn't codified as a rule and is subjective.

chschlue:
Leaving Vref floating _is_ an actual error.
`Input` looks perfectly fine to me, or am I missing something?

cpresser:
Layout proposal:
image

cpresser:
> Leaving Vref floating _is_ an actual error.
> `Input` looks perfectly fine to me, or am I missing something?

Yep. `Input` is good.

lorem-ipsum:
I changed the pins around as suggested by @cpresser and rebased against the current master.

Now looks like this:
DAC1220E

cpresser:
Thanks!

PR #2001, opened 2019-07-22
add PAT attenuator series from Mini-Circuits

lorem-ipsum:
This adds the series of PAT attenuators from Mini-Circuits. They use the AF320 footprint added in [this PR](https://github.com/KiCad/kicad-footprints/pull/1732). The symbol is based on the LAT-3 symbol, only the pin types have been set to passive.
Screenshot_20190722_222747

---


lorem-ipsum:
as the number of attenuators is increasing, should they be moved to `RF_Attenuator.lib`? I could start this with this PR and make another one for moving the existing attenuators.

aewallin:
If the right pin-types for LAT-xx is Passive for both input and output, then both LAT and PAT can use the same single symbol - and just add many aliases?

I agree that a separate rf_attenuators library is OK if there are tens of different attenuator aliases.

lorem-ipsum:
I think they should be passive on the LAT models as well. These are completely passive attenuators without any active components.

myfreescalewebpage:

- [ ] All datasheets link to be updated

Else no comment, just need to wait the footprint.

Cheers,
Joel

lorem-ipsum:
Hi @myfreescalewebpage ,

I updated the datasheet URLs to the standard www.minicircuits.com.

Cheers,
Hendrik

myfreescalewebpage:
Thanks, just need to wait the footprint. Thanks @chschlue for the missing tag.

PR #2003, opened 2019-07-22
Add TPA3251

evanshultz:
http://www.ti.com/lit/ds/symlink/tpa3251.pdf

Pins 9 and 10 were a little hard to figure out their type, but based on page 23 I chose Input.

image
image

I've generated a scripted footprint, but haven't submitted it yet pending the resolution of https://github.com/KiCad/kicad-footprints/issues/955.
image

---


evanshultz:
The CI failures are OK, I believe. `AVDD` and `DVDD` are internal regulator outputs that cannot source current and just need a cap to ground, making Passive a fine choice.

cpresser:
A few questions & remarks:
- [ ] Why did you split it this exact way? I would rather expect 3 units. A: all Controll-Stuff, B+C one channel each. Or just make one single unit. But having control pins split over two units seems weird.
- [x] I agree for the pin-types of `AVDD` and `DVDD`. Also the position bottom-left is good.
- [ ] `OSC_IOP` and `FREQ_ADJ` are outputs
- [ ] `OC_ADJ`and `OSC_ION` are bidir, 
- [ ] `FAULT`and `CLIP_OTW` should be on the right side.
- [ ] I would swap `OUT_B`and `BST_B` so they have the same order as `*_A`. Thats just my personal opinion that this looks better when wiring it up with the bootstrap capacitors.

evanshultz:
1. The logic is the unit A has IC-to-IC interface and 'passive' pins, while unit B has pins that interface with an external processor (or circuit). So would you prefer two units with just audio input, output, and power, with a third unit just for interface and 'passive' pins? Would you then prefer all interface pins on the same side? I'm trying to seek answer and explain my reasoning at this time. You may look at the IR4302 class D IC which has pins split unevenly.
2. Pages 15 and 16 intimates that `OSC_IOP` are bidirectional. Pin type changed. FREQ_ADJ is actually connected just to a resistor for setting the freq as we've used Passive for that. Probably the pin table shows as 'O' because it's a current source inside the IC.
3. `OSC_ION` pin type changed. OC_ADJ is same as `FREQ_ADJ`.
4. In this case I thought keeping all the pins that would go to an external process together was best. So they're all on the left side. (But yes, split across two units.)
5. I thought it was more clean the way it is so that the class D filter can be between the outputs without a bootstrap cap in the middle. But I agree this is rather subjective. No change made for now.

cpresser:
I re-checked the pin-types of the 'controversial' pins. The block-diagram arrows and the pin-table don't always match. And then there is also the example application (e.g. 10.2.1).
Anyway, I don't want to spend to much time discussing pin types. The current types are good enough.

Regarding the layout:
As suggested I looked at `IR4302`. That did help me understand why you did split the symbol this way. I agree that this spit does follow sane rules. Just because I would have done it different does not mean it needs to be changed :smile: 

So in conclusion, I am happy to merge as is. Or is there anything else?

evanshultz:
Thanks for looking again!

I can change things. It doesn't have to be this way. Maybe `IR4302` isn't a good example and no symbols should look like this. If you're OK so am I, but I'm not staunchly opposed to a change.

The footprint needs to be merged before this. And that depends on https://github.com/KiCad/kicad-footprints/issues/955. Sigh.

evanshultz:
Submitted the footprint at https://github.com/KiCad/kicad-footprints/pull/2468.

myfreescalewebpage:

myfreescalewebpage:
@cpresser the footprint has been merged

evanshultz:
I fixed the footprint name referenced by the symbol, so the only remaining Travis errors are fine and described above.

PR #2004, opened 2019-07-23
Add ATX 20pins and 24 pins

jneiva08:
ATX v1.x 20 pins
https://pinouts.ru/Power/atxpower_pinout.shtml

ATXv1

ATX v2.x 24 pins
https://pinouts.ru/Power/atx_v2_pinout.shtml

ATXv2

---


cpresser:
Hi,

a few questions and suggestions
* Is there a real datasheet? I found this one: https://web.aub.edu.lb/pub/docs/atx_201.pdf , but none for 1.x. It would be nice to have official ones so we can verify the pins/names
* why did you call them ATX12_? what does the 12 indicate?
* KLC says positive power should go to the top, but I think we can consider them as outputs. A LDO in the current lib has its output also on the right side. So I am okay with that.
* ~~PS_ON for the 20pin connector seems to be wrong. At least the .pdf above has it at pin16 as well. AFAIK there are ATX-PSUs that have a 20+4 pin connector, so they are compatible with both versions. That also suggests that the first 20pins should not change.~~ stupid me. Looked at the pin-numbering of the connector. The numbers of the pins in the 2nd row all shit by 2 when going from 20->24pin connector.
* Description and Keywords should be populated

chschlue:
@cpresser 
Not sure if anything more official than https://www.intel.com/content/dam/www/public/us/en/documents/guides/power-supply-design-guide.pdf#page=26 is freely available.

The 12 probably refers to ATX12V.

The names are wrong anyway, the 24-pin connector was introduced in ATX12V v2.2, so 20-pin is ATX <= 2.1. This also means your datasheet link is fine for checking the first symbol.

jneiva08:
@cpresser @chschlue , have renamed the symbol names to be easy to understand.

In the motherboard normally can find the vertical connector, you can find on last one are horizontal

cpresser:
Thanks for changing the datasheets. The new ones look way better. 
- [ ] Can you please add `#page=13` anchors to the .pdf link?
- [ ] Lets rethink the naming. They are both `ATX12V` since both have a 12V rail. How about just adding the pin-numers to name? Like `ATX-20` and `ATX-24`. Similar to those symbols: 
image
- [ ] `PWR_OK` should be on the right side, since its an output.
- [ ] Please move `PS_ON#` to the top of the left side.
- [ ] Please fix the other errors reported by travis
- [ ] Is there a reason you ordered the voltages on the right side the way you did? My first idea would be to sort them ascending, with `-12V` at the bottom. But any other scheme is also fine as log as it is not arbitrary.

jneiva08:
@cpresser 

Have reorganized the pins and changed the name, but atx-24 can't center it without going to 1.27mm grid.

Screenshots:
ATX-20
atx-20

ATX-24
atx-24

I don't know if the footprint filter is the best. and it is giving error because of datasheet i don't know why.

cpresser:
This is the change you are looking for to fix the travis error.
image
Delete the visible datasheet-field in the editor window to get this result.

The footprint-filters look good. I tested them im cvPCB and they worked fine.

jneiva08:
@cpresser , have already made the fix.

jneiva08:
Hi, @cpresser @chschlue , still need to make more changes?

cpresser:
Sorry for the delay.

I gave it another look and found a small issue with the `-5V`rail:
* Add a `-5V` pin also for the 20-pin version (on pin 18 afaik). According to wikipedia it was made obsolete in ATX 1.3 but was present for older PSU.
* Make the two symbols the same size, so one can be used as drop-in-replacement for the other
* Rename `-5V` to indicate it is not always available. perhaps `-5V/NC`. Otherwise designers might depend on the voltage to be available which might not be true.

The current naming-scheme is also compatible with the upcoming `ATX12VO` system. It would be called `ATX-10` according to the pin-count and thus not create a collision.
https://www.intel.com/content/dam/www/public/us/en/documents/guides/single-rail-power-supply-platform-atx12vo-design-guide.pdf

jneiva08:
@cpresser Have already made both symbols size and pin compatible, also renamed pin 18 from ATX-20.

cpresser:
looks good. thanks!

jneiva08:
thanks for your time

PR #2009, opened 2019-07-24
add RF switch HCM849A

hvraven:
This adds the RF switch [HMC849A](https://www.analog.com/media/en/technical-documentation/data-sheets/hmc849a.pdf).

The footprint does not match the one specified in the datasheet, but rather the one on the product page of Analog under [Sample & Buy](https://www.analog.com/en/products/hmc849a.html#product-samplebuy). As the dimensions in the datasheet seem to lack the nominal values I had issues creating a proper footprint for it. The one on the product page is a standard housing from Analog. However the dimensions do not exactly match (especially the pad size is different). The matching footprint cat be found at https://github.com/pointhi/kicad-footprint-generator/pull/398.

I put the NC pins towards the bottom as they should be connected to ground. The rule checker is not happy about having them visible and stacked.

Screenshot_20190724_082605

---


cp-aquila:
A really quick review:
1. The footprint-filter is set to a wrong package.
2. some of the travis-warnings will vanish when you set pin4 to passive as well.
3. The NC pins that can/should be connected are pretty weird. Not sure how to handle this. One option would be to not name them "NC" but still stack them. Then all travis errors will be gone. But that violates the rule that the pin names should reflect the datasheet 1:1.
4. I assume you put EN on the bottom, since it might get tied to GND?

poeschlr:
And a merge is required.

lorem-ipsum:
did the merge. About the comments:
1. this is somewhat on purpose and mentioned in the description
2. & 3. I know travis doesn't like this, but there is no way to make Travis like this. I could set the NC pins to passive, that would at least make some sense.
4. didn't think to much about it to be honest. With this argument it would probably be better to put in on top. Tying it to ground would make the device a bit pointless.

cp-aquila:
>     1. this is somewhat on purpose and mentioned in the description
The filter is set to "SOP*3.76x4.96mm*P1.27mm*" the footprint however is set to "Package_CSP:LFCSP-16-1EP_4x4mm_P0.65mm_EP2.4x2.4mm"
 
>     4. didn't think to much about it to be honest. With this argument it would probably be better to put in on top. Tying it to ground would make the device a bit pointless.
The Datasheet says about the EN Pin 
> "An enable input (EN) set to logic high will put the switch in an “all off” state."
So its actually a Not-Enable Pin. At least thats how I read it.

lorem-ipsum:
Ah. Sorry I did not get that on your first response. Thanks for spotting that, it is fixed now.

You're also right about the enable pin. Should be call `~EN` IMHO, but sticking to the datasheet is probably better.

evanshultz:
1. Perhaps change description to `High Isolation SPDT Non-Reflective Switch, DC-6GHz, Matched To 50 Ohm, CSP-16`.
1. Remove `EP2.4x2.4mm*` from the footprint filter.
1. I would prefer if VDD and GND were aligned vertically.
1. Would it make any sense to make each NC pin unique for layout reasons? Or separate 4, 6-8, and 13-16 into separate pin groups?
1. I somewhat feel like the AC impedance matching is excessive. But no strong feelings either way.

myfreescalewebpage:
@lorem-ipsum can you follow @evanshultz recommandations ? Then we can make a final check together and merge the symbol. Thanks. Joel

myfreescalewebpage:
@hvraven ping

myfreescalewebpage:
@hvraven ping
Thanks
Joel

myfreescalewebpage:
For the two point, ok I agree. NC pins like on #1990 is great. Finalize this and I will make a final review. Thanks

hvraven:
Stacked the NC pins on top of GND, now looks like this:
HMC849A

myfreescalewebpage:
Thanks, Just one comment before merging:

- [ ] The package in the description should be "LFCSP"

Joel

hvraven:
Hi @myfreescalewebpage ,

thanks for checking. Corrected the package description. 

Cheers,
Hendrik

myfreescalewebpage:

Issue #579, opened 2018-05-05
Issues with Interface_UART


evanshultz:
@jkriege2 
Couldn't https://github.com/KiCad/kicad-symbols/pull/548 have been an ALIAS? In fact, it seems like the following could all be ALIASes of a single symbol. Can you check me?
- MAX14783ExS: https://datasheets.maximintegrated.com/en/ds/MAX14783E.pdf
- MAX481E:  https://datasheets.maximintegrated.com/en/ds/MAX1487-MAX491.pdf
- SN75LBC176D: http://www.ti.com/lit/ds/slls067h/slls067h.pdf
- SP3481CN: http://www.icbase.com/pdf/SPX/SPX00480106.pdf

This lib also has some extended temp symbols.

SIPEX parts should be using Exar datasheets since they were acquired instead of some 3rd party datasheets.

Many descriptions do not have the package type at the end.

jkriege2:
Hi!

hmmm ... partly you're right. these could be ALIASes:
- ST485EBDR
- MAX14783ExS
- SP3481CN (+its ALIASes)
- SN75LBC176D
These not (as the symbols cover several footprints that are not available for each:
- MAX481 (unless we split it by package: DIP-8, SOIC-8, µMAX-8


In addition these should have the same style/internal schematic as the ones above:
- SP3481CP (+its ALIASes)
- SN75LBC176P

Best,
JAN

PR #2011, opened 2019-07-24
Add (revise) TI SN75176 RS-485 transceivers

chschlue:
Add (revise) TI SN75176 RS-485 transceivers

See also #2012 

http://www.ti.com/lit/ds/symlink/sn75176a.pdf
http://www.ti.com/lit/ds/symlink/sn65176b.pdf
http://www.ti.com/lit/ds/symlink/sn55lbc176.pdf

sn75176

---


evanshultz:
@chschlue 
Can you please clarify your intent? I only see additions in the Git diff. You aren't revising anything, right?

chschlue:
See #2012

chschlue:
The old symbols only have aliases for SN[5-7]5LBC176* components, split into two symbols for specific packages each.
The new one has aliases for all the functionally equivalent SN[5-7]5(LBC)?176* parts, plus decoration copied from MAX485.
IMHO, this way the arrangement is more logical.

I only just noticed I screwed up the LCCC variants.
Will remove them from this PR for now.

evanshultz:
Thank you for clarifying.

1. We don't accept temp variants, so please remove the 55* and 65* ones, leaving just 75* ones.
1. You didn't show a datasheet for `SN75176`, so please remove it unless you have something more to add. (Note that this and the above item leave 3 symbols to add.)
1. Perhaps the description can be in this form `Differential Bus Transceiver, RS-485 / RS-422 Half-Duplex, 5 Mbps, DIP-8/SOIC-8`. Perhaps add `, CMOS` before the packages for the LBC version.
1. The library is moving towards unique symbols for each package. Would you be willing to do that? This is more like the package-specific symbols #2012 is removing.

There are minor things which don't have to be done by I will document here:
1. There are 11 polylines but I believe only 10 should be required.
1. There is an extra rectangle.
1. There are no filled circles at the wire junctions by the A and B pins.
1. The line from the D pin doesn't go into the middle of the triangle.

chschlue:
> 1. We don't accept temp variants, so please remove the 55* and 65* ones, leaving just 75* ones.

LCCC package is only available as 55*, what to do about that? (I could just skip those, assuming no one will miss them anyway)

> 2. You didn't show a datasheet for `SN75176`, so please remove it unless you have something more to add. (Note that this and the above item leave 3 symbols to add.)

SN75176 doesn't exist. This was supposed to be kind of a generic name. Didn't know better back then.

> 3. Perhaps the description can be in this form `Differential Bus Transceiver, RS-485 / RS-422 Half-Duplex, 5 Mbps, DIP-8/SOIC-8`. Perhaps add `, CMOS` before the packages for the LBC version.

OK

> 4. The library is moving towards unique symbols for each package. Would you be willing to do that? This is more like the package-specific symbols #2012 is removing.

Didn't know that was intentional. I thought (part of) the point of having split symbols / footprints was so you could decide on a package later while keeping the symbol.

> 1. There are 11 polylines but I believe only 10 should be required.
> 2. There is an extra rectangle.
> 3. There are no filled circles at the wire junctions by the A and B pins.
> 4. The line from the D pin doesn't go into the middle of the triangle.

Ok.

Perhaps I better scrap this PR and #2012 and start over?
I still think the current 75*176 symbols are a mess, there are variants among them which only differ in tube vs. tape-and-reel packing and such.

Is there any kind of "canonical", backwards-compatible way of cleaning this up?

evanshultz:
1. Removing the LCC package is fine with me.
1. OK. Please remove this symbol.

If you want to start over, I'm fine scrapping this. If you want to make things more universal, and the current mess is probably just due to less strictness about merging symbols, please take care not to cause any changes to existing symbols. Show before and after images of symbols if you do. And also note that using unique symbols where pin names differ on the datasheet (even if the pin function is the same) is helpful for troubleshooting purposes.

Thanks!

chschlue:
@evanshultz 
> 1. We don't accept temp variants, so please remove the 55* and 65* ones, leaving just 75* ones.

Done.

> 2. You didn't show a datasheet for `SN75176`, so please remove it unless you have something more to add. (Note that this and the above item leave 3 symbols to add.)

Done.

> 3. Perhaps the description can be in this form `Differential Bus Transceiver, RS-485 / RS-422 Half-Duplex, 5 Mbps, DIP-8/SOIC-8`. Perhaps add `, CMOS` before the packages for the LBC version.

Done.

> 4. The library is moving towards unique symbols for each package. Would you be willing to do that? This is more like the package-specific symbols #2012 is removing.

Impossible without #2012 
Splitting into *D and *P symbols would clash with existing aliases.

> 1. There are 11 polylines but I believe only 10 should be required.

Fixed.

> 2. There is an extra rectangle.

Removed manually. Kept reappearing when using symbol editor. No idea why.

> 3. There are no filled circles at the wire junctions by the A and B pins.

Fixed.

> 4. The line from the D pin doesn't go into the middle of the triangle.

Fixed.

chschlue:
Found https://github.com/KiCad/kicad-symbols/issues/579 by chance.

PR #2012, opened 2019-07-24
Remove redundant TI SN75176 symbols

chschlue:
Remove redundant TI SN75176 symbols

See also #2011 

---


evanshultz:
We won't remove symbols in the middle of a major KiCad release, so I can't merge this. We don't want users to have symbols on their schematic that suddenly disappear if their libraries are updated. I didn't look closely, but perhaps your two PRs could you combined in a way that works?

chschlue:
That's why I split it into two PRs, so this one could be merged for 6.0 or whatever.
The old and new symbols don't interfere with each other, they're just not needed both in the long run.

PR #2013, opened 2019-07-24
Add PIC18F47J53

asukiaaa:
I add PIC18F47J53._SSOP

I'm not sure the rule of suffix for PIC18.
If SSOP is wrong please let me know.
I will correct the name.

Data sheet: http://ww1.microchip.com/downloads/en/DeviceDoc/30009964C.pdf

Screenshot from 2019-07-24 23-08-52


cpresser:
- [ ] Can you please use short names for the pins? See https://github.com/KiCad/kicad-symbols/pull/460#issuecomment-379525628
- [ ] You can stack the VSS pins since they are in the same power domain
- [ ] The name should be (See Page 562) "PIC18F47J53-I/SS"
- [ ] I can't see the pin types in the screenshot, they are all "???"
- [ ] VDDcore shold go to the top, so should VUSB
- [ ] The Description should end in ", SSOP-28"

Side-Question: Where do you buy the Chip? Mouser, Digikey & Arrow only list the TQFP and QFN variants.

asukiaaa:
Thank you for the feedback.
I updated the symbol.

> Can you please use short names for the pins? See #460 (comment)

I assign shorter name for each pins?
How is it?

>  I can't see the pin types in the screenshot, they are all "???"

I could see the types after changing language to English from Japanese.

Screenshot from 2019-08-20 01-22-29

> Side-Question: Where do you buy the Chip? Mouser, Digikey & Arrow only list the TQFP and QFN variants.

I use 28-Lead SPDIP package of PIC18F27J53 but the datasheet is PIC18F47J53.
https://www.microchip.com/wwwproducts/en/PIC18F27J53

Should I use PIC18F27J53 for name of this symbol?

chschlue:
@cpresser Not even Microchip sells this as SSOP
pic

@asukiaaa Why do you want to add a symbol for a nonexistent chip?

asukiaaa:
@calebns 
Because of this.

> I use 28-Lead SPDIP package of PIC18F27J53 but the datasheet is PIC18F47J53.
https://www.microchip.com/wwwproducts/en/PIC18F27J53

cpresser:
I updated the list above. More hints were added to items that are not yet to spec.

> Should I use PIC18F27J53 for name of this symbol?

Absolutely :smile:

asukiaaa:
Thank you for the feedback.
I updated the symbol.

cpresser:
- [ ] Description duplicates the package name "32K Flash, 3.5K RAM, 1K EEPROM, PIC18 Microcontroller ADC PWM CAN SPI I2C USART in SSOP28 package, SSOP-28"
- [ ]  I don't really like the arrangement of pins. Given that the whole library is a mess, this is rather hard to argue. But we should not include yet another new style.

How about something like this? image 
- That way the symbol is similar to "PIC18F23K20_ISS". 
- Moved U? and the symbol-name
- Placed power pins closer together
- Moved PortA to the top

asukiaaa:
Thank you for the feedback.

> Description duplicates the package name

I updated description to "128K Flash, 3.8K RAM, PIC18 Microcontroller ADC PWM I2C USART USB RTCC, SSOP-28".
I delete EEPROM information because I cannot find EEPROM size on datasheet.

> How about something like this?

It seems good.
I updated so.

Screenshot from 2019-08-24 15-11-59

asukiaaa:
Thank you for checking and merging.

PR #2016, opened 2019-07-24
add CD4097B

greyltc:
Here's a symbol for CD4097B. http://www.ti.com/lit/ds/symlink/cd4097b.pdf

It's very similar to CD4051B that's already in the Analog Switch library, except it's a differential analog 8x1 MUX (aka two 8x1 MUXs in the same package).

Here's what it looks like:  

image
---


evanshultz:
Welcome and thank you for contributing!!

1. Remove the footprint filter `SO*5.3x15.0mm*P1.27mm*` since it doesn't apply.
1. I don't see the DIP package available at http://www.ti.com/product/CD4097B/samplebuy, and Digikey shows no other vendors and TI's DIP package as obsolete. So perhaps it can be removed too?

I would argue that the X and Y pins should be Passive type, but since CD4051 is Bidirectional there's no point.

greyltc:
Hey, thanks for looking over the symbol.

1. I put that filter in because I think it's a valid match for the bottom package listed on the page here: http://www.ti.com/product/CD4097B/pinout-quality , `SOP (NS)`. The 16 pin version of this footprint is already in KiCAD (SOIC-16W_5.3x10.2mm_P1.27mm) but it seems that nobody has drawn the 24 pin version for KiCAD yet. Honestly though, I can't really figure out who/if any distributor actually sells this package. Do you still want me to remove it?
1. I'm not 100% sure if the DIP is manufactured anymore or not. I was able to find some stock [here](https://www.verical.com/pd/texas-instruments-interface---analog-switch-mux-CD4097BF-5063624?utm_campaign=octopart_2018&utm_currency=GBP&utm_keyword=CD4097BF&utm_medium=aggregator&utm_content=inv_listing&utm_source=octopart) and [here](https://rcfreelance.com/p/ic/182918/cd4097bf/?source=octopart) for example, since I can still buy that package today, I'm of the opinion that the DIP matching filter should remain.

evanshultz:
1. That package isn't available from TI nor is it in Active status at DigiKey. But based on your 2nd response above somebody could possibly troll the internet and get them or perhaps find some in their parts box.
1. The 2nd link is for two parts date coded 2001. Hobbyist only. At best.

I think that takes care of everything then. Thanks for making your first contribution here and hope to see you again soon!

PR #2017, opened 2019-07-24
PCM2902, ISD25xx added

EpeR1:
I have pushed some Audio IC symbols.

### PCM2902:
Doc: http://www.ti.com/lit/ds/symlink/pcm2902c.pdf
Created symbol:
pcm2902

### ISD2560/75/90/120
Doc: http://pdf1.alldatasheet.com/datasheet-pdf/view/83094/WINBOND/ISD2560.html
Symbol:
isd25

---


evanshultz:
@EpeR1 
Welcome and thank you for contributing to the libraries!

Since the connector is not really related and in a different library, could I ask you to split it off? It would really be nice to have all three symbols in unique PRs, but I'll review the first two together if you keep them here.

EpeR1:
@evanshultz 
Okay
Sorry for my mistake!
It is separated now.

cp-aquila:
Hi,

(i just did a quick look at the ISD2560 datasheet)

regaring the ISD2560 symbol, I think you should change the pin types. They are not all passive:
grafik

The pin names do not exactly match the datasheet. For example "Analog Out" is "ANA OUT" in the datasheet.

I am not sure about the right side which mixes input and output pins. 
Outputs should be on the right, inputs on the left.
http://kicad-pcb.org/libraries/klc/S4.2/
So P/R should move to the left, some other pins might need to move as well.
This will make the symbol bigger, but cleaner.

EpeR1:
Hi @cp-aquila 

Thanks for your review!
These pins were all passive, because I copied form my very-very old kicad library, and I forgot to update them.

evanshultz:
PCM2902 review:
1. Because the letter at the end is a revision (fixing bugs), I'd like to either add just the "C" version or leave off the version. What do you prefer?
1. Change `IN/OUT` in description to `Input/Output`.
1. Add `, SSOP-28` to the end of the description.
1. PCM2900 and PCM2902 need different symbols because of the pin name differences. So if you want a PCM2900 please add that symbol. Otherwise, please delete that ALIAS.
1. Add `P0.65mm*` to the end of the footprint filter.
1. The value should be at the top right.
1. The symbol is rather bigger than needed and perhaps could have the pins arranged differently. Also this symbol should have 100mil length pins. And we can be a bit more flexible with pin placement in order to make the symbol easier to wire up. What do you think of the below or something more like it?
image
```
# PCM2902
#
DEF PCM2902 U 0 20 Y Y 1 F N
F0 "U" -550 850 50 H V C CNN
F1 "PCM2902" 400 850 50 H V C CNN
F2 "Package_SO:SSOP-28_5.3x10.2mm_P0.65mm" 0 0 50 H I C CNN
F3 "" 0 0 50 H I C CNN
$FPLIST
 SSOP*5.3x10.2mm*P0.65mm*
$ENDFPLIST
DRAW
S -600 800 600 -800 0 1 10 f
X D+ 1 -700 -600 100 R 50 50 1 1 B
X VCCCI 10 700 -300 100 L 50 50 1 1 P
X AGNDC 11 0 -900 100 U 50 50 1 1 W
X VINL 12 -700 700 100 R 50 50 1 1 I
X VINR 13 -700 600 100 R 50 50 1 1 I
X VCOM 14 700 -400 100 L 50 50 1 1 P
X VOUTR 15 700 600 100 L 50 50 1 1 O
X VOUTL 16 700 700 100 L 50 50 1 1 O
X VCCP1I 17 700 -500 100 L 50 50 1 1 P
X AGNDP 18 100 -900 100 U 50 50 1 1 W
X VCCP2I 19 700 -600 100 L 50 50 1 1 P
X D- 2 -700 -700 100 R 50 50 1 1 B
X XTO 20 700 -100 100 L 50 50 1 1 O
X XTI 21 700 200 100 L 50 50 1 1 I
X AGNDX 22 200 -900 100 U 50 50 1 1 W
X VCCXI 23 700 -700 100 L 50 50 1 1 P
X DIN 24 -700 400 100 R 50 50 1 1 I
X DOUT 25 700 400 100 L 50 50 1 1 O
X DGND 26 -100 -900 100 U 50 50 1 1 W
X VDDI 27 -100 900 100 D 50 50 1 1 w
X ~SSPND~ 28 -700 200 100 R 50 50 1 1 O
X VBUS 3 0 900 100 D 50 50 1 1 W
X DGNDU 4 -200 -900 100 U 50 50 1 1 W
X HID0 5 -700 -200 100 R 50 50 1 1 I
X HID1 6 -700 -300 100 R 50 50 1 1 I
X HID2 7 -700 -400 100 R 50 50 1 1 I
X SEL0 8 -700 100 100 R 50 50 1 1 I
X SEL1 9 -700 0 100 R 50 50 1 1 I
ENDDRAW
ENDDEF
```

You may find more information about library guidelines at http://kicad-pcb.org/libraries/klc/.

EpeR1:
@evanshultz 

I updated them according to your recommendations.
I was a bit uncertain about the internal power supply pins, but yes, I also think, they are on a better place at the right side, and with type of passive.

I also updated the ISD25xx symbol, to 100mil's pin length, and setted up a smaller body.

EpeR1:
@evanshultz

I updated them according to your recommendations.
I was a bit uncertain about the internal power supply pins, but yes, I also think, they are on a better place at the right side, and with type of passive.

I also updated the ISD25xx symbol, to 100mil's pin length, and setted up a smaller body.

So, what do you think?

EpeR1:
@evanshultz 

I updated them according to your recommendations. 
I was a bit uncertain about the internal power supply pins, but yes, I  also think, they are on a better place at the right side, and with type  of passive. 

I also updated the ISD25xx symbol, to 100mil's pin length, and setted up a smaller body. So, what do you think?
.
pcm

isd25

EpeR1:
I updated again the PCM2902 symbol, for an easier wiring up.

XTI moved near to XTO and SSPND placed near to HID pins according to the dataset's wiring example

pcm29

evanshultz:
PCM2092C:
1. We don't need blank, A, and B versions. They're the same part. Since you only updated the blank version's description, maybe get rid of the ALIASes?
1. If XTI and XTO can spread apart a bit the crystal can fit between them nicely.
1. What about flipping pins 3 and 17 and also 26 and 4 so the USB pins are on the left side closest too the USB data pins?

ISD25xx:
1. Please use an official datasheet and I can review.

EpeR1:
Done.

Note: 
The ISD25xx's datasheet is only available here, because this IC, is a bit old construction, and has no official webpage.
http://www.profesormolina.com.ar/electronica/datasheets/downloads/isd2560.pdf

evanshultz:
PCM2902;
1. Just capitalize all pins and this is ready to go. Thanks!

ISD25xx:
1. Is this device obsolete? Is any vendor producing it now?

EpeR1:
PCM 2902: Updated.
ISD25XX: No, the production is discontinued.

evanshultz:
@chschlue @cpresser @myfreescalewebpage 
Should we merge ISD25xx? PCM2902 is ready to go so I could whack out the other symbol if y'all want to go that way.

cpresser:
I am not exactly sure how we should handle obsolete parts. 
* KLC does say nothing about them, 
* a quick search in the issues did not show anything useful
* There is an obsolete folder in the lib
* I see the use-case of hobbyists and retro enthusiasts for having symbols of obsolete parts
* There are already quite a few obsolete parts in the library, just search for obsolete (`TMS4C1050N`, `SAA7182`)
* I don't think that having obsolete parts in the lib has any downsides as long as they are clearly marked as obsolete in the description

Considering all of the above I am in favor of including obsolete parts. But they do require some changes:
- [ ] Add `obsolete` to the description
- [ ] The description for the 4 variants should differ

chschlue:
I'm mostly with @cpresser on this, except for the descriptions.
Marking symbols als obsolete only catches the ones that are already obsolete on addition while nobody is going to amend the descriptions as already included parts go out of production.

Having _some_ obsolete parts marked as such doesn't seem very helpful to me.
In fact, marking them doesn't seem very helpful to me at all. It's kind of like a fictitious encyclopedia entry. If it doesn't exist, why would you even search the symbol lib for it?

BTW, does anyone know why exactly we have `obsolete` _and_ `legacy` directories? They're pretty much invisible from a user's perspective anyway. We should probably get rid of them after the migration.

evanshultz:
I agree with the philosophy. Sounds good! Basically what I was thinking but much better written and explained than what I would have written.

In this particular case, I haven't reviewed the obsolete symbol so if we don't get a response can I prune it and merge the good symbol? I probably should have asked to separate them up front since the symbols aren't related.

I do! Old v4 libs with stuff that wasn't ported over for whatever reason for the v5 lib reorg (crufty things, mostly) are in /legacy. /obsolete was created for just this case: users wanting to contribute parts that are already obsolete.

chschlue:
> I do! Old v4 libs with stuff that wasn't ported over for whatever reason for the v5 lib reorg (crufty things, mostly) are in /legacy. /obsolete was created for just this case: users wanting to contribute parts that are already obsolete.

So they can be removed? (Git takes care of preserving them anyway)

evanshultz:
I'm not totally sure of the reason. @SchrodingersGat was the head librarian at the time and I don't recall all the details.

myfreescalewebpage:
Not sure we have to remove the obsolete parts on my side. I think also to enthusiasts or hobbyists who may need this part because they have them in stock and would like to create something with it.

I also think one day all actual parts will be obsolete (let say in few years at least for example) and at this time you will still be happy to open the schematics and still have all the symbol available (even if the cache of the schematic is here to handle this).

My opinion is that the obsolete directory should be retrieved and added with all the other symbols. Manufacturers usually indicate "Not recommend for new design" when the parts are obsolete, maybe this sentence can be added to the description or in a new field of the properties ?  Proposition below:

Capture

chschlue:
Looks like there's general agreement on keeping obsolete parts in the main library.

I'm still not convinced that maintaining life cycle information is feasible though and I'm against having mostly unreliable information in that regard in the lib.
Thoughts?

myfreescalewebpage:
I agree with you @chschlue that maintaining the status of the part in the library will be impossible.
I think this is a just an information we can add when the datasheet or the manufacturer explicitly indicate the part is no more recommended. Else, we just indicate nothing (no status info) which mean Active of we don't know.

EpeR1:
I am not in the position to make decision about this, but my opinion is:
If we signs that, the item is not recommended,  when the manufacturers says that, **at** the symbol's drawing/commiting time, can be informative.

evanshultz:
I'm fine with them in either place, but I also don't like maintaining any lifecycle info in our library. There are resellers who carry NOS parts for years after the original manufacturer discontinues parts, for example, I think we just leave it alone and the library's users can find whatever parts they're interested in. I think we're only a library of parts, not a part manager.

For clarity, I was only suggesting removing the obsolete part so that the other symbol could be merged. Only in the context of this PR was I suggesting to remove any symbols. But since @EpeR1 is here we can hopefully merge both. I'll take a look at ISD25xx and give feedback today.

myfreescalewebpage:
So what is the status here ?

According to me we should continue and I have the following comments right now:
- [ ] ISD2560 should be atomic
- [ ] ISD2560 datasheet found at http://kitsrus.com/pdf/isd_2560.pdf (the previous link is no more available)

Joel

evanshultz:
@myfreescalewebpage 
I still have a few more symbol PRs to wrap up before I focused on the footprints assigned to me. On this one, I'm happy with PCM2902 so if you want to do a review of ISD25xx that's fine. Go ahead. :)

myfreescalewebpage:
Alright, @EpeR1 I have the following comments for the ISD symbol:

- [ ] Can you update the datasheet links
- [ ] Can you fix the description to keep only the right duration, 60 75 90 or 120 seconds depending of the parts
- [ ] Fp filter for the DIP version: we need the width (7.62mm)
- [ ] Fp filter for SOIC version we need the size XxYmm of the footprint required, two variants are possible
- [ ] Fp filter TSSOP: to be removed, there is no TSSOP variant. There is a TSOP variant but it has 32 pins so it's not compatible with the symbol.
- [ ] AGC should be Passive
- [ ] MIC_REF should be Passive
- [ ] I'm not sure the position of the pins is very convenient. The sample schematic page 23 of the datasheet with control on the left and audio on the right is nice in my opinion
- [ ] I would really prefer to have atomic parts, this is finally just a copy/paste at the end. Moreover the device also exist with a 32 pins version, so we need to distinguish it and if we don't use the part numbering codes, then it will not be possible to add the 32 pin version. The part numbering codes are available page 35 of the datasheet

Joel

EpeR1:
Okay, I can make the separated version.
But I would like to use the Winbond's version of the pdf file.
The PDF on kitsrus.com is the older version, and a little bit inconsistent, because there is a 28-pin TSOP package on page 28, but never says about 28pin/TSOP is available, and no SOIC sizes/page present. 
The ISD company was acquired by Winbond, so most of this ics were manufactured by Winbond.
(DIP version is 15.24mm wide. (I have one piece of this IC))

EpeR1:
Done.

a

+1) The Winbond ended manufacturing the 32pin TSOP version.

myfreescalewebpage:
- [ ] Thanks @EpeR1 hoiwever, it was possible to keep the aliases. Aliases are used as soon the difference is only the description. For example for the DIP package you can make ISD2560P symbol, and then add ISD2575P, ISD2590P and ISD25120P as alias. This is better to not enlarge too much the size of the lib !

- [ ] SOIC package fp filter: need the pitch.

- [ ] TSOP package in the description: "TSOP-I-28"

Thanks
Joel

EpeR1:
Updated

myfreescalewebpage:
Great job done here! Last comment for me:

- [ ] ISD2560S: The footprint is a little to much long, you should use `Package_SO:SOIC-28W_7.5x17.9mm_P1.27mm` which is already in the library. Please modify the fp filter accordingly.

Joel

EpeR1:
done

myfreescalewebpage:
Thanks, great job !! No comment comment on this PR, merging.

PR #2018, opened 2019-07-25
Microchip PIC ICSP connector

EpeR1:
### Microchip ICSP Connector:
Microchip PIC In-Circuit Serial Programming/Debugging (ICSP/ICD) Connector
Separated from my Previous PR.

Doc: http://ww1.microchip.com/downloads/en/devicedoc/30277d.pdf
Symbol:

PIC_ICSP_ICD
---


diegoherranz:
Thanks for your contribution. I hope it's the first of many.

I've done a brief review of the symbol and I've got a few comments:

- There is a typo: "Microcip"

Regarding pin names:
- MCLR should mention VPP too. I can see it mentioned by Microchip as `~MLCR/VPP`.
- For the other pins, can you list the PGx option first? I think that it is normally listed first. And can you use a slash? E.g. `PGD/ICSPDAT`. That way, MCLR and PGx are first and VPP and ICSPx second, keeping consistency.

Regarding pin directions, if we think of this connector as used on a target board and not in a programmer/debugger:
- PGC should be output
- VDD should be power output and be on the right hand side.
- MCLR/VPP: if it was for the MCLR part, it could be open collector, but given it is VPP too which can be a high voltage, I'd say passive is the best option
- PGM should be output

Thanks!

EpeR1:
Hi @diegoherranz 

Thanks for your review!
I updated the symbol, according to your recommendations.
I was a bit uncertain about the type of PGC/ICSPCLK pin, the clock is generated by the programmer, so this pin is driven by external circuit.
I think I would be better to mark as input, as mentioned in documentation page 22.
http://ww1.microchip.com/downloads/en/devicedoc/30277d.pdf 

How do you think?

diegoherranz:
Thanks for the changes.

`PGC/ICSPCLK` is an input to the PIC microcontroller but it is an output from the programmer/debugger, so I think it is correct as is.

Regarding `VDD`, I made a mistake on my previous comment. It can be power output if the programmer is powering the target PIC microcontroller. But it can be power input if the target is self-powered to provide a reference voltage for the logic levels.
With all that in mind, I think that the best option is setting it as passive and placing it back on the top side of the symbol. Sorry for the confusion.

Regarding the footprint filters, what do you think about adding the pitch, since 2.54mm should be the standard defined by Microchip with their PICkits?
`PinHeader*1x06*P2.54mm*`
`PinSocket*1x06*P2.54mm*`

If anyone needs a different pitch, they can still do so if they know what they're doing but should make it easier for the most common case.

Thanks!

EpeR1:
Okay, Update is done.

icsp

diegoherranz:
Looks good!

Just one more thing. I've noticed that the similar (in function) connector for ARM micros is named `Conn_ARM_JTAG_SWD_10`.

Maybe for consistency we could call this one `Conn_PIC_ICSP_ICD`?

Everything else ready to go.

There are 2 errors in Travis. The centre of the symbol, but I think it is OK as is. And VDD being passive but as you say and as discussed above, I think passive is the best choice this tine.

Thanks!

EpeR1:
Yes, renaming is done.

diegoherranz:
Thanks for all your changes. Merging.

EpeR1:
Thank you so match!

PR #2024, opened 2019-07-26
Added Ti LM2705 micro Boost converter

Duckle29:
Added Texas Instruments LM2705 Micropower Step-Up DC-DC Converter With 150-mA Peak Current Limit.

---

[Datasheet](http://www.ti.com/lit/ds/symlink/lm2705.pdf)
screenshot


cpresser:
Hi and thanks for your contribution.
I was searching a little bit in the library and found `LMR10510XMF` which has the same pin-out and package. So you can make the `LM2705` an alias of that one and save some space in the library.

The symbol is slightly different (has VIN on the left), but thats an accepted style in the library.

Duckle29:
@cpresser Thanks for the review. I made the requested changes :) Only question I have is what to do about the different datasheet?

cpresser:
Thanks!
Regarding the datasheet, you can set it in the gui on the aliases tab. Then the line 1362 of you changeset should contain the datasheet.
image

myfreescalewebpage:
@Duckle29 any news here ?

Duckle29:
Aah sorry, must have left my mind. I added the datasheet. That said running the schlib checker, it complained about pin-offset, VIN being input instead of power and as mentioned vin being on the left. I'm assuming those are fine as the symbol is already in, however I could change vin to be a power input if you want? That seems like an easy / non-intrusive change?

cpresser:
I took the liberty of fixing two of the travis warnings with the github web-editor.
@myfreescalewebpage  can you please review them and then merge? Thanks!

cpresser:
@Duckle29 sorry for intruding on your PR. I did not read your comment properly the first time and decided to fix it since you don't want to address the warnings. Now that I read your comment again and figured I misunderstood your intention. Because I didn't read properly. Its all my fault - my bad.

Duckle29:
No problem at all 😃, it's quicker and easier that way :P

myfreescalewebpage:
@Duckle29 I'm doing a full review now:

- [ ] The pin 4 is not the same that LMR10510XMF, the LM2705 symbol can't be an alias of this device. You have to duplicate and change pin 4 name.
- [ ] The name of your symbol should be "LM2705MF" I think according to the datasheet

Joel

myfreescalewebpage:
@Duckle29 ping

Duckle29:
Because SHDN is a fair bit longer than EN, the symbol had to change. The least change in symbol size was putting VIN on top and SHDN middle left. This also mirror the way it's displayed in the datasheet

image

Duckle29:
image
Forgot I was on 25mil grid

PR #2028, opened 2019-07-27
Arduino Pro Mini

nerdyscout:
Arduino Pro Mini is a small PCB module containing a ATmega328p just like the Arduino Nano.
The [official Pro mini](https://store.arduino.cc/arduino-pro-mini) might be retired, but many vendors, eg sparkfun, still sell it.

image

footprint pull request: https://github.com/KiCad/kicad-footprints/pull/1690

---


matthijskooijman:
It seems this symbol can be (mostly) a drop-in replacement for the Arduino Uno symbol, is that indeed the case? The Uno has the same MCU and mostly the same pins, so it would be good if the overlapping pins are indeed in the same position.

matthijskooijman:
I also wonder: Would it be good to have a version that does not include the 6-pin FTDI serial header? Often when integrating a pro mini, you just connect the side headers and leave serial header to directly attach an FTDI adapter?

nerdyscout:
> It seems this symbol can be (mostly) a drop-in replacement for the Arduino Uno symbol, is that indeed the case? The Uno has the same MCU and mostly the same pins, so it would be good if the overlapping pins are indeed in the same position.

yes, this part is meant as replacement. For some reason the symbol was a bit to tall, but all pins should already have aligned.

myfreescalewebpage:
Hi @nerdyscout , thanks for contributing,

- [ ] Pin 25 and 26 are crossed I think, but will need to verify the pinout after the footprint is merged.

Cheers,
Joel

myfreescalewebpage:
PS : no symbol for the Sparkfun board ?

nerdyscout:
I would prefer to use this schematic: https://www.arduino.cc/en/uploads/Main/Arduino-Pro-Mini-schematic.pdf as it is the original source.
what do you think about changing the layout completely to make it drop in compatible to #1933 ? As there are Feather boards with the same MCU this might make sense.

myfreescalewebpage:
The schematic you refer is very nice, but it does not indicate where are the pins on the board physically.
We have spoken at #1933 of pin compatibility, why not but I prefer to have the arduino family on one side, with pin compatibility (if possible) and feather on the other side.
It seems complicated to have pin compatibility, VBAT does not exist on arduino, USB power does not exist, etc. But you can try if you want to have something with the same design.

nerdyscout:
maybe I should read 1933 now...
but I guess something like this would fit into the scheme
image

matthijskooijman:
When talking about compatibility, making the Pro Mini compatible with the Uno or the Nano (which unfortunately are nearly but not entirely compatible between them) would make more sense, since they're the same manufacturer, same chip and first building something on a Uno and then using a Pro Mini or Nano on a subsequent version is quite imaginable.

Note  the sure which of the Uno of Nano to prefer, probably the Nano since IIRC it's slightly bigger and offers two extra pins analog pins which, I think, the Pro Mini also has. Then we could make the Uno compatible with that later at a point where breaking compatibility is ok (or is that now?).

nerdyscout:
image image
currently I am checking if I missed some point, but these two quick drafts should fit each other and the Feather boards.

nerdyscout:
| ProMini | ProMicro |
| --- | --- |
| image | image |

ready for review, hope I got everything right.

PR #2032, opened 2019-07-28
Change footprint for the APA102 5050

### LINKED ISSUE #1910: APA102 Footprint incorrect
I printed out a PCB that used the APA102 symbol provided in the built-in library and was validating all of the footprints against the printout.

Pictured, a bit hard to see, is an APA102 using the library's footprint. You can see that the spacing on the pads doesn't match.

image

[The SparkFun library](https://github.com/sparkfun/SparkFun-KiCad-Libraries) seems to be correct when I checked it physically against the part.

wirehead:
Fixes #1910

 - change footprint from LED_SMD:LED_RGB_PLCC-6 to LED_SMD:LED_RGB_5050-6
 - see KiCad/kicad-symbols#1910 for discussion.

Since we don't have a datasheet and nobody from the manufacturer wants to give any more details, here's a board I fabbed at DirtyPCB's from a file generated in KiCad using the LED_RGB_5050-6 footprint with a real live APA102 part I ordered from DigiKey.

*sigh*


evanshultz:
@wirehead 
Thanks! Can you please update the footprint filter to `LED*RGB*5050*`?

Also, there's no picture that you referenced above.

wirehead:
image

evanshultz:
Thank you!

PR #2035, opened 2019-07-28
Fix 4068 IEEE symbol

chschlue:
Fix 4068 IEEE symbol

* add non-inverted output
* fix pin types

Datasheet: http://www.ti.com/lit/ds/symlink/cd4068b.pdf

Violations: kept style of whole 4xxx_IEEE lib

---


evanshultz:
Thanks for another fix PR! And sorry that things need fixing.

Can you please include before/after screenshots? That makes the review easier. Here it is:
image

This is a correction to a clear error, so it's welcomed as-is if you don't want to put in any more effort. But if you're willing, would you consider adding keywords (perhaps just `cmos nand and`) and a datasheet link? Other errors and warning aren't worth taking any action.

chschlue:
Will do.

BTW, do almost all the generic logic symbols look like that for a reason or has just nobody cared to polish them yet?

evanshultz:
There has been quite a bit of discussion about how to design the logic symbols, but it didn't get wrapped up for the KiCad v5 release. Everything needs to be ironed out and then the rest of the work done to update symbols for v6. So it's in a holding pattern right now. Because of the design of the symbol libraries and KiCad, the devs prefer not to make massive symbol changes during a major release.

evanshultz:
Thank you!

PR #2038, opened 2019-07-28
Added SN65HVD11D and Conn_ARM_SWD_10

pegasus17:
* Added SN65HVD11D to Interface_UART.lib
SN65HVD11D

* Added Conn_ARM_SWD_10 to Connector.lib based on Conn_ARM_JTAG_SWD_10,  removed the JTAG only part to get a clean SWD symbol
Conn_ARM_SWD_10


cp-aquila:
Thanks for you contribution.
Can you please split this up into two PRs since the two symbols don't have any relation?

The SWD-Symbol as is does violate KLC S4.6 http://kicad-pcb.org/libraries/klc/S4.6/ , can you please fix that.
Can you please provice a link to a datasheet of the SWD-Connector pinout?

The RS485-Transceiver might clash with this PR: https://github.com/KiCad/kicad-symbols/pull/2011 it is for similar parts.
The symbol also looks pretty similar to "SN75LBC176D", so one option would be to just define some new aliases instead of having a separate symbol.

chschlue:
I'm with @cp-aquila on this.
Almost all non-insulated TI RS485 parts could be merged into a single symbol with aliases, possibly even including MAX485, several AD parts and other pin-compatible types.
Maybe a librarian could elaborate on why this approach appears to be frowned upon?

On a different note: SN65HVD* are temp variants of SN75HVD*. Why not call the symbol 75?

pegasus17:
@cp-aquila and @chschlue for the RS485-Transceiver your right the symbol of PR: [2011](https://github.com/KiCad/kicad-symbols/pull/2011) is the same are mine also naming it SN75HVD* is no problem. I would go with that one - it's submitted earlier.

"SN75LBC176D" look similar but we you use one signel for controlling RE and DE you get crossing wires with are not connected.
TI-RS485

pegasus17:
@cp-aquila The SWD-symbol I have put into PR: [2058](https://github.com/KiCad/kicad-symbols/pull/2058) with a link to a datasheet of the SWD-Connector pinout.

chschlue:
Personally, I'd like like to combine all of these "standard" 8-pin transceivers. Just waiting for maintainer feedback whether it's worth the effort.

PR #2040, opened 2019-07-29
added diodes inc ZXBM5210

g-gg:
added the symbol for the reversible DC Motor Driver ZXBM5210 from diodes inc in SO-8
Screen Shot 2019-07-29 at 23 43 09
https://www.diodes.com/assets/Datasheets/ZXBM5210.pdf
---


evanshultz:
1. Add `, SOIC-8` to the end of the description.
1. Change pin text offset to 20mil. You may then be able to shrink the symbol.
1. Remove `?8` from the footprint filter.
1. We want to have `-S` added on the symbol to indicate there is no thermal pad. If you want both variants, make another symbol with an `-SP` suffix and then you need a pin 9, stacked on pin 8, for the thermal pad.

g-gg:
Shrinked version of the symbol (here in -SP variant)
Screen Shot 2019-07-30 at 22 09 16

Thanks, Evan!

evanshultz:
Thanks! This symbol looks very nice.
1. I forgot to mention that pins should be 100mil long. Sorry about that.
1. Please change the non-EP symbol to use the `SOIC` footprint and not the `SO` one.
1. Lastly, the symbol text is cramped. What about if VDD moves right or FWD moves down?

g-gg:
Thanks Evan for your friendly and accurate comments! You're absolutely right, I've rearranged and shortened the pins a bit, I hope you like the result.

Screen Shot 2019-08-01 at 21 46 23

evanshultz:
Thank you once again! This looks very nice. Since We require power pins to be on the top, please move VM there. I can see that it may be a bit unbalanced but we should stick to the guidelines at http://kicad-pcb.org/libraries/klc/.

g-gg:
Sure, that's possible. I've shifted the pins around a bit to make the symbol as compact as possible without (willingly) compromising the convention.

Screen Shot 2019-08-01 at 22 25 19

evanshultz:
Everything looks good. Thanks so much!

PR #2051, opened 2019-08-01
Add Sm420561N

DecadoPL:
image

image

https://datasheet.lcsc.com/szlcsc/Wuxi-ARK-Tech-Elec-SM420561N_C141367.pdf


cpresser:
- [x] ~~We already have similar symbols in the lib. For example "MAN3810A". Yours is slightly different. For consistency it would be nice to align this to the already existing symbols~~ Upon closer inspection, the library is inconsistent. Your symbol looks similar to "KCSA02-106" which is fine.
- [ ] Footprint does not exactly fit. "7SegmentLED_LTS6760_LTS6780" is 12.5mm, the datasheet says 12.6mm
- [ ] The Pin-out of the footprint is missing in the datasheet. I assume the numbering is correct, but there is no way to verify.

DecadoPL:
@cpresser 
I compared drawings from datasheets and i belive this component almost perfectly fits to "7SegmentLED_LTS6760_LTS6780" footprint. There is difference in height. SM420561N is 19.0mm and the other one is 19.05mm. But i think there is no need to create another footprint since both are made in tolerance of +-0.25mm. 
Drawing that shows SM420561N has weird problem with number "4" it looks like "1".
image
image

cpresser:
Quite interesting, your second screenshot says 12.6mm, but the symbol in the library has 12.5mm. I think a PR to fix that footprint is required.

>  Drawing that shows SM420561N has weird problem with number "4" it looks like "1".

I don't see that. Can you clarify?

poeschlr:
>>   Drawing that shows SM420561N has weird problem with number "4" it looks like "1".

>I don't see that. Can you clarify?

The datasheets distance between the two rows of pins kind of looks like it would be 15.21mm But such a measurement would be really odd. 15.24 would be a nice imperial number and quite typical for a lot of devices.

A similar observation can be made with the 2.54*4 entry on the same page where both "4" characters also kind of look like a 1. (The font uses a much thinner line for parts of the 4 that distinguish it from a 1. JPEG compression does the rest.)

cpresser:
Sorry for the long delay. 
The Issue is actually with the Footprint, not with the symbol. I will create a PR to fix the symbol (width is 12.6 not 12.5 mm)

PR #2052, opened 2019-08-02
Add MAX7219

DecadoPL:
image

image

https://datasheets.maximintegrated.com/en/ds/MAX7219-MAX7221.pdf


cpresser:
Thanks for your contribution.

- [ ] Can you please stack the GND pins
- [ ] Once they are stacked, it might be possible to make the symbol 200mil smaller in x direction
- [ ] I would seperate Iset from the SPI pins (move it closer to the top), then the SPI pins build one logical unit
- [ ] Move Dout to the right side, to allow daisy-chaining of multiple devices (this might conflict with the above suggestion)
- [ ] The +T suffix in the name is not needed
- [ ] The device is available in SO, DIP (and CERDIP) package. Unfortunately an alias can't have a different package. So in order to have both variants, the name needs to be generic (MAX7219
xxG), the footprint removed and the two packages added to the fp-filter. Or make two symbols that just have a different name and Footprint. Perhaps somebody else can shed light in how to handle this best.
- [ ] Pin Names are not exactly like in the Datasheet. Pin19 is 'V+', Pin12 is 'Load' (we could skip the !CS part here since it applies to another variant.
- [ ] Have you tried wiring it up to a 7-Segment-Display symbol? I am asking because I have no clue if it might make sense to rearrange the output-ports to make wiring in the schematic easier.

DecadoPL:
@cpresser 
Changed library acording to your list. About the last point, i used this symbol in schematic and it is ok. When connecting this many lines to 8 displays you need to use bus to make it clear. So yes, it was ok for me, otherweise i would propose different pins placement.

cpresser:
Thanks, only a few small thins remain. I updated the checklist above.

Proposed pin-layout:
image

chschlue:
The G suffix means 24 pins. Why not just call it MAX7219?

See https://www.maximintegrated.com/en/design/packaging/package-information/maxim-naming-conventions.html

cpresser:
> The G suffix means 24 pins. Why not just call it MAX7219?
> 

There is the "die-only" version of this thing, which would also be included in "MAX7219".
But I think we can ignore the die version, so removing the complete suffix is fine for me.

chschlue:
Naked dice are a little out of scope, I guess.

DecadoPL:
@cpresser looks like now your list is compleated :)
image
image
image

cpresser:
Merged, thanks!

PR #2053, opened 2019-08-02
Add Analog Devices LTC2942 and LTC2942-1

MacroYau:
Adds symbols for Analog Devices LTC2942 and (with LTC2942-1 alias) fuel gauge.

Datasheet: [LTC2942](https://www.analog.com/media/en/technical-documentation/data-sheets/2942fa.pdf) and [LTC2942-1](https://www.analog.com/media/en/technical-documentation/data-sheets/29421f.pdf)

LTC2942

---


cp-aquila:
Hi and thanks for your contribution,

- [x] Pin-Types and Names are good
- [x] Its in the correct library
- [x] Footprint matches
- [ ] Footprint-Filter should use * instead of ? after the pin-count
- [ ] GND Pin should move to the bottom
- [ ] Please set Pin name position offset to 20mil
- [ ] Datasheet field should be empty according to KLC, (S6.3 still has an old screenshot, so I am not sure how to do this properly)

Then a subjective thing, I would mirror the symbol, having sense+- on the left side, and the logic connections on the left side. Usually symbols are designed/read left-to-right. 
See http://kicad-pcb.org/libraries/klc/S4.2/
But what is actually input and output here is not 100% cleaer (the examples for the S4.2 rule could use improvement)
Anyway, I²C on the left side feels better for me.

MacroYau:
@cp-aquila Thanks for the feedback!

I have adjusted the pin placement as you suggested (and align with the way it is shown on the datasheet).

image

For the other items on your checklist:

> - [x] Footprint-Filter should use * instead of ? after the pin-count
> - [x] GND Pin should move to the bottom
> - [x] Please set Pin name position offset to 20mil

Fixed.

> - [ ] Datasheet field should be empty according to KLC, (S6.3 still has an old screenshot, so I am not sure how to do this properly)

I am confused about this. S6.3 says the datasheet URL should be added. I guess S6.2 only suggests the datasheet field on the symbol (like the package name in grey) should be invisible, which I cannot see it anyhow, even if I deliberately turn it on (I am running KiCad 5.1.2)? 

> - [x] The current-range should be included in the description

The LTC2942 device's range is determined by an external current sense resistor, while the LTC2942-1 (with current sense resistor built-in) alias has already included 1A range in its description.

cp-aquila:
> 
> > * [ ]  Datasheet field should be empty according to KLC, (S6.3 still has an old screenshot, so I am not sure how to do this properly)
> 
> I am confused about this. S6.3 says the datasheet URL should be added. I guess S6.2 only suggests the datasheet field on the symbol (like the package name in grey) should be invisible, which I cannot see it anyhow, even if I deliberately turn it on (I am running KiCad 5.1.2)?

I looked at other recently merged footprints to clarify this. 
So what you did so far is fine: the Datasheet is set according to S6.3 and thats all there is.

> > * [x]  The current-range should be included in the description
> 
> The LTC2942 device's range is determined by an external current sense resistor, while the LTC2942-1 (with current sense resistor built-in) alias has already included 1A range in its description.

My wording was not clever. I meant to say 'the description of the two parts should differ'. Which was already in your first commit, I just missed that.
But the package needs to be in the description.
I would also include "i2c" to the keywords

I will update the checklist in the post above.

MacroYau:
I have appended "DFN-6" in description and added "I2C" in the keyword field. Thanks!

cp-aquila:
Thanks, looks good; all my checks are green. I think this is ready for merge. @evanshultz ?

evanshultz:
@MacroYau 
Just one more thing: Remove `?6` and `EP1.65x1.35mm*` from the footprint filter so that the pincount filter in CvPcb will work and the user can pick a different EP size if they want.

@cp-aquila 
Great! Thank you!

MacroYau:
@evanshultz Done. Thanks for your review!

evanshultz:
And thank you!

PR #2055, opened 2019-08-02
Add Vishay Siliconix SiP32431 and SiP32432

MacroYau:
Adds the symbols for Vishay Siliconix SiP32431 (high-enable version) SiP32432 (low-enable version) load switches. Only the DR3 variant (SC-70-6 footprint) is included, as the official footprint library lacks the TDFN4 footprint used by the DNP3 variant (with different pin assignment).

Datasheet: http://www.vishay.com.hk/docs/66597/sip32431.pdf

SiP32431
SiP32432

---


evanshultz:

1. Remove the `?6` from the footprint filter.
1. Can you put the GND pin on the bottom? You may then be able to shrink the symbol, something like our linear regulators with an enable pin.
1. Perhaps stack the refdes and value above the symbol instead of next to each other.
1. Stacked GND pins need the non-visible pins to be Passive type.
1. The IN and OUT pins should be Passive type, since we don't know they are a current source. That is the intended application of this part, but we can't know how it will be used. This is the same as our other analog switches (at least the modern ones).
1. Please name pin 6 as `NC`.

MacroYau:
@evanshultz Much appreciated for your feedback! I have revised the symbols based on your recommendations:

evanshultz:
Looks good. Thanks for the contribution as well. Hope to see you again soon!

PR #2058, opened 2019-08-02
Added Conn_ARM_SWD_10

pegasus17:
Added Conn_ARM_SWD_10 to Connector.lib based on Conn_ARM_JTAG_SWD_10, removed the JTAG part to get a clean symbol for ARM Cortex-M SWD.

Conn_ARM_SWD_10

Datasheet: http://infocenter.arm.com/help/topic/com.arm.doc.faqs/attached/13634/cortex_debug_connectors.pdf

---


diegoherranz:
Thanks for your submission.

Honestly, I'm struggling to see the value of this symbol over the existing one including JTAG (Conn_ARM_JTAG_SWD_10).

The only differences I can see are:
1. Hiding the `SWO/TDO` pin: even if you're not using JTAG, the SWO pin is used quite frequently so I think it should remain.
2. Hiding the `GNDDetect` pin: similarly to 1, even in SWD more this pin is needed and should be kept.
3. Removing the JTAG names from SWDIO and SWDCLK: the SWDx pins were listed first already so I think it was sufficiently clear.
4. Hiding the `NC/TDI` pin.

I'm not sure that 3 and 4 are significant enough to create and maintain a different symbol.

I don't want to reject this straight away so I'd like to ask the @KiCad/librarians  for their input. I'll leave this open for ~a week to allow time for anyone interested on commenting.

There are  also a few other issues with this submitted symbol: e.g. location of hidden pins and GND pins not stacked. But that could be fixed and is completely separate from the discussion above.

I sincerely appreciate your effort helping with the KiCad libraries and hope to see more of your submissions in the future.

Thanks!

PR #2059, opened 2019-08-03
Add AVR-UPDI-6

chschlue:
* Add AVR-UPDI-6, https://www.microchip.com/webdoc/GUID-9D10622A-5C16-4405-B092-1BDD437B4976/index.html?GUID-9B349315-2842-4189-B88C-49F4E1055D7F
AVR-UPDI-6

---


matthijskooijman:
Heh, seems I missed one AVR programming connector when I added them, thanks for making the collection complete :-)

Some remarks:
 - Why is the pin named UPDI_DATA and not just "DATA" like with the AVR-PDI-6 symbol? Might be more consistent, and looks nicer because it is shorter.
 - Would it make sense to put the UPDI_DATA pin one position up so it is in the same position as the data pin in the AVR-PDI-6 symbol?
 - There is still a small rectangle left over from the clock pin that should be removed.

Some more remarks, which really apply to all of the AVR-connector symbols (and should either be left as-is in this symbol, or fixed in all of them):
 - Pin length should be 100mil ([KLCS4.1](http://kicad-pcb.org/libraries/klc/S4.1/)).
 - The NC pins are inside the box, not on the box border ([S4.6](http://kicad-pcb.org/libraries/klc/S4.6/)), though I guess this is needed to keep them on the 100mil grid. Changing pin length to 100mil should allow conforming to both the grid and on-box-border requirements.
 - Should VCC/GND really be power input? They might be used as such when this connector is on the programmer side, but might be power output on the device-to-be-programmed side. Often, they are just used to sense the voltage level (VCC) or serve as a ground reference (GND) and no current should flow at all. Maybe they should just be passive?

chschlue:
> Heh, seems I missed one AVR programming connector when I added them, thanks for making the collection complete :-)
You're welcome.

> * Why is the pin named UPDI_DATA and not just "DATA" like with the AVR-PDI-6 symbol? Might be more consistent, and looks nicer because it is shorter.

Don't blame me, Microchip came up with that.

> * Would it make sense to put the UPDI_DATA pin one position up so it is in the same position as the data pin in the AVR-PDI-6 symbol?

Yes, it would. But it would also make the symbol larger and kill pin compatibility as well.

> * There is still a small rectangle left over from the clock pin that should be removed.

Right, missed it.

> Some more remarks, which really apply to all of the AVR-connector symbols (and should either be left as-is in this symbol, or fixed in all of them):

> * Pin length should be 100mil ([KLCS4.1](http://kicad-pcb.org/libraries/klc/S4.1/)).
> * The NC pins are inside the box, not on the box border ([S4.6](http://kicad-pcb.org/libraries/klc/S4.6/)), though I guess this is needed to keep them on the 100mil grid. Changing pin length to 100mil should allow conforming to both the grid and on-box-border requirements.

All true, but might be hard to fix before v6. Feel free to take a look.

> * Should VCC/GND really be power input? They might be used as such when this connector is on the programmer side, but might be power output on the device-to-be-programmed side. Often, they are just used to sense the voltage level (VCC) or serve as a ground reference (GND) and no current should flow at all. Maybe they should just be passive?

Right, don't know what I was thinking.

matthijskooijman:
> > Why is the pin named UPDI_DATA and not just "DATA" 

> Don't blame me, Microchip came up with that.

Hm, I see. But is matching the datasheet the most important thing here? Consistency between similar symbols also seems relevant (and as an added advantage, it would allow fixing the next point). Also, is this really the official specification for this connector? This is just the pinout of one particular programmer (though it does seem to be pretty much the only definition of this pinout and more generally programmer pinouts seem to be the most official sources for AVR programming). 

> Yes, it would. But it would also make the symbol larger and kill pin compatibility as well.

Ah, right, one more argument for shortening the name to DATA? :-)

> All true, but might be hard to fix before v6. Feel free to take a look.

I think these can be fixed without breaking compatibility? The pins become shorter and the box a bit bigger, but the pin positions can remain the same (except for the NC pins, but moving those should be harmless?). Or am I missing something here?

> > Should VCC/GND really be power input?

> Right, don't know what I was thinking.

Well, you only copied this from the AVR-PDI-6 symbol, I suppose. Looking through git, it seems I originally added them as passive, but @evanshultz changed them to Power Input in commit 9d919e3d80f4ecbe4484224472b31ef7bca0ac1b. @evanshultz, do you really think these should be power input? Also see my previous comment.

evanshultz:
I'll leave it to you guys who are more familiar than me. If the connector itself must be connected to a power source for the programmer to work, it would make sense for it to be Power Input. I'm not sure about the pin type in my PR and it could be wrong. Pin types should be assigned based on what makes ERC most useful without giving false positives/negatives.

Also, changing pin length and symbol outline at once will result in pin ends at the same location, so that is fine to do anytime. Just confirmation of the above.

chschlue:
Some programmers use these to power the uC during programming / debugging, others just sense levels as most AVRs can be supplied with 1.8V to 36kV.
I strongly lean towards passive.

matthijskooijman:
> Some programmers use these to power the uC during programming / debugging, others just sense levels as most AVRs can be supplied with 1.8V to 36kV.

Agreed. Also, this connector could be used fora device or a programmer, in which the direction of flow (either power or sensing) would reverse, so I would also definitely change these to passive, which should be fine in all cases (but might need a power flag when you expect your device to be powered through the programming connector).

chschlue:
AVR-UPDI-6

Changed as discussed. Will also update existing AVR ISP connectors.

@matthijskooijman and @fauxpark as acting head of MCU_Microchip department: Any thoughts?

fauxpark:
It could be made a little smaller, I think - the NCs, being invisible, can go on the left and overlap the GND/VCC text. Other than that, looks good 👍

chschlue:
It's the same size as the other 6-pin connectors in #2164 right now, that's why it's larger than necessary.

matthijskooijman:
Changes looks good. I would suggest also enlarging the box on the left side, so the box is again symmetrical around the origin.

evanshultz:
Seems OK. Merging this one also.

PR #2061, opened 2019-08-04
Fixed STM32F042F*: renamed pin 15 from VSSA to VSS

pegasus17:
* Renamed pin 15 from VSSA to VSS. According the datasheet the F-variant (TSSOP20) does not have a ground for the analog domain. 
Datasheet: https://www.st.com/resource/en/datasheet/stm32f042f6.pdf
* Also fixed  Violating S3.1

STM32F042F4Px
STM32F042F6Px


evanshultz:
@pegasus17 
Thanks for making this contribution and fixing things.

However, there is an annoying problem with KiCad in which different versions change the line ordering. I don't know why the devs periodically decide to change things, or allow it to vary.

Anyway, what it means is that this library hasn't been touched for a while and your version of KiCad introduces a huge number of changes, mostly line re-ordering. It's very difficult to check your changes. Would you be willing to make one commit that saves the file with no changes in KiCad, and then a second commit which includes just the fixes you are making? That will be much easier to review. Thank you!

pegasus17:
@evanshultz 
I have re-done the fix in the two steps you suggested. It should be now easrier to review.

evanshultz:
Thanks! Re-naming the pin is fine, but changing the locations of other pins is not because it will cause a disconnection is schematics that are opened that were designed with the old symbol in mind. If you can revert changing the VDD pin location and symbol size, even though it creates an error, this will be OK. So to be clear, I should only see pin 15 renamed to get this PR merged.

pegasus17:
No problem. I have reverted the change in the symbol size and pin location.

evanshultz:
Even with all the messiness, I don't see anything wrong in the diff (but it is noisy). The branch opens in KiCad 5.1.2 and I see the pins renamed to `VSS` on the two symbols that should be affected. So I'll merge. Thanks for your effort to correct this and hope to see you again!

PR #2065, opened 2019-08-09
add NCP718 linear regulators as alias to lp5907

lorem-ipsum:
This adds the [NCP718](https://www.onsemi.com/pub/Collateral/NCP718-D.PDF) in SOT-23-5 package as alias to the existing [LP5907](https://www.ti.com/lit/ds/symlink/lp5907.pdf).

---


evanshultz:
We could select the A or B version instead of using the 'x' wildcard in the PN. Or add both versions. This would better help to define the parts but this works and is OK by KLC (for now).

Thanks!

PR #1976, opened 2019-07-12
Commonalize Transistor_FET

evanshultz:
From https://github.com/KiCad/kicad-symbols/issues/1975.

Changes:
-Use a common symbol design
-Set various parameters per KLC
-Align FP filters

Here are examples of the unified style:
image
image

There are a lot of lines changed, so I'm not sure the best way to review this.

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

Issue #1617, opened 2019-03-08
Cleanup of descriptions and keywords


evanshultz:
While reviewing https://github.com/KiCad/kicad-symbols/pull/1616, I looked at the DCM file changes. Here is the part submitted and the two after it:
```
$CMP Si7141DP
D 60A Id, 20V Vds, 1.9 mOhm Ron, PowerPAK-8
K P-Channel MOSFET
F https://www.vishay.com/docs/65596/si7141dp.pdf
$ENDCMP
#
$CMP Si7336ADP
D 30A Id, 30V Vds, 2.4 mOhm Ron, PowerPAK-8
K N-Channel MOSFET
F www.vishay.com/docs/73152/si7336adp.pdf
$ENDCMP
#
$CMP SiA449DJ
D -12A Id, -30V Vds, P-Channel MOSFET, PowerPAK SC70-6
K P-Channel MOSFET
F http://www.vishay.com/docs/62644/sia449dj.pdf
$ENDCMP
#
```

You can see one difference in that the type of transistor (NPN, PNP, N-channel, P-channel, IGBT, etc.) is in the description on one but not the others. And the keywords contain just the transistor type.

Now that we will be heading towards v6, perhaps it's a good time to think through our descriptions (I'm using transistors above but the scope can be library-wide) and figure out if there are more rules and structure we should put in place for descriptions and keywords?

Lastly, I want to be clear that the contributor did everything right and this PR is fine. It's a framework and consistency issue I'm surfacing here.

myfreescalewebpage:
I agree with you @evanshultz, if some description/keywords guidelines can be added to very large families, such as transistors, regulators, etc to help search when making a schematic, that's a good idea !!!

PR #2066, opened 2019-08-09
add BSP89 as alias and fix symbol of BSP129

lorem-ipsum:
This adds the [BSP89](https://www.infineon.com/dgdl/Infineon-BSP89-DS-v02_02-en.pdf?fileId=db3a30433b47825b013b4b8a07f90d55) as alias to the existing [BSP129](https://www.infineon.com/dgdl/Infineon-BSP129-DS-v01_42-en.pdf?fileId=db3a30433c1a8752013c1fc296d2395f). It also fixes the gate pin of the BSP129. The length was set to 100mils, causing the pin to end before the gate.

Screenshot_20190809_135412

---


cpresser:
This seems to be linked to https://github.com/KiCad/kicad-symbols/pull/1976

electroniceel:
I don't think BSP89 is a correct alias of BSP129:

BSP89 is enhancement mode, BSP129 is depletion mode. These are different symbols.

The current symbol for BSP129 shows an enhancement fet which is wrong. I have just uploaded symbols for depletion mode FETs (including a fixed BSP129) in #2095 .

lorem-ipsum:
I made the BSP89 a unique symbol instead of an alias to the BSP129. It is a copy of the current symbol of the BSP129 which is correct for the BSP89 (but not for the BSP129). Also added the enhancement mode to the description.

evanshultz:
@lorem-ipsum 
Thanks!

Because the description for FETs is all over the place, I'll go ahead and merge this. But https://github.com/KiCad/kicad-symbols/issues/1617 is capturing this issue (and for other symbol types too) and perhaps we can standardize things later. For example, always include the FET type?

PR #2068, opened 2019-08-10
Add Hirose MS-156C3 rf testpoint

drid:
Add Hirose MS-156C3 rf testpoint

Datasheet https://www.hirose.com/product/document?clcode=CL0358-0340-0-00&productname=MS-156C3&series=MS-156C&documenttype=2DDrawing&lang=en&documentid=0000954375

Footprint: https://github.com/KiCad/kicad-footprints/pull/1773
image

---


poeschlr:
The connector lib is for generic symbols of different connector types. I suspect that there will be similar parts out there so my suggestion would be to remove the datasheet and footprint assignment. Set the footprint filter to fit the prefix of the footprint and rename the symbol to CoaxialSwitch_Testpoint (or similar)

Alternative we make a specialized lib for things like that and have fully specified symbols. I am however not convinced that this would be the right move at this point in KiCads development.

poeschlr:
The footprint filter should be generic as well

poeschlr:
I think you misunderstood what i meant with "generic footprint filter". I would suggest to use the filter `CoaxialSwitch*`

poeschlr:
Thanks

PR #2071, opened 2019-08-13
Added PCA9600D & PCA9600DP.

arnmulligan:
Datasheet: https://www.nxp.com/docs/en/data-sheet/PCA9600.pdf

PCA9600D
PCA9600DP


diegoherranz:
Thanks for your contribution. I hope it's just the first of many :)

A few comments:
- [ ] Can you set RX and RY as input?
- [ ] And TX and TY as open collector?
- [ ] This one is more a personal opinion. What would you think about having the TX pins on top of the RX to match the diagram/symbol in the datasheet? I find it a bit more natural too. As I said, this is a personal opinion so I'm OK if you don't want to change it.

Note: On SDA slaves, we use Bidirectional for SDA and input for SCL. In this case, so believe that Bidirectional for both SX and SY is the right choice. So nothing to change there.

Thanks!

arnmulligan:
>     * [ ]  Can you set RX and RY as input?

Done.

>     * [ ]  And TX and TY as open collector?

Done.

>     * [ ]  This one is more a personal opinion. What would you think about having the TX pins on top of the RX to match the diagram/symbol in the datasheet? I find it a bit more natural too. As I said, this is a personal opinion so I'm OK if you don't want to change it.

Disagree that it is more natural. Agree that it matches the datasheet. Personally not bothered either way, so I've done it to your taste.   :)

Thanks!

diegoherranz:
Thanks for the changes. Merging.

PR #2076, opened 2019-08-14
Fix 74LVC1G66

### LINKED ISSUE #2074: 74LVC1G66 Bilateral Switch Has Missing Datasheet and Unspecified Footprint
`74xGxx:74LVC1G66` has a 404'd datasheet link. 

Additionally, according to [this datasheet](https://assets.nexperia.com/documents/data-sheet/74LVC1G66.pdf) the pinout changes with different package types. I.e. for SOT353-1 and SOT753 packages, `Vcc` is pin 5. For SOT886, SOT891, SOT1115 and SOT1202, `Vcc` is pin 6, and pin 5 is `N/C`. 

No indication as to pinout is made in the description of the part, nor is a default five-pin footprint set. This may result in mistakenly assigning `Vcc` to pin 5 when using six-pin footprints.

chschlue:
* Fix datasheet link
* Fix pin types

DS: http://www.ti.com/lit/ds/symlink/sn74lvc1g66.pdf

Fixes #2074 

74LVC1G66
74LVC1G66_6

---


evanshultz:
Power pins are hidden in the logic library. It's gross, but leave that alone so the symbols on existing schematics don't change. The symbol is for the 5-pin package so if users already have this symbol in their schematic, we can hopefully assume they want this package, and since it's working for them we don't want to break things.

Also, it's ugly to cross over text with the lines. I think you can graphically leave the symbol alone.

Because the `SC*` FP filter works for the existing symbol, maybe we can leave that symbol alone except to update the description as you have to clearly indicate it's for a 5-pin package. I think noting the pincount in the description for both is a good solution for this issue until this library is totally reworked and perhaps better ideas pop out.

And then on the 6-pin symbol, we need to work on it because the FP filter is overly broad. Using `X*SON*` (both datasheets) and `SOT*886*`/`SOT*891*`/`SOT*1115*`/`SOT*1202*` (NXP only) should capture those types of packages using our library footprint names. The number is part of these specific packages' names, and not a pincount, so we can include it.

The trailing `_6` in a symbol name isn't something I can recall seeing before and isn't salient. Following the convention that Rene has clarified at https://github.com/KiCad/kicad-website/pull/407 it may be reasonable to call this one `74AUC1G66_SON` since all of the packages are Small Outline No-lead packages while the existing symbol matches packages with 5 gullwing leads.

@poeschlr 
This one is a bit tricky. Would you mind having a look and poking holes in my suggestions above?

chschlue:
> The trailing `_6` in a symbol name isn't something I can recall seeing before and isn't salient. Following the convention that Rene has clarified at [KiCad/kicad-website#407](https://github.com/KiCad/kicad-website/pull/407) it may be reasonable to call this one `74AUC1G66_SON` since all of the packages are Small Outline No-lead packages while the existing symbol matches packages with 5 gullwing leads.

That convention applies to _fully specified_ symbols.
Is it reasonable to append specific package names to _generic_ symbols? What's generic about that?

chschlue:
> Power pins are hidden in the logic library. It's gross, but leave that alone so the symbols on existing schematics don't change. The symbol is for the 5-pin package so if users already have this symbol in their schematic, we can hopefully assume they want this package, and since it's working for them we don't want to break things.

This one currently is especially gross. GND is at least on the symbols "bounding box", so to speak, but VCC is both outside the symbol and on a 50 mil grid.

poeschlr:
The logic libs really need fixing but it is not a good idea to do it during a minor release cycle. And remember the symbol file format will change so i guess once that is there then we might have other ideas how to properly fix these libs (would make fixing them now kind of redundant anyways)

evanshultz:
Could you note in the description the symbol is for the 5-pin package? This could be done by putting `, SOT-23-5/SC-70-5` at the end.

chschlue:
Sure

evanshultz:
Thanks!

PR #2082, opened 2019-08-17
ESP32-PICO-D4 added

waschhauser:
Push of ESP32-PICO-D4
https://www.espressif.com/sites/default/files/documentation/esp32-pico-d4_datasheet_en.pdf

Screenshot from 2019-08-17 17-08-12

Checking symbol 'ESP32-PICO-D4':
  Violating S4.4
     - Pin CLK (31) @ (900,500) is type BIDI : suggested INPUT

CLK can be used as input and output

Screenshot from 2019-08-17 17-12-56

---


diegoherranz:
Thanks for your contribution.

A few comments from a first review:

* I think this may belong to the `RF_Module` library, along with other ESP32 modules, because:
  * It includes crystal, flash, filters, etc. So it's not an MCU alone, but a module even if the package looks like a chip (System-in-Package).
  * Espressif lists it as a module: https://www.espressif.com/en/products/hardware/modules
  *  What do you think? Could you move it there if you agree?
* EN shouldn't be negated, it is active-high
* Can you try to make it somewhat similar to existing ESP32 modules symbols? For example:
  * SENSOR_VN next to SENSOR_VP
  * EN on top left
  * Maybe others? 
* Espressif doesn't recommend individual decoupling capacitors for VDDA or VDDA3P3 (see 7. Peripheral Schematics in the datasheet) and there is decoupling inside the package (see 6. Schematics in the datasheet). So, I think that those power pins could be stacked.
* Datasheet says: "Pins IO16, IO17, CMD, CLK, SD0 and SD1 are used for connecting the embedded flash, and are not recommended for other uses. For details, please see Section 6 Schematics."
  * I would group those pins in the symbol and keep them separate
* I would separate the LNA_IN pin from the rest and place it at top right or bottom right given its special importance and different characteristics (RF).
* I would place VDD_SDIO on the right-hand side (maybe at the bottom) given it's an output.

Many thanks!

waschhauser:
Hi Diego,

I've got some questions.

> Espressif doesn't recommend individual decoupling capacitors for VDDA or VDDA3P3 (see 7. Peripheral Schematics in the datasheet) and there is decoupling inside the package (see 6. Schematics in the datasheet). So, I think that those power pins could be stacked.

Does it mean I have to create a big stack that contains all 5 pins (in this case what is the name of the pin) or one stack for the **VDDA** pins and one other stack for the **VDDA3P3** pins?

The next question is related to the order of the **SENSOR_*** pins.
I decided to take a look to the "ESP-WOORM-32U" symbol.

image

And in the datasheet it looks like the following
image

For the ESP32-PICO-D4 it looks like the following
image

What is the recommended order of this pins in this case?
- SENSOR_VP
- SENSOR_CAPP
- SENSOR_CAPN
- SENSOR_VN

Is is OK to place the **SENSOR_*** pins on the right side because this pins are not used inside the chip?

After adjusting the **IO_*** pins in a similar way as it was done for the "ESP-WOORM-32U" the two pins **SD2** and **SD3** were left. What should be done with this pins?

image

Sorry it's the first time that I try to create a symbol.

Regards,
Richard

diegoherranz:
> Does it mean I have to create a big stack that contains all 5 pins (in this case what is the name of the pin) or one stack for the VDDA pins and one other stack for the VDDA3P3 pins?

One stack for the VDDA pins and one other stack for the VDDA3P3 pins.
Also, see [KLC S4.3](http://www.kicad-pcb.org/libraries/klc/S4.3/), namely the special case at the bottom.

> What is the recommended order of this pins in this case?
SENSOR_VP
SENSOR_CAPP
SENSOR_CAPN
SENSOR_VN

I hadn't seen that you may need caps between VP and CAPP and between VN and CAPN.
In that case, leave them in that order

> Is is OK to place the SENSOR_* pins on the right side because this pins are not used inside the chip?

The SENSOR_* signals are connected to the ESP32 chip inside the PICO package, so they can be used. I think they are fine on the left-hand side.

> After adjusting the IO_* pins in a similar way as it was done for the "ESP-WOORM-32U" the two pins SD2 and SD3 were left. What should be done with this pins?

They are GPIO9 and GPIO10. Maybe place them in that order?

> Sorry it's the first time that I try to create a symbol.
Nothing to be sorry for. You're making the right questions :)

Thanks again!

waschhauser:
Thank you for your comments.

I hope I have integrated all review comments.

But please take a closer look to the stacked pins.

Thx,
Richard

diegoherranz:
Thanks for the changes. The stacking pins look good!

- Just to make it even more similar and consistent with the other ESP32 modules symbols already in the library, can you rename U0TXD to U0TXD/IO3 and similar (e.g. SD2/IO9, etc.)
- Can you order the pins on the left which are not recommended to use: 16, 17, SD0, SD1, CLK, CMD
- Can you make the description and keywords following the same format / similar to the other ESP32 modules in the library?

Thanks!

waschhauser:
I have one question to the order of the pins 16, 17, SD0, SD1, CLK, CMD.
Why to use this order?

My idea was to use the oder who they are connected to the flash.

Picture how the 6 pins of the ESP-WOORM-32U are connected to the flash:
image

Picture how the 6 pins of the ESP32-PICO-D4 are connected to the flash:
image

Then I receive the following order

| Flash | ESP-WOORM-32U | ESP32-PICO-D4 |
|-------|---------------|---------------|
| 2     | SWP/SD3       | IO17          |
| 5     | SDI/SD1       | SD1           |
| 7     | SHD/SD2       | CMD           |
| 3     | SWP/SD3       | SD0           |
| 6     | SCK/CLK       | CLK           |
| 1     | SCS/CMD       | IO16          |

diegoherranz:
> I have one question to the order of the pins 16, 17, SD0, SD1, CLK, CMD.
Why to use this order?

It is arbitrary. I was simply trying to:
- Keep SD0 and SD1 together
- Keep CMD and CLK together
- Keep IO16 and IO17 together
- CLK and CMD at the bottom same as other ESP32 modules.

So that was why I suggested that order.

> My idea was to use the oder who they are connected to the flash.

Have in mind that the order of pins for that memory which Espressif used in their schematics is arbitrary too. It could be different, for instance see the symbol for 25LCxxx in the KiCad library which has the same signals. So I'm not sure trying to match the order of their memory symbol is necessary.

25lcxxx

Thanks!

waschhauser:
Thank you for the explanation.

I have made the changes.

image

image

diegoherranz:
Thanks for all the changes.

- There are extra zeroes on IO01, IO03 and IO09. They should be IO1, IO3 and IO9 same as every other single digit pin without the zero.

A couple of comments regarding the description.

- The SoC bit is which chip is used inside the module. See the "Chip Embedded" column on https://www.espressif.com/en/products/hardware/modules. For instance the ESP32-WROOM-32 module uses the ESP32-D0WDQ6 chip. For this ESP32-PICO-D4, given the smaller System-in-Package (SiP), it is probably just the silicon die inside. See that it simply says ESP32 in this case.
- Regarding, the package at the end, your QFN-48 was correct. The other ESP32 modules simply say SMD because they are not a standard package.

So I think it should be:

RF Module, ESP32 SoC, Wi-Fi 802.11b/g/n, Bluetooth, BLE, 32-bit, 2.7-3.6V, external antenna, QFN-48

- Regarding the keywords, it is external antenna, but not U.FL. The connector, if any, is up to the user. So:

RF Radio BT ESP ESP32 Espressif external antenna

We're nearly there :)

Thanks!

waschhauser:
I integrated your review comments.
Maybe it was not a good idea to make the modifications early in the morning.

image

diegoherranz:
Thanks again for the changes.

Just on thing pending: IO09 still has an extra zero. Can you fix it?

Thanks!

waschhauser:
:-(

I removed the "0" from the IO9

image

diegoherranz:
Looks good. Merging.

Many thanks for your contribution and all the changes done. I hope you continue contributing to KiCad.

waschhauser:
And thank you for the many reviews too

PR #2085, opened 2019-08-17
add RF switch MASWSS0178

hvraven:
This adds the RF switch [MASWSS0178](https://cdn.macom.com/datasheets/MASWSS0178.pdf).

Screenshot_20190817_210058

---


myfreescalewebpage:

- [ ] The footprint filter should be `MSOP*3x3mm*P0.65mm*`
- [ ] I suggest to have the same design than the "MSW-2-20" in the lib (except the drawing inside of course)

Cheers,
Joel

lorem-ipsum:
Hi @myfreescalewebpage ,

thanks for the review. I fixed the footprint filter. I have a question about adjusting the symbol. I do like the idea of having them interchangable, but the MSW-2-20 symbol is rather large due to the long pin names used by Mini-Circuits. When using the same outline the symbol will be unnecessarily large. Should I use the same vertical positions, but compressed in the horizontal direction?

Cheers,
Hendrik

myfreescalewebpage:
Good comment. "Should I use the same vertical positions, but compressed in the horizontal direction?" I think yes, it is a good idea.
Joel

lorem-ipsum:
I changed the drawing and pin layout as suggested. I kept the combination of closed and open switch from the MSW series. There is no documentation on the switch state if both CTL pins are low, so there is no correct layout for it.

MASWSS0178

myfreescalewebpage:
I disagree with the drawing inside the symbol.
With my previous comment I was thinking to have the same pin position to have something consistent across this kind of symbols. However, the switches were great before and are no more as shown in the datasheet now....

myfreescalewebpage:
@lorem-ipsum ping ?

myfreescalewebpage:
@hvraven ping

myfreescalewebpage:
@hvraven ping
Thanks
Joel

hvraven:
Hi @myfreescalewebpage ,

I just understood what your issue with the drawing was, of course you're totally right. What do you think of this version? It is not exactly pin compatible with the other switches (RFC is one position down), but it allows it to keep the style of the original drawing.
Screenshot_20200826_103805

myfreescalewebpage:
It's not bad, ok for me. Merging.

myfreescalewebpage:
Thanks for the update of this PR :)

PR #2087, opened 2019-08-17
add HCPL-0201 opto-coupler (and variants)

lorem-ipsum:
This adds the [HCPL-0201](https://docs.broadcom.com/docs/AV02-0674EN) and some variants as aliases for different packages.

Screenshot_20190817_212240

---


matthijskooijman:
IIUC, when multiple footprint options are available for a part, each footprint option should get its own part (rather than aliases), even if the pin numbering is the same for all footprints. This makes the parts fully specified, so you can just put them in your schematic and then the default footprint will be correct (with your current PR, you still need to pick the correct footprint even when you have already chosen it through the schematic symbol).

See also https://github.com/KiCad/kicad-website/pull/407 and https://github.com/KiCad/kicad-website/issues/386

matthijskooijman:
Looking more closely at the datasheet, I also see you:
 - Also included the HCPL2202, which has a different pinout (output on pin 6 instead of 7), so that should definitely be a separate symbol.
 - Missed a number of aliases from the datasheet. It specifies HCPL-2201, HCPL-2202, HCPL-2211, HCPL-2212, HCPL-2231, HCPL-2232, HCPL-0201, HCPL-0211, HCNW2201, HCNW2211, but you only included a number. Since these are mostly just non-functional variations, it would be easy to include them as aliases (except for the dual versions, which would need another symbol).

lorem-ipsum:
I expanded it into different parts for each possible footprint. My original plan was just to add the ones with identical symbol to the HCPL-0201, but I missed the pin different between the HCPL-xxx1 and HCPL-xxx2 versions.

matthijskooijman:
Thanks for updating, I have had a closer look at your symbols.

AFAICS, you now included all single parts from the datasheet. There are two variants of the pinout, one of which is available in three packages and the other only in one package, for a total of four symbols (and a bunch more aliases for differnt isolation ratings).

I had a look at the descriptions, but there seems to be an inconsistency.

In the datasheet, some parts are listed as 5kV/300V, with a footnote "c" stating they can be 10kV/1000V under some conditions:

image

However, in your symbol descriptions, it seems you have used 10kV/1000V for the HCNW2211 and 5kV/300V for the others. I'm not sure which of these would be best (perhaps to lower ones to be on the safe side?), but they should at least be consistent.

Furthermore:
 - The datasheet and footprint fields are on top of the symbol now. They are invisible, but for readability inside the symbol editor I would suggest moving them outside of the symbol.
 - The output pin is a bit funny. It seems you used a 150mil pin rather than 100mil, extending the pin inside of the body rectangle, so the inverting circle on the pin is actually inside the body. It also means that the pin number for the output pin is closer to the body than the other pins. Did you copy this from another symbol? I'm note sure what the conventions are exactly, but I think it would better to just use a normal 100-mil pin, with either a manually drawn circle inside the body (e.g. like the HCPL-063A), or with the inverting circle outside of the body.
 - I doublechecked the footprint associations, those look correct to me.
 - Electrical types look good to me.

lorem-ipsum:
I used the values from the second table on page 13 "Switching Specifications". There they use the higher current of 5mA. I am not quite sure how to understand the comment in the first table.

The symbol drawing is copied from the H11L. I will change it to a hand-drawn circle instead. I can also move the invisible fields to the outside. So far I never bothered about those.

matthijskooijman:
> I used the values from the second table on page 13 "Switching Specifications". There they use the higher current of 5mA. I am not quite sure how to understand the comment in the first table.

Ah, I see. Maybe the extra isolation really only applies to the HCNW2211 and HCPL-2232 but the first table fails to reflect that? Since your current descriptions match the (more detailed) table on page 13, I think they are good as they are right now.

> The symbol drawing is copied from the H11L. I will change it to a hand-drawn circle instead. I can also move the invisible fields to the outside. So far I never bothered about those.

Thanks. If the librarians agree that this is indeed better, maybe the H11L should also be fixed?

lorem-ipsum:
Checking the truth table again on page two the inversion itself is an error for these symbols. I fixed the pin length and added a line to fill the gap. Currently I am thinking it would be a good idea to draw the stylised Schmitt-Trigger as in the symbols on page 2 of the datasheet. Do you have any opinions on this?

matthijskooijman:
> Checking the truth table again on page two the inversion itself is an error for these symbols. I fixed the pin length and added a line to fill the gap. 

Good call, should have seen that. Thanks for updating.

> Currently I am thinking it would be a good idea to draw the stylised Schmitt-Trigger as in the symbols on page 2 of the datasheet. Do you have any opinions on this?

I think some optos have this, but not consistently. I'm not sure what the best style is here. Maybe following the lead of the datasheet? See also #2115 for a slightly related discussion.

lorem-ipsum:
I now added a drawing of the Schmitt-Trigger, similar to the one in the datasheet
Screenshot_20190927_181432

evanshultz:
I recall some conversation discussing not showing the triangles but just the lines that make up the Schmitt trigger symbol.

Ah! https://github.com/KiCad/kicad-symbols/issues/2115 This doesn't mean it's consensus, just that the people in that issue (and I) prefer no triangles. They're not always shown in part documentation and can result in clutter, while just the two 'L' lines are always shown and can convey the concept clearly.

lorem-ipsum:
I removed the arrows, now it looks like this:
Screenshot_20191002_134735

myfreescalewebpage:
Hi @lorem-ipsum I have performed a new review here, no comment to add. Merging. Joel

PR #2089, opened 2019-08-17
add LDO regulator NCV8674

lorem-ipsum:
This adds the linear regulator [NCV8674](https://www.onsemi.com/pub/Collateral/NCV8674-D.PDF) from OnSemi.

Screenshot_20190817_214326
---


cpresser:
Hi,

- [ ] Use "TO-263-3_TabPin4" as Footprint (and Filter). The Filter "TO\*263\*" will also match packages with just two pins (e.g. "TO-263-2_TabPin1"). The other option is to choose "TO-263-3_TabPin2", but thats not 100% according to the datasheet.
- [ ] Can you please add Pin4 to the GND-Stack so it matches the footprint.

evanshultz:
@cpresser 
I think this might be OK. Here's the default footprint assigned:
image

That matches with the datasheet pinout and doesn't have the middle pin (which the datasheet indicates isn't present).

The footprint filter is loose enough to allow for the footprint you mentioned above with 4 total pins (two are pin 2) if the users prefers. Perhaps for greater device availability?

@lorem-ipsum 
We generally don't allow automotive parts as they're usually duplicates. However, this datasheet only includes the NCV-prefix part. Is there a non-automotive version?

cpresser:
@evanshultz I agree that the default FP ("TO-263-2") matches. My main concern is the filter also matching FPs that are not compatible.

poeschlr:
@cp-aquila  You do not need to worry about filtering the pin count. This is done by a separate kicad feature.

lorem-ipsum:
In this case there appears to be no non-automotive variant. I searched for the 8674 part number, but couldn't find anything.

PR #2090, opened 2019-08-17
add EMI filter Schaffner FN406

lorem-ipsum:
This adds the EMI filter series [FN406](https://www.schaffner.com/products/download/product/datasheet/fn-406-ultra-compact-emc-filter/) from Schaffner. Footprint can be found in this PR https://github.com/KiCad/kicad-footprints/pull/1783.

Screenshot_20190817_220222

---


chschlue:
[G1.1](http://kicad-pcb.org/libraries/klc/G1.1/) doesn't mention pin names but I'm not sure if ' is a good idea.

poeschlr:
@chschlue Pin names are only used to communicate to the reader of a schematic. So they can use any character supported by kicad. And it is a good idea to have them agree with the datasheet.

The reason asset names are restricted is because different operating systems have different restrictions for file names and it is not yet clear if the symbols will eventually end up as one symbol per file like the footprints.

---

@lorem-ipsum looks good, thanks

chschlue:
@poeschlr If file name restrictions are the only reason then arbitrary characters in the description field should also be ok, right?

poeschlr:
Maybe @evanshultz can say more about it as i am not really experienced on the symbol side (I practically worked only on footprints the last few years.)
I at least am not aware of any rule restricting characters for description fields. (I am however not sure if kicad supports the full utf character set. But it should be quite obvious if we encounter one that does not work.)

evanshultz:
As @poeschlr said, we should stick with the minimal characters in KLC for best compatibility. We know other characters render fine in some OSes but KiCad runs on a lot of systems and can change. And regression testing of character set may not be done. In addition, changing character handling "under the hood" was actually being discussed recently so you never know...

That all being said, I think we have some Greek letters and probably some other stuff in symbol descriptions throughout the library. I pinged the dev mailing list to ask if they have any official guidance on the allowable character set to see if G1.1 can be expanded.

chschlue:
@poeschlr @evanshultz https://github.com/KiCad/kicad-website/issues/431

PR #2091, opened 2019-08-17
Add symbol for FTDI FT4222HQ

jonas-schievink:
This adds the FTDI FT4222HQ USB to Quad-SPI / I²C bridge.

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1786

screenshot-2019-09-29-18:55:27

---

  - ~~https://www.mouser.de/ds/2/163/DS_FT4222H-461024.pdf~~
  - https://www.ftdichip.com/Support/Documents/DataSheets/ICs/DS_FT4222H.pdf


chschlue:
* Wrong FP, see https://www.ftdichip.com/Support/Documents/TechnicalNotes/TN_166%20FTDI%20Example%20IC%20PCB%20Footprints.pdf#page=22
* Official DS link is https://www.ftdichip.com/Support/Documents/DataSheets/ICs/DS_FT4222H.pdf
* / and ² are disallowed

chschlue:
Some more things:
- [ ] [No double negations](http://kicad-pcb.org/libraries/klc/S4.7/) (pins 2 & 3)
- [ ] By swapping pins 9 & 10, QSPI bits would be in right order
- [ ] UGND and DGND should be stacked
- [ ] EP is missing

DS says
> The centre pad on the base of the FT4222H is internally connected to GND and the PCB should not have signal tracking on the top layer under this area. Connect to GND.

without mentioning which one. Do you know more?

jonas-schievink:
> Do you know more?

No, unfortunately. Any existing schematics I found just connect all the grounds together anyway. Should I add an extra pin for the pad?

poeschlr:
There are two questions in this discussion:
- Should the center pad have its own pin in the symbol? The answer is yes. No matter what the center pad is used for it always needs its own pin in the symbol as it would not be possible to connect with generic symbols otherwise.
- Should the pin for that center pad be stacked with some other pin? The answer is only yes if we can find out with which ground it must always be connected. (My guess is that the datasheet means that all ground pins are internally connected. The reason to have multiple grounds is such that one can have better control over return currents for specific signals. I guess it is not always clear which one of these grounds is best connected to the center pad on the pcb.)

jonas-schievink:
> Wrong FP, see https://www.ftdichip.com/Support/Documents/TechnicalNotes/TN_166%20FTDI%20Example%20IC%20PCB%20Footprints.pdf#page=22

I'm a bit confused since the datasheet also contains a package outline, which is quite different from the document you linked:

https://www.ftdichip.com/Support/Documents/DataSheets/ICs/DS_FT4222H.pdf#page=40

In particular:
- DS lists leads as 0.25x0.4, the TN as 0.3x0.75 (the latter should still work, but other footprints are using the exact package sizes and aren't this much oversized)
- DS lists EP as 3.5x3.5, the TN as 3.3x3.3

chschlue:
DS describes the package, TN describes the recommended FP for that package.

QFN and SOIC example:
qfnex
qfnex2

You can also see the package outline (purple) and the pads (red) sticking out in the TN:
qfnex3

OTOH, [F6.3](http://kicad-pcb.org/libraries/klc/F6.3/) forbids footprint EP to be smaller than package EP...

jonas-schievink:
Okay, I guess I'll try to comply with the TN, except I'll make the EP as large as in the datasheet.

chschlue:
You can use a 3.45mm EP, that's within the package tolerance.

poeschlr:
I am not really convinced we can trust the suggested footprints to be honest. They do not list the dimensions of the package in the same document so it could easily be the case that these are out of sync with each other. I would suggest to go with what IPC suggests which will result in a package with 3.5mm ep size.

Use https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON to generate the footpirnt. (Enter the dimensions as found in the datasheet into the fitting yaml file. run with python3 ipc_noLead_generator.py size_definitions/gfn/vqfn.yaml

jonas-schievink:
@poeschlr yeah, I tried doing that in https://github.com/pointhi/kicad-footprint-generator/pull/414 / https://github.com/KiCad/kicad-footprints/pull/1786

jonas-schievink:
The TN seems to specify these dimensions because the footprint is listed as being for FT4222HQ, FT245RQ, and FT232R. The EP of the FT245RQ and FT232R is 3.3x3.3mm, so there the footprint matches fine.

poeschlr:
Ah i should have read more than the latest responses. Well sorry for the confusion. You based the generated footprint on the wrong data however. See comment on the script pull request.

jonas-schievink:
The footprints have now been fixed and merged. Is there anything left that I need to do here?

chschlue:
Please use only the GPIO pins' main function as pin names.

Otherwise this should be good to go.

chschlue:
Oops.

Sry, one more thing: Replace commas in description with spaces.

chschlue:
Thanks

PR #2092, opened 2019-08-18
Add the TI TS3A27518E 6-channel 1:2 analog mux

jonas-schievink:
This adds the WQFN and TSSOP packages. There's also a BGA variant using "24-pin BGA Microstar Junior", which not in the footprint library. I might add that in a later PR, but for now these 2 variants should be fine.

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1785

screenshot-2019-08-19-13:59:48

---

  - http://www.ti.com/lit/ds/symlink/ts3a27518e.pdf


chschlue:
* Add packages to the descriptions
* WQFN EP is too large
* BGA is EOL, can probably left out
* : is disallowed

jonas-schievink:
> WQFN EP is too large

True, 2.6mm instead of 2.45mm. Do you usually create an entirely new footprint in this case?

> : is disallowed

Do you mean in the symbol description? (the guidelines don't mention this, they only say what's disallowed in symbol *names*)

chschlue:
> True, 2.6mm instead of 2.45mm. Do you usually create an entirely new footprint in this case?

Easier than it sounds, QFN footprints are scripted.
You basically add it to https://github.com/pointhi/kicad-footprint-generator/blob/master/scripts/Packages/Package_NoLead__DFN_QFN_LGA_SON/size_definitions/qfn/wqfn.yaml , generate it, create an FP PR and cross-reference them (generator, FP and symbol PRs)
 
> Do you mean in the symbol description? (the guidelines don't mention this, they only say what's disallowed in symbol _names_)

That should probably be clarified.

chschlue:
- [ ] [NC pin may be connected to GND](http://kicad-pcb.org/libraries/klc/S4.5/)
- [ ] EP should be connected to GND

jonas-schievink:
I think I didn't do the N.C. pin right. What electrical type should it get? Is "Not Connected" right?

chschlue:
You could set it to invisible, type NC and place its end on the symbol outline.
That way it's out of the way but still able to be connected if one really wants to.

poeschlr:
NC pins **must** be set to invisible and placed inside the symbol outline (or on the outline itself depending on if the outline is on the grid for pins)

See: https://github.com/KiCad/kicad-website/pull/377 for a clarified version of the NC pin rule set.

chschlue:
Apart from the fact that _invisible and placed inside the symbol outline_ is the current state of this PR, you can't expect anyone to follow **future** KLC versions.

poeschlr:
I only clarified that "should" simply is not strong enough. The KLC rules were sadly written in polite voice. In reality however this is a "must" rule. (which is what i intended to both clarify in the rule update and in my comment.)

jonas-schievink:
> See: [KiCad/kicad-website#377](https://github.com/KiCad/kicad-website/pull/377) for a clarified version of the NC pin rule set.

Thanks, that's helpful.

I've fixed the remaining warnings in the CI script and also updated the screenshot to reduce confusion.

chschlue:
Keywords must be separated by spaces. Everything else looks nice.

PR #2093, opened 2019-08-19
MCU_Module: Sparkfun Pro Micro

nerdyscout:
Arduino look-a-like Module but with an 32u4 microcontroller.
Symbol is therefore drop in compatible

Module: https://www.sparkfun.com/products/12640

image

[footprint pr](https://github.com/KiCad/kicad-footprints/pull/1690)

---


fauxpark:
Hi, some notes:
 - `RESET` is an active low input pin, so should be named `~RESET`, and placed on the left side
 - The GPIO pins should all be Tri-state per the 32U4 datasheet
 - `D0` and `D1` should go first in the pin names
 - `A0-A3` should be moved up to align with `D0-D3`
 - The `GND` pins should be stacked per [S4.3](http://kicad-pcb.org/libraries/klc/S4.3/) (I like to make the lowest pin number the visible one)
 - `VIN` should be renamed to `RAW` to match the silkscreen
 - The refdes, name and footprint positions need adjustment IMO - the ATmega symbols all place the footprint at (0,0) in italics, with the refdes in the top left, name at bottom right

nerdyscout:
> Hi, some notes:
> 
>     * `RESET` is an active low input pin, so should be named `~RESET`, and placed on the left side
I agree, but the original Arduino_Uno seems the be without ~ . I suggest to fix it there as well.
 
>     * The GPIO pins should all be Tri-state per the 32U4 datasheet

 
>     * `D0` and `D1` should go first in the pin names

done.

>     * `A0-A3` should be moved up to align with `D0-D3`

doing so the part would not be drop in compatible to an Arduino Uno with the same MCU. Therefore I disagree.
 
>     * The `GND` pins should be stacked per [S4.3](http://kicad-pcb.org/libraries/klc/S4.3/) (I like to make the lowest pin number the visible one)

Power pins of existing Arduino Uno symbols are not stacked as well. I would like to keep it the same. fixing the old smybol will break existing schematics, therefore I decided not to stack here.
 
>     * `VIN` should be renamed to `RAW` to match the silkscreen
 
>     * The refdes, name and footprint positions need adjustment IMO - the ATmega symbols all place the footprint at (0,0) in italics, with the refdes in the top left, name at bottom right

chschlue:
If this is supposed to be a drop-in replacement, why don't you just alias Arduino_Uno?

I don't think adding faulty symbols for compatibility with existing faulty symbols is the way to go.

poeschlr:
Using an alias is only an option if the same footprint is used. So if there is a difference on the footprint side then keep it as a separate symbol. If not then an alias would be better suited.

chschlue:
The way I understand this and the accompanying FP PR, both are supposed to be drop-in compatible, so I don't quite get the point.

poeschlr:
Same land pattern does not always mean same footprint. They might differ in the 3d model for example. This is why one needs to keep this in mind when deciding between alias and separate symbol.

matthijskooijman:
Also, the Uno symbol has *more* pins, so it is not a 100% drop-in replacement. But IMHO it would be good for the overlapping pins to be at least in the same position, so if you only use pins that are in both symbols, it is easy to switch between both.


image

matthijskooijman:
Looking at the Uno footprint, it seems to have 3V3 where this footprint has VCC. But I think the Pro Micro runs at 5V, so shouldn't VCC align with the 5V pin of the Uno?

fauxpark:
The Uno symbol has many of the same KLC issues that I mentioned. IMO, in its current state, it should not really be used as a basis for new Arduino compatible module symbols such as the Pro Micro, but I will let the librarians decide that. After all, it will be quite a while until the pins can be moved approaching the next major version.

nerdyscout:
> have 3V3 where this footprint has VCC. But I think the Pro Micro runs

The Pro Micro can be both - 5V or 3V3. see: https://cdn.sparkfun.com/datasheets/Dev/Arduino/Boards/Pro_Micro_v13b.pdf
To keep the pins separate I had now to to 2 different symbols.

image
image

myfreescalewebpage:
Hi @nerdyscout , thanks for contributing,

I haven't read all the comments, just making a fresh review.

Arduino_Pro_Mini
- [ ] Seems you have more pins than the number of pins on the module, why ?

Sparkfun_Pro_Micro_3V3
- [ ] Dx and Ax pins should be Bidirectional
- [ ] You should order pins Dx (D15/D16/D14 not ordered)
- [ ] Name of pin 21 is VCC

Sparkfun_Pro_Micro_5V
- [ ] Dx and Ax pins should be Bidirectional
- [ ] You should order pins Dx (D15/D16/D14 not ordered)
- [ ] Name of pin 21 is VCC
- [ ] VCC pin should be at the same place of 3V3 version

Cheers,
Joel

nerdyscout:
Hi @myfreescalewebpage, 
thank you a lot for your comments. This is a old pull request... I agree with most of your points but to stay compatible to the current Arduino_Nano_v3.x it should stay like it is. The problem is the current MCU_Modules do not conform to the KLC.

* initially I had them as bidirectional, but according datasheet they are tristate. see comments above.
* order of D15 D16 D14 are this way to stay drop in compatible with Arduino_Nano_v3.x and its SPI pins
* module is available as 3V and 5V, original Arduino module does have both pins. Therefore I created two independent symbols with pins in different locations
* gnd not stacked as original doesn't either

chschlue:
> * initially I had them as bidirectional, but according datasheet they are tristate. see comments above.

Tri-state will throw an ERC warning when connected to an output while bidirectional doesn't, which is why they should be bidirectional. See discussion at https://github.com/KiCad/kicad-symbols/pull/774#issuecomment-406773658 for example.

myfreescalewebpage:
The comment of @chschlue is correct, thanks for changing.
The other symbol already in the lib are not very nice I know. But it's not a reason to design the new one badly. It is good to enhance the symbols to conform to the KLC when we can. It's so more difficult later...

chschlue:
Oops.
The PR I mentioned above is actually the one where the pin types got screwed up, not the later one where they got fixed again. ;)
Still, bidirectional is correct.

myfreescalewebpage:
@nerdyscout any news here ?

nerdyscout:
> @nerdyscout any news here ?

Not really... I understand your point not to add bad symbols just because it has been done in the past. 
Adding new symbols independently from current symbols would work, but it is not really nice.
Having drop in compatible symbols would require breaking changes to old symbols (which, I guess, it would be the perfekt time now to do)

myfreescalewebpage:
So, do you want to stop working on this PR ? Old symbols can be updated now or later (and it's good to do it)

nerdyscout:
I guess I will come back to this when v6 is released, or at least in '[v6] MCU-* ' it was decided to write an module generator.

myfreescalewebpage:
Ok, no problem.
Just want to indicate you all the unmerged symbol pull request will be lost and will need to be done again because of change of the symbol file format.
Joel

nerdyscout:
this symbol is now part of #2028

PR #2095, opened 2019-08-20
Add several depletion mode FETs

electroniceel:
This adds generic depletion mode FET symbols to Device and specific FET models to  Transistor_FET.

Also the currently existing BSP129 symbol, which shows an enhancement mode fet, is replaced with the correct symbol for a depletion mode FET.

The symbols all use existing footprints like SOT-23, SOT-223 and SOT-89.


evanshultz:
@electroniceel 
Can you post a screenshot of the symbol? It's always a handy reference to have that here in the PR body, plus it's requested as part of the PR template above.

https://github.com/KiCad/kicad-symbols/pull/1976 standardizes this library using a minimal set of primitives to make up the symbols. On Monday I'll check this submission to see if it can be simplified or is already good in this respect.

electroniceel:
depletion-fet-symbols
This is a screenshot of all the new graphical symbols as I used them for the generic symbol entries.

The symbol entries for the individual parts just copy the respective graphical symbol and add the footprint, datasheet and so on.

evanshultz:
@electroniceel 
Apparently my comment didn't get committed...

Please add dots where the wires intersect, like for the pin 4 wires, as you can see in the `C3M0030090K` symbol in the library.

There should be three circles and nine polylines in these symbols, but the commit includes more polylines. Can you please review that? MASTER should now contain the simplified FET symbols so perhaps you can get inspiration from there. Let me know if you would like help with this.

I'm familiar with 3-terminal depletion mode FETs, but I haven't come across the 4-terminal symbols. Out of my own laziness, can you provide an example datasheet? Not all of the packages listed above would support a 4-terminal part anyway.

Also, you need to resolve the branch conflict.

electroniceel:
@evanshultz 
Thanks for reviewing my patches.

> Please add dots where the wires intersect, like for the pin 4 wires, as you can see in the `C3M0030090K` symbol in the library.

The dots make sense, they are needed for a correct schematic.

But why is the line to pin 3 in the `C3M0030090K` going left, next to the gate? I'd much prefer the pin 3 being right next to pin 2 in the symbol as you'd usually connect them to the same signal. And even if not, they logically belong together, so they should be next to each other. 

I copied the style I used from Q_NMOS_GDSD in Device.lib.

> There should be three circles and nine polylines in these symbols, but the commit includes more polylines. Can you please review that? MASTER should now contain the simplified FET symbols so perhaps you can get inspiration from there. Let me know if you would like help with this.

I must admit I haven't looked at the file level, just in the graphical editor. There could be unnecessary lines in there. I will fix that up once the best way forward is decided (see below).

> I'm familiar with 3-terminal depletion mode FETs, but I haven't come across the 4-terminal symbols. Out of my own laziness, can you provide an example datasheet?

The 4th terminal is the pad, for example on SOT-223 parts:
https://www.infineon.com/dgdl/Infineon-BSP129-DS-v01_42-en.pdf?fileId=db3a30433c1a8752013c1fc296d2395f

> Also, you need to resolve the branch conflict.

I'm wondering what the best way to proceed is. When creating the patches I didn't see #1976 at first.
 
Maybe split this pull request? First the generic variants for the device.lib and if that is merged and the dust settled after #1976, create a separate pull req for the individual model going into Transistor_FET.lib.

Or do you prefer keeping everything in this pull req?

evanshultz:
https://www.wolfspeed.com/media/downloads/1185/C3M0030090K.pdf shows that the fourth pin is an electrically distinct pin used for source sense. I didn't know the purpose of the fourth pins above so I was using this as an example, but from your comment above I don't believe it applies.

The BSP129 datasheet shows the tab is the same net as pin 2. You can discard the 4-terminal symbols unless there are other packages which may require them, as is done in the library already. You will find a variety of SOT-223 footprints which have two pin 2 pads so the symbol only needs 3 pins. (I realize these may be situations where the user would prefer to have the tab floating, as just one example, but the library currently doesn't support this and it would be easiest to ignore it here.)

Split PRs or keep together is your call. Either way is OK for me and I appreciate whichever method you choose to help improve the library.

Thanks!

electroniceel:
> The BSP129 datasheet shows the tab is the same net as pin 2. You can discard the 4-terminal symbols unless there are other packages which may require them, as is done in the library already. You will find a variety of SOT-223 footprints which have two pin 2 pads so the symbol only needs 3 pins.

Hmm, is it Kicad lib/symbol policy to put the tab distinction into several footprints instead of the library?

There are currently 2 footprints for SOT-223, one with tab on 4, one with just 3 pin numbers and tab on 2. So both ways would have a footprint to match.

SOT-223 may be easy in this regard, but look for example at TO-252: There are now 4 different footprints needed, just to accommodate all the tab variants.

I prefer to have the tab assignment explicit in the symbol and just one, generic footprint with the tab as 4th pin.
* you just need one footprint, Getting footprints right is more complicated, because they must be verified for production, paste layer adapted and so on. So you want to reduce variants of footprints more than variants of symbols.
* you can denote stuff like floating the tab or connecting a sense wire to pin 2 while connecting the power trace to the tab
* most parts have the pin and the tab connected internally. But I've seen parts where tab and pin aren't internally connected and the device requires the user to externally connect them. This was not in bare FETs, but ICs. These require a footprint with separate pin 2 and tab. You would just use the same method for these as for the others.

Now I want to contribute to an existing lib, so I can't force my opinions on others. So is there consensus among kicad lib maintainers that the pin/tab assginment should be done on footprint level or is it open to the contributors? In the latter case I'd opt to use 4 pin parts.

> (I realize these may be situations where the user would prefer to have the tab floating, as just one example, but the library currently doesn't support this and it would be easiest to ignore it here.)

What do you mean by "the lib currently doesn't support this"? Is this about a missing feature in Kicad lib format, or do you speak about policy regrading not putting the pad on pin 4?

evanshultz:
I know I read a clause in [KLC](http://kicad-pcb.org/libraries/klc/) about not duplicating symbols for pinout but instead handling this with footprints, but I went looking for it before I posted today and couldn't find it. It's a big document so I assume it's just me.

I mean the current library doesn't support discrete symbol pins for physical leads that are internally shorted. So not multiple symbols and fewer footprints, it's the other way around. And I agree there are valid situations where I've wanted this. You made another argument above.

Ultimately I think the answer is that the library is the way it is today and maintaining that is easier than opening a discussion that will block this PR from being merged. Opening a discussion in parallel, while merging this, is a good place to solicit thoughts and comments from other users as well. So if you want to change the way things please feel free to make a separate issue, for easier tracking, and see what happens. Maybe everybody agrees with you?

poeschlr:
@evanshultz wrote:
> I know I read a clause in KLC about not duplicating symbols for pinout but instead handling this with footprints, but I went looking for it before I posted today and couldn't find it. It's a big document so I assume it's just me.

Sometime in the past we had a rule that stated that the footprint always uses the numbering as in the datasheet (or in extension industry standards) and that the symbol must be drawn to fit that (if necessary duplicating the symbol)
I can no longer see this rule explicitly in the KLC so i guess it was removed with KLC version 3 (or it got reworded in such a way that i do not recognize it right now.)

Requiring footprints to be duplicated (as the quote above seems to suggest) would not make any sense to me. Most components use standardized packages that have a defined standardized numbering. So the KLC is written to fit these. 

We now have added an (as of yet) unwritten exception for Audio connectors. This is only possible for parts whose package is specific to only one application (a footprint for an audio connector is very unlikely to be reusable for anything else). Transistors however use generic packages that are also used to house devices of very different functionality.

myfreescalewebpage:
Hi @electroniceel , I try to finish this review now

BSP149 and alias
- [ ] Descriptions should be terminated by the package name `, SOT-223`

DN2470 and alias:
- [ ] I disagree with the footprint. The package has only 3 pins, this should be `Package_TO_SOT_SMD:TO-252-2` and the symbol has no pin 4 
- [ ] Footprint filter to be corrected

Else seems to be good. 

Cheers,
Joel

myfreescalewebpage:
@electroniceel ping

PR #2096, opened 2019-08-20
Add MAX1968 / MAX 1969 TEC drivers

chschlue:
Add MAX1968 / MAX 1969 TEC drivers

DS: https://datasheets.maximintegrated.com/en/ds/MAX1968-MAX1969.pdf

FP: https://github.com/KiCad/kicad-footprints/pull/1788

MAX1968xUI

---


chschlue:
Help wanted.

Not sure what to make of the footprint.

https://pdfserv.maximintegrated.com/package_dwgs/21-0108.PDF
https://pdfserv.maximintegrated.com/land_patterns/90-0146.PDF
Maxim calls this TSSOP although I thought TSSOP with EP was HTSSOP.

No matching FP exists. Should I call the new one SSOP or HTSSOP?

poeschlr:
Step one ignore the suggested footprint document as it references an outdated standard (The standard is based on pcb design and reflow requirements -> the machines used there are now a bit more precise.).

Head over to https://github.com/pointhi/kicad-footprint-generator/tree/master/scripts/Packages/Package_Gullwing__QFP_SOIC_SO and use it to generate the footprint (use the metric dimensions given in the datasheet.)
[Documentation for the generator (work in progress but should help)](https://github.com/pointhi/kicad-footprint-generator/blob/b9cba75fa7ec5314015ed94a0cd9a1de350b7c0c/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/Readme.md)

---

regarding name: The datasheet uses TSSOP (page 1) so why is this even a question?

chschlue:
> regarding name: The datasheet uses TSSOP (page 1) so why is this even a question?

@poeschlr I don't think the question is as silly as you make it out to be.

Given that
1. There are lots of HTSSOP-1EP in the lib
2. There is just one TSSOP-1EP in the lib
3. There is no fitting generator yaml for TSSOP with an exposed pad
4. JEDEC MO-153 says that only thermally enhanced TSSOPs have an exposed pad (duh!)
5. And in turn calls thermally enhanced versions HTSSOP

I was wondering if a standardized name or something a single manufacturer made up should take precedence (or if this deviation calls for an [F2.3](http://kicad-pcb.org/libraries/klc/F2.3/) manufacturer specific name, which I think is ridiculous in this case).

That's why this is even a question.
Look, I don't _have_ to contribute anything, I'm just trying to help here. No need to treat me like an idiot.

poeschlr:
The reason why i asked "why is this a question" is because i do not have the time to research this stuff (We maintainers have hundreds of contributions to take care of and are also just volunteers who do not need to do this stuff.) 
All you provided were the two options and the datasheet. (Yes i could have worded it differently and will keep that in mind for the future.)

---

With this new information the HTSSOP name really makes a lot more sense and i would be ok with using it.

chschlue:
Alright. No hard feelings.

> We maintainers have hundreds of contributions to take care of and are also just volunteers who do not need to do this stuff.

I am aware of that and I think you're doing great work there.

I kept the question short so as not to waste anybody's time and because I assumed it was (kind of) common knowledge that this kind of package is usually called HTSSOP.
May well have been that somebody would just come by and answer something like "we decided years ago to always use the datasheet name no matter what." Solved in 30 seconds.

chschlue:
Next question:
Is the new library Driver_TEC OK? Couldn't find anything fitting.

cp-aquila:
> Next question:
> Is the new library Driver_TEC OK? Couldn't find anything fitting.

I also could not find any library that suits this type of device. Not even remotely.
Making a new lib for two parts also seems overkill. I don't expect that we will get more TEC drivers to fill this library.
So my conclusion is 'no clue how to handle this'.

chschlue:
Driver_Miscellaneous?

cp-aquila:
That does sound reasonable.
But can you name one other possible symbol that would go into that library? I can't come up with just one that does not fit into one of existing libs.

chschlue:
Actually, a lib for miscellaneous drivers should probably simply be called Driver.

Driver_TEC has some growth potential, there are others, e.g. https://www.analog.com/media/en/technical-documentation/data-sheets/ADN8833.pdf

myfreescalewebpage:
Hi @chschlue , thanks for contributing,

A comment I have during my review:

- [ ] Footprint filter should be `HTSSOP*4.4x9.7mm*P0.65mm*`

Else no comment.

Cheers,
Joel

chschlue:
Oops.
Fixed.

myfreescalewebpage:
Thanks. Merging.

PR #2099, opened 2019-08-24
Reference_Voltage: add aliases for MAX62xx

fenugrec:
Aliased to closely-related MAX63xx parts (max6350, max6341, max6325).

---


cpresser:
Thanks for the addition.
Can you please change the description so that the two variants (63xx and 62xx) are alike?
Right now, the user has to look up the datasheet so see what is the difference between the two variants.

> 1ppm/°C Low-Noise +2.5V Voltage Reference, SO-8/DIP-8
> 0.5ppm/°C Low-Noise +2.5V Voltage Reference, SO-8/DIP-8

fenugrec:
Hi,
I would argue that anyone designing something with these parts has probably already closely studied the datasheets, but yes, I could change the descriptions.
However I'm not sure it's strictly correct to rename 63xx 0.5ppm. While the 63xx *does* have variants that are spec'd at 0.5ppm/C, the datasheet title still calls that whole series 1ppm/C.

The 62xx datasheet has a more generic title even, with no mention of ppm.  Some 62xx variants go down to 1ppm/C. Another subtle difference is the aging drift, 20 (62xx) vs 30 ppm/1000h (63xx), but I don't think that kind of detail belongs in the component name in kicad ...

Considering those details, I thought it would be best to follow (closer) the datasheet titles, which are arguably not great but still authoritative.

If I haven't convinced you, then ok, I'll change both component descriptions.

poeschlr:
please change the descriptions for them to be clear

fenugrec:
Done, but I disagree with changing the 63xx description.

cpresser:
> Done, but I disagree with changing the 63xx description.

As long as the user can distinguish the variants by looking at the description and sees the major differences everything is good.

And you are right, looking at 'features' on both datasheets I can see 1ppm/°C, it is just the text on the left which differs. My bad.
What is the major difference between the two series. You have studied the datasheets already, that might save me the time to search for it.
If there is no difference between the series, than I am fine with having the same description.

fenugrec:
As I understand it :
- The 63xx series has a variant that guarantees < 1ppm/C (typical 0.5), so they decided to say "1ppm" in the DS title.
- The 62xx best variant only guarantees < 2ppm/C (typical 1.0), so marketing probably said "make sure you mention 1ppm somewhere on the front page".

Maybe a better solution would be thus : 
1. keep original 63xx description that says `1ppm/°C Low-Noise Precision...`  : follows DS title and max guaranteed spec;
1. describe 62xx as `2ppm/°C Low-Noise Precision...`, where 2ppm is the max guaranteed spec as compared with the 63xx.

evanshultz:
I would prefer the typical value be in the description, as it is now. Looks good to me.

PR #2100, opened 2019-08-24
Add HT75xx-1 voltage regulators

twam:
Datasheet: https://www.holtek.com/documents/10179/116711/HT75xx-1v250.pdf
Screenshot:
image

---


fenugrec:
I'm new here, but you should consider using wildcards instead of adding every voltage, since there are many. See for example https://github.com/KiCad/kicad-symbols/commit/0897998bb58e2f70f5606abdcd11f70c4d344478 
the 'xxx' part of the MPN will match voltages.

chschlue:
You added the wildcard symbol instead of replacing.

twam:
> You added the wildcard symbol instead of replacing.

Damn it. Still too early ;)

chschlue:
Can you add FP filters for SOT-23-5, SOT-89 and TO-92 (and remove default FP)? This appears easiest to catch all as Holtek doesn't seem to use package suffixes.

twam:
What would be the correct filter for SOT-23-5? If I put SOT*23* it will match against 3-pin SOT-23 which is wrong and with SOT*23*5 it will not show the 5-pin version as only 3 pins are used.

chschlue:
SOT?23?5*

chschlue:
Oh, and pin 2 is VIN and 3 is VOUT

poeschlr:
You do not need to include the pin count. Pin count is filtered by a separate filter of kicad.

chschlue:
> You do not need to include the pin count. Pin count is filtered by a separate filter of kicad.
> Edit: you will however need to include the not used pins in the symbol (set to NC, invisible, inside the outline) I do not believe that it would be wise not to include them in this case.

[KLC disagrees](http://kicad-pcb.org/libraries/klc/S4.5/)

> Edit2: i also disagree with the idea of removing default footprints. We prefer fully specified over generic in this cases for a reason.

I couldn't find any manufacturer package suffix (in fact I also had a hard time finding anyone who actually sells these parts (and a hard time figuring out why anyone would want to use them)), so my suggestion seemed a KLC-compliant way to match all three variants without making up suffixes for the symbol names.

poeschlr:
@evanshultz do we really not include NC pins in voltage regulators?

RS components has the sot89 variant and gives it the name [HT75xx-1-SOT89](https://uk.rs-online.com/web/p/low-dropout-voltage-regulators/8850721/)
This name would also fit with what we suggest in the KLC.

chschlue:
SOT89 is the easy one, try and find TO92 and SOT23 ;-)

Calling it HT75xx-1-SOT89 fully specifiying it that way is fine with me, but like I said, the KLC example seemed to be a simple way to catch all three packages.

KLC needs to be updated to reflect what is actually done and not done...

poeschlr:
Well the name for the others would follow the same example. I kind of expected that there is no need for me to list them explicitly.

And no examples of a specific rule do not mean other rules are not to be followed. The rule you look at only looks at footprint filter naming in isolation. Other rules determine that fully specified is to be used where feasible. This should be even clearer once https://github.com/KiCad/kicad-website/pull/407 is merged. (This rule update clarifies how we librarians use the rule set right now.)

twam:
The part is available at LCSC and will be also available for the most likely very popular assembly serviced by JLCPCB. That's at least the reason why I'm using it. :)

Name with SOT89 suffix and no other footprint filters? Or include them?

chschlue:
Name it HT75xx-1-SOT89, set default FP to SOT-89 and add a single filter to SOT?89*.

Then add symbols for SOT-23-5 and TO-92 if you like.

chschlue:
Is this finished now or are you going to add the other variants?
Either way, please change pin names to reflect the datasheet.

twam:
I wouldn't add the other variants for now, as they are quite uncommon as you already pointed out.

chschlue:
Thanks.

PR #2104, opened 2019-08-28
Add MAX31865

matthijskooijman:
This adds the MAX31865 RTD temperature sensor amplifier.
Datasheet: https://datasheets.maximintegrated.com/en/ds/MAX31865.pdf

image

This adds the SSOP and TQFN versions. They are mostly identical, except:
 - Pin numbering is different
 - TQFN has an exposed pad connected to GND (represented as a hidden stacked pin in the symbol).

There is one remaining checklib violation:

     - Pin SDO (14) @ (-700,200) is type TRISTATE : suggested OUTPUT

However, the SDO pin is a MISO pin that is tristated (to support multiple SPI devices), so TRISTATE is more appropriate.


chschlue:
- [ ] Analog power pins are called VDD and GND in the DS. Perhaps also flip them with the digital ones so analog power is on the "analog side" of the symbol
- [ ] Force+/- are power outputs

matthijskooijman:
> Analog power pins are called VDD and GND in the DS. Perhaps also flip them with the digital ones so analog power is on the "analog side" of the symbol

We added the "A" to make it clearer that these are the analog power pins, in contrast with the D power pins (based on the pin table, which describes them as "Analog supply voltage" and "Analog ground"). I've looked in the KLC for any rules regarding pin names, but that only specifies rules for active-low pin names, I could not find any rules for pin names in general. Is the convention to always match the datasheet? If so and you'd rather seem them as VDD/GND, I can change that.

Swapping the pin locations is a good idea, I'll change that.

> Force+/- are power outputs

Are they really? They are not "Power and ground pins" ([KLC S4.4](http://kicad-pcb.org/libraries/klc/S4.4/) I think, since the they are not always enabled. Also, FORCE+ is not a (regulated) voltage output, but only a resistor-limited current output. If anything, the BIAS pin could be seen as a power output, since it supplies 2.0V permanently (though the maximum current output is really limited, only 5mA or so).

The KLC is a bit vague on the electrical type of pins (does not specify the options exhaustively), but AFAIK the point of these is mostly to facilitate ERC, right? Looking at the default configuration, it seems that the only difference is that Power output to bidir gives a warning (instead of ok) and power output to tristate gives error (instead of warning). Others are identical between power output and output.

image

Also, I think that each power input must be driven by exactly one power output (or power flag), but I'm not so sure these pins actually provide proper power for this (e.g. if you would connect the FORCE+ to some other IC's VCC pin, things would not actually work, so should that not trigger an ERC error?

I'm always a bit confused about the meaning of the electrical types. The KLC is not exhaustive and a quick Google shows no other official documentation for them. I did find [this forum post](https://forum.kicad.info/t/please-explain-electrical-type-when-create-new-ic-pin/742/3) that has a list. AFAIU, the question is mostly whether a pin can be driven (output and friends), is high-impedance (input) or just passive (not driven, not high-impedance).


image

Looking again at the right side pins of this chip, I would say:

 - REFIN+, REFIN-, FORCE2, RTDIN+ and RTDIN- are inputs, since they are (through switches or multiplexing) connected to the ADC.
 - ISENSOR and FORCE+ are (through a switch) connected to each other, but not inputs and not driven, so these should be passive?
 - BIAS could perhaps be power output, since it supplies a fixed voltage, or maybe just output (since it is not really intended to power other devices)?
 - FORCE- is essentially open-collector (since it switches between GND and floating), but I'm not sure if that is appropriate here?

chschlue:
Please stick to datasheet names.

As to the pin types: These are always good for an endless discussion and should have read more than the first half datasheet page.
I would say:
* REFIN+/-, RTDIN+/- and FORCE2 are voltage-sensing inputs
* ISENSOR and FORCE+: like you said, passive.
* BIAS and FORCE-: might be a pair of power outputs as they actually supply power to the RTD, OTOH, they aren't power outputs in a general sense. Output/input isn't a reasonable combination either as that would imply that FORCE- is high-Z. FORCE- may technically be an open collector/drain, but isn't used in the logic output sense here.
If I don't miss anything, BIAS and FORCE- as power outputs shouldn't upset ERC if this part is used as intended. Otherwise, there's only passive left.

matthijskooijman:
Thanks for your considerations. I have updated the symbol per your comments:
 - Renamed power pins according to datasheet
 - Swapped digital and analog power pins to put them at the right side of the symbol
 - Your proposal for electrical types looks good. I tried power output for BIAS and FORCE- in an actual design, which satisfies DRC, so made them power output.

I think that resolves all open comments.

chschlue:
> I think that resolves all open comments.

Yes, thanks. Time for new comments :)

- [ ] Please replace the first A (denoting temp range) with an x
- [ ] TQFN variant calls for a new footprint. Yours would be ok EP-wise but isn't thin. Existing TQFN-20s don't fit. So it's got to be TQFN-20-1EP_5x5mm_P0.65mm_EP3.25x3.25mm.
.

evanshultz:
Just looking at the image in the original post, this symbol looks excessively large.

chschlue:
I'd like to keep the pin grouping on the right, but it could be narrower.

matthijskooijman:
>  Please replace the first A (denoting temp range) with an x

How can you tell? In the datasheet I can only see:

image

Or is this standard with Maxim parts? If so, should the trailing "P" also be an x?

>  TQFN variant calls for a new footprint. Yours would be ok EP-wise but isn't thin. Existing TQFN-20s don't fit. So it's got to be TQFN-20-1EP_5x5mm_P0.65mm_EP3.25x3.25mm.

I had assumed thin vs non-thing would not matter for the footprint (also since there was no TQFN footprint either), but it indeed seems that the TQFN footprint used is *slightly* smaller. I'll see about adding the proper footprint.

> Just looking at the image in the original post, this symbol looks excessively large.

Good point. I think I can indeed make it a bit more narrow, maybe also make it 100 mil smaller at the bottom.

chschlue:
> Or is this standard with Maxim parts? If so, should the trailing "P" also be an x?

https://www.maximintegrated.com/en/design/packaging/package-information/maxim-naming-conventions.html
TLDR; first A -> temp, second A -> package type, P -> pin count

> I had assumed thin vs non-thing would not matter for the footprint (also since there was no TQFN footprint either), but it indeed seems that the TQFN footprint used is _slightly_ smaller. I'll see about adding the proper footprint.

The height matters for the 3D model, so TQFN packages should actually be using TQFN FPs.

chschlue:
@matthijskooijman ping

matthijskooijman:
@chschlue Thanks for the reminder, I'll have another look tomorrow.

matthijskooijman:
I've rebased on top of master, and fixed the review comments above (see commit message). The screenshot in the first post is also updated.

This now refers to a non-existent footprint, I'll have to look adding that one later.

matthijskooijman:
I've rebased on top of master, and fixed the review comments above (see commit message). The screenshot in the first post is also updated.

This now refers to a non-existent footprint, I'll have to look adding that one later.

matthijskooijman:
I've added the needed footprint (and a bunch of related ones) in https://github.com/KiCad/kicad-footprints/pull/2008

chschlue:
Thanks.

PR #2108, opened 2019-08-30
Add R_Pack04_Split

DecadoPL:
During creating device with many resistors around microcontroller it was very hard to use rigid R_Pack arrays. Because of that i've created R_Pack array containing 4 separate resistor symbols.

image
image
image


DecadoPL:
Travis is failing but this symbol is copy of oryginal R symbol. Tell me what should i do with it.

poeschlr:
This does contain more than one contribution. Please make a separate branch for every contribution and make sure that branch starts at the current HEAD of master.

PR #2109, opened 2019-08-30
Add R_Pack04_Split

DecadoPL:
During creating device with many resistors around microcontroller it was very hard to use rigid R_Pack arrays. Because of that i've created R_Pack array containing 4 separate resistor symbols.

This is continuation of PR#2108 that has been closed because i had to delete the old branch. 

image
image
image


cpresser:
Just a few thoughts:
- This looks like a useful addition for me. I want to use that for my future designs :)
- I have now clue why this symbol throws the off-grid-travis errors, but `R` which has identical pin locations does not
- The Description should contain "DIP"
- FP-Filter should allow both concave and convex

evanshultz:
@cpresser 
Why does the description need to include `DIP`? I don't find any DIP resistor array footprints in the library currently, though they are available (depending on pinout, our normal DIP footprints might be fine). And this PR includes the pinout for a DIP package.

Specifying concave or convex in the FP filter would preclude DIP packages because those terms are for SMD R arrays.

Also note that we have a symbol generator which makes our existing R array symbols and I expect it would be the preferred solution to use that. See https://github.com/KiCad/kicad-library-utils/tree/master/schlib/autogen. We wouldn't dismiss a good PR in this case but just FYI.

cpresser:
> @cpresser
> Why does the description need to include `DIP`? I don't find any DIP resistor array footprints in the library currently, though they are available (depending on pinout, our normal DIP footprints might be fine). And this PR includes the pinout for a DIP package.

I think it would be nice it this looks consistent:
image

poeschlr:
Do resistor arrays not use the default DIP packages? (Which we have in the lib)

cpresser:
> Do resistor arrays not use the default DIP packages? (Which we have in the lib)

The filter for `R_Pack_4` is
`DIP*`
`SOIC*`

But my usual usecase is something like `R_Array_Convex_4x0603`

evanshultz:
Concave/convex packages are mutually-exclusive with DIP packages.

DIP and convex/concave can share a pinout, but SIP needs a unique pinout.

poeschlr:
We might then want to add R_Array* as a footprint filter (Assuming footprints like these are in the name. And assuming they have the same pinout as DIP)

And of course SIP needs different pin numberings (I am however uncertain if pure resistor arrays are even available in SIP. I thought SIP was more a thing for resistor networks.)

DecadoPL:
@cpresser 
I'm glad to hear You like it :)

@ALL
What it the final decision? How should i change that?

cpresser:
> Concave/convex packages are mutually-exclusive with DIP packages.

@evanshultz 
Can you elaborate? I dont quite understand why.
Taking the  `R_Pack4` symbol as reference, DIP, SOIC and R_Array footprints all match and work. Or am I missing something here?

@DecadoPL 
My suggestions are:
FP-Filter: `DIP*` + `SOIC*` + `R_Array_Concave*` + `R_Array_Convex*` (using `R_Arrax*` would also match the THT-SIP packages that have a different pinout).
Don't add the 'DIP-Package' to the description. It already has 'parallel topology' which is enough info.

I would also suggest to expand the current `R_Pack*` symbols fp-filters with `R_Array_Con***` and remove the text 'DIP Package' from the description. I will open a seperate PR for this once this is merged to avoid conflicts.

cpresser:
> And of course SIP needs different pin numberings (I am however uncertain if pure resistor arrays are even available in SIP. I thought SIP was more a thing for resistor networks.)

Just for reference: Bourns sells these 
image
They are also available as different network topologies in the same package (https://www.mouser.de/datasheet/2/54/600M-777112.pdf).

evanshultz:
You can't put DIP in the description if the FP filter only matches convex/concave. With the list of filters you give above it should be OK. But replace underscores with asterisks.

DecadoPL:
@evanshultz @cpresser 
Did what You said. I think it is ready to merge now.
image

cpresser:
There is missing a star at the end of 'convex'

DecadoPL:
@cpresser 
Fixed!

cpresser:
Thanks for the PR and that last missing byte :)

PR #2110, opened 2019-08-30
Isolator: Fix names and descriptions for HCPL-061/063/261/263 

matthijskooijman:
Some had anode and cathode labels swapped, or used VO1 rather than VO
where there was only a single output. These were probably copy-paste
errors. The pin numbering and drawings are correct.

Datasheet: http://www.farnell.com/datasheets/1266137.pdf
Here's the HCPL-061A/N and HCPL-063A/N, the HCPL-261A/N and HCPL-263A/N have been fixed in the same way.

image
image


matthijskooijman:
I rebased on top of master, improved the commit messages and added one more commit that fixes the descriptions (which I also found to be broken).

chschlue:
Good job, just a little confusing that you inserted screenshots of the broken symbols ;-)

matthijskooijman:
> just a little confusing that you inserted screenshots of the broken symbols ;-)

Oh, not sure how that happened, woops :-)

Thanks for merging!

PR #2111, opened 2019-08-30
Add 6N137 and related optoisolators.

matthijskooijman:
This adds the 6N137, VO2601, VO2611, HCPL2601 and HCPL2611 single
optocouplers (all functionally equivalent), and the related VO2630,
VO2631, VO4661, HCPL2630 an HCPL2631 dual optocouplers (also
functionally equivalent). For the dual version, also a _Split version
is added.

Datasheets: https://www.vishay.com/docs/84732/6n137.pdf and https://www.onsemi.com/pub/Collateral/HCPL2631-D.PDF

Some remarks:
 - There are multiple vendors that have these chips. For the 6N137, they all seem to use the same model number, for the others they seem to use different letters (I found VO26xx for Vishay and HCPL26xx for Fairchild/onsemi), so I just added all versions as aliases.
 - I based the symbol drawings on the Vishay datasheet, and the overall design and pin placement on the (very similar) HCPL-063A.
 - From the HCPL-063A I imitated the smaller pin names, placed on the outside, since there is no free space on the inside. I considered omitting the pin names altogether (since the diagram makes the pinout clear), but that would leave the VCC/GND pins unlabeled. On the _Split version, I used normal (inside 50mil) names for VCC/GND and omitted the others (which are clear from the diagram), imitating the 7402 symbol.
 - Some of these are available in alternative footprint (In particular Gull-wing DIP and bended-lead-DIP). However, I think there are no separate footprints for bended-lead DIP and the gull wings fall into to the FP filter already. It seems other optos are in the same position and only supply the main DIP version.

image
image

And the three units of the split version:
image
image
image

---


matthijskooijman:
Hm, I just realized that these parts are effectively the same as the HCPL-261/HCPL-263, so are better off as aliases of those. The diagram is a bit different on them, but actually better reflects the workings of these parts than the diagram I copied from the Vishay datasheet.

PR #1889, opened 2019-05-29
Add Analog ADN4651

cronJ:
Add Analog ADN4651

Created from [https://www.analog.com/media/en/technical-documentation/data-sheets/ADN4650-4651-4652.pdf](https://www.analog.com/media/en/technical-documentation/data-sheets/ADN4650-4651-4652.pdf)

image

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

PR #2112, opened 2019-08-30
Isolator: Fix ADN4651 and ADN4652 order in the files

matthijskooijman:
These seem to have been accidentally swapped in a merge resolution as
part of #1889 and now cause uncommitted changes whenever the Isolator
library is saved by the symbol editor. Swapping them restores the
intended sort order and fixes this.

I am omitting screenshots, datasheet, etc. since this only swaps symbols, it does not change them in any way.


antoniovazquezblanco:
Thank you!

PR #2113, opened 2019-08-30
Isolator: Add 6N137 and related optocouplers

matthijskooijman:
This adds:
 - 6N137, HCPL-2601, HCPL-2611, VO2601 and VO2611 (single optocouplers in
   DIP) as aliases of functionally equivalent HCPL-261A.
 - HCPL-2630, HCPL-2631, HCPL-4661, VO2630T, VO2631T and VO4661T (double
   optocouplers in DIP) as aliases of functionally equivalent HCPL-263A.
 - HCPL-0600, HCPL-0601, HCPL-0611, VO0600, VO0601 and VO2011 (single
   optocouplers in SOIC) as aliases of functionally equivalent
   HCPL-061A.
 - HCPL-0630, HCPL-0631, HCPL-0661, VO0630T, VO0631T and VO0661T (double
   optocouplers in SOIC) as aliases of functionally equivalent HCPL-063A.
 - CPC-5002 (double optocoupler in DIP) as an alias of functionally equivalent HCPL-263A.

This only adds aliases, since these optocouplers are functionally
equivalent to the existing ones (only differences are in things like
isolation ratings, input current, and for the VOXXXX and CPC5002 also
the manufacturer).

This is somewhat related to #2110 in that it uses alias descriptions similar to the the ones fixed by that PR.

Datasheets can be found here:

 - https://docs.broadcom.com/docs/AV02-0940EN
 - http://www.vishay.com/doc?84607
 - https://www.vishay.com/doc?84732
 - http://www.ixysic.com/home/pdfs.nsf/www/CPC5002.pdf

I am omitting screenshots, since this PR only introduces aliases, not complete symbols.


chschlue:
Looks good so far.
I noticed that all the existing symbols you touched got anodes and cathodes wrong. Would you be willing to fix that?

matthijskooijman:
Thanks for the review.

> I noticed that all the existing symbols you touched got anodes and cathodes wrong. Would you be willing to fix that?

Already did, see #2110 :-)

chschlue:
Great, missed that.

PR #2114, opened 2019-08-30
add LTC1562 quad active RC-filter in SSOP-20

aewallin:
LTC1562 active filter (four op-amps) in SSOP-20 package
https://www.analog.com/media/en/technical-documentation/data-sheets/1562fa.pdf
LTC1562-2 alias: https://www.analog.com/media/en/technical-documentation/data-sheets/15622fa.pdf

LTC1562xxG

---


cpresser:
A few thoughts - no full review yet.
- [ ] Shutdown should be on the left since it is an input pin
- [ ] I would like to see AGND to the right of GND, thats what we do for other chips.
- [ ] Can you please rebase this so its a trivial merge? The current branch has conflicts.


Do you have an example screenshot that shows the symbol in use? I am wondering if having all outputs on the left and inputs on the right works for this symbol. Currently it looks similar to the datasheet which is not bad per se, but does not conform to KLC.

aewallin:
thanks for the comments.

- AGND to bottom right
- SHDN to lower left
- should now merge
- I would prefer the datasheet layout as most users will use one of the example circuits presented in the datasheet (for e.g. low-pass, band-pass, high-pass application)

chschlue:
The alternative is a multi-unit symbol, one unit per filter section and a fifth for power and SHDN.

aewallin:
> 
> 
> The alternative is a multi-unit symbol, one unit per filter section and a fifth for power and SHDN.

Are you suggesting that the "DeMorgan"-mechanism can/should be used here?

For practical reasons I think following the datasheet makes sense. I agree that if you are teaching active filters in a course or a textbook you would draw explicitly op-amps.

chschlue:
No, I'm suggesting the "Number of Units: 5" and "All units are not interchangeable" mechanisms can be used ;)
Example: First part in the lib (alphabetically), the non-IEEE 4001.

aewallin:
@cpresser ping! anything further I can do to help this get merged? many months now since the last comment. thanks.

cpresser:
> @cpresser ping! anything further I can do to help this get merged? many months now since the last comment. thanks.

Basically you can do nothing but join the reviewers team :) But asking after that much time is okay.

I did some more checks:
- [x] I think the current layout is plausible, at least from my experience I would rather copy stuff from the datasheet. So having individual symbols is not a real advantage.
- [ ] Keywords are space separated, please remove the `,`
- [ ] Footprint-filter should also contain the pitch (`SSOP*20*5.3x7.2mm*P0.65mm*`)
- [x] Pin-numbers, names and electrical type are good.

cpresser:
Thanks!

PR #2116, opened 2019-08-31
Fix pins for BD139 and BD140

RavenFire:
Swap the Base and Emitter pin numbers to align with the datasheet.

Datasheet:
http://www.st.com/internet/com/TECHNICAL_RESOURCES/TECHNICAL_LITERATURE/DATASHEET/CD00001225.pdf

Original symbols:
bd139_old

New symbols:
bd139_new


chschlue:
This fix won't do.
These symbols each several aliases for which the old pinout was correct, see https://www.onsemi.com/pub/Collateral/BD238-D.pdf

I also suggest using ST's speaking DS URLs like https://www.st.com/resource/en/datasheet/bd140.pdf

jkriege2:
Please cross-check against our footprint ... I have the feeling that in the DS they number 1-2-3 right to left, while the KiCAD footprint number 1-2-3 left to right!

jkriege2:
see e.g. https://www.onsemi.com/pub/Collateral/BD139-D.PDF

chschlue:
True, ST has it the wrong way round, missed that.
In that case the symbols are correct but the DS links should be updated.

evanshultz:
@RavenFire 
Please see and check the above conversation. Would you instead transform this PR into a doc update and revert the pin number changes?

chschlue:
@evanshultz 
The pin swap is fine but we probably shouldn't link to datasheets from the only manufacturer that numbers pins RTL to avoid confusion.

chschlue:
Please use MF DS links if possible, i.e.
https://www.onsemi.com/pub/Collateral/BD135-D.PDF
https://www.onsemi.com/pub/Collateral/BD136-D.PDF

chschlue:
@RavenFire reopen if you are still on this.

PR #2117, opened 2019-09-01
Added STM8S208RB

aris-kimi:
Hi. I have added STM8S208RBT6 under MCU_ST_STM8.
I have described all pins' functions inside symbol's body.
I don't know why those additions and deletions. Please help!! Or if possible assign this to someone who can make it happen further!!

stm8s208RB


aris-kimi:
Datasheet: https://www.st.com/resource/en/datasheet/stm8s208rb.pdf

cpresser:
Thanks for you contribution.
However, I do have a few thinks that should change:
- Can you try to match this to the existing symbols? Like this one? image
- The library usually does not allow to have all alternative functions on the pins. It is okay to add one (the main) alternate function though.
- Can you do a screenshot with white background so we can read the blue text?
- Please sort the Ports

There are some rules we follow when adding new parts so the library is consistent. http://kicad-pcb.org/libraries/klc/
But we can help you with those.

myfreescalewebpage:
@aris-kimi thanks for your contribution. Can you follow the advice of @cpresser so we continue the review ? Thanks, Joel

aris-kimi:
stm8s

aris-kimi:
What do you think? I kept reset vcap and references at bottom left, if it is not acceptable inform me to change it.

myfreescalewebpage:
Hi @aris-kimi , here is my review:

- [ ] The name of the device should be "STM8S208RB" (I don't know what is T6, is it tube packaging?)
- [ ] The format of the package at the end of the description is not correct, it should be `24 MHz, 128 KB Flash, 2KB EEPROM, 10-bit ADC, 2 UARTs, SPI, I²C, CAN, LQFP-64`
- [ ] You should group PA1 and PA2 with the other pins of the port PA
- [ ] The main function should be on the border of the symbol, thus name of pin 2 should be "PA1/OSCIN" and name of pin 2 should be "PA2/OSCOUT"
- [ ] VCAP should be Input
- [ ] Concerning the place of VCAP, OK, other devices are done like that, but VREF+ and VREF- go to the top and bottom of the symbol
- [ ] You can easily reduce the width of the symbol, there is actually plenty of place lost!

Cheers,
Joel

aris-kimi:
stm8s

aris-kimi:
@myfreescalewebpage Thank you Joel. I had corrected them all, T6 or T3 is about temperature but i removed it. I have also added R8 and R6 wich have 64 and 32kb memory as aliases. Pin PI0 is not in order but i think it is better like this. Tell me if not to make it like it was.

myfreescalewebpage:
The result is very nice, thanks for all those fixes !

"T6 or T3 is about temperature" => OK. We discard temperature so it is ok with the current name of the symbol. Thanks also for the two aliases.

No more comment, merging.

Joel

PR #2118, opened 2019-09-01
Add ATSAMD51J18/19/20A

fauxpark:
sym

ARM Cortex M4F MCU, 256/512/1024kB flash, in TQFP and VQFN.
Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/60001507E.pdf

---


cpresser:
Hi,

- [ ] I think !RESET should go to the top, to be more consistent with other ATSAMx5x parts ( "ATSAME53J18A-A") 
image

- [ ] Pins should be 'Bidirections", not Tristate according to KLC S4.4
- [x] Pin-Types for VSW, VDDIO?, VDDCORE, VDDANA are good
- [ ] Pin 7 is GNDANA and should not be stacked
image

fauxpark:
- Not sure what you mean by this, you say `~RESET` should be placed at the top, which I have done, but you show an image of the opposite? Having the pin at the top is consistent with and in the Atmel/Microchip AVR parts that I previously did a lot of work to clean up.
- Per the datasheet, [S7.3.3](http://ww1.microchip.com/downloads/en/DeviceDoc/60001507E.pdf#_OPENTOPIC_TOC_PROCESSING_d136e48132), these pins are capable of being tri-stated. This too is consistent with the AVR symbols.
- Fair point on `GNDANA`, I misremembered the negative power pin stacking rules (which incidentally are not in KLC). And also as you show Fig. 7-1 says they are separate circuits.

chschlue:
> * Per the datasheet, [S7.3.3](http://ww1.microchip.com/downloads/en/DeviceDoc/60001507E.pdf#_OPENTOPIC_TOC_PROCESSING_d136e48132), these pins are capable of being tri-stated. This too is consistent with the AVR symbols. IMO KLC should be updated to take this into account.

This has been taken into account. A bidirectional pin is one that can push, pull and sense logic levels in its high impedance state -- tri-state + input basically.

> * Fair point on `GNDANA`, I misremembered the negative power pin stacking rules (which incidentally are not in KLC).

KLC does not differentiate between positive and negative power pins.

poeschlr:
For a generic symbol: Bidirectional is the correct pin type to for fully configurable gpio pins. I however do not agree with the logic by @chschlue. It has nothing to do with them being in high impedence when in input mode. 
It is simply the case that bidirectional is the only one that can be used (input can not be used because it can be configurated as output. Similarly output can not be used because it can be configurated as input. Threestate can not be used because the pin can be configurated as push pull and as input, Same logic for open collector.)
In reality there simple is a feature missing to properly work with generic MCU symbols. One would need a way to tell ERC which pin is used in which mode. Hopefully something like this will come in v6.

fauxpark:
@evanshultz has told me previously to use the Tri-state pin type for GPIO pins that are capable: https://github.com/KiCad/kicad-symbols/pull/774#issuecomment-406645213

> KLC does not differentiate between positive and negative power pins.

Again, I was [previously advised](https://github.com/KiCad/kicad-symbols/pull/1894) to *not* stack *positive* power pins, and that's what I misremembered.

chschlue:
This doesn't make any sense to me. A GPIO pin that isn't tri-state capable cannot be used as an input.
That there may or not be an explicit difference between input and tri-state logically / firmware-wise is a completely different question.

chschlue:
I just realized I could have been more precise.

A 1Wire pin for example is bidirectional but not tri-state, but what is
commonly known as a GPIO pin is necessarily bidirectional and tri-state
capable.

So calling all non tri-state capable uC IO pins bidirectional is like
calling only non-white polar bears "bears" and the others "horses".

poeschlr:
@chschlue you might want to look more deeply into how pad output stages can look like. There are many things that can be used as input and output while not being truly three state capable. Three state not only requires that this pin can be in high impedance mode but that it can be in this mode while being configurated as output (there must be a separate output stage enable and output logic enable option).

---

Regarding threestate for I/O pins: if evan really suggested the use of threestate then it was a mistake. Threestate clearly states in its description that this represents an output pin. If the pin can be used as input as well then threestate must not be used! (This is not about ERC at this point but about the logic of it. The only kicad electrical type that can be used for a pin configurable for both input and output is bidirectional.)

poeschlr:
There is another option available that we so far never use. We could set user configurable pins to "unspecified" which will always result in an ERC error. This would force users to make a project specific symbol.

fauxpark:
Good point. The datasheet for the Microchip 8051 symbols for instance do not mention the word "tristate" anywhere; the GPIOs sense input solely with the pullup. I was cleaning those up at the time too, and there seemed to be a difference between them and the AVR symbols, thus they were made Bidi and the AVRs Tristate.

So, this means instead the 8051s should be Output and the AVRs Bidi, yes? Are pin type changes acceptable at this point in the dev cycle?

cpresser:
This has GNDANA on the right of GND-Stack (others r"SAM[DE]5\d" have it to the left), but that's something purely cosmetic that we can fix in the future. 
Or does somebody think that needs to be changed now?
Otherwise this seems to be ready to merge.

chschlue:
That would be a breaking change after this is merged.

Different topic: Microchip really doesn't provide actual part datasheets for these, just the 2000 page complete manual, right?

fauxpark:
That *is* the datasheet. It just happens that there are a lot of different configurations and variants that don't warrant separate datasheets (they all share the same ARM core).

RE the GNDANA pin placement - there is one other ATSAM symbol with it on the right so it is already inconsistent.

chschlue:
> That _is_ the datasheet. It just happens that there are a lot of different configurations and variants that don't warrant separate datasheets (they all share the same ARM core).

Just asking. I'm not familiar with SAMs but I know they provide complete and summary datasheets for other MCUs, the latter consisting of what is usually considered to be a datasheet. So I was wondering if linking to the full programming manual in a symbol library might be overkill.

> RE the GNDANA pin placement - there is one other ATSAM symbol with it on the right so it is already inconsistent.

That's not a valid argument for adding more inconsistency :-)

fauxpark:
Sorry if that sounded rude, that wasn't my intention.

> That's not a valid argument for adding more inconsistency :-)

I just don't see the point in caring too much about it at the moment. The convention with the AVR symbols is to have the analog power pins to the right, so IMO moving it would be adding inconsistency. Although, since AREF pins are input and thus go on the left, perhaps *they* are wrong... 😄

chschlue:
No offense taken.

Re anal pin position: That actually _is_ an argument.

cpresser:
To add even more confusion, when https://github.com/KiCad/kicad-symbols/pull/2146 is merged 2 more parts will have GNDANA to the left of GND.

chschlue:
It's not too late change #2146

cpresser:
> It's not too late change #2146

True - and I am open for that. But that does not solve the issue of having different styles, there is already one symbol which has it on the left.

I feel like we are bikesheeding here. But I also fail to see what the solution is.

chschlue:
I feel the new symbols should match the hundreds of existing AVRs and the
existing SAMs should be adjusted later. I also feel the bikeshed should be
painted taupe.

PR #2124, opened 2019-09-03
fix minicircuits ww2 library links

lorem-ipsum:
As pointed out by @aewallin Mini-Circuits moved their datasheets from ww2.minicircuits.com to www.minicircuits.com. This fixes the existing symbols in the library


antoniovazquezblanco:
Thank you!

PR #2127, opened 2019-09-03
[Interface_USB] Add Microchip MCP2210 USB-to-SPI converter

maximeborges:
Add Microchip MCP2210 USB-to-SPI converter with all variants


maximeborges:
I just saw that #2094  already add this component and variants, but on mine I used the symbol from the MCP2200 as reference so the components looks almost the same, with more descriptive second/third functions on GPIO pins.

diegoherranz:
Hi, @maximeborges.

I didn't have much time lately to look into this. During that time, there has been quite some progress on https://github.com/KiCad/kicad-symbols/pull/2094, to the point where it is nearly ready to be merged. I'll close this one and carry on there. Bear in mind that I'm trying to make the symbols slightly similar to yours, with SPI pins and USB pins on opposite sides (see https://github.com/KiCad/kicad-symbols/pull/2094#issuecomment-541296759).

Many thanks for your submission. I hope we see more pull requests by you in the future :)

PR #2131, opened 2019-09-06
BAV99 fixes

evanshultz:
https://assets.nexperia.com/documents/data-sheet/BAV99_SER.pdf

BAV99 was a two-unit symbol with a duplicated pin 3. This is dangerous, and we fixed this with opamps having duplicated power pins, because it can allow an unintentional net short.

It's now a single-unit symbol like our other dual-unit diodes:
image

I also make the dot at the internal wire intersection for BAV99S be on both units (it was missing on the second unit):
image

---


cpresser:
BAV99 does not have the package-name in the description.
Besides that I found no other problems.

evanshultz:
@cpresser 
I didn't even notice that was always missing. Added now. Thanks!

cpresser:
Thanks! 
Out of curiosity, why is there a 2nd commit which is empty (e811612)?

evanshultz:
Thanks!

Because that's what I had at the time, I used the web interface to make this change. When the page didn't save I hit the 'commit' button again.

PR #2137, opened 2019-09-09
Add microMELF diode symbols

evanshultz:
Simple copies of the existing diode symbols with an updated package and some ALIASes.

http://www.vishay.com/docs/85597/bzm55.pdf
image

http://www.vishay.com/docs/85566/mlc4148.pdf
http://www.vishay.com/docs/85545/bav300.pdf
image

---


lorem-ipsum:
MLC and BZM look good. Two comments about the BAV-series:
- BAV303 has a lower current in the description. Couldn't find any differences in the datasheet
- Wouldn't it make more sense to use the reverse voltage in the description instead of the repetitive peak reverse voltage?

evanshultz:
- Oops. Fixed!
- I've changed it as you mentioned. KLC doesn't really control descriptions and there is a conversation about it at https://github.com/KiCad/kicad-symbols/issues/1617. Without any guidance or consistency I selected the value that I like to see, which is why the descriptions for many parts vary greatly within a library.

Thanks for checking!

cpresser:
I dont quite understand the 4148 datasheet. It says 
> Electrical data identical with the devices 1N4148 
and 1N4448 respectively

But then there is this table:
image

Anyway, this looks good. Drawing and Footprints+Filter look good. As do description, keywords and data-sheet links.

evanshultz:
Thanks!

I believe the datasheet is saying that 1N4148 and MCL4148 are the same electrically, just like 1N4448 and MCL4448. Not that MCL4148 and MCL4448 are the same electrically.

PR #2138, opened 2019-09-09
Move MCP6561 to Comparator.lib

evanshultz:
I noticed that [MCP6561](http://ww1.microchip.com/downloads/en/devicedoc/22139c.pdf) somehow ended up in with opamps when it's a comparator. The symbols are copied from [MCP6566](http://ww1.microchip.com/downloads/en/DeviceDoc/DS20002143E.pdf) with the output pin type changed.

This should wait for v6, I believe, since it's moving symbols between libraries.

image
image
image
image

---


antoniovazquezblanco:
This may lead to user mistakenly using a comparator when wanting to use an opamp. On the other hand moving it before v6.0.0 may break designs 🤔

poeschlr:
As this fixes a bug it is allowed before v6 (As you say right now it could lead to users selecting the wrong symbol.)

evanshultz:
Alright. The milestone was already removed so it's fine with me to merge once someone has time to review.

Issue #2041, opened 2019-07-30
Transistor_FET: TP0610L and VP0610L can't be used as aliases for 2N7000


frantony:
2N7000 is a **N-channel** MOSFET. *P0610L are **P-channel** MOSFETs.

evanshultz:
2N7000 is extremely common so I think that's understood.

The *P0610L datasheet (https://www.vishay.com/docs/70209/70209.pdf) confirms this. Oops.

@frantony 
Thanks for catching this and taking the time to report it. I've marked this as a bug.

PR #2143, opened 2019-09-10
Fix xP0610L and BS250

evanshultz:
Fixes https://github.com/KiCad/kicad-symbols/issues/2041

- TP0610L and VP0610L were an ALIAS of 2N7000 but they are P-channel; make them an ALIAS of BS250 (same datasheet)
- Sadly, BS250 had the S and D pins swapped

https://www.vishay.com/docs/70209/70209.pdf

image

---


cpresser:
BS250 is not the same as TP0610L and VP0610L. They are in the same datasheet, but have a different pinout. So they cant be an alias.
image

* BS250 has DGS
* TP0610L and VP0610L have SGD

TO-92 should work for both types, TO-29L in the library seems to same as TO-226

evanshultz:
Yesterday was not my day for making changes...

Thanks for catching and noticing my error. Fixed now. I also corrected BS250's drain-source voltage.

PR #2144, opened 2019-09-10
Add Lattice ICE40HX4K-TQ144

joshtyler:
Adding ICE40HX4K-TQ144 Lattice FPGA
[Datasheet](http://www.latticesemi.com/view_document?document_id=49312)
[Pinout](http://www.latticesemi.com/view_document?document_id=49385)
Unit A
image
Unit B
image
Unit C
image
Unit D
image
Unit E
image

---


joshtyler:
I'm still getting a warning about S3.1 being violated, but I think this is because https://github.com/KiCad/kicad-library-utils/blob/master/schlib/rules/S3_1.py is calculating the center as the center of all pins over all units together, and also ignores the graphical outline.

Each unit is visually well centred (whilst keeping pins on a 100mil grid).

cpresser:
Did you use some sort of scripting to generate this? If yes, can you provide the sources? That might be easier to review instead of comparing each pin to the datasheet.

joshtyler:
I used KiPart to generate the symbol (https://github.com/xesscorp/KiPart)

I reformatted the xlsx file provided by Lattice (link in original message) into this csv file which was processed by kipart:
[kipart_import.zip](https://github.com/KiCad/kicad-symbols/files/3602999/kipart_import.zip)
(csv is zipped because Github doesn't allow csv upload)

The pins in the CSV are still in the same order as in the xlsx so that file should be easy to verify.

The generated part directly imported into KiCad is commit bd1978b. I then tidied it to match KLC, and moved a couple of pins (CDONE and ~CRESET) to a different unit, but didn't re-number any pins.

cpresser:
I started reviewing this.
To bad the HX1K and HX4K have different IO and we can't just define an alias.

But the Non-IO Pins seem to be identical or at least similar:
image
image

- Can you try to match the style of that last unit?
- The pin names are different. Looking at the family-data-sheet it looks like the ones in HX1K are more correct (CDONE, !CRESET_B, ...)

joshtyler:
Thanks for reviewing this!
Yes it is a shame that the parts differ. [There is a migration guide](http://www.latticesemi.com/view_document?document_id=49426), but as you will see there are significant differences in pinout.

I've reformatted the last unit to closely match the HX1K
image

As for the pin names:
- CDONE was already named CDONE (the 'C' just happened to line up precisely with the greyed out 'NC' in my screenshot).
- CRESET_B, I believe that my naming is correct. CRESET_B is Lattice's name for the pin. It is an active low configuration reset (with the _B designating that it is active low), therefore according to [KLC S4.7](http://kicad-pcb.org/libraries/klc/S4.7/) the correct thing to do is to prefix it with ~ and remove the _B

cpresser:
Not sure if you closed this intentional?
Maybe because both your PRs use your master branch? If you want to continue with this I suggest you use one branch for each PR.

joshtyler:
No, it wasn't intentional - sorry!

I accidentally made this PR from master, and then couldn't change it to a different branch, because Github doesn't have that feature. I then sorted out my local repository, including moving this part to its own branch, but thoughtlessly force-pushed master.

My other PRs are on their own branches already, so I've merged the commits onto master, and I can now re-open this. We are now back to where we were before - sorry again!

cpresser:
* I did check all pins (names & numbers) with a script, just to be sure. :heavy_check_mark: 
* `~CRESET` vs. `CRESET_B`: stripping the suffix is according to KLC :+1:  
* NC-Pins should be on the outline of the package, but that doesn't work here since the pins are 150mil long. So lets keep them where they currently are :heavy_check_mark: 

What I could not figure out yet is if we actually want to stack or split the `VCCIO` and `VCC` pins.
I assume every pin gets get a dedicated decoupling cap in a 'proper' design. But, I could not find any hint in any of the data-sheets on how to handle decoupling caps. The schematic-checklist does have details about `VCCPLL`, but not on the other supply pins.
Do you have more info on that?

joshtyler:
Great, thanks!

I'm afraid I don't have any information on the the VCC and VCCIO decoupling. I did look quite hard when first making the symbol, and so I'm fairly sure Lattice simply have not documented it. 

You're correct that a 'proper' design would probably allocate a cap to each VCC pin (advisable, but perhaps not quite so critical for VCCIO), as well as probably a bit of bulk decoupling for the chip too.

I stacked the pins for two reasons. Firstly to follow KLC S4.3 (Power supply pins must be stacked unless the datasheet specifies the need for decoupling capacitors on every pin.), but also to allow easier migration to other iCE40 parts, which have also stacked the pins.

cpresser:
Giving that Lattice did not specify to add decoupling to each pin, the current symbol follows KLC S4.3.
I just wanted to make sure that neither of us missed that.
Everything looks good. Lets merge!

PR #2146, opened 2019-09-11
Fix Pin7 on ATSAME5x devices

cpresser:
Pin7 is GNDANA and should not be part of the GND stack.

http://ww1.microchip.com/downloads/en/DeviceDoc/60001507E.pdf
See https://github.com/KiCad/kicad-symbols/pull/2118, there is a screenshot of the datasheet for a similar device (same family)

image
image

Screenshots for reference, but this is very simple and can be reviewed by looking at the diff :)

---


evanshultz:
For consistency and, from what I've seen, convention, could you make pin 22 (the lowest-numbered one) the visible GND pin on `ATSAME53J18A-M`? It's not in KLC but in practice I can only recall seeing this. Probably it should be there...

cpresser:
As discussed in https://github.com/KiCad/kicad-symbols/pull/2118 I moved the GNDANA pins to the right of GND.

As a side-note I also checked the ATSAME70, it does have pins named GNDANA, GNDPLL, .. and so on. But the datasheet says they should be connected to the GND-Plane, so its fine to stack them with GND.

evanshultz:
I missed this. Thank you.

PR #2148, opened 2019-09-12
Add W9812G6KH SDRAM

joshtyler:
Add W9812G6KH SDRAM
[Datasheet](https://www.winbond.com/resource-files/da00-w9812g6khc1.pdf)
Pinout is p.4 of datasheet, table showing variants is p.3

The symbol is based off IS42S16400J-xC which is pin compatible, only changes are naming to some pins (GND->VSS, GNDQ->VSSQ, BAx->BSx), and properties such as value and datasheet.

image

---


myfreescalewebpage:
Hi @joshtyler , thanks for contributing,

- [ ] It seems the pinout and package is the same for all the symbol, in this case you can make only one and then create aliases for all the others
- [ ] Pin length should be 100mil
- [ ] You should stack VSSQ pins
- [ ] You should stack VSS pins
- [ ] Footprint filter should be `TSOP?II*22.2x10.16mm*P0.8mm*`
- [ ] The pin style of CLK should be Line (we do not use clock style in the official lib)

Cheers,
Joel

joshtyler:
Thanks for the review @myfreescalewebpage.

Regarding pin length and stacking of pins, I left the pin length at 200mils, and didn't stack pins to allow easy migration between this part and IS42S16400J-xC (which already exists in the library and is pin compatible).

Is it better to make these changes, or keep the ability to swap between the two parts?

myfreescalewebpage:
No it's better to make the changes to go in the direction where the symbol are conformed to the KLC.
If you want to update also the old symbols, this is always welcomed too :)
Joel

joshtyler:
Sorry for the delay, I've now made the requested changes. 

I also stacked the power pins, because KLC S4.3 says they must be stacked unless the datasheet states one decoupling capacitor per pin (which it doesn't).

Stacking the pins left whitespace at the top and bottom, so I've reformatted the shape slightly.
image

myfreescalewebpage:

PR #2149, opened 2019-09-13
Change GPIO pins back to Bidirectional for Microchip AVR symbols

fauxpark:
As pointed out in #2118, the Tri-state pin type for these symbols' GPIO pins is wrong, and they should just be Bidirectional instead.

---


chschlue:
https://github.com/KiCad/kicad-symbols/issues/387

cpresser:
image

poeschlr:
@cpresser are you aware of https://github.com/KiCad/kicad-library-utils/blob/master/schlib/comparelibs.py ?

cpresser:
> @cpresser are you aware of https://github.com/KiCad/kicad-library-utils/blob/master/schlib/comparelibs.py ?

Tbh I have not looked at it into detail before.
I just tested it and in this case I don't think it is useful. I already know that basically all symbols in the lib have been changed.

Anyway, related question: Is there also a clever way to feed two git commit-ids to comparelibs?

PR #2159, opened 2019-09-14
Add CH340E

joshtyler:
Add CH340E USB to UART converter.
There doesn't appear to be an english datasheet available from the manufacturer directly. The best source I can find is [here](https://www.mpja.com/download/35227cpdata.pdf). All package and pinout information is in sections 3/4.

The symbol looks like it is laid out in a slightly odd manner, this is because the pin locations are copied from the existing CH340G, to allow easy migration between the two.
image

---


myfreescalewebpage:
Hi @joshtyler , thanks for contributing,

- [ ] Where have you seen V3 is pin 10 ?
- [ ] TNOW in place of UART control ine is a bit strange, I propose below the USB connections, what do you think ? Else I have other solution but we lost the compatibility with CH340G.

Capture

Cheers,
Joel

joshtyler:
Hi Joel,

Thanks for the review!

V3 is shown as pin 10 in the package diagrams at the top of page 2 of [the datasheet ](https://www.mpja.com/download/35227cpdata.pdf). I have just noticed that the pin tables at the bottom of that same page do contridict this and show V3 as not present (which I assume is why you are asking). I've done some research and can't find any further information unfortunately.

I have built this in hardware, and can confirm that connecting pin 10 to 3V3 does work in real life (although leaving it disconnected _may_ also work).

image

TNOW below the USB data lines works for me. I originally put it on the right hand edge since it is an output, but I think your solution is better. The only thing I would suggest is leaving a space between the data lines and TNOW to show it is a different functional group. I've gone ahead and made this change to the symbol in my last commit.
The part now looks like this:
image

joshtyler:
I've just looked at the Chinese datasheet [hosted by the distributor LCSC](https://datasheet.lcsc.com/szlcsc/1811151452_WCH-Jiangsu-Qin-Heng-CH340E_C99652.pdf), and it appears to show pin 10 as V3 in the datasheet:
image

Likewise[ the manufacturers product page](http://www.wch-ic.com/products/CH340.html) shows pin 10 as V3.

joshtyler:
I realised you didn't leave a space for TNOW because it will conflict with CH340G. I've moved it again, so it now matches your suggestion exactly.

Sorry for all the comments!

image

myfreescalewebpage:
Hello @joshtyler 
Thanks for all those details, great job !
No more comment, merging your PR.
PS : if you think all CH340x have been added and you wish to submit an enhancement of the symbol to have something better in design for example for the TNOW pin, it's possible and we will review it, but this will only be merged for V6 of KiCad (modifications of symbols are done only during major release).
Joel

PR #2160, opened 2019-09-14
Add TPS22917 Load Switch

sudhi345:
Add TPS22917 load switch
This modification adds ultra low power load switch from Texas Instruments.
---
- [ ] URL to the [datasheet](http://www.ti.com/lit/ds/symlink/tps22917.pdf) for the symbol
Screenshot from 2019-09-14 16-19-47
- [ ] An example screenshot image
- [ ] Ensured that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [ ] Give a reason behind intentional library convention rule violation - S4.2
It's more convenient to have Vin pin on the left side.


cpresser:
Hi,
there seem to be two additions mixed up in this one PR. I also found the component "TPS62056DGS" in the changeset.
Can you please make one PR for one part.

sudhi345:
Hello,
The additions are under different branches and PRs but they all are being shown here.

A capacitor is connected between pin CT and Vin to change output slew rate and QOD is to discharge the output. So pin CT is on left side and pin QOD is on right side.

evanshultz:
@sudhi345 
You have pushed branches for unreleated parts into the same PR. And the PR title, screenshot, and datasheet above are for just one part. Please don't do that because reviewing is harder.

Separate the additions into branches, which can be grouped if the additions are related, and then push separate PRs for each branch. That will make reviewing easier and faster.

And thanks for contributing to the official KiCad library. Welcome!

sudhi345:
Thanks for your suggestions. I will make separate PRs.
Closing this PR.
Thanks again.

PR #2162, opened 2019-09-14
Fix 74LS192/193

### LINKED ISSUE #2161: Incorrect electrical type for pins 12,13 on 74LS192/193
The symbol for the 74LS192, 74LS193 list pins 12 (~CO) and 13 (~BO) as inputs.  These pins should be outputs.  See http://www.ti.com/lit/ds/symlink/sn74ls193.pdf

This is causing ERC to errantly report that the pins are not driven when connected to pins 5 and 4 of the next 74LS193 (when chaining them together).

chschlue:
Fixes #2161

* Update DS links
* Fix pin types
* Fix pin names
* Update library version

http://www.ti.com/lit/ds/symlink/sn74ls193.pdf

74LS192
74LS192_IEEE

---


chschlue:
That's "Power input" with the power cut

evanshultz:
Ugh, this symbol was especially mangled but all the logic libs are awful. The fix looks OK to me.

In the future, if the file version is changing you can make one commit to update the file (like moving one line and then moving it back) and then a second commit to implement the intended change. This makes review so much easier because everything isn't mixed together into one commit.

chschlue:
I noticed the version part too late, will do.

PR #2164, opened 2019-09-15
Fix AVR ISP connectors

chschlue:
F'up to #2059

* Fix pin lengths
* Fix pin types
* Fix description

https://www.microchip.com/webdoc/GUID-9D10622A-5C16-4405-B092-1BDD437B4976/index.html?GUID-9B349315-2842-4189-B88C-49F4E1055D7F

AVR-ISP-6
AVR-ISP-10
AVR-JTAG-10
AVR-PDI-6
AVR-TPI-6

---


matthijskooijman:
Changes looks good to me. As with #2059, I would suggest also enlarging the box on the left side, so the box is again symmetrical around the origin.

evanshultz:
Making symbol outline symmetric and centered are weak rules and not enforced if aesthetically there's no need or it makes things worse, and here is an example. I prefer smaller boxes. But, I'll merge either way.

I'm not all that knowledgeable about these symbols but I don't see anything to comment on. I'm trusting you guys are experts and did the detailed checks.

Is there an order between this and #2059? Does it matter which is merged first?

matthijskooijman:
> Making symbol outline symmetric and centered are weak rules and not enforced if aesthetically there's no need or it makes things worse, and here is an example. I prefer smaller boxes. But, I'll merge either way.

I guess it's a matter of taste. I was going to suggest that having a symmetrical box might lead to slightly less surprise when mirroring or rotating, but rotation is still slightly wonky because of the even number of pins on the right side (which makes the box off-center vertically). I have not strong preference about this in any case, as long as we're consistent :-)

chschlue:
@evanshultz 
They're adjacent in the files, so git might complain either way. But a conflict in #2059 is easier to resolve, so this one first, please.
BTW, any chance you could take a look at https://github.com/KiCad/kicad-templates/pull/25?

evanshultz:
Alright. Merging this one then.

PR #2173, opened 2019-09-16
Add SG5032/SG7050 Oscillators

joshtyler:
Add SG5032/SG7050 Oscillators.
[Datasheet](https://support.epson.biz/td/api/doc_check.php?dl=brief_SG5032CAN&lang=en)

image

image

5032 parts are the same but with a different footprint assigned.

CBN is an alias of CAN since the only difference is the frequency range. CCN is a separate symbol because one pin is different.

The symbols violate KLC S3.6, but this is to make the symbols interchangeable with other pin-compatible oscillators.

---


cpresser:

We can save a little space in the lib by using aliases:
* `SG5032CCN` can be an alias of `SG-8002LB`. (5x3.2mm, OE input)
* same for `SG7050CCN` and `SG-8002CA` (7x5mm, OE input)

Pin4 does not have the correct name. It is `V+` but should be `Vcc`. The existing symbol is not exaclty in line with the data-sheet. We can fix that in a separate PR and don't need to handle that now.

That leaves the `CAN` and `CBN` parts. They seem to be an exact copy of the `SG-8002*` parts with  pin names changed. That is perfectly fine.

There is just one thing I would like to be improved. It is not mandatory since KLC has no explicit rule for it (yet). But it will help the user, so I think it should be changed.
- [ ] Add the frequency-range to the description so the parts do actually differ.

joshtyler:
Thanks for the review!
Requested changes have been made

cpresser:
Thanks for splitting this up in separate commits. That made the review really simple :heart:

Issue #2163, opened 2019-09-14
PZTA42 has 3 pins in symbol, 4 pins in footprint


gkeeth:
PZTA42 is a SOT-223 NPN BJT ([datasheet](https://www.onsemi.com/pub/Collateral/PZTA42T1-D.PDF))

The pin layout of the physical part is BCEC, i.e. 1-Base, 2,4-Collector, 3-Emitter. See screenshot from datasheet:

![pzta42_pinout](https://user-images.githubusercontent.com/4452267/64914389-51084d00-d71e-11e9-804e-a210842e5edc.png)

 

The current Kicad symbol is basically an NPN-BCE, i.e. 1-Base, 2-Collector, 3-Emitter. 
The symbol assigns SOT-223 as the default footprint.

![pzta42_symbol](https://user-images.githubusercontent.com/4452267/64914395-8876f980-d71e-11e9-9711-69404e7935b7.png)![pzta42_footprint](https://user-images.githubusercontent.com/4452267/64914402-9462bb80-d71e-11e9-915e-8b74bb201ff1.png)

The problem is SOT-223 is a 4-pin footprint, but the symbol has only 3 pins. Pin 4 in the footprint should correspond to the collector.

I can think of two solutions:
1. change the PZTA42 symbol to be essentially Q_NPN_BCEC, which would match the existing SOT-223 footprint
2. Keep the PZTA42 symbol as is, but change the default footprint to SOT-223-3_TabPin2. This will effectively connect pins 2 and 4 in the footprint rather than the schematic. 

I think 1 is the better solution because then the kicad symbol will match the datasheet. It is a symbol pin change, which will break any schematics currently using this symbol, but is this a big problem if it's already broken?

2 has the advantage of not breaking existing schematics (I think) but then the symbol doesn't match the datasheet, and it will use a non-standard footprint for no real reason. 

I'm happy to submit a pull request if I can get some guidance on which solution to go for.

gkeeth:
I dug around a little more, and this affects all SOT-223 transistors in the library. 
These parts also have 3 pins in the symbol but specify the standard (4 pin) SOT-223 footprint:
BCP51
PZT2222A
PZT3904 (alias BCP56)
PZT3906 (alias BCP53, PBSS301PZ)
PZTA42

evanshultz:
Option 2 matches the convention used in the library, and won't break existing schematics, so let's go with that.

[...]

PR #2175, opened 2019-09-17
Update SOT-223 BJTs to use SOT-223_TabPin2 footprint

gkeeth:
This pull request updates all SOT-223 BJTs to use the SOT-223_TabPin2 footprint by default instead of the vanilla SOT-223 footprint. This fixes issue #2163 

All of the SOT-223 BJTs currently in the library have their collector on physical pins 2 & 4, but they have a 3-pin symbol in the library. Therefore when the default SOT-223 footprint is used, the tab (pin 4) cannot be used.

By switching the default footprint to SOT-223_TabPin2, the tab also becomes pin 2. This matches the pins in the footprint to the pins in the symbol. 

Note that the [datasheets](https://assets.nexperia.com/documents/data-sheet/PZT2222A.pdf) for these parts invariably numbers the pins 1-4, while the kicad symbols only have 3 pins. However, the functionality matches & all pins are usable as intended. (see discussion in #2163)

---


evanshultz:
This should work for now. Thank you!

PR #2176, opened 2019-09-17
Added LD1086

JeppeSRC:
Added LD1086 regulators

[Datasheet](https://www.mouser.se/datasheet/2/389/ld1086-974047.pdf)

img

---


cpresser:
Hi,

can you do a matrix of which parts you added?
I don't see any aliases for other voltages.

image

Those are the parts added in this PR:
```
New 'Regulator_Linear.lib:LD1086D2T12'
New 'Regulator_Linear.lib:LD1086DT25'
New 'Regulator_Linear.lib:LD1086PU'
New 'Regulator_Linear.lib:LD1086x'
New 'Regulator_Linear.lib:LD1086x18'
New 'Regulator_Linear.lib:LD1086x33'
New 'Regulator_Linear.lib:LD1086x50
```

Is it possible to split this PR up into single PRs, one for each package? That makes it easier to merge and review.
I did check other parts in `Regulator_Linear`. The LD1086 seems to be different then the existing ones since it has the output on pin2 - others have GND on pin2. So making a new symbol (instead of aliases to existing ones) is the way to go.

poeschlr:
Using too many pull requests for a symbol change might not be the best option as it will require merges. So i only advice for it if the contributor is experienced with git. (Which we really can not assume here. Most contributors have their first contact with version control to contribute something.)

JeppeSRC:
I added the following:
1.8v (LD10866x18)
2.5v (LD1086DT25)
3.3v (LD1086x33)
5v (LD1086x50)
12v (LD1086D2T12)
ADJ (LD1086x)
ADJ DFN8 (LD1086PU)

It wasn't really possible to just make one and alias the rest since they all have different package combinations. The 'x' is for the different packages ie LD1086x33 could be LD1086DT33.
And for those that that only come in one package like the 2.5v on I created one with the full name containing the package identifier.

And it's not the first time I'm using git :)

cpresser:
> It wasn't really possible to just make one and alias the rest since they all have different package combinations. The 'x' is for the different packages ie LD1086x33 could be LD1086DT33.

That is the correct way. Because an alias always has the same symbol as the parent part, new symbols are required for different packages. That's a limitation of KiCad which might be resolved with v6.
One issue here is the wildcard 'x' in the name. In your example two letters are replaced with one 'x'. But there should be two xx in the name.

My suggestion is to make one symbol per package. Then add an alias for each voltage option. If we name Pin2 "ADJ/GND" (as in Figure2) we also won't need a separate symbol for the adjustable voltage version.

The only version that comes in all packages is the Adjustable version. So basic symbols should be ones marked (light grey) in this screenshot.
image

When adding aliases, just add every part in the respective column to the symbol.


**Example**
Symbol: LD1086DTTR
FP + Filter: DPAK
Aliases:
* LD1086DT18TR
* LD1086DT25TR
* LD1086DT33TR
* LD1086DT50TR
* ~~LD1086DTTRY~~ (Automotive part, don't add this)
* LD1086BDTTR (same as LD1086DTTR? do you know the difference?)

JeppeSRC:
Alright I'll make the changes. About `LD1086BDTTR (same as LD1086DTTR? do you know the difference?)`. I've no idea I've been trying to find out but I came out empty.

JeppeSRC:
There was no fitting DFN8 footprint available.

[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/1871)
[Generator PR](https://github.com/pointhi/kicad-footprint-generator/pull/431)

cpresser:
Alright, I made an of the names/aliases:

Symbol | Alias | Footprint + Filter
-------- | -------- | --------
LD1086V-DG   :heavy_check_mark:| LD1086V18-DG, LD1086V33-DG  :heavy_check_mark: | `TO-220-3_Vertical` `TO?220*` :x: 
LD1086DTTR  :heavy_check_mark:| LD1086DT18TR, LD1086DT25TR, LD1086DT33TR, LD1086DT50TR  :heavy_check_mark: | `TO-252-3` :heavy_check_mark: 
LD1086PUR :heavy_check_mark:| | Pending FP :white_check_mark: 
LD1086D2MTR :heavy_check_mark:  | LD1086D2M33TR :heavy_check_mark:  | `TO-263-3_TabPin2` :heavy_check_mark: 
LD108D2TTR :x: | LD1086D2T18TR, LD1086D2T33TR, LD1086D2T50TR, LD1086D2T12TR :heavy_check_mark:  | `TO-263-2`
`TO?263?2*` :heavy_check_mark:

- [ ] The LD1086V-DG is not an atomic part. it matches both vertical and horizontal TO-220 packages. So the fp-field should be empty.
- [ ] `LD108D2TTR` is missing a '6' in the name.

cpresser:
While looking at the diff i spotted one more minor issue.
- [ ] The Datasheet points to mouser. Please use the direct link to ST: https://www.st.com/resource/en/datasheet/ld1086.pdf

I also made a mistake above. I did interpret [KLC S5.1](http://kicad-pcb.org/libraries/klc/S5.1/) wrong. Sorry about that.
This is an fully specified symbol. The TO*220*3* just allows variants of the footprint, but no other footprints.
- [ ] Re-Add the default footprint for  LD1086V-DG

I also did see that some of the filters contain question marks. Please replace them with '*'
- [ ] Filters should contain * instead of ?

chschlue:
What's wrong with `TO?`?
`TO*23*` for example matches TO-223 when TO-23 is wanted, `TO?23*` doesn't.

Pin counts should be removed, though.

cpresser:
> What's wrong with `TO?`?
> `TO*23*` for example matches TO-223 when TO-23 is wanted, `TO?23*` doesn't.
> 
> Pin counts should be removed, though.

I somehow was under the assumption that we want `*`everywhere. I should read KLC more carefully.

Anyway, I could use help with the filter for `LD1086D2TTR`. 
image

`TO?263*` also matches `TO-263-3_TabPin2` which we don't want here. The part will fit that FP, but it has an extra pad that we don't want and which is taking away precious board space.
The only allowed footprint for that symbol should be `TO-263-2`.

I also don't understand why `TO-263-2` has the `-2` in its name. It has 3 pins. Is the name wrong?

evanshultz:
@cpresser 
Because there is a default footprint the FP is really only used if the user wants to deviate. In that case, we'll leave it in their hands even if the trailing asterisk could allow selection of goofy footprints.

I dislike the footprint naming for these packages as well. No intuitive at all for me. But it hasn't gotten done yet. Probably would take a lengthy conversation.

chschlue:
> `TO?263*` also matches `TO-263-3_TabPin2` which we don't want here. The part will fit that FP, but it has an extra pad that we don't want and which is taking away precious board space.
> The only allowed footprint for that symbol should be `TO-263-2`.

Useless, but not harmful in this case. With enough solder you can reach the stub pin 2 from that extra pad :)
The trailing asterisk is useless right now but, say, a TO-263-2_ThermalVias FP isn't unreasonable.

> I also don't understand why `TO-263-2` has the `-2` in its name. It has 3 pins. Is the name wrong?
> It seems to violate http://kicad-pcb.org/libraries/klc/F3.5/
> 
> Edit: the above also applies to `TO-252-2`

I don't really like the naming either, but I think I get the rationale.
TO-252-2 is also known as a two lead DPAK, presumably carried over from its THT ancestors like TO-220 whose tabs are in many cases not isolated but still not considered to be additional pins.

cpresser:
FP filter for the DFN8 part should not include pin-count and EP-Size
`DFN*1EP*4x4mm*P0.8mm*`

JeppeSRC:
@cpresser FP filter is fixed, was there anything else?

cpresser:
Nope, we are good. Thanks for you patience.

PR #2179, opened 2019-09-17
Add a few BJTs

gkeeth:
This pull requests adds several related BJTs: MMBTA42, MMBTA92, PZTA92. It also fixes PZTA42, which already existed in the lib.

These are the SOT-23 and SOT-223 versions of MPSA42 (NPN) and MPSA92 (PNP), which are already in the library. I copied the existing MPSA42/92 symbols and updated the datasheet links, footprint, footprint filter, and the package type in the description.

I also fixed a few minor issues with PZTA42:
- updated the datasheet link, which was a dead link to a Fairchild part
- changed the collector current given in the description from 0.2A to 0.5A, which is what the new On Semi datasheet gives (and matches all of the other package variants of this part).
- slightly reworded the description to match the other package variants of this part.

---


gkeeth:
I should have included datasheet links:
[MMBTA42](https://www.onsemi.com/pub/Collateral/MMBTA42LT1-D.PDF)
[MMBTA92](https://www.onsemi.cn/PowerSolutions/document/MMBTA92LT1-D.PDF)
[PZTA42](https://www.onsemi.com/pub/Collateral/PZTA42T1-D.PDF)
[PZTA92](https://www.onsemi.com/pub/Collateral/PZTA92T1-D.PDF)

gkeeth:
I don't know what I was doing when I made these symbols, but I got the pin numbering totally wrong. 

Here's a commit that fixes the pin numbering. Compare the symbols with the datasheets:

Screenshot_20190919_185614
Screenshot_20190919_185140

Screenshot_20190919_185636
Screenshot_20190919_185220

Screenshot_20190919_185300
Screenshot_20190919_185324

Screenshot_20190919_185401
Screenshot_20190919_185538

Sorry for the excessive number of commits on this. I can squash it down if that would be better.

evanshultz:
Our standard procedure is to squash merge so no problem about so many commits here.

PZTA92:
Please uncheck Place Names Inside to match PZTA42
image

PZTA42/92 have varying stated currents depending on the vendor. [NXP says 100mAdc](https://assets.nexperia.com/documents/data-sheet/PZTA42.pdf) and [MCC says 200mAdc](https://www.mccsemi.com/pdf/Products/PZTA42(SOT-223)-A.pdf). Sadly NXP doesn't have a graph, but both ON Semi and MCC show the current gain has basically collapsed by 50mA Ic at 10V collector-emitter. So while the part may be able to pass 100mA or 500mA, it's not all that useful as an amplifier. My point is that the current rating is more marketing and less engineering. PZTA92 shows a similar variation of manufacturer DC collector current ratings. So I'm not opposed to merging this with showing 0.5Adc since it matches the linked datasheet, it's notable that this parameter varies by manufacturer on the first page of the datasheet though the parts themselves may behave much more similar than one would expect.

gkeeth:
Presumably I should also uncheck "Place Pin Names Inside" for MMBTA42 and MMBTA92, right? In that case, should I also uncheck it for MPSA42/MPSA92? I see now that most BJTs have it unchecked, but not the MPSA42/MPSA92 that I based these on.

I understand what you mean about marketing vs. engineering. I think many transistors have similarly optimistic headline specs in the datasheet (MOSFET max drain current while Tc=25C...) but that's one hazard of reducing a datasheet's worth of characteristics to a single line. At least this way the symbols are consistent with the linked datasheets and with each other.

gkeeth:
The previous commit updates PZTA92, MMBTA42, and MMBTA92 as above, but doesn't touch MPSA42/MPSA92. If I should make those changes as well, and roll them into this PR, let me know.

evanshultz:
Let's not touch other symbols at this time so we don't affect existing schematics. Thank you.

PR #2187, opened 2019-09-21
Add BCV[24]9 NPN Darlington

chschlue:
DS https://assets.nexperia.com/documents/data-sheet/BCV29_49.pdf

BCV29

---


cpresser:
- [ ] The pin-numbering of the NXP-Package seems to differ from the one already in the lib.
image
You seem to have fixed this by inverting the pin numbers. I think that is a bad idea because pin numbers will be shown on the schematic. 
Perhaps this actually needs a different package.

- [x] The Drawing is exactly the same as BC517 (exact same coordinates of graphical elements, just different pin-out). :+1: 
- [x] Description and Keywords are good :+1:

cpresser:
Something else I noticed while looking at this PR is that `BDW93` has a slightly different drawing. It has one more polyline because the pin is 200 instead of 225mil. Is that something that we want to change?

chschlue:
I think there's a BJT rework already in progress like it's been done with FETs the other day.
Here I just used Device:Q_NPN_Darlington

chschlue:
AFAIK, NXP and STMicro are the two major manufacturers (offenders?) labeling pins right to left while JEDEC and the rest of the world (including kicad-footprints) does it the other way round.

There is no SOT-89-3_British or anything in the lib, so I assume what I've done here is fine.

cpresser:
I looked for other parts in the lib and found `BCX51` which has the same issue - and the same solution you choose. Which indicates that is is okay to merge it as it is.

But I still don't like it. The user will see the wrong pin numbers in the schematic.
Perhaps somebody else has useful input on this issue.

chschlue:
No problem, I'm not in a hurry.

evanshultz:
IMO, there's no sense in duplicating footprints for pin number variations if there's no user benefit. So renaming in violation of the datasheet, when the symbol itself shows the electrical function, is OK with me.

In some cases we have also used pin names, like G, D, and S for a MOSFET if that allowed just three symbol pins to connect to many footprint pins. Some benefit there. Or with audio connectors where a few simple symbols can then work with many footprints regardless of what the datasheet pin numbers are.

While I don't see anything in KLC, we have used IEC symbols (see IEEE Std 315) in the past. While it's not such a big deal, it might be nice to settle on the Darlington transistor symbol style in this library. Here is what is shown in that doc:
image

poeschlr:
The named symbol pin number stuff is not feasible for things like mosfets that come in standardized packages also used for other devices. It only works for highly specialized things like audio connectors where basically every device that can use this symbol will need a highly specialized footprint anyways.

For the interested: https://forum.kicad.info/t/idea-for-pin-mapping-hard-wired-items-ie-mosfets/18586/ (I am on the fence here as i fear any such suggestion might make the fully specified workflow harder than necessary. And i personally favor fully specified over generic for a vast majority of parts including transistors)

The sad truth is that some packages like the sot-23 and as it seems also sot-89 do not use a standardized numbering scheme on the footprint side. This is something where we will not be able to find a nice solution. (Not even more powerful kicad features can save us from this mess.)

PR #2190, opened 2019-09-22
Adding LM21212-2

ki5libs:
Adding LM21212-2
[Datasheet](http://www.ti.com/lit/ds/symlink/lm21212-2.pdf)

LM21212-2

---


myfreescalewebpage:

- [ ] The footprint has an exposed pad, but not centered. You will probably need to submit it because I'm not sure we have it in the library. Page 36 of the datasheet.
- [ ] EP should be stacked with PGND according to the datasheet

Always a pleasure to review your pull request, they are good quality !

Cheers,
Joel

ki5libs:
Hi @myfreescalewebpage 

thanks for review and good words. But I really miss the situation with footprint. And now I made this [PR](https://github.com/KiCad/kicad-footprints/pull/2190) based on existing similar footprint from Texas. 

Actually I'm not sure with a name. I used `Texas_PWP0020AA` supposed that it's OK, because of existence `Texas_PWP0020A` in the library. But maybe something like `Texas_PWP0020AA_4.4x6.5mm_P0.65mm_EP3.32x4.55mm` would be better?

myfreescalewebpage:
I'm not used to review the footprint, other are clearly better to this exercice, I let them give you the right advice on your Footprint PR.

PR #2194, opened 2019-09-22
Add MAX6035 vrefs, SOT-23/SOIC-8

chschlue:
DS https://datasheets.maximintegrated.com/en/ds/MAX6035.pdf

MAX6035xxUR25
MAX6035xSA25

---


cpresser:
- [x] Pin names and numbers are good
- [ ] Keyword and description are rather sparse. I think you should mention 'high-voltage,  precision  micropower voltage reference'
- [ ] Naming of devices
I found this table for the device/ordering information. Is there more info somewhere?
image
The SO-8 device does not seem to have variants. So the `x` could be removed from the name and just go with `MAX6035ESA25`.
Similar the SOT-23 devices all have the `A` common in their name. Its just the first letter after `MAX6035` that differs.

- [ ] Why did you choose to add only 3 NC pins instead of all 5?
Since there is no Pin8 the pin-count matching of footprints might fail. Perhaps it is enough to change Pin1-NC to Pin8-NC for that to work again.

chschlue:
> * [x]  Pin names and numbers are good

Yay!

> * [ ]  Keyword and description are rather sparse. I think you should mention 'high-voltage,  precision  micropower voltage reference'

What's the point of repeating words that are already in the description? The keywords only aid searching if they contain terms that aren't already there.
And in this case, I simply couldn't think of more than `vref` but I'm open to suggestions.

> * [ ]  Naming of devices
>   I found this table for the device/ordering information. Is there more info somewhere?
>   image
>   The SO-8 device does not seem to have variants. So the `x` could be removed from the name and just go with `MAX6035ESA25`.
>   Similar the SOT-23 devices all have the `A` common in their name. Its just the first letter after `MAX6035` that differs.

https://www.maximintegrated.com/en/design/packaging/package-information/maxim-naming-conventions.html

E and A denote the temp range. No point replacing it with x later if Maxim chooses to add a -55°C/+125°C.

> * [ ]  Why did you choose to add only 3 NC pins instead of all 5?
>   Since there is no Pin8 the pin-count matching of footprints might fail. Perhaps it is enough to change Pin1-NC to Pin8-NC for that to work again.

Because I'm dumb. Or was tired. Or both. Will fix.

cpresser:
> What's the point of repeating words that are already in the description? The keywords only aid searching if they contain terms that aren't already there.
> And in this case, I simply couldn't think of more than `vref` but I'm open to suggestions.

I have no good reasoning but 'other parts in the library do have those repetitions'.
Right now KLC has no explicit rules for what to put into keywords. There was an issue somewhere discussing that, but right now I can't find it.

> https://www.maximintegrated.com/en/design/packaging/package-information/maxim-naming-conventions.html
> 
> E and A denote the temp range. No point replacing it with x later if Maxim chooses to add a -55°C/+125°C.

Good to know. In this case the `x` as you choose it makes perfectly sense.

chschlue:
> I have no good reasoning but 'other parts in the library do have those repetitions'.

I know many keyword lists are basically the description minus punctuation.
I'm starting to think there is no explanation other than "that's how it's always been done".

Maybe @evanshultz can shed light?

evanshultz:
http://kicad-pcb.org/libraries/klc/S2.2/ says temp should be wildcarded and it doesn't matter if there is only one temp variant. So like it's done is great.

I don't often see any usefulness in keywords if the description is good. And most of it is legacy. So without any rules or guidelines I usually try to make the description looks nice since it's more prominent in the UI of KiCad and just take whatever the contributor puts in the keywords.

cpresser:
Thanks everybody!

poeschlr:
Regarding keywords: we could test if the description is used for the search feature. If it is then we can possibly even ignore them in the future. Otherwise it is a nice way to define which search terms will yield the specific asset.

chschlue:
The search feature definitely uses the description. It yields results for search terms that appear nowhere but in the description.

Keywords are nice to have in cases like this (searching for `vref` yields all symbols containing that keyword, including this one, plus one that has VREF in its description, for example) but simply repeating the description is totally useless.

poeschlr:
So the keyword field does not have a purpose in v5. Interesting.
I would then say that we definetly should not really invest time into fixing any keyword fields. And i see no reason to invest time into specifying their use in this case either. Maybe we could say that it simply contains additional search terms that would make the description unwieldy but are reasonable to expect a user to use to find the thing in question.

PR #2195, opened 2019-09-23
Add LTC4317 Dual I2C/SMBus Address-Translator

cp-aquila:
Add LTC4317 Dual I2C/SMBus Address-Translator

Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/4317fa.pdf
grafik

I choose to put ENABLE at the top because those pins are most likely tied to vcc. Thus the READY pins are also on top on the opposite side.
Exposed Pad: "EXPOSED PAD (PIN 17) IS GND, PCB CONNECTION OPTIONAL", I choose to add it to the GND stack.

---


chschlue:
How about spreading XOR* so R_Smalls fit inbetween?

cp-aquila:
> 
> 
> How about spreading XOR* so R_Smalls fit inbetween?

I thought about that, but I think it makes the symbol quite big. 
Not sure if we want that.

I could also mix up (interleave) the positions of XOR* a little so they are better suited for fitting a R_Small between them.
grafik


Right now my wiring looks quite messy, but it works:
grafik

What do you think?

cp-aquila:
How about this?
grafik

chschlue:
Better, I think.

diegoherranz:
I'm happy to assign this one to myself but, @chschlue, you started providing some comments. Do you want to take this one?

Either way, a few comments:

- Exposed Pad: datasheet says "_Exposed pad may be left open or connected to device GND_"
I would connect it to GND too if I used this device but I think we need to give the user the option. I would include EP as a separate pin.

- READY1/READY2: these are open-drain. Can you set them as open collector?

Thanks!

chschlue:
Feel free to take it.

diegoherranz:
Thanks for the changes.

After reviewing it, I think there is only one thing pending:

- [ ] The footprint filter shouldn't include the size of the exposed pad. See [KLC S5.2](http://kicad-pcb.org/libraries/klc/S5.2/)

Checked done, which all look OK:

- Description
- Datasheet
- Pinout
- Pin types: even if the SCL pins in I2C can be bidirectional due to clock stretching, we tend to set them as just their main direction, obliviating clock stretching, so it looks OK.
- Pin arrangement
- Footprint: nearly identical to the recommended one in the datasheet. Just the EP is 0.01mm longer, which is perfectly within the tolerance suggested by the datasheet.

Thanks!

diegoherranz:
That was quick! Merging.

Thanks!

cp-aquila:
Wow, that was fast. Thank you very much!

>     * Footprint: nearly identical to the recommended one in the datasheet. Just the EP is 0.01mm longer, which is perfectly within the tolerance suggested by the datasheet.

I also noticed that. 
The FP also does not have rounding of the EP.
There are some more parts in the library, all from Linear, that use the same FP. 

grafik

All of them use the same `LTC DWG# 05-08-1706`, so replacing the FP should be okay.
Perhaps I will redo it using the footprint-generator, but thats a different story.

PR #2197, opened 2019-09-24
Fixed IGBT Arrow direction (#2196)

### LINKED ISSUE #2196: Many IGBT "Emitter" Arrows seem inverted
I am uncertain. but I think several of the IGBT symbols have the Emitter Arrow pointing toward the gate.

jgladen:
Fixes #2196 

* Inverted "Emitter" arrows for affected *IGBT* devices

Fixed simple issue


evanshultz:
IEEE Std 315 doesn't show an IGBT symbol that I can find, but at least this is more correct.

Thanks for taking the time to catch this error, report it, and then submit a fix! And welcome!

PR #2201, opened 2019-09-24
Add BZX384 zener diode

evanshultz:
https://assets.nexperia.com/documents/data-sheet/BZX384_SERIES.pdf

image

The trailing `xx` is a little weird since there can be two or three characters for the voltage, but I kept it since that's how zener families are done in the library. And the zener tolerance is to be considered wildcarded as well. Let me know if you want any changes here.

---


cpresser:
- All other Zeners in the lib seem to have this exact drawing, so this one is good. While looking at this I also checked `Device:D_Zener` which has a slightly different drawing (one less poly-line, instead: longer pins meet in the center)
- Why did you not include the dash in the name? How about `BZX384-xxxx`. Four wildcards, since that covers the longest possible name. I would go for at least 3 wildcards, there is no device that has less than 3 characters after the dash.

evanshultz:
- I have cleaned up and commonalized FET symbols, but not other types yet. Thanks for the note. There are two other PRs by my which touch the diode library (one assigned to you) which I hoped to get merged before I cleaned up that library.
- Updated symbol name.

PR #2202, opened 2019-09-25
 add ADA4610 operational amplifiers

hvraven:
This adds all variants of the [ADA4610 operational amplifier series](https://www.analog.com/media/en/technical-documentation/data-sheets/ADA4610-1_4610-2_4610-4.pdf). Where possible aliases to existing symbols were used. Otherwise copies from existing symbols were created (using LM2904/LM2902 as reference for dual/quad symbols) and adjusted. The quad CSP version has some NC pins which were buried in the first unit. Pads were stacked with V- as given by datasheet.

Only one extra footprint is required (for the dual LFCSP variant), which can be found here: https://github.com/KiCad/kicad-footprints/pull/1876

Screenshot_20190925_144149

---


lorem-ipsum:
Thanks for the review. Integrated the changes and pulled master to fix a merge conflict.

What is the difference between the non compliant symbols and the compliant symbols? On first look they seem to be fairly similar.

chschlue:
In the op amp lib, they mostly lack default footprints or package suffixes (or both).

chschlue:
I just noticed that NCS2325DM uses OnSemi_Micro8 instead of MSOP-8_3x3mm_P0.65mm, seemingly for no good reason.
Changing this would save another symbol.

Is an en passant default FP change acceptable, @poeschlr?

chschlue:
Any feedback on this, @poeschlr?

poeschlr:
I am not sure which question i should answer. If the idea is to use an on semi symbol as the basis of an alias and that symbol right now has a on semi specific footprint then i see two options:
- make a new symbol pointing to a generic footprint instead of reusing the onsemi symbol
- point the on semi symbol to a generic footprint after checking that the new footprint is indeed for the same package (check the dimensions of it including the tolerances)

chschlue:
I'll try to clarify.

NCS2335DM comes in an MSOP-8 package which On also refers to as Micro8.
On's recommended footprint (https://www.onsemi.com/pub/Collateral/846A-02.PDF) doesn't fit an IPC MSOP-8 footprint, but AFAIK standard footprints are preferred for standard package.

So my question is: Would it be acceptable to change NCS2325DM's default footprint right here and right now, thereby improving it in the lib convention sense while also opening it up for other MSOP-8 op-amp (ADA4610-2xRM in this case) aliases **or** should the existing symbol rather not be touched at this point?

I would add MSOP-8 to the generator scripts as the 0EP variant is apparently not scripted right now.

poeschlr:
Compare the package drawings (not the suggested footprint)
If these two are the same then yes the generic footprint should be used. if not then the packages are truly different and need different footprints.

chschlue:
It's a metric pitch, 8-pin, small outline, JEDEC MO-187 variation AA package aka MSOP-8_3x3mm_P0.65mm, to be precise.

I checked that before I even started asking if changing default FPs mid-release was acceptable, of course.

chschlue:
@lorem-ipsum please add ADA4610-2xRM to NCS2325DM

chschlue:
@lorem-ipsum ping

lorem-ipsum:
changed ADA4610-2xRM to be an alias and resolved the merge conflict

chschlue:
Thanks

lorem-ipsum:
fixed the merge conflict again

myfreescalewebpage:
Footprint merged, @hvraven can you solve the branch conflict here please ?

hvraven:
rebased against the current master

myfreescalewebpage:
You have travis error about the footprint name to be fixed.
@chschlue I let you doing final review and merge except if you ask me to do it.

chschlue (approved):
Thanks

hvraven:
Oh, sorry for not checking this properly. Should be good now

PR #2212, opened 2019-09-29
Adding CBTL02043A, CBTL02043B

ki5libs:
Adding CBTL02043A, CBTL02043B
[Datasheet](https://www.nxp.com/docs/en/data-sheet/CBTL02043A_CBTL02043B.pdf)

CBTL02043x

---


myfreescalewebpage:

- [ ] Pin offset should be 20mil (see travis error)
- [ ] Pin length should be 100mil
- [ ] VDD pins should not be stacked

Cheers,
Joel

ki5libs:
Hi @myfreescalewebpage 

I've changed pin offset and length. Could you tell me the reason why VDD pins shouldn't be stacked? I couldn't find any information about separate decoupling capacitors for these pins. So according to KLC I suppose to stack it.

myfreescalewebpage:
In case of no information we give the user the opportunity to do it anyway, so by default the VDD pins are not stacked, except if the datasheet explicitly indicate capacitor are not required on all pins.

ki5libs:
@myfreescalewebpage I'm sorry, but [S4.3 KLC](https://kicad-pcb.org/libraries/klc/S4.3/) tells exactly the opposite:
`Power supply pins must be stacked unless the datasheet specifies the need for decoupling capacitors on every pin.`
Fixing it is a matter of minutes, but on my opinion (and looks like KLC has the same view) user could place several capacitors to single pin on the schematic and then locate it on the PCB any way he wants. And if I'm wrong why KLC doesn't explain it?

myfreescalewebpage:
I was pretty sure of the previous position, but thanks to remind my this point (or maybe the KLC changed recently). Merging current proposition.

PR #2221, opened 2019-10-01
add LM74700-Q1 ideal diode controller

florolf:
Add the LM74700-Q1 ideal diode controller.

Symbol passes the [linter](https://github.com/KiCad/kicad-library-utils). Datasheet link: http://www.ti.com/lit/gpn/LM74700-Q1

Screenshot:

2019-10-01-224539_1354x959_scrot

The pin positions are chosen such that default-sized symbols fit easily:

2019-10-01-224808_705x820_scrot

---


evanshultz:
Thanks for this contribution!
1. I only see this in automotive, but let's still remove the `-Q1` from the symbol name.
1. End the description with the package, in this case `, SOT-23-6`.
1. Let's make pin 1 Passive type.
1. I'm unsure about pin 6. I'm thinking it should be Power Input to help catch a mistake where the net on that pin is not a power source. What do you think?

florolf:
Thanks for looking into the PR.

> 1. I only see this in automotive, but let's still remove the `-Q1` from the symbol name.

There's no non-automotive version at the moment, so I thought we should consider the suffix as a part of the proper name. But sure, let's drop it.

> 2. End the description with the package, in this case `, SOT-23-6`.

Done.

> 3. Let's make pin 1 Passive type.

Okay, I was unsure about this one.

> 4. I'm unsure about pin 6. I'm thinking it should be Power Input to help catch a mistake where the net on that pin is not a power source. What do you think?

Your right. From the point of view of the chip, this actually *is* a power input: It powers the charge pump as well as the logic.

evanshultz:
Thanks. Please don't force push as it makes it harder for the librarians to review your changes.

PR #2182, opened 2019-09-20
Add Sc4503

DecadoPL:
@chschlue 
This is continuation of https://github.com/KiCad/kicad-symbols/pull/2010 
Last PR was from "unknown repository" so i had to create new PR.
I renamed the EN pin to ~SHDN~. Didn't use full name ~SHDN~/SS because that would require biger symbol and i dont see any reason to do that since the name ~SHDN~ gives a clear information about this pin.
image
image

https://www.semtech.com/uploads/documents/sc4503.pdf

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [X] Provide a URL to a datasheet for the symbol(s) you are contributing
- [X] An example screenshot image is very helpful
- [X] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [X] Check the output of the Travis automated check scripts - fix any errors as required
- [X] Give a reason behind any intentional library convention rule violation.

PR #1906, opened 2019-06-06
Added NCP1529

Trubochkin:
- Added NCP1529 (Regulator_Switching)

- Datasheet: https://www.onsemi.com/pub/Collateral/NCP1529-D.PDF

- Fotprint: https://github.com/KiCad/kicad-footprints/blob/master/Package_TO_SOT_SMD.pretty/SOT-23-5.kicad_mod

- Screenshot: 

SharedScreenshot

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

PR #2222, opened 2019-10-01
add symbols for the AOZ128xCI family of switchers

florolf:
Add symbols for the AOZ128xCI family of switchers. Passes the linter with the exception of S4.2, which is caused by the flow-through style of powering DC-DC regulators that is also common in the KiCad standard library. Exemplary datasheet: http://aosmd.com/res/data_sheets/AOZ1280CI.pdf (all of them are linked in the symbols).

Compatibility with the standard library SOT-23-5 footprint has been proven both in manual assembly and with various manufacturing services.

Screenshot:

2019-10-01-231935_1377x642_scrot

---


cpresser:
Thanks, this looks good!
- [x] description, keywords and data-sheet (for all aliases)
- [x] pin-types and names
- [x] Footprint and FP-Filter (side-note: the text in your post is `SOT-23-5`. I assume that's a typo since datasheet and symbol both have `sot-23-6`)
- [x] pin positions

We actually have different styles regarding the `VIN` pin even in recent commits to the library
* #2182 (Vin on the left)
* #1988 (Vin on top)
* Your PR is pretty similar to `NCP1529`(#1906), which was just recently merged, so its good to go.

PR #2226, opened 2019-10-03
Addition of symbol for APA-106-F5 5mm Addressable RGB LED

jamesrlucas:
Addition of APA-106-F5 Addressable RGB LED. Datasheet can be found here: https://cdn.sparkfun.com/datasheets/Components/LED/COM-12877.pdf

**NOTE:** that the LED Pins are backwards compared to the datasheet - The datasheet counts pin 1 as being furthest from the flat, however the 4 Pin LED footprint in the library has pin 1 next to the flat. I have confirmed this by testing on the bench. I can grab a photo if required.

image

---


cpresser:
Hi and thanks for your contribution.
* Regarding the pin-numbering on the symbol: Adjusting it to the existing footprint is okay if the footprint has a reasonable numbering scheme. This seems to be the case here.
* Usually we also require that the names of the pins reflect the data-sheet names exactly. The data-sheet has actually both variants `DOUT` and `DO`, so sticking with the shorter `DO` is fine here. It also has the benefit being short allowing the symbol to have the same size as the other `APA*` symbols.
* Pin3 however is `VDD` in the datasheet.
* The Footprint-filter does contain the number 4, referring to the pin-count which should not be there. `LED*D5.0mm*RGB*`is better.

jamesrlucas:
Hello,

I've fixed the VDD Pin name & the footprint filter as requested.

cpresser:
It seems like you deleted APA102 from the lib and commited that. Can you please fix that?
Other than that, this is ready to be merged.

jamesrlucas:
I can see how I did that - I didn't save as a copy *facepalm*. All fixed now!

PR #2231, opened 2019-10-03
Add SRC4392 audio sample rate converter

qberthet:
First pull request, hope I don't mess up

Adds SRC4392 sample rate converter in TQFP version, QFN will follow soon if the symbol is ok.

Here is the [Datasheet](http://www.ti.com/lit/ds/symlink/src4392.pdf)

Screenshot from 2019-10-03 22-36-17

Known KLC violations: 
- Some inputs/outputs are placed on the wrong side, either to match the port they are related to or to avoid a too big symbol. Will correct if needed
- 3 GND pin are stacked and thus set as passive
---


evanshultz:
1. The symbol name should end in `xPFB` to indicate the TQFP package.
1. Ref des should be on the top left of the symbol.
1. Let's keep the DGND pins separate. They all have unique uses in the pin table and figure 68 shows an application-specific use. (If they were still stacked the visible one in the stack should stay Power Input type but that won't apply when you unstack.)
1. Pin 23 should be Open Collector type.
1. You can put the power pins next to each other and slim the symbol.
1. I'm not sure about the pin placement either, but maybe put pins 19 and 21 at the bottom of the right side of easy access to ground and then make a chunk of I2C/SPI pins and the rest of the control pins together above it.

qberthet:
Thanks for the review.

1. Changed for SRC4392IPFBx as it seems that this is the name for TQFP variant
2. Done
3. Done (Thanks for the stacked pin info)
4. Done
5. Done, but kept 1v8 pin slightly appart
6. I have moved pin 19 and 21 down, but kept the CPM pin at the bottom: as it is used to select between I2C and SPI, it may also need to be grounded. Will update if you don't agree.

Updated screenshot:

Screenshot from 2019-10-15 13-09-43

evanshultz:
I resolved the branch conflict, fixed the symbol name (wildcard the temp grade), and bumped the value to the right side (ref des on left, value on right). Should be good to go. Thanks for all your great work and sorry for the huge delay, @qberthet !

PR #2236, opened 2019-10-05
add STUSB4500 USB PD controller

lorem-ipsum:
This adds the USB power delivery controller [STUSB4500](https://www.st.com/resource/en/datasheet/stusb4500.pdf) in the QFN variant. Pins are sorted to the sides based on towards the USB side (left) and towards the device side (on the right). The I2C interface is located on the left as well, mainly due to space reasons. Also it is used for programming during setup and not necessarily needed when the operating.

Screenshot_20191005_172221

---


cpresser:
Hi,
I did not look at the datasheet yet, but a few things that seem odd:
- ALERT should be on the right side (output)
- VSYS on the top
- Most devices in the Library have RESET as the uppermost pin on the left side
- I would put VREG* outputs to the top of the right side, for easy wiring up to supply symbols

lorem-ipsum:
- I put ALERT to the I2C pins as it is described as I2C interrupt pin.
- I first had it there, then moved it to the side as it is not necessarily used and is grounded then. But I can move it back up.
Moved the others around, now it looks like this
Screenshot_20191005_194348

myfreescalewebpage:

- [ ] The footprint filter should be `QFN*4x4mm*P0.5mm*`
- [ ] Pins 21 and 23 should be on the top of the symbol (Power Output)

I trust in you for the position of the pins, if you have a sample schematic with this symbol it's also very nice!

Cheers,
Joel

lorem-ipsum:
I changed the footprint filter and moved the pins around.

I have a simplistic schematic using the chip, but it is not using the I²C interface. But reading the datasheet again ALERT is an interrupt pin for the I²C interface, so I would like to keep it next to the other I²C pins.

myfreescalewebpage:

PR #2239, opened 2019-10-08
Fixed symbol with boundingbox offset

fabikroeger:
Fixed symbol with boundingbox offset at part A of the symbol

Before fix: 
P4080_before

After fix:
p4080_after

I did not touch anything else on that symbol so i do not fill any checkbox below!

---


evanshultz:
Unfortunately there are many symbols that are old and with problems. This one especially need a lot of love but that won't happen during KiCad 5.x. Thanks again!

PR #2246, opened 2019-10-14
Add Raspberry_Pi_4 connector

patrickelectric:
- Based over Raspiberry_Pi_2_3 symbol
- Pins configuration based in preliminary datasheet
    - https://www.raspberrypi.org/documentation/hardware/raspberrypi/bcm2711/rpi_DATA_2711_1p0_preliminary.pdf

Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

---

image


chschlue:

Please refrain from force-pushing (at least after somebody has started reviewing). Makes it really hard to follow changes.

patrickelectric:
@chemicstry Sorry, just trying to be approved by travis

cpresser:
Writing this post kinda hurts, because this is you first contribution and I am gonna ask if we even need it:

What is the difference between this symbol and the existing Raspberry_2_3 symbol? I assume there is nothing but more alternate pin functions. However, the current rule of the library is not to add more than one alternate function. Some older parts still have them. For example `PIC18F13K50-EP`. New parts should stick to the rule: Only add alternative functions for the most common usage scenarios. In this case the `ID_SCL` and `ID_SDA pin, perhaps others.

From what I understand right now it would be better to just add an alias for the existing `Raspberry_Pi_2_3` symbol.

patrickelectric:
Hi @cpresser! 
I don't have a deeply knowledge of kicad symbol rules so I'm not aware of such things, if there is a list or something can you please share it ?

The difference between this and `Raspberry_Pi_2_3` is that the old raspberries does not provide the number of peripherals that the new ones does, both Pi 2 and 3 are retrocompatible. 
The old one contains: 1 UART, 1 I²C and 2 SPIs
You can use the old `Raspberry_Pi_2_3` symbol with the new Raspberry Pi 4, but the other way around is not true since:
The new raspberry contains: 6 UARTS, 6 I²C, 5 SPIs and more.

Some of the configurations available in the Raspberry Pi 4 can matches the Raspberry Pi 2/3, but it's not the default configuration.

Let me know if this PR is still valid or how can we change the old symbol to be improved in a more clear way for the users that want to create schematics for the Raspberry Pi 4.  Otherwise they will need to open the Datasheet, check the GPIO in the header, figure out the GPIO that matches the expected configuration and etc..

chschlue:
What's your suggestion for a Raspi 4 symbol that's distinct from Raspberry_Pi_2_3 and contains at most one alternate pin function?
A picture says more than a thousand words in this case, so just update the screenshot above.

patrickelectric:
I'm thinking to replace/update the `Raspberry_Pi_2_3` symbol with a `Raspberry_Pi_2_3_4` maybe and use the symbol of this PR, does it sound valid or do you have a different idea ?
image

poeschlr:
Maybe having a symbol for Rasperry_Pi_2 and then assigning everything else as alias is a better long term solution. For now i guess we leave the Rasperry_Pi_2_3 symbol unchanged and assign aliases for Pi_2, Pi_3 and Pi_4. (plus create an issue for the v6 transfer that the Pi_2_3 symbol can be deleted.)

myfreescalewebpage:
@patrickelectric just for your information there are plenty of other Raspberry PR pending made by another contributor. This is just a warning to avoid duplicated work. Maybe you can help too ! They are all referenced by an issue I have done: https://github.com/KiCad/kicad-symbols/issues/1772
Joel

PR #2247, opened 2019-10-16
Add LP5036 LED driver

rleh:
Add LP5036 LED driver

Datasheet: http://www.ti.com/lit/ds/symlink/lp5036.pdf

Footprint (Pull-Request): KiCad/kicad-footprints#1922 and pointhi/kicad-footprint-generator#445

image


cpresser:
Hi,
it seems like there is no footprint for this yet.
Did you also submit a PR for that? If yes, please put a link here.

rleh:
Footprint Pull-Request is here: KiCad/kicad-footprints#1922 and pointhi/kicad-footprint-generator#445

rleh:
The footprint has been merged.

I force-pushed my commit to restart Travis-CI and detect my typo in the footprint name.
Now fixed!

cpresser:
- [ ] `SCL` should be an input
- [ ] Please move `IREF`, `VCAP` to the bottom of the left side. They will be connected to GND via resistor, cap, so putting them there will make the connections short.
- [ ] Similar to the above point, I suggest to move the `ADDR?` pins to either the top or bottom of the left side for easier wiring up.

rleh:
Updated symbol:
grafik

cpresser:
One more thing, the pin-stack should have the pin with the lowest number visible and set as power.
That does not seem to be codified in the KLC. But we do handle it that way.

rleh:
I updated the symbol again, pin 37 (lowest number) is now visible.
I had thought about which GND pin should be visible, but found nothing in the KLC. Thanks for the info!

cpresser:
Thanks. #Leistungswolf

PR #2248, opened 2019-10-16
Fix SG-8002LB pinname

cpresser:
Pin4 is called Vcc, not V+

This came up when working on https://github.com/KiCad/kicad-symbols/pull/2173

https://support.epson.biz/td/api/doc_check.php?dl=brief_SG5032CAN&lang=en


evanshultz:
The datasheet above doesn't point to the right device. https://support.epson.biz/td/api/doc_check.php?dl=brief_SG-8002LB&lang=en is the right datasheet and it confirms the pin name change. Can you update the datasheet?

And what about the other SG-8002 devices? Can you also fix their pin name and datasheet URL?

cpresser:
The datasheet in this post was a copy&paste error.
I still had it around for the #2173 PR.

I think I got all SG-8002 devices now. Some actually had the wrong datasheet.
Some look like its wrong, but one datasheet has info on most of the parts.

evanshultz:
Thanks!

PR #2251, opened 2019-10-18
Explode sodimm 200

yankee14:
[Datasheet](https://www.te.com/commerce/DocumentDelivery/DDEController?Action=srchrtrv&DocNm=1473005&DocType=Customer+Drawing&DocLang=English)

[Associated footprint PR](https://github.com/KiCad/kicad-footprints/pull/1925)

This symbol used to be split so that the first 40 pins formed a unit, and the remaining 160 pins formed another unit. I needed it to be split further so that the symbol fit on an 8.5x11 inch landscape page. I split the symbol into 5 groups of 40 pins each, and arranged the top/bottom borders as shown below so that the symbol can still be used as one giant connector, if desired:

scap1

I also added a footprint filter that references a PR I am about to submit.

Screenshots of each unit:
sd1
sd2
sd3
sd4
sd5

---


myfreescalewebpage:
Hello @yankee14 

I suggest you should create a second symbol for the split version called "SODIMM-200_split".
Also for the footprint filter, just keep *SODIMM*, this is volunteer to have something generic, has done with mini pcie for example.

Will review the symbol after that.

Joel

myfreescalewebpage:
Thanks, no comme to do, merging.

PR #2253, opened 2019-10-20
Add Potentiometer AD5254

asukiaaa:
Screenshot from 2019-10-20 20-53-32

datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/AD5253_5254.pdf


cpresser:
Nice symbol, i like the graphics...
- [ ] but it has the American style resistors. All other digi-pots use a different style. See `MCP4251-xxxx-P`.
- [ ] Pin15 is called `DGND`
- [ ] I would prefer to have all potentiometer outputs on the right side, similar to `X9258`
- [ ] Please add `, TSSOP-20` to the description
- [ ] It would be nice to also have the alias for 64-position part

asukiaaa:
Thank you for the feedback.

> but it has the American style resistors. All other digi-pots use a different style. See MCP4251-xxxx-P.
 Pin15 is called DGND
 I would prefer to have all potentiometer outputs on the right side, similar to X9258
 Please add , TSSOP-20 to the description

I updated about them.
Screenshot from 2019-11-05 20-15-02


> It would be nice to also have the alias for 64-position part

I'm not sure the expected values for alias.
Could you show some example?

cpresser:
Thanks for you changes. Looking good.

If you want to add the alias, go to the alias-tab (press `+` on the bottom).
Like so:
image
The resulting diff might look like so: 
image

asukiaaa:
Thank you for the explanation.
I understood about alias.
I add alias to AD5254 about AD5253 like this.
Screenshot from 2019-11-07 19-49-59

asukiaaa:
Thank you for checking and merging.

PR #2254, opened 2019-10-20
Adding symbol for nRF24L01 module - SMD version

critBit95:
This symbol is based on nRF24L01_Breakout symbol - only one additional GND pin is added.

image

Some additional information:

1. In my opinion, the name of the symbol is a little bit confusing - the device is a standard module, not a breakout. Right now, I've named my "Breakout_Smd" to avoid further confusion, and renaming existing device. However mabye renaming is worth considering.
2. I wasn't able to find datasheet for the module - however it's sold by many Aliexpress vendors, such as [this one](https://www.aliexpress.com/item/32949511170.html). They all seem to include the same image on their auctions:

image

I have verified it with a module I have on-site.

3. The corresponding footprint PR: https://github.com/KiCad/kicad-footprints/pull/1930


cpresser:
- [ ] Pin-Length should be 100mil
- [ ] I think you are right about the name. Feel free to suggest another name
- [ ] Since this is a module, it should go into `RF_Module`
- [ ] A real datasheet would be nice. Try to get one from the ali sellers or other source.

critBit95:
Sorry for my late work. Regarding the issues you pointed:

- Pin length fixed.
- For now i've renamed the modules to NRF24L01-MOD-SMT and NRF24L01-MOD-THT. I don't know if it's unambiguous eneough, but since there's no official name, I can't think of anything better.
- moved the symbols to "RF_module" section.
- I think, in this specific case, we might want to violate the rule - as the second pad seems to serve as mounting reinforcement rather than electrical connection. And there's no guarantee, that in every project the pin will have to be connected. Sometimes it may even be outside the PCB. I think, it's better to violate S4.3 than to create potential for rather confusing DRC errors. However, it's your call.
- I am not able to find anything that would serve as datasheet for these modules. For now i've removed the previous, broken link.
- Pin name changed.
- I think leaving IRQ pin as it currently is, will make the schematic clearer.

After resolving the aforementioned issues, i'll update the fooprints too.

cpresser:
Thanks for the changes. I did update the checklist in the above post.
Regarding the data-sheet: There seems to be none, I also could not find one. But we want some kind of documentation link.
Did you try to contact one of the sellers and request a better data-sheet?
Worst-case is that we put a link to a shop-website. But those tend to change pretty fast, so its not a good option.

critBit95:
All right, i'll try to contact some Ali vendors.

poeschlr:
@cpresser correction: worst case is we do not add this to the official lib. An asset without documentation is useless to our users so it is essential that a documentation can be added. This documentation must live somewhere where it is expected to be reachable in at least a few years. (One option could be that the documentation is then uploaded here and the asset links to this discussion. But this is to be avoided if at all possible.)

To be honest i find it disturbing that people sell components without providing proper documentation. How can i know if the component does what i expect it to do? How can i tell if the component is within its specifications without knowing them?

chschlue:
@poeschlr 
I'm all for a professional-grade official lib, **but...**

Additions of this kind come up regularly, most probably because KiCAD, being free as in beer, will always have a large hobbyist user base, meaning there will always be demand for "Alibaba style" components.
How about establishing some general guidelines regarding this matter?
(Like a) summarily rejecting poorly documented additions or z) supporting some kind of parallel "in demand but crappy" lib or one of the 24 other options in between)

poeschlr:
In this particular case there is not even any documentation provided for what package pad has which function. So the question is how would one determine this? (how would we check it for correctness) Does one get documentation if one buys this part? If so can this documentation be shared?
And also how would a user determine that the part they buy does indeed fit the version we have?

Or would you suggest that simply by providing bad documentation suddenly we ignore our normal rules for what to check?

chschlue:
> Or would you suggest that simply by providing bad documentation suddenly we ignore our normal rules for what to check?

Certainly not.
I'm suggesting that badly documented PRs should not lead to endless discussions over and over again, wasting several people's time.

critBit95:
Since I wasn't able to find anything remotely resembling proper documentation, I'm closing the PR.

PR #2257, opened 2019-10-20
Add LT3756xMSE buck/boost/sepic 100V LED driver

MatthewStephenson:
Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/375612fb.pdf

This is a straight copy of the existing symbol for the pin-compatible LT3755xMSE, with the relevant data fields changed.

image

Rule 4.4 will fail the travis checks because VREF is actually an output on this chip.


cpresser:
Thanks for your contribution.
Pinout seems to be identical:
image
image

We prefer you use the alias feature instead of adding a 1:1 copy. That saves some bytes in the library and makes it load faster.
Also, if there is a bug with the symbols, fixing does only require one change.

MatthewStephenson:
Thanks for the feedback, I had forgotten about the alias feature. I've aliased the variants as well, as they all had identical pinouts.

Driver_LED.dcm
```diff
@@ -219,7 +219,43 @@ $ENDCMP
 $CMP LT3755xUD-2
 D 40 Vin, 75 Vout, DC/DC LED Controllers, Improved Open led Status pin, QFN-16
 K Boost Buck Buck-Boost SEPIC Flyback
-F https://www.analog.com/media/en/technical-documentation/data-sheets/37551fd.pdf
+F https://www.analog.com/media/en/technical-documentation/data-sheets/375612fb.pdf
```
cpresser: This data-sheet should not change.

MatthewStephenson:
Oh, that was a silly mistake. Sorry for that, and well spotted.

cpresser:
Thanks!

PR #2261, opened 2019-10-22
Update datasheet links for MCU_Microchip_SAME parts

Lohrer:
MCU_Microchip_SAME now use one datasheet for the SAM E70/S70/V70/V71 family (https://www.microchip.com/mymicrochip/filehandler.aspx?ddocname=en589960), and one for the SAM D5x/E5x family (https://www.microchip.com/mymicrochip/filehandler.aspx?ddocname=en599585).

### DIFF
--- a/MCU_Microchip_SAME.dcm
+++ b/MCU_Microchip_SAME.dcm
@@ -3,97 +3,97 @@ EESchema-DOCLIB  Version 2.0
 $CMP ATSAME51J18A-A
 D SAM E51 Microchip SMART ARM Cortex-M4F-based MCU, 256K Flash, 128K SRAM, TQFP-64
 K 32-bit ARM Cortex-M4F MCU Microcontroller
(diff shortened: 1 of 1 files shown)


chschlue:

Please use canonical URLs instead of redirections, i.e.
http://ww1.microchip.com/downloads/en/DeviceDoc/60001507E.pdf
http://ww1.microchip.com/downloads/en/DeviceDoc/SAM-E70-S70-V70-V71-Family-Data-Sheet-DS60001527D.pdf

Lohrer:
Switched. I assumed the redirects would be longer-lasting but based on the SAM E51 links for example it seems at worst they'll just become outdated.
The end result is now that the SAM E51/E54 datasheet links are updated to the latest revision and the SAM E53/E70 links are functional again.

chschlue:
Thanks for the fix.

PR #2269, opened 2019-10-27
added AS13704

Canwail-Modular:
added ALFA As13704
Symbol is based on that of the LM13700
http://www.alfarzpp.lv/eng/sc/AS13704.pdf
grafik

KLC Scripts tell me that 3.2 and 3.6 are violated, but that is the case for almost every device, so that seems to not be an issue with the symbol itself. 

---


cpresser:
Thanks for you contribution.
It looks pretty good. A few things should be improved though:
- [ ] Add the NC pins (7, 18). If they are present the footprint-filter based on pin-count will work properly. I suggest you add them to UnitA, according to KLC S4.6
- [ ] The Travis-CI datasheet-error should be fixed.

Canwail-Modular:
grafik

I added the NC pins as seen in your image above and moved the datasheet and info text down a notch.

cpresser:
Thanks!

PR #2270, opened 2019-10-28
Microchip SAM E and SAM L corrections

Lohrer:
The kicad-library-utils scripts reported a few errors for a few SAM E/L parts.
- Fixed pin name position offset for SAM E/L parts, set to preferred 20 mils. (S3.6)
- Fixed ATSAML21G16B-AUT footprint filter to TQFP (S5.1)
- Fixed ATSAML21G16B-AUT power pin stacking (S4.3)
  - Stacked power pin was not labelled as passive.
- Fixed off-center symbol for ATSAME54N19A-A (S3.1)
  - The symbol was extended 50 mils on the bottom to re-center.


cpresser:
Thanks!

evanshultz:
@cpresser 
Wasn't merging this introducing a breaking change since the GND pin positions moved? I suppose too late now, but the off-center clause/check is often ignored in the interest of aesthetics.

cpresser:
Yes. My bad, I didn't consider the GND pins do break stuff. 
I suppose the ERC will catch this if there are disconnects. But still, that was a mistake.

I don't think that it is necessary do undo the change on 'ATSAME54N19A-A` as it is not really critical and the new symbol is 'better'.

PR #2276, opened 2019-10-29
[Inteface_USB/MCP2210] Swap USB and SPI pins positions

maximeborges:
As discussed in https://github.com/KiCad/kicad-symbols/pull/2094#issuecomment-546597392, here is my pull request to swap the USB and SPI pins on the MCP2210.

Actual version:
MCP2210 old
With pins swapped:
MCP2210

The new symbol is a bit longer but more coherent with the functionalities of the component: USB and control (reset/osc) on the left, SPi and GPIOs on the right:

Example with components and nets:
image


chschlue:
@diegoherranz assigned you because #2094 was yours. Feel free to yell at me if you're not ok with that.

diegoherranz:
I think it looks better. However, the RST will be probably controlled by the same IC that the SPI and GPIOs, as @evanshultz, already mentioned on https://github.com/KiCad/kicad-symbols/pull/2094#issuecomment-536630285. Is that case on your example? That is, same chip controlling SPI and RST?
Following that, RST could be at the top right. On the other hand, RST on the top left feels natural... What do you think?

Also, VDD and VUSB could be swapped, following the same that was mentioned on point 7 here https://github.com/KiCad/kicad-symbols/pull/2094#issuecomment-539723435

Thanks!

maximeborges:
Usually it's used as an USB Master to SPI Slave, so for most common cases it would be connected to VDD IMO. My test point is to force a reset from an external debugger connector (I should actually add a resistor there btw...).

Anyway, I've always seen reset on the top left on correctly designed symbols in KiCAD, so I don't think there is any reasons to move it elsewhere.

I don't have any opinion on swapping VDD and VUSB, but I can do it. In the datasheet we can see that VDD is the input of the internal LDO, and VUSB is the output. So for coherence I would keep VDD on the left and VUSB on the right.

diegoherranz:
Sorry for the delay.

Regarding the RST, I think I'm trying an impossible, hehe! Because I think it would be ideal on the same side as the SPI and GPIO but also top left is where you would normally expect it.
If SPI and GPIO would move to the left hand side, and the USB to the right hand side, the symbol would become very long.
I think it looks OK as is, and is an improvement, in my opinion.

> In the datasheet we can see that VDD is the input of the internal LDO, and VUSB is the output. So for coherence I would keep VDD on the left and VUSB on the right.

Fair enough.

Merging. 

Note: Please note that while this is a breaking change, the symbol hasn't made it into any release so I think it is OK (asked here https://github.com/KiCad/kicad-symbols/pull/2094#issuecomment-547316583).

Thanks!

PR #2279, opened 2019-11-01
Fix Vishay-MBLS pin order

cpresser:
This is the fix for https://github.com/KiCad/kicad-footprints/issues/1869

Datasheet: https://www.mouser.com/ds/2/427/mbl104s-245342.pdf


evanshultz:
@chschlue 
I haven't seen Dan in a while. You might as well go ahead and take this one.

chschlue:
Yes, I just thought it would be nice to knock first.

PR #1019, opened 2018-09-29
Added LPC11U35FHI33-501

tmineno:
## Added LPC11U35FHI33-501, NXP CortexM0+ MCU

Datasheet: https://www.nxp.com/docs/en/data-sheet/LPC11U3X.pdf
lpc11u35fhi33-501

------------
Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required

PR #2283, opened 2019-11-02
Add LPC11U35FHI33-501

DanSGiesbrecht:
# Add LPC11U35FHI33-501, NXP CortexM0+ MCU

(Continuation of #1019)

Datasheet: https://www.nxp.com/docs/en/data-sheet/LPC11U3X.pdf

symbol

My original review checklist:
- [x] Please update RefDes positions.
- [x] Please remove commas separating the keywords, and replace with spaces.
- [x] As per the other PR, please change `KB` -> `kB`.
- [x] Please add `QFN-32` to the description.
- [ ] Please consider adding aliases to account for the multiple flash configurations (drop-in replacements for this symbol, not all of them)
  - **The datasheet does not mention parts which could be aliases.**
- [x] Please modify the footprint filter:
  - [x] I think that `*QFN*:` can be removed. I have not seen the library name commonly used in the filters.
  - [x] The number of pins should be removed. CvPCB uses the number of pins on the symbol to determine this.
  - [ ] The EP size should be removed to make the filter more robust.
    - **I chose to leave this in**
  - [x] I think that `?` should be replaced with `*`, to account for potential changes to footprint names, for robustness.
- [x] Because there is only one GND/VSS connection, please change pin 33's name to `VSS` instead of `PAD`. This allows for more clarity to the designer.

---


chschlue:
Regarding unchecked items:

> * [ ]  Please consider adding aliases to account for the multiple flash configurations (drop-in replacements for this symbol, not all of them)
>   * **The datasheet does not mention parts which could be aliases.**

Aren't all QFN variants pin compatible?

>   * [ ]  The EP size should be removed to make the filter more robust.
>     
>     * **I chose to leave this in**

Why?

DanSGiesbrecht:
> * [ ]  FP should be VQFN-32-1EP_7x7mm_P0.65mm_EP4.7x4.7mm
> 
> Note: I wonder why it hasn't come up before that almost everything about the FP is wrong. Am I missing something?

Due to a somewhat poor datasheet naming, there are two packages called `HVQFN33`; one with a 5x5mm body, the other with 7x7mm body:

image

---

**LPC11U35FHI33** calls out the 5x5mm package:

image

I don't believe that 'everything about this footprint is wrong', but please let me know if I missed something.

chschlue:
Ah!
My mistake. Remind me to avoid NXP like the plague...

chschlue:
> * I believe that the EP size should be explicitly defined.

I'm sure you left it in because you believe it should be left in :)

Not sure whether I should ask why you left it in or why it's generally not supposed to be left in.
The only sense I can actually make of this rule is for the filter to match SMD variants.

chschlue:
One more thing: While most of the FP is indeed correct, it's still missing a V.

PR #2284, opened 2019-11-03
Added TPS23753A

ppielorz:
Hi,
I've added TPS23753A PoE interface and converter controller.
Here is the datasheet: http://www.ti.com/lit/ds/symlink/tps23753a.pdf

Screenshot:
image

I had sone doubts when I was drawing this symbol:
- There are many pins marked as input/output in datasheet but their placement isn't matching KLC - I've left them on different sides according to basic implementation of first page of datasheet,
- The datasheet says that this is TSSOP14 chip. However the one I found in modules have correct general dimensions but pad dimensions differs a bit (1,55x0,3mm in datasheet vs 1,45x0,45mm in KiCAD)
- Also pin directions are not clear in this chip - would it be better to leave them as passive maybe?

---


chschlue:
Note: This far from a full review.

> * The datasheet says that this is TSSOP14 chip. However the one I found in modules have correct general dimensions but pad dimensions differs a bit (1,55x0,3mm in datasheet vs 1,45x0,45mm in KiCAD)

That's fine. The TSSOP-14 footprint in the lib is a generic one following IPC recommendations, not TI's suggested PW package footprint.
Following up on that: This symbol should be called TPS23753APW (adding its package suffix.)

> * Also pin directions are not clear in this chip - would it be better to leave them as passive maybe?

It looks this symbol is going to need a lot of passive pins. Its power pins, for example, are usually going to be connected to a bridge rectifier's passive output pins.

PR #2285, opened 2019-11-04
Tps6303x

cnieves1:
Added TPS63030 and TPS63031.
Footprint depends on: https://github.com/pointhi/kicad-footprint-generator/pull/453.
Datasheet: http://www.ti.com/lit/ds/symlink/tps63030.pdf
imagen

imagen


cnieves1:
Travis CI failed with invalid footprint (pending PR, see first message), and:
```
 Violating S4.2
    Positive power pins should be placed at top of symbol
     - Pin VIN (5) @ (-400,200)
```
This PR uses the same symbol as TPS63000 (alreday in the library) with different footprint. 
I hope the required change is not needed.

chschlue:
VIN on the left is fine.
But power input for VINA doesn't make sense to me, given how it's supposed to be used. Should be output or passive.

cnieves1:
Why? In the datasheet (page 4) it is specified as power input:

imagen

chschlue:
tps1
tps2

cnieves1:
Agreed. Changed VINA to passive.

chschlue:
Thanks.

PR #2288, opened 2019-11-05
Added BelFuse SI-60062-F.

cnieves1:
Datasheet: https://belfuse.com/resources/drawings/magneticsolutions/dr-mag-si-60062-f.pdf
Footprint depends on PR: https://github.com/KiCad/kicad-footprints/pull/1954
Same symbol as RJ45_Amphenol_RJMG1BD3B8K1ANR, changed footprint, alias
and datasheet link.

imagen

imagen


jneiva08:
@cnieves1 have checked the pin order and don't make much sense.

Have found the following links:
https://componentsearchengine.com/SI-60062-F/Bel
http://ww1.microchip.com/downloads/en/DeviceDoc/61166A.pdf page 22

SI-60062-F

cnieves1:
You were right. They should be correct now.
L3 and L4 were exchanged in the footprint PR, to have a consistent numbering from left to right.

imagen

myfreescalewebpage:
Hi @cnieves1 , thanks for contributing,

- [ ] The tx/rx pins are called "T" in the datasheet, why choosing "R" ?

Else ok with the pinout according to the current latest revision of the footprint. Will check again when the footprint will be merged.

Cheers,
Joel

cnieves1:
At the time I draw it, there was another RJ footprint I based this footprint PR on, with pin names R[1-8].
Isn't this right?

Captura de pantalla de 2020-04-29 08-42-17

Captura de pantalla de 2020-04-29 08-42-09

myfreescalewebpage:
It's probably correct for the other device, but you should simply create your new symbol and your new footprint according to the right datasheet.

myfreescalewebpage:
@cnieves1 ping

cnieves1:
Hi Joel,
Sorry for the long delay... Life got complicated... I will try to fix it...

myfreescalewebpage:
Thanks for the news, I keep an eye on it

cnieves1:
Hi Joel,
In case you missed it, the requested changes are done, if I am not mistaken...

Thanks

myfreescalewebpage:
Yes I know, waiting the footprint before the symbol can be merged :)

myfreescalewebpage:
Personal note: pin name to be checked again after integration of the footprint

myfreescalewebpage:
@cnieves1 can you update the pin names here following the integration of the footprint ? Then I think this will be ok.

cp-aquila:
Pin13 should be called `SH`: https://kicad-pcb.org/libraries/klc/S3.7/

myfreescalewebpage:
@cp-aquila yes we already change this on the footprint, I just wait @cnieves1 to make the symbol according to the footprint (see my previous comment)

cp-aquila:
Sorry for the noise, I looked at an outdated screenshot.

myfreescalewebpage:
No problem :)

myfreescalewebpage:
@cnieves1 number of pin SH should be "SH", not 13, to match with the footprint
Joel

cnieves1:
@myfreescalewebpage  Sorry! I changed pin name instead changing pin number... Fixed now. Thanks!

myfreescalewebpage:
Thanks. May I require one more change, please: the pin length should be 100mil for this symbol. Other RJ45 symbol have 150mil but it's an error. I would like to improve all of them and starting with yours is a good point. Just set to 100mil and increase a little bit the size of the border outline to not draw again all the symbol.

myfreescalewebpage:
Would like to have this before going to GitLab, accept the previous defect because it's already present for most RJ45, I know we have to make a global rework of them

cnieves1:
Thanks Joel! I'm on a trip with no access to kicad, so I could do nothing...

myfreescalewebpage:
No problem

PR #2291, opened 2019-11-08
Added symbols for LEM CKSR series

almghandi:
Push of
CKSR_15-NP alias CKSR_6-NP, CKSR_25-NP, CKSR_50-NP
[https://www.lem.com/sites/default/files/products_datasheets/cksr_series.pdf](https://www.lem.com/sites/default/files/products_datasheets/cksr_series.pdf)

One symbol for the current transducer of the CKSR series with Aliases for primary currents of 6, 15, 25, 50 A

This is **my first time trying to contribute to KiCAD** and using git. I tried to follow all the Conventions, please let me know if I did something wrong. I also made the footprint and added the corresponding 3D model

Edit: 
Link to **footprint** pull request: [https://github.com/KiCad/kicad-footprints/pull/1958](https://github.com/KiCad/kicad-footprints/pull/1958)
~~Link to **3D model** pull request: [https://github.com/KiCad/kicad-packages3D/pull/615](https://github.com/KiCad/kicad-packages3D/pull/615)~~

Edit2:
Outdated screenshot removed


cpresser:

It looks pretty good:
- [x] Description and Keyword are good
- [x] It is in the correct library
- [x] Pin-Numbers are correct
- [x] Vref can be an input or output. Which makes it hard to decide if it should go to the top or right side of the symbol. Top seems to be fine, in most scenarios this will be connected to some kind of supply symbol. The pin-type of passive seems ok: it won't produce a ERC error in either usage scenario. 

but there are a few things we can try to improve:
- [ ] The supply-voltage pin is called `Uc` (instead if `+Uc`)
- [ ] Next time, please add a link to the footprint pull request in the first post (https://github.com/KiCad/kicad-footprints/pull/1958)
- [ ] The checklist from the PR-Template is also pretty useful, please use it next time.
- [ ] I would like to see the simplified internal diagram in the symbol. That is just a suggestion though, not mandatory as the other current sensor in the lib also follow this style (e.g. `HO40-NP`). The connections between the in- and out-pins will be pretty useful for the circuit designer. That also means that the in and out-pins need to move to top and bottom respectively. You don't need to add the rest of the internal schematic as shown in the data-sheet, that would be to much. 
- [ ] a footprint-filter is missing (lets wait for the footprint to be merged for this)

almghandi:
I updated the description and pin name. And also I added the simplified internal diagram. Here a screenshot of the changes:

image

cpresser:
Thanks, i like the internal diagram. Well done.

However, I would slightly move the pins on top. The goal is to have the `GND` and `Uc` aligned.
image

cpresser:
@almghandi are you still active on this PR? If yes, I will give the footprint a review so we can get this merged.

almghandi:
@cpresser yes please! In the meantime I can also confirm that this footprint fits and works on my PCB.

P1100997_git


If the footprint is ok, I am also going to align GND and Uc as you suggested for the symbol.

cpresser (approved):

almghandi:
@cpresser Here you can have a look at the PV inverter, together with  the thesis that should explain theory and design process:
[https://cloud.owncube.com/s/nMoemmMnfAjG3cK](https://cloud.owncube.com/s/nMoemmMnfAjG3cK)

Should I create the symbol for CKSR 75-NP in a new pull request?

PR #2265, opened 2019-10-26
modified Arduino Nano symbol, added Nano Every alias

5N44P:
Hi, i've modified the Nano symbol. Changes:
- Hidden duplicated pins (rst and gnd)
- Swapped left and right pins, to be consistent with the convention already in place with microcontrollers
- Added an alias for the new Nano Every, which is pin compatible. Reference: https://content.arduino.cc/assets/NANOEveryV3.0_sch.pdf

image

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

PR #2292, opened 2019-11-09
added Arduino Nano Every

valerionew:
Not much changed in the new Arduino Nano Every, but a couple pins were renamed, so i tought it would be easier to use with the new names, grouped by function.

See also #2265, closed to do this modifications, and #2136 which generated some conflicts with the previous PR.

image

Reference for renaming:
https://content.arduino.cc/assets/NANOEveryV3.0_sch.pdf
image

The footprints stays the same, even though we could add an smd footprint, for mounting the module with the castellated holes on the sides.

---


myfreescalewebpage:
Hi @5N44P , thanks for contributing,

- [ ] RESET pins should be stacked

Else, no comment to do.

Cheers,
Joel

5N44P:
Should be good now

5N44P:
(#2293  should also be mentioned here)

myfreescalewebpage:
Thanks for the issue you just indicated, it's not a bad comment, at least for the RESET pin. Maybe we can keep GND pins stacked because it is usually recommenced to connect all GND pins on modules, but keep RESET pin unstacked.

What do you think ?

myfreescalewebpage:
@valerionew ping see my last question

valerionew:
@myfreescalewebpage Sorry if took a while, i was traveling the past days.
I think the best solution is the one i just commited, with stacked grounds and unstacked reset pins

myfreescalewebpage:
I agree with you, it's very nice.
You have a stacking error, see travis log.
Else will be ready to merge.
Joel

valerionew:
@myfreescalewebpage yeah my bad, should be good now!

myfreescalewebpage:

PR #2306, opened 2019-11-15
Added TSM2302CXRFG, alias of BSS138

cnieves1:
Added TSM2302CXRFG, alias of BSS138.
Datasheet:  https://www.taiwansemi.com/products/datasheet/TSM2302CX_E1608.pdf
imagen
Please notice that pin 2 and 3 are exchanged if we compare them with SOT-23 footprint.
imagen
imagen


diegoherranz:
Can you name it TSM2302CX following the discussion on https://github.com/KiCad/kicad-symbols/pull/2307#issuecomment-555747430?

Thanks!

diegoherranz:
This one doesn't merge cleanly now. Can you fix that?
Also, I've noticed that the order of Id and Vds in the description is the opposite compared to most of the other FETs. Can you change that too? Could you change it for the TSM2301ACX too?

Thanks!

cnieves1:
Done for TSM2302. TSM2301 Id and VDs data order seems to match other P-channel transistors...

diegoherranz:
> TSM2301 Id and VDs data order seems to match other P-channel transistors...

Sorry, I'm not sure what I was looking at.

Can you fix the Id to be 3.9A again? You had it at 3.9A but it changed back to 3.2A when you fixed the merging problems.

Thanks!

diegoherranz:
Thanks. Merging.

PR #2307, opened 2019-11-15
Added TSM2301ACXRFG, alias of TP0610T.

cnieves1:
Added TSM2301ACXRFG, alias of TP0610T.
Datasheet: https://www.taiwansemi.com/products/datasheet/TSM2301A_C15.pdf
imagen
imagen
imagen


diegoherranz:
Thanks for your submission. Just a few minor comments:

- I believe we normally use the Id current at 25C (-2.8A in this case). Can you change that?

- Can you remove the suffix and leave it as 
without suffix? TSM2301A. That's what we do on most (all?) other transistors.

Note: Same as you mentioned on https://github.com/KiCad/kicad-symbols/pull/2306, Taiwan semi is using a weird pin numbering for sot-23 (don't ask me why!), but the assignment is correct.

cnieves1:
Done.
What is the criteria in a general case? Usually the suffix is related to footprint, and the pin numbers also depend on footprint, so I guess the suffix is important...
I could not find any info about this in KLC.

Thanks,

poeschlr:
The datasheet does not make it clear what part of the part number encodes which information. My guess is that some of it will encode the physical package, some the transport package (reel, tray, ....), some might encode the temperature rating, and other parameters.

The symbol (or alias) name should only include relevant info. Meaning in this case the base part number plus whatever specifies that it is in sot-23 package. everything else should be replaced with wildcard "x". Wildcard chars at the end of the name should be removed. 
KLC rules: G2.x and S2.x

diegoherranz:
> The datasheet does not make it clear what part of the part number encodes which information

Agree.

Having a look at 
https://www.taiwansemi.com/en/products/parametric-search/mosfet it seems that every SOT-23 FET includes the CX suffix. That suffix is not used for any other package size. And Taiwan semi includes the CX in the part number column of that table, leaving out the RFG.

Having said all that, I think we should go for TSM2301ACX. Could you change it, @cnieves1? Sorry I didn't figure this out the first time, asking for two changes...

Thanks!

diegoherranz:
Thanks for your work. It looks OK, so merging.

By the way, please note that this FET is not recommended for new designs (NRND). TSM650P02CX is suggested as an alternative.

PR #2309, opened 2019-11-16
Added 74XG123

rs-development:
I matched pins and position with 74LS123.

http://www.ti.com/lit/ds/symlink/sn74lvc1g123.pdf

image


rs-development:
Changed the name, so now it is a replacement of an old styl symbol.

rs-development:
image

image

chschlue:
Breaking changes like this are not possible until the next major release.
Also, active low pins should keep the bar instead of the circle ([S4.7](https://kicad-pcb.org/libraries/klc/S4.7/))

rs-development:
Updated the label for pin A. How is the workflow regarding the next release?

chschlue:
This isn't carved in stone yet, but the overall plan is to migrate to GitLab after 5.1.6 is released and start work on v6 libs, including relaxing rules regarding changes to existing symbols.

rs-development:
> including relaxing rules regarding changes to existing symbols

Is there a public discussion about that? T would like to participate to optimize the current workflow for library management and consistency.

chschlue:
Sort of... the topic has come up here and there.
https://github.com/KiCad/kicad-symbols/issues/2345 is probably the "official" thread.

myfreescalewebpage:
@chschlue what is the status here ? I can continue the review if you want

chschlue:
@myfreescalewebpage you're welcome to do so.

myfreescalewebpage:
Just check, for me no comment to do.
Merging.

PR #2310, opened 2019-11-17
added LTC3105

rs-development:
https://www.analog.com/media/en/technical-documentation/data-sheets/3105fb.pdf

image

I have created 2 Parts for 2 Footprints, please check the pinning!


rs-development:
Don't really agree with MPPC as Input, it does output current. You're sure? @chschlue

chschlue:
True, it does source 10µA.
But then, input can be connected to resistors, diodes, capacitors and power symbols.
Output connected to ground throws an ERC error.

Am I missing something here?

rs-development:
Correct it can be connected to GND to disable MPPC function. So Output is not ok.

chschlue:
Thanks.

PR #2311, opened 2019-11-18
Fix the output pins of LTC6362* devices which were marked as power in…

gbrown128:
…puts.

Not much to say here. The output pins were erroneously marked as power inputs, turned up as an ERC error.

I've skimmed over the remainder of this file and can't find any other devices where this mistake has been made.

Checked the [datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/6362fa.pdf) thoroughly and can't imagine any reason to refer to these pins as power inputs.


evanshultz:
Yep. Looks like an accident. Thanks for catching this and submitting a fix!

PR #2315, opened 2019-11-19
Fix output pin position to be on 100mil grid.

cnieves1:
Output pin was not within 100mil grid. This commit fixed it.
However, it is not the only symbol in 74xGxx that has this problem, so I don't know if this is intentional or not...


evanshultz:
I personally would say this isn't worth fixing one at a time because one item everybody is hopefully for with v6 are using the symbol generator to remake all logic symbols.

chschlue:
I would say it doesn't hurt either, especially because this particular fix is only one click away and there's a non-zero probabilty that nothing remarkable is going to happen to the logic libs in time.

evanshultz:
It does make the pins not align any longer with any other logic gate in the library that is 2-in, 1-out. And swapping and OR for an AND, etc. can happen during development. At least it has to me. I'd rather keep homogeneity as it is until the whole library is revamped. My opinion.

chschlue:
Fair enough.

PR #2317, opened 2019-11-19
Fix LT3755 ISN/ISP current-sense pin ordering

MatthewStephenson:
This change fixes the vertical order of the ISP and ISN current-sense pins so that the positive current-sense pin is uppermost on the diagram. This error caused me to have PCBs manufactured with these pins incorrectly wired, because I didn't think to check that the positive sense pin was at the top.

Datasheet link: https://www.analog.com/media/en/technical-documentation/data-sheets/LT3755-3755-1-3755-2.pdf

Before this change it was necessary to cross the current-sense wires as shown, because the ISN (negative) pin was uppermost:

image

After this change the wires don't need to be crossed:

image

This change is consistent with the diagrams in the datasheet which have ISP uppermost:

image

---


poeschlr:
Could you move the two pins a bit down such that ERC will complain with not connected on at least one pin if users update the lib without noticing this? (Kicad does not include symbols in schematic files and this change is easily overlooked in the rescue dialog.)

MatthewStephenson:
Moved the pins down as requested. It did cross my mind whether this would be a bit dangerous for people with existing designs!

image

cpresser:
There are a few variants of this chip? Can you please fix all of them?

myfreescalewebpage:
@MatthewStephenson following @poeschlr comment can you do the same you have done with LT3755xUD on the LT3755xMSE  symbol ? (position of pins) Joel

myfreescalewebpage:
@MatthewStephenson ping

PR #2330, opened 2019-11-23
FSUSB42MUX: Addition of USB2.0 switch

myfreescalewebpage:
Add FSUSB42MUX device to the library. Design based on USB3740 already in the library.

---

https://www.onsemi.com/pub/Collateral/FSUSB42-D.PDF

Capture

- [NA] If there are matching footprint PRs, provide link(s) as appropriate


myfreescalewebpage:
Hello @cpresser thank for looking at my symbol! Any comments ?

cpresser:
Sorry for the delay. I grabbed this because it seemed simple and wanted to do it right away. But I closed the browser tab and did forget about it. My bad!

Anyway, lets do some checks
- [x] pin electrical types
- [x] Data-sheet, description, keywords
- [x] Pin numbering
- [x] package and FP filter
- [x] Overall layout / pin locations
- [ ] PartName: I am not sure if we should wildcard some parts of the name (the last `X`?) I am not sure what all the letters stand for, I can't find it in the data-sheet.
- [ ] GND and VCC Text overlapping the pins on the right side. That looks ugly, But I don't see a really good solution. According to S3.2 it is allowed to change the Text size. To me changing the text size in this case does not feel right. What do you think?

myfreescalewebpage:
@cpresser thanks for the review.
- The last "X" in the part name is not a wildcare (we use xx in lowercase for that in the lib). Instead, it's a letter which is part of the real name of the device.
- Yes the letters are not properly displayed on the right. I made myself the same remark but was waiting your position. The main interest to keep this design is because it is the same than the USB3740 (similar device in the lib). I have reduce the text size to 1mm, this give the following result:

Capture

cpresser:
Thanks, the smaller font looks good.

Regarding the wildcard thing: That was a misunderstanding. I was asking if we want to replace any of the `MUX` with a lower case `x` to wildcard them. 
I don't know what the letters are for (I could also not find a onsemi reference for that), but I think it can be assumed that they somehow represent the package and perhaps the temperature range.
But since we don't know for sure if the letters are code for some properties, lets keep them in the name.

myfreescalewebpage:
Yes we don't know exactly today.
Thanks for the review.

PR #2340, opened 2019-11-27
New symbol for FDMT MOSFET series

SchrodingersGat:
The existing FDMT MOSFET symbols are an alias of a MOSFET in a 5x6mm package.

- These are in an 8x8mm package, not a 5x6mm package
- The footprint has a different pinout

Here is an example of the correct device:

https://www.digikey.com/product-detail/en/on-semiconductor/FDMT800120DC/FDMT800120DCCT-ND/5360116

This PR creates a new symbol for these and sets the correct footprint.

image


myfreescalewebpage:
Hello @SchrodingersGat no comment on the symbol, can you fix the datasheet links of FDMT800100DC, FDMT800120DC, FDMT800150DC, FDMT800152DC and FDMT80080DC at the same time, it seems they are broken since it's now on ON Semiconductor website ? Joel

myfreescalewebpage:
@SchrodingersGat  ping

myfreescalewebpage:
Merging, my previous comment was related to neighbor devices.

PR #2344, opened 2019-11-27
Fix pins for ADT1.5-1

### LINKED ISSUE #2343: ADT1.5-1 wrong pin assignment
ADT1.5-1 which is an alias of ADT1-1WT has different CT and NC pins  
ADT1.5-1 has pin 2 as secondary CT and pin 5 not connected

drid:
ADT1.5-1 which is an alias of ADT1-1WT has different CT and NC pins  
ADT1.5-1 has pin 2 as secondary CT and pin 5 not connected
Fixes #2343
image
image

https://www.minicircuits.com/pdfs/ADT1.5-1+.pdf

---


evanshultz:
@drid 
Sorry for this mistake and thanks for reporting it.

We also need to reference https://www.minicircuits.com/pdfs/ADT1-1WT.pdf to check this.

I see the pin table is different so they can't be the same symbol.
ADT1-1WT:
image

ADT1.5-1+:
image

But there are no dots so we can't fully check this. You will need to add dots to both symbols to fully verify them. Can you please do that?

Here is the referenced footprint:
image

drid:
Is something like 
image
 ok or should i use the inverted style on pins?

drid:
or even better
image

evanshultz:
@drid 
The symbol needs the NC pin, pin 5, added so the pin count filter in CvPcb works. Can you please add it? And I believe this is also missing from the `ADT1-1WT` symbol. Then I think this is good.

Sorry for the long, long delay. And thank you for fixing this!

drid:
Added NC pin 5.
There seems to be some inconsistencies with ADT1-1WT aliases and CT pins.
I think i'll do a separate MR for these

evanshultz:
Thanks!

ADT1.5-1 looks great now, but there is one visual thing: Please push the NC pin a bit right so it's under the part symbol and thus not easy to accidentally connect. It could be at [-100,0] or even [0,0]. Also, please make that pin 100mil long.

In addition, ADT-1AT touched in this PR and it's not right:
image

Pin 2 should be the secondary CT. See https://www.minicircuits.com/pdfs/ADT1-1WT.pdf. And there's no NC pin. It was wrong before but I hesitate to merge this when I still see problems in a symbol that was modified. Would you mind correcting it too?

Lastly, it's easier to incrementally review if you don't force push. Thank you.

drid:
It looks like ADT1-1WT is wrong regarding pin 2/5 but all its aliases are correct.
So ADT1-1WT should be an alias of ADT1.5-1

evanshultz:
I didn't check all the ALIASes, but if so then yes, we needs a new symbol. It's not required to merge this PR since it's fixing another symbol, but if you are willing to do it that would be great.

BTW, I'm not sure about all the ALIASes but I looked at the first one, https://www.minicircuits.com/pdfs/ADT1-1WT-1.pdf, and it doesn't match the screenshot I posted above. Pin 4 should be the dot on that part.

drid:
I checked all the aliases and it they have to be split into 3 parts if we want to be precise.
I'll do that now

drid:
image
image image

evanshultz:
Thank you! I verified the pinout of all parts and it looks good to me too.

Only one small thing: NC pins on `ADT1-1WT-1` and `ADT1.5-1` aren't named `NC`. That will match the other symbols.

drid:
Just to clarify, NC pins should me named NC or ~ ?

evanshultz:
Normally we name them `NC`. However, there's no rule that I know of. And I just looked at https://kicad-pcb.org/libraries/klc/S4.4/ and the next few KLC clauses and don't see anything. So I will merge them with a tilde.

Thank you so much for fixing so many things, and sticking with this PR to get it finished even after I dropped the ball!

PR #2352, opened 2019-12-03
Addition of LM20 Temperature Sensor

jamesrlucas:
Addition of LM20 Temperature Sensor in SC70 package

I've only created the SC70 package verison, as there is no footprint for the DSBGA variant - I'll add it to my to do list, but it may be some time before I get to it!

Datasheet Link: http://www.ti.com/lit/ds/symlink/lm20.pdf

image

---


cpresser:
Hi and thanks for your contribution!


There are a few small things that should get fixed
- [ ] According to the datasheet is is not mandatory to connect pin2 to GND. There might be a use-case where the user does not want to measure the temperature of the ground-plane, but of the housing of the sensor. In that case Pin2 is left floating. Or am I mistaken here? If not, put the two `GND` pins next to each other.
- [ ] Rename `Vout`to `Vo`, and `+Vs` to `V+` to match the exact names in the data-sheet
- [ ] You did wildcard part of the MPN. Can you explain why? I could not find it in the data-sheet.
- [ ] I am not sure about the footprint-filter. I checked some other parts in the lib that also come in `SC-70` housing. There are at least 5 different variants of filters for the same package :cry: I would suggest to use `SOT?353*`

jamesrlucas:
All done. 

I made Pin 2 a passive pin (so it wouldn't fail DRC if the designer chooses to leave it disconnected) Travis doesn't like this, should I have it as a power input?

I've also removed the wildcard, and created an alias instead - there are LM20B & C variants, with different accuracy specifications. I didn't put the accuracy in the description, as it's not fixed, it varies with temperature...

cpresser:
Thanks for the changes. I have two more requests:
- [ ] The description should differ between the parts. There is no need to do an alias if they have the same specs in the description: But you are right, it is hard to find a few simple words to describe the difference. This graph here is already pretty weird: 
image Perhaps just take this as typical rating (which is also the best rating):
image
- [ ] I am not quite sure how to handle Pin2. Naming it `GND` follows the data-sheet and is correct. But if gives the designer no hint on why Pin2 is different than Pin5. What do you think of renaming it to `NC/GND`? I am not sure if I like it myself, perhaps there are good arguments for or against renaming. Or perhaps you have a better idea for a different name?

jamesrlucas:
Okay, I'll add the typical rating.

I agree about differentiating between pin 2 and 5, how about `DIE` or `GND/DIE` for Pin2 as it connects to the die? Or use `SUB` instead, meaning substrate?

cpresser:
I am in favor of `GND/DIE` for pin2.

PR #2353, opened 2019-12-03
Addition of OPA1692 as a variant

jamesrlucas:
Addition of OPA1692 Audio Opamps

I've added these as variants of LM2904 - however is this the right way of doing this? (some of the footprints listed under the parent are not compatible with the OPA1692.) Should I create a separate part with only the correct footprints?

Datasheet Link: https://www.ti.com/lit/ds/symlink/opa1692.pdf

image

---


chschlue:
Please add these aliases to fully specified symbols (symbols with package suffix and the correct default FP) like NCS2325D and OPA2197xDGK, respectively.

jamesrlucas:
Done :)

chschlue:
One more thing: add VFA and whatever you can think of that people might use to search for this part to the keywords. `dual` is unnecessary as that word is already in the description.
Looks good otherwise.

jamesrlucas:
Okay - I've added VFA to the keywords instead of dual, I can't think of anything else that isn't already in the description or keywords that I would search for...

chschlue:
Thanks

PR #2359, opened 2019-12-04
Added symbol for LMZ23605, LMZ23603, LMZ22005, LMZ22003.

cnieves1:
Datasheets:
  http://www.ti.com/lit/gpn/LMZ22005
  http://www.ti.com/lit/gpn/LMZ22003
  http://www.ti.com/lit/gpn/LMZ23603
  http://www.ti.com/lit/gpn/LMZ23605

imagen
imagen

imagen


cpresser:
This already looks pretty good, but some changes are required:
- [ ] This part needs a custom footprint. The exposed-pad is smaller than `TO-263-7`
- [ ] The Description should not contain to-263. If the part has a standard package (which is not the case here), the package should be the last item in the description list
- [x] ~~The exposed-pad is missing in the symbol~~ Not really, the footprint has the EP as Pin4 which is correct.
- [ ] Swap `EN` with `SYNC`. The EN pin will usually be connected to the positive supply, where SYNC in a typical application is connected toward the GND rail.
- [ ] Why did you choose open-collector for `SS/TRK`? I think it should be passive and on the left side, how about between EN and SYNC?
- [ ] Question (most likely not an issue): Do you know what the `TX` in the name means? Could it be a temperature-range indicator? If yes, then we should wildcard it.

cnieves1:
I think all issues are fixed now.
Footprint depends on: https://github.com/pointhi/kicad-footprint-generator/pull/467
imagen

I think TZ suffix in the name has something to do with the package.
If you go to:
http://www.ti.com/packaging/docs/partlookup.tsp
and look for TI part number "TZ/NOPB", only devices with NDW or NDY package are shown.

If you search for some devices:
  LMZ23605TZ/NOPB  is -40 to 85ºC
  LMZ14201TZ-ADJ/NOPB is -40 to 125ºC
  LMZ14201EXTTZ/NOPB  is -55 to 125ºC.

cpresser:
Thanks, that looks better.
- [ ] I would like you to compress the symbol more. Or is there a good reason to have the big spaces between the pins?
- [ ] The footprint-filter should not contain the pin-count. That is handled by another kicad-feature

cnieves1:
Done:
imagen

cpresser:
Thanks, that looks good. Lets wait for the footprint and then get this merged.

PR #2361, opened 2019-12-04
Addition of MAX15012 & MAX15013 Half Bridge Drivers

jamesrlucas:
Addition of MAX15012 & MAX15013 Half Bridge Drivers. 

'A' and 'C' Variants are added as aliases of HIP2100_SOIC and HIP2100_EPSOIC symbols respectively.
'B' and 'D' variants are created using HIP2100 parts as templated, the only change was to add the bar over the (inverted) low input terminal

The symbols do have different names compared to the datasheet, is it necessary to create new symbols to cover this?

Datasheet Link: https://datasheets.maximintegrated.com/en/ds/MAX15012-MAX15013B.pdf

image

---


myfreescalewebpage:
Hello @jamesrlucas thanks for contributing

- [ ] The accuracy and temperature should be discard in the name of the device. For example MAX15012BASA should be "MAX15012xxSA"
- [ ] The aliases you have done with the HIP symbol can't be made. The pin names are too different. Please make your own symbols.
- [ ] Package at the end of the description, do not write EPSOIC, just SOIC

Will do a complete review after those changes.

Joel

myfreescalewebpage:
@jamesrlucas ping

jamesrlucas:
Apologies for the delay... All done! Should I change the `VDD` / `GND` pins to the top and bottom of the symbol? (as per KLC) Or would it be better to maintain the pin position with equivalent parts already in the library?

myfreescalewebpage:
No worries, thanks for the update.
I still disagree with the name of the symbol, Maxim details are at https://www.maximintegrated.com/en/design/packaging/package-information/maxim-naming-conventions.html, see also my previous comment. The two first letters should be x, but the two last are relevant (package type and count).
Yes VDD and GND should be at the top and bottom for these devices.
Joel

chschlue:
@myfreescalewebpage In this case, the first letter affects package type and pin functions, so it has to stay if I'm not missing anything. Unfortunately, Maxim's SOIC-8 and SOIC-8-1EP are both enoded by 'SA' suffix.

jamesrlucas:
The first letter relates to the pin functions (A is non-inverting Low side driver, B is inverting, C & D are the same as A & B but with thermal pad). I've left the `x` for temperature range, and added `SA` for package type and pin count.

I've changed the pin layout to match those of the IR MOSFET Drivers (IR2011, IR2101 etc)

Going off on a tangent, is this a library that needs a bit of a review for consistency? I'm happy to redraw/sort out symbols to get them consistent & conforming to KLC. (I've got a bit of free time at work right now!)

chschlue:
Sure, but mind that breaking changes (like moved pins) can only be merged into the v6 libraries.

myfreescalewebpage:
Just read your comments, agree with the first letter, I was focused on the two last ones !
I think the PR is now ready to be merged, thanks for the job here !

PR #2365, opened 2019-12-06
Create TPD6S300A

eclecticc:
Create TI https://www.ti.com/lit/ds/symlink/tpd6s300a.pdf

image

Uses the footprint at https://github.com/KiCad/kicad-footprints/pull/1992

---


eclecticc:
Note that the NC pins are bidirectional rather than passives because TI calls them out that way in the datasheet and notes that they need to be connected to ground.

myfreescalewebpage:
Hi @eclecticc , thanks for contributing,

- [ ] Description should ends with package name: `USB Type-C Port Protector: Short-to-VBUS Overvoltage and IEC ESD Protection, WQFN-20`
- [ ] Footprint filter should be `WQFN*1EP*3x3mm*P0.4mm*` only
- [ ] NC pins should be stacked with GND because TI requires they should be connected to ground (this is the rule we have in KiCad)
- [ ] D1 and D2 should be Passive
- [ ] VBIAS has only a capacitor, can't be Power Input, I suggest should be Passive to avoid ERC errors in KiCad. Also should be moved elsewhere than the top for example below FLT
- [ ] FLT should be Open Collector

Capture

Cheers,
Joel

eclecticc:
Updated

image

eclecticc:
Footprint at https://github.com/KiCad/kicad-footprints/pull/1992

myfreescalewebpage:
@eclecticc thanks. Just need to wait the footprint now before merging the symbol.

myfreescalewebpage:
Footprint abandoned

PR #2369, opened 2019-12-06
Change HIP2100 HB pin to Passive type

evanshultz:
The HIP2100 symbols had the HB pin as Power Input type. I would argue that's an annoying choice at any rate because it requires a PWR_FLAG on that pin when there's usually a diode (and maybe a resistor too) in series between a power source and the bootstrap pin. But in this IC the bootstrap diode is on-chip and thus current doesn't even flow from external parts to the pin. So making this pin Passive is a lot easier implementation-wise.

Edit: Forgot to link the datasheet. https://www.renesas.com/us/en/www/doc/datasheet/hip2100.pdf

---


myfreescalewebpage:
@evanshultz agree with you, no comment.

PR #2372, opened 2019-12-08
Audio:SSI2164: Convert to multi-unit symbol

michd:
This PR changes the existing symbol for the SSI2164 in the Audio library to use multiple units. The symbol for each VCA unit is modeled after the symbol used in the datasheet. For the "E" unit (power, mode pin), I had to improvise a bit; I hope it's adequate.

Part datasheet: http://www.soundsemiconductor.com/downloads/ssi2164datasheet.pdf

Running checklib.py on it reports violations of:

- S3.1: symbol slightly off-center. This was deliberate, to aim the Voltage control pin on the VCA units at the center of the triangle slope. The pin ends are still on the 100mil grid.

- S4.1: The script gave no further info, but I can see myself that the pin lengths are no equal; this, again, is to accommodate the triangular shape.

Example screenshot:

ssi-2164-symbol-changes

---


evanshultz:
@michd 
Hi. This won't be accepted because it's changing the pins (and more!) on a symbol and that would break schematics using the existing symbol. We strive for library continuity and allow changes only at a major release of KiCad, like v5 to v6. (Fixes are a different story and can happen anytime).

However, we have put a `_Split` suffix on symbols which used multiple units. If you would like to retain the existing symbol as it is and add this one as `SSI2164_Split`, I or someone else can take a look at it.

Thanks for contributing!

michd:
@evanshultz Hi, thank you for your response.

After I submitted this pull request I browsed some other PRs/issues and saw mention of the same; I figured I'd keep this PR around in case it'd be useful for v6.0. I do like the `_Split` alternative though.

I'll amend my pull request tomorrow to make this one a _Split addition and leave the original one alone.

michd:
@evanshultz Hi there, I've now pushed the suggested change, adding the split version as its own symbol as to not conflict with existing schematics.

PR #2378, opened 2019-12-11
Fix CD4053 pinout (14/15 swapped)

dlharmon:
Data sheet as linked in the part http://www.ti.com/lit/ds/symlink/cd4052b.pdf

Part with same pinout and clearer data sheet: https://assets.nexperia.com/documents/data-sheet/74HC_HCT4053.pdf

## Before
Pins 14 and 15 are swapped.
Screenshot from 2019-12-11 11-47-57

## After
Swapped pins 14, 15 to match data sheet
Renamed channels from X,Y,Z to A,B,C as in the data sheet
The datasheet's "ax OR ay" is shortened to just "A"
I left the right side pins "A0 and A1" rather than the datasheet's "AX" and "AY" as I felt it was clearer how it operates that way.
Screenshot from 2019-12-11 11-59-16

---


dlharmon:
I checked CD4052 and CD4051. Those were OK. Thankfully, I caught this before sending the board out.

evanshultz:
Good catch. Thanks for noticing and submitting this!

Pins 9-11 are now A-C which is awkwardly the same as the other A-C pins. So if you're going to use the Nexperia pin names please use it for all pins.

Side note: When the logic libs are updated we will have to pick a convention for pin names because this is just one example where datasheets differ for the same part.

dlharmon:
Nexperia is very different:
Screenshot from 2019-12-11 13-14-36

Any suggestions? There's only room for a single character on 14,15,4 and I was wanting to avoid resizing the symbol and moving pins. My old self created symbol had the control lines as "AS", "BS", "CS". The TI names certainly don't fit.

evanshultz:
I was thinking to use S1-S3 for pins 9-11 so that all pins names match the Nexperia datasheet.

dlharmon:
We may be looking at different Nexperia datasheets. I'm not sure of what you are suggesting as none of the other pins match that.

This old Onsemi data sheet is what I think the original symbol was derived from:
https://www.onsemi.com/pub/Collateral/MC14051B-D.PDF

I could simply swap the numbers on pins 14 and 15. It doesn't seem that bad that A,B,C, controls X,Y,Z. The original is confusing when looking at the TI datasheet referenced in the symbol.

evanshultz:
Page 4 of the Nexperia datasheet you linked above shows this pin names for pins 9-11:
image

As long as the same pin name isn't re-used I think it will be understandable enough for now. And the main point is to get the fix into the library which you caught.

dlharmon:
Changed pins 9-11 to NXP naming.

Screenshot from 2019-12-12 11-28-12

evanshultz:
Thanks!

PR #2373, opened 2019-12-09
Added symbol for Vl53l1x ToF distance sensor

cnieves1:
Datasheet: https://www.st.com/resource/en/datasheet/vl53l1x.pdf
Footprint depends on:
  https://github.com/KiCad/kicad-footprints/pull/1995

imagen

imagen

PR #2379, opened 2019-12-11
Add Microchip ATF1502AS(V,L) and ATF1504AS(V,L)

whitequark:
Add Microchip ATF1502AS/ASL/ASV CPLDs, a true 5 V (with AS/ASL suffix) 32 macrocell CPLD in active production.

Datasheets: http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-0995-CPLD-ATF1502AS(L)-Datasheet.pdf http://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-0950-CPLD-ATF1504AS(L)-Datasheet.pdf
Screenshot:
Screenshot_20191211_203231

---


cpresser:
Hi, thanks for this addition.
- [x] Pin names are ugly: They don't have distinctive name, but that can't be helped. Then there is the issue that we don't like more than one alternate function per pin. Here that is only an issue with Pin40, and it does not clutter up the symbol to much. So I think we can keep it that way.
- [ ] Description should contain the package name, see KLC S6.3 - 1
- [ ] Please fix the Travis-CI errors: S3.3 and S3.1 are simple (the symbol will become slightly asymmetrical because of the number of pins). For S6.3 remote the data-sheet from inside the editor-window, then the `F3` line in the .lib file should have an empty string.
- [ ] More Travis-CI: The GND-Pin stack should set all but one pin to invisible, passive. The pin with the lowest number should be the visible power-input pin
- [ ] This is a new lib, it needs to be added to `sym-lib-table`. Are there other devices that will go in that same lib? If not, having a lib just for one part seems strange.

whitequark:
Thanks for the review!

> Description should contain the package name, see KLC S6.3 - 1

Done

> S3.3

Done

> S3.1 (the symbol will become slightly asymmetrical because of the number of pins)

Do I have to? It'll just become ugly for no real reason.

> For S6.3 remote the data-sheet from inside the editor-window, then the `F3` line in the .lib file should have an empty string

Done

> The GND-Pin stack should set all but one pin to invisible, passive. The pin with the lowest number should be the visible power-input pin

Done, and for VCC too

> This is a new lib, it needs to be added to `sym-lib-table`.

Done

> Are there other devices that will go in that same lib?

Yes, there are a few: ATF1502, ATF1504, ATF1508, and some of them have dual IO voltage variants as well. I think there's around a dozen unique symbols in this series, considering package and pinout variations, for three non-discontinued devices.

whitequark:
>     Component ATF1502AS has metadata errors:
>     - Missing DOCUMENTATION FILE NAME field on 'Properties > Description' tab

I have no idea what this Travis error means.

whitequark:
I made two changes:
  * I added package code for the ATF1502AS* devices, which is 44A for TQFP-44.
  * I also added ATF1504AS*-44A, which is a variant that is 100% pin compatible with ATF1502AS*.

Microchip clearly specifies the package code as 44A. However, their ordering code is weird: e.g. ATF1502AS-7**A**X**44**, where the bold parts comprise package code, and the rest is speed/temperature grade. I opted to use the package code directly because of potential for confusion between the letter X in the ordering code, and the wildcard x as recommended in KLC S2.2.

cpresser:
Thanks for the quick changes. Can you please not force push next time? That allows me to review individual commits - sometimes it is easier to just look at the files/changes directly instead of opening the graphical symbol-editor.

- [ ] Regarding the datasheet-travis-error, you need to fill this field: 
image There is a similar field in the aliases tab that also needs a data-sheet for each alias. Those end up in the `.dcm` file. My wording in my first post was obviously not good enough - sorry about that.
- [ ] The Footprint-filter should be `TQFP*10x10mm*P0.8mm*` otherwise it will also match other package-sizes.
- [ ] Centering the symbol/Reordering of pins: What do you think of reordering the pins? KLC wants them grouped by function. Perhaps in this groups (page 20 datasheet): 
image That would increase the size of the symbol slightly and give the option to center it so the KLC 3.1 error is fixed. This could be one of the few symbol-librarys with 0 travis-ci errors :)

cpresser:
> Microchip clearly specifies the package code as 44A. However, their ordering code is weird: e.g. ATF1502AS-7**A**X**44**, where the bold parts comprise package code, and the rest is speed/temperature grade. I opted to use the package code directly because of potential for confusion between the letter X in the ordering code, and the wildcard x as recommended in KLC S2.2.

I am not fully convinced that this is a wise choice. With that option the part-number does not match the MPN at all since the `A` comes after the `44`.

The correct name would be `ATF1502AS-xAx44`. I don't think its to bad to not choose it. I would rather stick to S2.2 here.

whitequark:
> Can you please not force push next time? That allows me to review individual commits - sometimes it is easier to just look at the files/changes directly instead of opening the graphical symbol-editor.

Will do.

> * What do you think of reordering the pins? KLC wants them grouped by function.

Normally, I would agree, but these CPLDs are a bit of a special case in that the JTAG pins are shared with macrocell I/O, and in fact it is reasonable and very common--many vendor HDL examples are doing it--to disable JTAG entirely and reuse the JTAG pins for logic to avoid losing what can be, for smaller packages, a significant part of device resources to JTAG. (I don't know how is one supposed to reprogram the device afterwards--probably high voltage programming. It's certainly possible but you have to sign an NDA before Microchip tells you as little as which pins are used for HVP...)

> The correct name would be `ATF1502AS-xAx44`. I don't think its to bad to not choose it. I would rather stick to S2.2 here.

The first wildcard is a decimal number, e.g. it can be `7` or `10`. Does this fall under the semanics of wildcards as defined in S2.2? If so I'll change it.

whitequark:
> * There is a similar field in the aliases tab that also needs a data-sheet for each alias. Those end up in the `.dcm` file.

Actually, I'm not quite sure I understand the purpose of this rule. The one symbol I'm adding refers to 2 or 4 unique devices through aliases; these devices have 4 different datasheets. Why am I only allowed to put one of them in the library?

cpresser:
> > * There is a similar field in the aliases tab that also needs a data-sheet for each alias. Those end up in the `.dcm` file.
> 
> Actually, I'm not quite sure I understand the purpose of this rule. The one symbol I'm adding refers to 2 or 4 unique devices through aliases; these devices have 4 different datasheets. Why am I only allowed to put one of them in the library?

I think there was/is a misunderstanding:
Each alias has its own entry in the dcm file. And thus each alias can have its own datasheet. Different devices should have different data-sheets.
The UI allows to set the data-sheet for all aliases in the properties-dialog of the parent device.

whitequark:
OK, I'm really confused now. What I did at first is I filled the datasheet in both General > Datasheet as well as for every alias, which is as I understand what you're suggesting I do. But it violated F6.3. Why?

cpresser:
> OK, I'm really confused now. What I did at first is I filled the datasheet in both General > Datasheet as well as for every alias, which is as I understand what you're suggesting I do. But it violated S6.3. Why?

Because there was also the `F3` entry in the `.lib` file set. But that should be empty.

whitequark:
> Because there was also the `F3` entry in the `.lib` file set. But that should be empty.

Sorry, but your requests are impossible to fulfill. You said:

> [ ] Regarding the datasheet-travis-error, you need to fill this field:

However, if I fill this field, then the F3 field in the `.lib` file always gets filled as well. And besides, if I do not fill this field (such that the F3 field in the `.lib` is cleared), then there would be no datasheet specified for the base device, only for the aliases, which makes no sense.

whitequark:
If I manually edit the `.lib` file to clear the F3 field while keeping the `.dcm` file as it is, then KiCAD recreates the F3 field in the `.lib` file each time I save the library. I can do it for the PR, but it is very frustrating that I had to figure this out myself, and either the KCL checker or KiCAD itself are obviously broken.

I am using KiCAD 201911250038+8bd2765~86~ubuntu18.04.1 from the js-reynaud PPA. I know those are nightly builds, but AFAICT I do not have any other options because the last time I tried non-nightlies there was a bug that made them completely unusable on my machine (something wxwidgets-related, I do not recall exactly).

whitequark:
> Regarding the datasheet-travis-error [...]

All done

> * [ ]  The Footprint-filter should be `TQFP*10x10mm*P0.8mm*` otherwise it will also match other package-sizes.

Done

> The correct name would be `ATF1502AS-xAx44`. I don't think its to bad to not choose it. I would rather stick to S2.2 here.

Done

> (I don't know how is one supposed to reprogram the device afterwards--probably high voltage programming. It's certainly possible but you have to sign an NDA before Microchip tells you as little as which pins are used for HVP...)

I am now completely certain as to how the device is reprogrammed after carefully reading the datasheets, studying proprietary programming adapter schematics, looking at measurements of the input buffer parameters, and asking a colleague to capture traces from a proprietary programmer capable of HVP. Specifically, +12 V on OE1 is used as an override to force enable JTAG, on this device and every other in the family.

Because of this, I have added an alternate function on the schematic to OE1, now marking it as "I/OE1/Vpp". I understand that this is somewhat of an edge case, this functionality not being explicitly documented by the vendor (all they say is, paraphrasing, "any of the four dedicated input pins may receive +12 V during HV programming"). However, I think there is a compelling reason to do so: an enterprising designer aiming to support in-circuit HVP (or, perhaps, field reworking a bricked device without desoldering it) would have to make sure that her circuit can withstand +12 V on that pin, and having it clearly marked on the schematic helps a lot.

I am not sure if the name "Vpp" is ideal here. The high voltage is not used for the actual EEPROM programming, only to enable JTAG.

cpresser:
I did review your changes and I like them. Thank you very much, the commit messages are well worded and precise. Yay :heart: 
Naming the pin with the additional `Vpp` does violate KLC, but it is pretty useful. So I am in favor of keeping it. As a person that worked with microchip parts in the past `Vpp` also seems like a reasonable name. What do you think of `HVP`? Better, worse?

The `HVP` is just an idea. Perhaps you like it, if not, I will go ahead and merge it as is.

---

> However, if I fill this field, then the F3 field in the `.lib` file always gets filled as well. And besides, if I do not fill this field (such that the F3 field in the `.lib` is cleared), then there would be no datasheet specified for the base device, only for the aliases, which makes no sense.

The base device also has an entry in the `.dcm`file. You can see an example here (just a single device, no alias): https://github.com/KiCad/kicad-symbols/pull/2373/files

Regarding the nightly build, I don't have any useful advice since I am using stable (as it is available for my distro via backports)
The KLC kinda says that one should use stable (G1.8). We also have seen some other issues with nightlies. If some behavior changes in the next KiCad stable we will adjust the KLC accordingly (with the usual lag).

---

That leaves your question with the number of `x` characters for the wildcarding. KLC has no exact rule for this. I did some digging in the library, looking at existing parts. The conclusion is that we already have both possible styles:
1. Single `x` for numbers with 1 or more digits
2. Double `x` for numbers that could also be a single digit

whitequark:
> What do you think of `HVP`? Better, worse?

To me it seems like `Vpp`-style pin names are reasonably common (at least in the rather small set of ICs with high voltage programming that still have relevance today) whereas I'm not sure if I've ever seen `HVP`-style names used.

> The KLC kinda says that one should use stable (G1.8). We also have seen some other issues with nightlies.

Yes. I agree this probably isn't the fault of either KLC or KiCAD, just an unfortunate combination of circumstances.

> The conclusion is that we already have both possible styles:

I am fine with having either one or two `x` per field.

PR #2381, opened 2019-12-12
Standardize testpoint-like power symbols

evanshultz:
Most of our power symbols look like this:
image

However, some (LINE/NEUT/PRI*/V*) looked like the testpoint symbol:
image

I see two problems with the VDD and VSS symbols above:
1. The power symbols aren't consistent and look like a testpoint. If I see a power symbol in the schematic, it should be clear what it is.
1. Positive and negative voltages are not differentiated.

This PR updates the testpoint-like symbols to look like the first screenshot.

Note that there is other clean up to be done here, like simplifying the symbols and standardizing text sizes, but I'll do that in another PR.

---


myfreescalewebpage:
@evanshultz totaly agree with you. No comment on the symbols you have modified, but I suggest also to standardize +3.3VADC / +3.3VDAC / +3.3VP / +VDC / -VDC / HT / VDC. What do you think ?

evanshultz:
Thanks for looking, Joel.

+VDC,  -VDC, and HT are bigger and I'm a little hesitant to make changes right now to them. Even if they are going to be smaller and don't affect the pins. Maybe @poeschlr would like to comment?

I'm happy to update the others but I don't know why they're different. Do you (or anybody else) have any idea if the uniqueness of these symbols has a special meaning that should be retained?

myfreescalewebpage:
> +VDC, -VDC, and HT are bigger and I'm a little hesitant to make changes right now to them

Yep, sure. The update you have done is almost very nice, it's still possible to decide later for the other.

> Do you (or anybody else) have any idea if the uniqueness of these symbols has a special meaning that should be retained?

No I don't, I never use these specific symbols.

evanshultz:
@myfreescalewebpage 
No comments in a couple days so let's go. I standardized the ones that were only slightly different than the common symbol, and left alone the three big symbols we discussed above. OK?

myfreescalewebpage:
@evanshultz current status is convenient I think. Merging.

PR #2383, opened 2019-12-13
Updated jumper FP filters

evanshultz:
Both solder jumpers and non-solder jumpers had FP filters for solder jumpers. I removed the solder jumper FP filter for the latter and added a non-solder jumper FP filter (of which we have none right now) and also a FP filter for testpoints with multiple pads which would work for jumpers.

We could add jumper footprints the same or similar to the testpoint footprints we have now, but I wasn't sure if that would be accepted having the same footprints twice.

---


myfreescalewebpage:
@evanshultz no comment, merging.

PR #2385, opened 2019-12-14
Add TestPoint_Small

0xdec:
Added a really small testpoint that's super useful for compact schematics, or just sprinkling on pins that would normally be NC.
TestPoint_Small

An example of the symbol being used, once on top of a wire junction and once not:

---

### DIFF
--- a/Connector.dcm
+++ b/Connector.dcm
@@ -1698,6 +1698,12 @@ K test point tp
 F ~
 $ENDCMP
 #
+$CMP TestPoint_Small
+D test point
+K test point tp
+F ~
+$ENDCMP
+#
 $CMP UEXT_Host
 D Universal EXTension (UEXT) is a connector layout which includes power and three serial buses: Asynchronous, I2C, and SPI
 K UEXT, SPI, UART, I2C
--- a/Connector.lib
+++ b/Connector.lib
@@ -15373,6 +15373,23 @@ X 1 1 0 0 0 U 50 50 1 1 P
 ENDDRAW
 ENDDEF
 #
+# TestPoint_Small
+#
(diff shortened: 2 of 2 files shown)


myfreescalewebpage:
@0xdec thanks for this symbol, no comment to do, merging.

evanshultz:
@poeschlr 
I question the KLC compliance of this symbol. It appears to violate https://kicad-pcb.org/libraries/klc/S3.5/ and https://kicad-pcb.org/libraries/klc/S4.1/ because the pin is inside the symbol and the pin length is 0mil (S4.1 says shorter pins are allowed but it doesn't explicitly allow zero-length pins, though other `TestPoint*` symbols use 0mil pin lengths). Do you have input on this?

poeschlr:
Well the normal test point symbol already has a very short (possibly the shortest viable) pin length. So if one needs a smaller one then this might be an alternative as it can then be placed directly "on" a wire. It might however be a dangerous thing as it might not work nice with junction dots (I am not sure if placing this symbol in the middle of a wire already creates a junction dot which in turn might then cover up the symbol. Especially in printouts). Meaning i am not sure if it is a good idea to have this symbol in the official distribution.

evanshultz:
@poeschlr 
You can see in the image above that this symbol has a hollow center and the junction dot appears in the middle. So a green dot with a red circle around it.

As the head librarian, I'm not sure who else would be able to determine if this part can be in the library but you. And provide guidance if that's because this symbol is in violation of KLC or KLC needs updated.

0xdec:
@poeschlr I see your point about the junction dot in printouts. I made it intentionally sit outside the junction dot, but looking at the screenshot above it's a pretty skinny red ring. I could make the ring larger and thicker if you think that might be better.

PR #2386, opened 2019-12-16
Diode_Bridge: Add aliases of B40R: RB151..RB157

piotr-wiszowaty:
Diode_Bridge: Add aliases of B40R: RB151..RB157

Datasheet: https://www.rectron.com/public/product_datasheets/rb151-rb157.pdf

RB151_157

---


evanshultz:
Thanks! The physical package and the pinout match so it should be a valid ALIAS. The voltage value shown doesn't seem to be consistent in this library so let's go with it.

PR #2390, opened 2019-12-17
Adding the C8051F380/1/2/3/4/5/6/7/C family (QFN32)

JonathSpirit:
Adding the silicon-labs "Full Speed USB Flash MCU Family" (C8051F380/1/2/3/4/5/6/7/C) in QFN32

(not sure about the footprint :/ )

https://www.silabs.com/documents/public/data-sheets/C8051F38x.pdf

image

image

---


myfreescalewebpage:
Hi @JonathSpirit , thanks for contributing,

- [ ] Package at the end of the description has not the right format. Should be "QFN-32"
- [ ] Footprint is to be completed
- [ ] Footprint filter is not correct (see other devices in the lib with the same footprint and also travis output)
- [ ] Pin length should be 100mil (pin name have only 2 chars)

Else no comment to do.

Cheers,
Joel

JonathSpirit:
Hi thanks for the review,
I applied all the changes :
image

image

image

myfreescalewebpage:
Thanks.
- [ ] The packages at the ends of aliases description are still not correct
- [ ] Footprint filter should be `QFN*1EP*5x5mm*P0.5mm*`

JonathSpirit:
oups didn't see the alias description, and should i put : 

> QFN\*1EP\*5x5mm\*P0.5mm\*3.3x3.3mm\*

for the filter ?

myfreescalewebpage:
No it's `QFN*1EP*5x5mm*P0.5mm*`, we never indicate the EP size in the footprint filter. If you found some devices where this rules is not correct this is because they are old and should be updated.

JonathSpirit:
Thanks for the answer, I applied the change

myfreescalewebpage:

PR #2393, opened 2019-12-20
add Hamamatsu S597x photodiode in TO-18 THT

aewallin:
S597x series photodiodes, in TO-18 package.
https://www.hamamatsu.com/resources/pdf/ssd/s5971_etc_kpin1025e.pdf

- The CASE pin is abbreviated "CS"
- There are other TO-18 footprints called LaserDiode_ which might also work, but here I suggest the generic TO-18 footprint.
- An alternative to naming it S597x would be to add one part S5971 and aliases S5972 S5973 (but these are all with identical pins and footprints)

S597x

---


myfreescalewebpage:
Hi @aewallin, I also check this one now :)

- [ ] Will be better to have "S5971" + some 2 aliases particularly because there are several packages options. Indicate in the description the difference between the devices.
- [ ] Package at the end of the description: "TO-18-3"
- [ ] Footprint filter should be `TO?18*`
- [ ] Name of pin 1 should be K and name of pin 2 should be C probably
- [ ] You can make pin 1 and pin 3 names invisible has done for the diodes, the symbol is enough clear for the user

Cheers,
Joel

aewallin:
I don't know how to make only two pin-names invisible - so now they are all invisible. Otherwise I tried to do the changes as suggested.

myfreescalewebpage:
Haven't think that it was not possible to hide only a part of pin names, but the result is nice.
Just the comment about the footprint filter to be fixed: no need of the leading '*'

myfreescalewebpage:
Thanks, merging.

PR #2397, opened 2019-12-22
Rpi 40pin run tv

jstjst:
Remaining RPi Zero symbols, related issue: #1772
grafik


myfreescalewebpage:
@jstjst thanks for theses symbols. No specific remarks except one: Can you cross pin number of TV and Run ? Run was pin 41 on previous symbols and its good to have the same logic.

jstjst:
@myfreescalewebpage I am not sure what you mean. Run is pin 41 on all symbols, expect the symbols without Run that have only TV.

myfreescalewebpage:
Sorry you're right, I was tired !!

cpresser:
This PR has pending footprint, but no link to the footprint PR. Am I to stupid to see it?
If not @jstjst can you please add that link?

myfreescalewebpage:
@cpresser the author is working on it as far as I know, and the symbol still wait for the link

PR #2398, opened 2019-12-22
Add AS5047D Magnetic Position Sensor

einball:
This commit adds the AMS AS5047D magnetic position sensor. The footprint is SSOP14 which is already used by other AMS sensors.

Bildschirmfoto vom 2019-12-22 13-21-02

I violated KLC S4.7 (Naming convention on LOW active Pins) and named the SPI Chip Select pin CSn due to consistency with other AMS magnetic sensors with an SPI interface. 

[Link to datasheet](https://ams.com/documents/20143/36005/AS5047D_DS000394_2-00.pdf)
---


aewallin:
pin14 name is I/PWM in the datasheet

pin12 seems to be power output

if  TEST is never used for anything and always connected to GND, then it can be stacked with GND.

einball:
Thanks for the thorough review!

I have corrected pin 14 (I/PWM) and stacked pin 5 (Test) with GND, since it actually is always connected to GND.

As for pin 12:

- This pin breaks out the internal LDO, so it is a power output for 5V input.
- The datasheet also says that for 3V3 operation, this pin (and pin11, VDD5V) is used as a supply pin.
- The library already features the AS5048A/B and AS5045B symbols which list VDD3V3 as power input

To keep the library consistent I did not change the pin type.

einball:
After reading KLC S4.3 carefully, I unstacked the Test pin from GND, since they do not share the same name or pintype.

myfreescalewebpage:
Hi @einball , thanks for contributing,

- [ ] The package at the end of the description is not correct (should be TSSOP, not SSOP)
- [ ] Name of pin 5 is TEST and this pin should not be on the bottom of the symbol. I suggest to move it on the left.
- [ ] Name of pin 11 is VDD
- [ ] Name of pin 12 is VDD3V3. This pin is Power Output.

Cheers,
Joel

einball:
Hello @myfreescalewebpage , thanks for reviewing it :+1: 

I named these pins according to the other devices to keep consistency between the different sensors - They all had the TST pins on the bottom.

The recent commit includes all changes you requested and moves VDD3V3 to the left side of the device.

myfreescalewebpage:

PR #2400, opened 2019-12-22
Add STM32H750 aliases

florolf:
Add aliases from some of the STM32H753 parts to their corresponding
STM32H750 parts. Equivalence was established on the basis of CubeMX
data, which currently only covers a subset of the announced STM32H750
parts.


myfreescalewebpage:
@florolf thanks, no comment to do, merging.

PR #2401, opened 2019-12-23
SST26VFxxxB added.

serkanxselcuk:
SST26VFxxxB added.
Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/SST26VF064B-SST26VF064BA-2.5V-3.0V-64-Mbit-Serial-Quad-IO-Flash-Memory-Data-Sheet-DS20005119J.pdf
image

image
image

Footprint PRs:
https://github.com/KiCad/kicad-footprints/pull/2018
https://github.com/KiCad/kicad-footprints/pull/2017
Existing Footprints:
SOIC-16 and SOIJ-8

---


serkanxselcuk:
image
**KLC S4.2**
Pins 6,5,2,3 grouped right side as similar function
Pins 1,7 placed to left side as inputs

image
**Exception to KLC S4.2**
Pin1 could be placed to right side, but it would not be pin compatible with other symbols.

cpresser:
Hi,
thanks for your contribution.
Can you elaborate why you did distribute the pins the way you did? I am asking because we already have a few different layouts for QSPI chips in the library, this PR would add another one.

image
image

serkanxselcuk:
> Hi,
> thanks for your contribution.
> Can you elaborate why you did distribute the pins the way you did? I am asking because we already have a few different layouts for QSPI chips in the library, this PR would add another one.

Hi @cpresser ,
There was no KLC or common usage for QSPI as I know. I am not sure, if it is the best solution to place CLK pin at left side, just because it is an input.

I find it better to place all pins for a bus system as neighbor pins. This why I preferred to set all SPI pins at one side. The other pin which is not used for SPI but QSPI (SIO3) at other side to get symbol smaller.

Actually I would prefer to set SIO0-3, CLK and ~CE~ pins at left side, so all pins with QSPI or SPI bus wires can be placed together. But it would cost of size in Y direction. (Left side because it is a slave device)

image

I understand KLC input at left, output at right for motor controller (microcontroller communication at left, motor outputs at right) or dc-dc converter (Vin at left, Vout at right) as like adc (analog input at left, digital output at right). But not forcing all pins defined as input at left, output at right.

cpresser:
Hi @serkanxselcuk,
thanks for your detailed answer.

> As like we cannot separate RX/TX in UART or SCL/SDA in I²C; I would place SPI and QSPI pins together at one side. Because of that, I would prefer option 2.

Yes. KLC explicitly states that pins with similar function should be grouped. All of those pins are QSPI, so they can be on one side.
It introduces a new style, but I think it is superior compared to the existing ones. So lets continue with that.

As a side-note, I think the existing spi flash memory symbols should be reworked to conform to s
the same standard.

serkanxselcuk:
Hi @cpresser,
I grouped QSPI pins, like we spoken before.
image

cpresser:
@serkanxselcuk Sorry I am really busy with other stuff right now. Our Hackerspace is moving within the next two weeks.

cpresser:
This took really long, thanks for your patience.

- [ ] PinNumbers: The `SO` part has VSS on Pin 10.
- [ ] PinTypes: `SIO3` should be bidir
- [x] PinNames the datasheet has some places where 'HOLD' is not shown as inverted/active-low. but according to Figure 4-3 it is inverted.
- [x] Pin Positions
- [ ] Naming: The operating frequency and temperature-variant are missing from the name. Please add `-xxxx` to the names before the package-suffix. Both need to be wildcarded. And we can't omit this part because it is not at the end of the name (KLC S2.2)

Related Footprint-PR:
https://github.com/KiCad/kicad-footprints/pull/2018

cpresser:
Hi @serkanxselcuk are you still active with this PR? If not, I would like to mark this and the related footprint PR as abandoned.

myfreescalewebpage:
@cpresser fp has been merged for this symbol

myfreescalewebpage:
@serkanxselcuk seems the fp need to be updated in the symbol following run of travis

PR #2405, opened 2019-12-27
Add OPA858, OPA859, and OPA855 high-speed opamps, WSON-8

aewallin:
http://www.ti.com/lit/ds/symlink/opa858.pdf
and
http://www.ti.com/lit/ds/symlink/opa855.pdf

The following might be discussed:
* The ~PD is now pointing up, but I noticed some other op-amp symbols have a ~DIS pin and it is pointing down. I think having it close to V+ (where it is connected, for normal operation) makes sense.
* The datasheet describes the FB-pin as Input, so I have made it Input-type
* The PAD pin is stacked with V- (datasheet suggests V- is the only way to connect PAD)
* parts are simply named OPA858 and OPA855, since they apparently come only in one package. There are full names at the back of the datasheets, such as OPA855IDSGR and OPA858IDSGT which would be an alternative.

OPA858

---


Amplifier_Operational.dcm
```diff
@@ -1356,6 +1356,18 @@ K single opamp wideband low-power
 F http://www.ti.com/lit/ds/symlink/opa847.pdf
 $ENDCMP
 #
+$CMP OPA855
+D 8-GHz Gain Bandwidth Product, Gain of 7-V/V Stable, Bipolar Input Amplifier, WSON-8
+K opamp, bipolar
```
chschlue: Please remove commas from keywords, add VFA.
Bipolar is not needed as it is already in the description.

Amplifier_Operational.dcm
```diff
@@ -1356,6 +1356,18 @@ K single opamp wideband low-power
...
+F http://www.ti.com/lit/ds/symlink/opa855.pdf
+$ENDCMP
+#
+$CMP OPA858
+D 5.5-GHz Gain Bandwidth Product, Gain of 7 V/V Stable, FET Input Amplifier, WSON-8
+K opamp, fet
```
chschlue: Same as above

Amplifier_Operational.dcm
```diff
@@ -1356,6 +1356,18 @@ K single opamp wideband low-power
 F http://www.ti.com/lit/ds/symlink/opa847.pdf
 $ENDCMP
 #
+$CMP OPA855
```
chschlue: OPA855xDSG

Name should contain package suffix, but not temp rating, packaging information and such.

Amplifier_Operational.dcm
```diff
@@ -1356,6 +1356,18 @@ K single opamp wideband low-power
...
+D 8-GHz Gain Bandwidth Product, Gain of 7-V/V Stable, Bipolar Input Amplifier, WSON-8
+K opamp, bipolar
+F http://www.ti.com/lit/ds/symlink/opa855.pdf
+$ENDCMP
+#
+$CMP OPA858
```
chschlue: OPA858xDSG

Amplifier_Operational.lib
```diff
@@ -1955,6 +1955,31 @@ X Flag 8 0 -300 200 U 50 50 1 1 O
...
+# OPA858
+#
+DEF OPA858 U 0 5 Y Y 1 F N
+F0 "U" 150 250 50 H V L CNN
+F1 "OPA858" 150 150 50 H V L CNN
+F2 "Package_SON:WSON-8-1EP_2x2mm_P0.5mm_EP0.9x1.6mm" -100 600 50 H I C CNN
+F3 "" 400 400 50 H I C CNN
+ALIAS OPA855
+$FPLIST
+ WSON*P0.5*
```
chschlue: WSON\*1EP\*2x2mm\*P0.5mm\*

Amplifier_Operational.lib
```diff
@@ -1955,6 +1955,31 @@ X Flag 8 0 -300 200 U 50 50 1 1 O
...
+$ENDFPLIST
+DRAW
+P 4 0 1 10 -200 200 200 0 -200 -200 -200 200 f
+X V+ 7 -100 300 150 D 50 50 0 0 W
+X ~PD 8 0 300 200 D 50 50 0 0 I
+X PAD 9 -100 -300 150 U 50 50 0 0 P N
+X NC 2 -200 -200 150 U 50 50 0 1 N N
+X ~ 6 300 0 100 L 50 50 0 1 O
+X FB 1 -300 0 100 R 50 50 1 1 I
```
chschlue: FB pin cannot be an input, even if the datasheet says so.

Amplifier_Operational.lib
```diff
@@ -1955,6 +1955,31 @@ X Flag 8 0 -300 200 U 50 50 1 1 O
...
+F3 "" 400 400 50 H I C CNN
+ALIAS OPA855
+$FPLIST
+ WSON*P0.5*
+$ENDFPLIST
+DRAW
+P 4 0 1 10 -200 200 200 0 -200 -200 -200 200 f
+X V+ 7 -100 300 150 D 50 50 0 0 W
+X ~PD 8 0 300 200 D 50 50 0 0 I
+X PAD 9 -100 -300 150 U 50 50 0 0 P N
```
chschlue: Change name to V- (same as pin 5)

aewallin:
* changes made as suggested
* FB pin type is now Passive
* added also new alias OPA859

chschlue:
Thanks.
Am I missing something? Do you see a particular reason why FB shouldn't be an output?

aewallin:
* I tried looking for other opamps (analog AD80xx or ADAx) with FB-pins, but did not find examples in the current kicad libs.
* for a clean looking schematic drawing one could stack FB with OUT? (but this hides the fact that high-speed parts use a special FB pin...)

chschlue:
Having a separate net for FB leads to uglier schematics but nicer designs.
I don't think stacking is a good idea, but neither is having it between inputs. How about opposite ~PD?

aewallin:
I would suggest FB as a parallel output to OUT:
this makes the function clear (?) - and the schematic is drawn as usual with feedback-resistors etc. connecting close to the output-tip of the triangle.
OPA858xDSG

chschlue:
Seems reasonable to me. The line leading to pin 1 could be thinner.
@evanshultz Any feedback from the triangle department?

aewallin:
line width changed. I don't know of any more pending changes that I could do.
OPA858xDSG

evanshultz:
The symbol looks nice to me. A separate pin is definitely the right way to go since you don't want a net between pins 1 and 6.

`OPA859xDSG`'s description doesn't match the others with the information given or the use of commas. However, that actually matches the datasheet description so I'm OK with it.

@chschlue 
I just looked at the datasheets and the images above, but it looks good. If you checked everything else in depth then it's ready to merge. :)

aewallin:
thanks!

PR #2407, opened 2019-12-28
DRV8662: fixed pin length and stacked pins to avoid ERC error

Shackmeister:
Datasheet: http://www.ti.com/product/DRV8662

All pins changed to 100mil length
GND (Pin 4,5,6) was stacked.
SW (Pin 7 & 8) was stacked.
VBST (Pin 10 & 11) was stacked, due to the 2 Power outputs, creating ERC error.
Moved NC (Pin 9) on to the outline (according to KLC)

Skærmbillede fra 2019-12-28 02-43-16

---


cpresser:
Old vs. New (shitty renderer)
image

- [ ] I would suggest to put `Vpump`and `Rext` to the bottom of the left side since both of them are connected to GND via passives.
- [ ] Move to the `Driver_Haptic` library?

Shackmeister:
@cpresser Thanks for the review.
Moving the pins around definitely makes sense, I will try to make a better layout. Originally I didn't want to touch to much, since my main focus was to avoid the ERC error.

As far as I'm aware "Haptic" means something with touch. I dont believe a buzzer driver fits there. I agree motor driver isn't the best either. Perhaps we should have a Driver_Buzzer or Sound library?

Shackmeister:
@cpresser Moved pins around, and moved to Driver_Haptic library (TI calls it Piezo Haptic Driver in datasheet)
Sorry for colorscheme

Skærmbillede fra 2020-01-07 21-36-50

cpresser:
> Sorry for colorscheme

/me uses a similar one. It has better contrast.

Pins look good.

One more thing I found, the footprint-filter has `*N20*` in it. Afaik that's the pin count, which should not be included in the filter.

evanshultz:
Why would we change the pin locations, breaking existing schematics? Could the package outline be swelled to make the 100mil pins in the same place, or else not change pin length? @poeschlr has been clear not to make breaking changes in the middle of a KiCad major release, unless there's a bug, and it seems like resolving the ERC error can be done without any breaking changes. Stacking pins should still be safe if one of the existing pin locations is still present.

Shackmeister:
Alrighty, I will keep it as close to the original symbol as possible.
I also added EP (Pin 21)  to the GND pin stack, In the layout example it's shown connected to GND.

Skærmbillede fra 2020-01-10 01-12-13

evanshultz:
All the pin termination is in the same place now so I'm happy with that. No more input from me. Sorry for the interruption, @cpresser .

Shackmeister:
@cpresser 
I'm not sure we can do very much more without breaking compatibility, perhaps I will do round 2 sometime when we can break designs again

cpresser:
> All the pin termination is in the same place now so I'm happy with that. No more input from me. Sorry for the interruption, @cpresser .

You are very welcome to help. Useful input is always appreciated.

@Shackmeister 
Here is a new render, looks okay to me. Travis is green as well.
pr2407

PR #2415, opened 2019-12-31
power_protection: Add NUP4114 ESD diode array

plaes:
Adds NUP4114 ESD diode array symbol to the **Power Protection** library

nup4114

- Datasheet: https://www.onsemi.com/pub/Collateral/NUP4114-D.PDF

I'm not really 100% sure with the footprint matches and these are currently written out like this:
```
 SOT?563*
 SC?88*
 TSOP?6*
```


evanshultz:
Hi and welcome!

1. Please put the ref des and value both on the right side of the symbol and right-side up.
2. I'm not sure if this symbol veers into the overly complex clause of https://kicad-pcb.org/libraries/klc/S3.4/, but many of these types of parts have complex symbols. So OK.
3. Change the description to `ESD Protection Diode, ` and then add the package at the end. You can also add more info before the comma if you like, such as `USB 2.0`.
4. That package will be specific to each package instead of having one symbol for all packages. Which package(s) do you want to add? And also, do you know the difference between the two SC-88 package? Everything looks identical to me.

evanshultz:
Also, please don't force push so it's easier to review. We will squash merge.

evanshultz:
@plaes 
Are you still around and willing to implement the requests I made above?

PR #2418, opened 2020-01-02
Amplifier_Current: Add ZXCT1010

charliebruce:
This part is quite similar to other parts in the ZXCTxxxx series. As described in datasheet:

> It is an enhanced version of the ZXCT1009 offering reduced typical output offset and improved accuracy at low sense voltage. 

This was simply created by cloning and modifying the symbol for ZXCT1110. The violation of rule S3.6 (Pin name position offset) seems to have been present on ZXCT1110 too (see below).

```
schlib % python3 checklib.py ../../kicad-symbols/Amplifier_Current.lib -c ZXCT1110
Checking symbol 'ZXCT1110':
  Violating S3.6
schlib % python3 checklib.py ../../kicad-symbols/Amplifier_Current.lib -c ZXCT1010
Checking symbol 'ZXCT1010':
  Violating S3.6
```

It looks like this has been done intentionally on the original symbol, to ensure that the pin name is closer to the pin than the other edge of the symbol.

Alternatively, would renaming the ZXCT1110 to add a wildcard (ZXCT1x10) be appropriate here - is a difference in output offset and accuracy big enough to be considered a functional change? Or would it be better to just close this PR unmerged? Users can override the name of a ZXCT1110 in a specific schematic if they require tighter-specification parts on their project - is that the usual way of working in KiCad, or would users expect both variants to be available? I'm not sure what the precedent is.

Datasheet: https://www.diodes.com/assets/Datasheets/ZXCT1010.pdf

Screenshot 2020-01-02 at 04 54 58


evanshultz:
Hi and thanks for contributing!

Because the package is different, yes it will need a new symbol.

I also believe pin name offset is 10mil for aesthetic reasons. That's fine to keep.

We wouldn't want to change an existing symbol name, since that would break schematics using that symbol. I would have accepted a wildcard if this was a new submission, but even then adding a new symbol is a bit more clear. So adding a new symbol is fine.

Everything seems in order to I'll merge. Thank you!

evanshultz:
The ALIAS feature allows a symbol to be shared with a unique symbol name and description. This would actually be the way to handle this. My mistake. I guess I haven't gotten my brain working in 2020 yet. Have you seen the ALIAS feature? Would you be willing to make another PR to make ZXCT1010 an ALIAS of ZXCT1110?

PR #2031, opened 2019-07-27
Split FT232H into FT2232HL/FT2232HQ

twam:
* Splits FT232H into FT2232HL (LQFP-64, basically previous FT2232H) and FT2232HQ
  (QFN-64 with exposed pad)
* Fixes KLC issues

Datasheet: https://www.ftdichip.com/Support/Documents/DataSheets/ICs/DS_FT2232H.pdf
image
image

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1754

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #1999, opened 2019-07-22
Small update of FTDI symbols

jneiva08:
Have made small change to FTDI symols, the changes are:
- Replace datasheet links of webpage to pdf
- Fix S3.6
- Moved Package name to be more easy to read

The FT245BM is absolete and it is replaced by FT245BL

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #2420, opened 2020-01-03
Split FT232H into FT2232HL/FT2232HQ

twam:
This is the part of #2031 which was not done by #1999.

* Splits FT232H into FT2232HL (LQFP-64, basically previous FT2232H) and FT2232HQ
  (QFN-64 with exposed pad)

Datasheet: https://www.ftdichip.com/Support/Documents/DataSheets/ICs/DS_FT2232H.pdf

FT2232HL:
image

FT2232HQ:
image

---


evanshultz:
This PR removes the symbol `FT2232H`, which we can't allow because it will break that symbol if it's used in existing designs. That is why this wasn't done at https://github.com/KiCad/kicad-symbols/pull/1999.

evanshultz:
I didn't see your comment at https://github.com/KiCad/kicad-symbols/pull/1999#issuecomment-570571089 first so I didn't realize the reason to change. Good catch. There is a mistake. Let me look at this again. Normally we don't allow symbol changes between major releases, but since this is fixing a bug (CvPcb can't match to the QFN or packages) I'll go for it.

evanshultz:
Page 57 of the datasheet shows that the thermal pad must be grounded and all other pins are identical. Looks good. Thanks! Sorry again for not catching the intent of this PR the first time I saw it.

twam:
That was quick! Big thanks.

PR #2422, opened 2020-01-03
Added symbol for the L293E quad push-pull DC motor driver with extern…

farmergaetano:
Added symbol for the L293E quad push-pull DC motor driver with external sense resistor connection.

Please refer to [ST Microelectronics product page](https://www.st.com/en/motor-drivers/l293e.html) and the [datasheet](https://www.st.com/resource/en/datasheet/l293e.pdf).

This is my first pull request with Git so feedback welcome!


aewallin:
* a screenshot is helpful
* add the datasheet URL
* add some keywords
* add a description (e.g. from datasheet "PUSH-PULL FOUR CHANNEL DRIVER" (but not all caps)), with the footprint at the end e.g. DIP-20
* I don't understand why existing symbols in the library have multiple GND-pins - shouldn't they be stacked? (see L293 and L298P)

farmergaetano:
Here is the screenshot, as requested.

L293E-screenshot

I have stacked the ground pins as per the example at [this Electronics Stack Exchange answer](https://electronics.stackexchange.com/a/138827), which came from the Kicad forum.

poeschlr:
The size of this symbol can be reduced quite a bit. Or is there a special reason for having the pins this far apart?

@aewallin Stacked ground pins are the norm in the official lib so unless there is a good reason not to stack them then they should be stacked. See https://kicad-pcb.org/libraries/klc/S4.3/

aewallin:
looking at the KLC errors:
* the stacked and hidden GND pins should be type Passive (S4.3.8)
* the footprint filter should be something like DIP*W7.62mm* (see S5.2.2)

myfreescalewebpage:
Hi @farmergaetano , thanks for contributing,

- [ ] Add a space after the comma and before the package in the description
- [ ] Use capital letters for pin names
- [ ] You can remove the underscore in the pin names
- [ ] Name of pin 1 should be "EN12"
- [ ] Name of pin 11 should be "EN34"
- [ ] SENSEx pins should be Passive
- [ ] VSS should be on the top
- [ ] The position of the pin can be arranged to have something easier to wire on schematic, what do you think of this:

Capture

Cheers,
Joel

farmergaetano:
I've updated based upon your suggestions. The reason for the pin placements is to allow readability of the schematic for a typical application. This is shown in the attached screenshot, where I have utilised a symbol with the latest changes.

Screen Shot 2020-04-29 at 10 28 29 am

myfreescalewebpage:
The schematic clearly helps to decide. Thanks for sharing this! I agree it's nice.
No more comment, merging.

PR #1585, opened 2019-02-27
Add SKY13575-639LF.

matthuszagh:
http://www.skyworksinc.com/uploads/documents/SKY13575_639LF_203270D.pdf

image

Footprint PR: https://github.com/KiCad/kicad-footprints/pull/1443

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #2423, opened 2020-01-05
Correct FP for SKY13575-639LF (point to manufacturer specific)

poeschlr:
Problem discovered in https://github.com/KiCad/kicad-footprints/pull/1913
Original symbol pull request https://github.com/KiCad/kicad-symbols/pull/1585

---

I decided in the footprint pull request https://github.com/KiCad/kicad-footprints/pull/1443 that this part requires a specialized footprint that is not really doable with the script so one was added but the symbol still pointed to the script generated one (that should have never been added)


evanshultz:
@poeschlr 
Do you want a generic QFN footprint to be matched by the FP filter? If a specific footprint is in the library, why do we need the generic footprint that doesn't match the datasheet?

poeschlr:
good point

evanshultz:
Thanks!

PR #2429, opened 2020-01-07
Make 555 timers fully specified

poeschlr:
- Split the original symbol into one per package
- Renamed symbols (and aliases) to include the package suffix
- Cleaned up documentation and keywords

**This is a design breaking change** I am however quite sure it is worth it. Especially as some of the old aliases already had the package suffix there but were assigned to a symbol that was generic. Plus the footprint filters used in that symbol did not fit all aliases.

Summary of the symbols we had in some form in the lib with their packages and relevant part numbers plus datasheets.
https://docs.google.com/spreadsheets/d/1VMY6TS7239a-nVVFeL2KGmgWQvnMle-uqJrocWEhT4M/edit?usp=sharing


poeschlr:
@evanshultz fixed what needed fixing and answered the remaining questions below:

---

>Usually I recall seeing the package show DIP and not PDIP, but that's fine.

I used PDIP because this is what is used in the datasheets (and it also explains why most of them use the suffix P)

> Missing , 555 compatible in LMC555xTP and TLC555xPW descriptions.

These two are in unusual packages (for 555) so i did not add the compatibility text.

> I believe NA555, SA555, and SE555 are high-temp or specially-qualified variants of NE555 so should they be removed or wildcarded not kept as individual ALIASes?

Not all of them are available in the same packages so it would get quite complex to wildcard them.

> TLC555xPW says VSSOP in the description but the datasheet calls it TSSOP. While we seem to only have a fitting footprint named VSSOP, I don't see why the description can't match the datasheet package name and if I was looking at the datasheet and the description it would be nice to match.

Are you sure? It is connected to Package_SO:TSSOP-14_4.4x5mm_P0.65mm and uses TSSOP in the description.

evanshultz:
Either it looks good or I got confused (the last one). Thanks!

PR #2430, opened 2020-01-08
Added MCP65R4x comparators

kajusK:
Added MCP65R4x comparators in push-pull and open-drain variants
Datasheet - http://ww1.microchip.com/downloads/en/DeviceDoc/22269B.pdf

Rule violation:
------------------
  Violating S3.6
    Pin name position offset
    Pin offset outside allowed range
     - Pin offset (5) should not be below 20mils
------------------
Copied from other comparator symbols to match the symbol dimensions

------------------
checking ruleS4.4
  Violating S4.4
     - Pin Vref (5) @ (0,-200) is type OUTPUT : suggested INPUT
------------------
Vref is output of the reference voltage from the IC
2020-01-08-174747_1920x1080_scrot


myfreescalewebpage:
Hi @kajusK , thanks for contributing,

For both symbols:
- [ ] Vref should be Power Output and I suggest to have it on the top.
- [ ] Vref pin length should be 200mil
- [ ] The description should be terminated by the package name: `, SOT-23-6` at the end

MCP65R41
- [ ] Footprint filter should be `SOT?23*`

Cheers,
Joel

kajusK:
Hi @myfreescalewebpage,

I've fixed the issues you found, thank for for the review.

Jakub

myfreescalewebpage:

PR #2433, opened 2020-01-09
Fix LM319 'ground' pin type

evanshultz:
I discovered that LM319's `Gnd` pins were of type Open Emitter. While this isn't totally wrong, it causes an ERC error when connected to ground. That is quite a common use case. LM311, which has a similar output stage, uses Passive type for it's corresponding pin so I switched LM319 pins to that same pin type.

Marking as a bug since this pin types generated a bogus ERC error when the part is used properly in a common way.

Also, I updated the LM311 datasheet link to one that better match the pin names in our symbol.

---


diegoherranz:
No issues seen. Merging.
Thanks!

PR #2434, opened 2020-01-09
Fix UC3842 Vref pin type and name

evanshultz:
https://www.st.com/resource/en/datasheet/uc3843b.pdf

The Vref pin was set as Power Input type, when it's really a regulator output that can supply substantial current. I changed it to the proper type, Power Output, so that it can supply other ICs or circuits directly without a PWR_FLAG on the net.

Marking as a bug since this pin will give a bogus ERC error unless it's connected to a Power Output pin, which just isn't correct.

I also renamed the pin. It was REF, but the datasheet shows VREF. Must have just been an oversight on my part when I originally submitted the symbols.

---


diegoherranz:
Thanks for fixing. Looks OK. Merging.

PR #2441, opened 2020-01-12
Add Pericon PI6C5946002 1:2 6GHz/12Gbps CML Clock/Data Buffer

whitequark:
Add Pericon PI6C5946002 1:2 6GHz/12Gbps CML clock/data fanout buffer.

I've added a new library for this chip, called "Buffer", because I could not find anything that is appropriate for it. The closest thing we currently have is Amplifier_Buffer, but this IC is not an amplifier because it does not have gain. The category of clock/data buffers is very broad (simple buffers, fanout buffers, zero-delay buffers, ...) and contains many ICs, so I think it is appropriate to create a new library for it, even if for now it'll just be this one chip.

Datasheet: https://www.diodes.com/assets/Datasheets/PI6C5946002.pdf
Screenshot:
Screenshot_20200112_200906
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/2049

I'm not sure about the VREF-AC pin. It is technically a power output, see the datasheet below:
Screenshot_20200112_200941
But I don't know if it is appropriate to use "power output" for a pin that is essentially a special purpse voltage reference.

The VREF-AC pin is located on the left side in spite of being an output because in all valid uses of the IC it will either float or be connected to VTH (and decoupled), and having it on the right seems really awkward.

---


aewallin:
if VTH and VREF-AC are used together, should they be closer together/adjacent?

needs new footprint with 1.6mm square EP (other EP sizes already exist)

whitequark:
> if VTH and VREF-AC are used together, should they be closer together/adjacent?

I tried to make the symbol a bit like the way it's drawn in the datasheet:

Screenshot_20200114_074654

Also, it's more symmetric this way :)

> needs new footprint with 1.6mm square EP (other EP sizes already exist)

Already pending review: https://github.com/KiCad/kicad-footprints/pull/2049

whitequark:
Thanks for the review. I've applied all changes.

whitequark:
Updated. Does this NC placement look OK? I'm not sure if that's what you meant.
Screenshot_20200209_172045

chschlue:
That's it.

chschlue:
Thanks.

PR #2443, opened 2020-01-13
Sensor_Temperature: add TC1047

ChristianHirsch:
Add [TC1047](http://ww1.microchip.com/downloads/en/DeviceDoc/21498D.pdf) analog temperature sensor.

image

I know that the checker gives a warning about violating rule S4.4, but the VOUT pin gives a voltage output depending on the temperature. Thus, I would not define it is a power output.


chschlue:
These can be aliases of MCP9700T-HTT as TC1047xNB and TC1047AxNB, respectively.

ChristianHirsch:
thanks @chschlue for this hint. I fixed that and they are aliases now.

Sensor_Temperature.dcm
```diff
@@ -462,6 +462,18 @@ K I2C Temperature Sensor
 F https://www.silabs.com/documents/public/data-sheets/Si7050-1-3-4-5-A20.pdf
 $ENDCMP
 #
+$CMP TC1047xNB
+D Precision Temperature-to-Voltage Converter, 10mV/ºC, -40ºC to +125ºC, ±2.0ºC (max), 2.7V to 4.4V, SOT-23B
```
chschlue: Change package to just SOT-23

Sensor_Temperature.dcm
```diff
@@ -462,6 +462,18 @@ K I2C Temperature Sensor
...
+K temperature sensor analog
+F http://ww1.microchip.com/downloads/en/DeviceDoc/21498D.pdf
+$ENDCMP
+#
+$CMP TC1047AxNB
+D Precision Temperature-to-Voltage Converter, 10mV/ºC, -40ºC to +125ºC, ±2.0ºC (max), 2.5V to 5.5V, SOT-23B
```
chschlue: Same as above

Sensor_Temperature.dcm
ChristianHirsch: although SOT-23B is mentioned in the datasheet?

chschlue: Yes.
SOT-23B seems to be Microchip's way of referring to TO-236 variation AB (which differs from AA in height tolerance), but that's neither salient nor does the lib contain different FPs for the two variants.

ChristianHirsch:
fixed that.

chschlue:
Thanks.

PR #2444, opened 2020-01-13
Regulator_Linear: add TC2014 LDO family

ChristianHirsch:
Add [TC2014, TC2015 and TC2185](http://ww1.microchip.com/downloads/en/DeviceDoc/21662F.pdf) family of CMOS LDOs with shutdown and reference bypass.

image


diegoherranz:
Thanks for your submission.

I think these could be aliases of LP2985-1.8. Could you convert them to aliases if you agree, not to add extra symbols?

Thanks!

ChristianHirsch:
@diegoherranz if you are ok with having `ON/~OFF` instead of `~SHDN`, I'd be fine with that..

diegoherranz:
In my opinion, it is OK. It is basically the same.
For LDOs, I think we have accepted slightly different pin names in the library (e.g. IN vs VIN) in order to unify symbols.
I think that the benefit of not duplicating symbols outweighs the slightly different pin names. I don't see that causing any confusion to the user.
Thanks!

ChristianHirsch:
fixed that.

diegoherranz:
Sorry for the delay. It looks good.
Just a very minor comment, could you say SOT-23-5 in the description fields instead of SOT23-A?
I know that is what Microchip says, but I think it makes it easier to search for parts if we unify on SOT-23-5 as we say on many other parts in the library.

Thanks!

ChristianHirsch:
@diegoherranz changed the footprint in the description.

diegoherranz:
Looks good! Thanks for the small tweak. Merging.

Issue #2446, opened 2020-01-13
Add open collector symbol


hildogjr:
Some parts / symbols listed above could have added the "open collector symbol". This will improve the schematics (main in a printed version):
7405 and 06 http://www.ti.com/lit/ds/symlink/sn74ls06.pdf
74596
7433
7403
...

evanshultz:
This topic is covered at https://github.com/KiCad/kicad-symbols/issues/2115, but apparently you _really_ want your own issue about it.

hildogjr:
Since this could be easy to fix than #2115 that is a more deep discussion, I use this to registry a pull request adding the "open collector" symbol to the logical parts.

hildogjr:
A implementation to better behavior is being discussed at https://gitlab.com/kicad/code/kicad/-/issues/4049

PR #2450, opened 2020-01-18
Add open collector symbol to 74 ports

hildogjr:
removed bck files


hildogjr:
This pull request add the open collector symbol to several logic ports (Fix #2446):
Screenshot from 2020-01-17 22-01-49

Add open collector symbol to some analog comparators, moving it close to the output pin with correspondent open collector behaviour
Screenshot from 2020-01-17 22-03-41
:

chschlue:
FYI: I proposed splitting comparator symbols into open collector and open drain types at https://github.com/KiCad/kicad-symbols/pull/2467.
The problem with your changes to comparators is that it also marks aliases which are actually push-pull parts.
Perhaps you could narrow this PR to logic only while I add OC symbols to #2467?

hildogjr:
Some different parts (not variation of the same one) are marked as alias. Could you give me enunciate some of the part problems?

chschlue:
Example:
ADCMP350/354/356 are aliases but 350/354 are open drain while 356 is push-pull.

PR #2456, opened 2020-01-20
Add 62256 based on the existing 628128

jdmarble:
62256_DIP28_SOP28

To create this, I copied the 628128_DIP32_SSOP32 and renamed it 62256_DIP28_SOP28.
I changed the symbol properties to match the new part.
I removed the extra pins, fixed the numbering, and recentered the part.

Like the 628128, there are multiple manufacturers of the 62256:

* [Samsung KM62256C](https://www.futurlec.com/Datasheet/Memory/62256.pdf)
* [Hitachi HM62256B](https://web.mit.edu/6.115/www/document/62256.pdf)
* [Cypress CY62256](https://ecee.colorado.edu/~mcclurel/Cypress_SRAM_CY62256.pdf)

Rather than follow the letter of KLC G2.1 and use a fully specified symbol,
I matched the precedent set by the 628128 and made this a generic symbol.

Similarly, I maintained conformity with the existing 628128 with the I/O pin naming.
The naming varies depending on the manufacturer,
so I used the same names as the 628128.

---


myfreescalewebpage:
Hello @jdmarble thanks for this symbol

I understand the way to create this symbol. However, this is no more recommended to do like that and I would really prefer to have one device per package, with the correct footprint associated.

In this example this will be "KM62256CLP" and "KM62256CLG". If you know other manufacturers providing the same device you can add some aliases.

On the symbol itself, no comment to do.

Joel

jdmarble:
Thanks for the guidance!

I've made the KM62256CLP (with aliases) fully specified. I did not create a KM62256CLG version because there are no matching SOP packages.

myfreescalewebpage:
I have an issue while opening the library, see screenshot below. This is sometimes the case when merging master as you have done.

Capture

There is missing liens at the end of your device definition in the lib file at line 1256:
ENDDRAW
ENDDEF

jdmarble:
Thanks for finding that. Looks like I fixed it an another problem with the footprint filter.

myfreescalewebpage:
Just one remaining comment

- [ ] Description of the HM62256BLP should be terminated with the package name as done for the others

Else, very good job done here !
Joel

myfreescalewebpage:

PR #2463, opened 2020-01-23
Rename STM32L4x1 VDDUSB pin into VDD

aurel32:
The STM32L4x1 SoCs do not have USB support and thus no VDDUSB pin.
A simple VDD pin is used instead. This probably comes from a copy and
paste issue from other another SoC in the family.

---

See the change in image for one of the SoCs:
stm32l4x1-vddusb

Link to the datasheet:
https://www.st.com/resource/en/datasheet/stm32l431cb.pdf


myfreescalewebpage:
@aurel32 thanks for pointing this. The modification break existing design and can be merge only for the next major release (V6). I wuld suggest to stack also the VSS pins at the same time. Joel

aurel32:
> @aurel32 thanks for pointing this. The modification break existing design and can be merge only for the next major release (V6). I wuld suggest to stack also the VSS pins at the same time. Joel

In the meantime, would that work simply replacing the VDDUSB pin by a VDD one without grouping them? Designs that do not place the "VDDUSB" pin connected to the "VDD" ones are broken anyway.

myfreescalewebpage:
If you want yes we can separate in two pull request and first fix the name of the pin only, with no move. This can be merged now without waiting V6.

evanshultz:
I believe all of these symbols are auto-generate by a script at https://github.com/KiCad/kicad-library-utils/tree/master/schlib/autogen/stm32. It seems the URL format is changed and I didn't try to track down if this particular symbol is, but fixing the script is the best solution if that is the case. FYI, the script was written by @Ratfink .

myfreescalewebpage:
Good information from @evanshultz !

myfreescalewebpage:
We are currently discussing of a new scripting method for all MCU, the STM generator will no more be up to date if the goal is achieved, I propose to merge the pull request right now for this fix.

myfreescalewebpage:
@aurel32 no comment on the VDDUSB pins, just one request if you have time to do it: several symbols have KLC error S5.2 because of wrong fp filter. `ST_WLCSP` should be `ST*WLCSP`.

If you have not the time to do it I will merge your PR like that because it's not a regression.

Thanks
Joel

myfreescalewebpage:
Not sure the author will respond to the previous question so merging right now.

PR #2465, opened 2020-01-24
Added dependent source symbols to Spice_simulation

warnerjon12:
Added 6 new symbols with spice mappings. The spice mappings
use the Spice_Params attribute, and so cannot be used as
intended in older versions of the software. The symbols
may still be opened in older versions without issues.
These symbols are intended to be used with the changes in
https://gitlab.com/kicad/code/kicad/merge_requests/77.

---


warnerjon12:
The Travis build is complaining that there are fields after the DATASHEET field, because there are Spice fields. This is the case with existing symbols in the Simulation_SPICE library as well, so that shouldn't be an issue.

warnerjon12:
Per [KLC G1.8](https://kicad-pcb.org/libraries/klc/G1.8/), these changes are not ready for merge yet, but can be merged later when https://gitlab.com/kicad/code/kicad/merge_requests/77 is resolved and the changes are included in a stable release.

evanshultz:
Please don't force push to make reviews easier.

This PR does not need to wait for your MR to be merged. They are usable now.

Please use the source terminology from http://ngspice.sourceforge.net/docs/ngspice-manual.pdf. Start at page 92. So an `F` source is named named `CCCS` and described as a `Linear Current-Controlled Current Source`. I'm not quite sure what's the best way to name V- and I-controlled ASRCs, though. Ideas?

For 2-pin symbols let's do something more to differentiate them from the existing sources, like perhaps using a diamond or hexagon instead of a circle around the source. But match whatever you can from the existing source symbols. For 4-pin symbols also do that and then make a box around the entire symbol since we don't allow floating pins. We can brainstorm this right now.

warnerjon12:
Okay, I can rename them. While the symbols themselves are usable now, the Spice models are not usable until the referenced merge request is completed. This is what I meant.

warnerjon12:
Alright, it is done. I decided on IDEP and VDEP for the ASRCs (DEP as in dependent).

warnerjon12:
I decided it was better just to connect the controlling pins to the two sides of the diamond than to box the whole thing.

warnerjon12:
History cleanup with rebase

myfreescalewebpage:
@evanshultz ping

PR #2468, opened 2020-01-27
Add LP3869x linear regulators

evanshultz:
https://www.ti.com/lit/ds/symlink/lp38691.pdf
https://www.ti.com/lit/ds/symlink/lp38693.pdf

image
image
image
image

Two notes:
- Page 20 of the datasheet talks about the thermal pad of the WSON. It's unconnected electrically but strongly recommended to be grounded.
- The WSON footprint oddly doesn't mention the thermal pad size, but it's correct and used by two symbols (`LMR10510YSD` and `LMR62421XSD`) so I went with it.

---


myfreescalewebpage:

myfreescalewebpage:
Hi @evanshultz !

- [ ] The aliases of "LP38691SD-1.8" have datasheet link "https://www.ti.com/lit/ds/symlink/lp38693.pdf"
- [ ] The "LP38693DT-1.8" has datasheet "https://www.ti.com/lit/ds/symlink/lp38691.pdf"
- [ ] LP38691SD-1.8 the expose pad is usually called "EP" in the lib, I was almost sure we do this even if the datasheet called it something else, isn't it ?

Else no comment on the symbols

Cheers,
Joel

evanshultz:
Fixed the first two. How did I do that?

For the last one, I think EP is standard for footprints (see https://kicad-pcb.org/libraries/klc/S5.2/) but I'm not aware of any standard for symbols. It does vary. Check out `TLC5947DAP` and `TLC5940PWP` as a couples examples I found quickly. But if you want I can change it. Probably this should be standardized, but there's a lot of cleanup to get it tidy.

myfreescalewebpage:
Yes I know there is plenty of things to cleanup like that. I personally always ask the expose pad to be called "EP", except if it is stacked with GND of course!

evanshultz:
Alright. I can change it, but did you notice on page 4 it's explicitly called 'DAP'?

myfreescalewebpage:
Yes I notice that. I think "EP" naming should really be standardized to symbol as well.

evanshultz:
OK. Done.

myfreescalewebpage:
Thanks. I have just notice in the travis log that the pin offset is 10 mil instead of 20. Is it intentional?

evanshultz:
Yes. To give a bit more clearance for the 'standard' regulator symbol pin text, going down to 10mil was used. I didn't find the PR about this but I recall seeing this when Jan standardized these symbols a few years ago. And right now I see about double the number of symbols using 10mil in the library. So clearly it's still mixed, but 10mil is intentional.

myfreescalewebpage:

PR #2477, opened 2020-02-04
symbols of ONFI NAND Flash

nerdyscout:
this PR includes several symbols for NAND Flash specified here:
http://www.onfi.org/-/media/client/onfi/specs/onfi_4_0-gold.pdf?la=en

all of those symbols should be drop in compatible, currently a few NC pins on each symbols are missing.
It would be nice to hear if this design pattern is ok or major changes are nedded. otherwise it would mean a big effort changing things afterwards. if everything is fine more symbols will follow.

2 channel package designed as 2 independent units. Power pins are duplicated as it is good practice to connect power even if one channel is not used.
grafik

TC58* is an alias of BGA-SDR_63x8
grafik

much more flashes match these symbols and will follow up.

best regards!

---


poeschlr:
Please be a bit more careful with your git history.

Also take a look at the travis output. In particular the requirement regarding visibility of power input pins. See rules 
- https://kicad-pcb.org/libraries/klc/S4.6/
- https://kicad-pcb.org/libraries/klc/S4.3/

nerdyscout:
Hi @poeschlr 

sorry about that git history, I'm not sure how this happened.

I don't really get your comment about the power pins... as far as I understand the KLC I am allowed to stack these and hide all besides one. So what do you suggest to change?

poeschlr:
Read the rules i linked carefully. The hidden ones are not allowed to have the pin type power input as that would convert them into global labels.

nerdyscout:
i made a new, clean pull request - with more advanced symbols - for this topic
https://github.com/KiCad/kicad-symbols/pull/2601

PR #2479, opened 2020-02-05
Added symbol for LF351 OpAmp

johnfwhitmore:
I've been away from electronics for a while but back again and probably rusty and things might have changed in the process. This is more or less a direct copy of the LM741 in terms of the symbol and the pin designation as both are pretty much the same. I could possibly have created a new alias in the LM741, but I think the last time I did this, probably over a year ago, it was suggested to add new symbol. If that's wrong I can certainly change things.

Datasheet:   http://www.farnell.com/datasheets/1880969.pdf

Screenshot from 2020-02-05 17-18-00


johnfwhitmore:
I think that's passing all checks now. If this should be two different symbols for each footprint let me know and I'll change it.

cpresser:
Welcome back :)

> I think that's passing all checks now. If this should be two different symbols for each footprint let me know and I'll change it.

Yes, that would be better. We call them 'fully specified symbols' (https://kicad-pcb.org/libraries/klc/G2.1/). In this case you would make two symbols: `LM351D` and `LM351D`. They can have the same drawing, but have different footprints and filters set.

Also, the datasheet should point to a manufacturer, not a distributor.

johnfwhitmore:
Good to be back. Was very busy most of last year so couldn't find time, sorry.

cpresser:
Way better :)
- [ ] Please add a proper description. Like `Wide bandwidth, single JFET, operational amplifier, SOIC-8`
- [ ] Keywords also could use improvement. Perhaps `single opamp jfet`

johnfwhitmore:
This branch now has conflicts so I'll rebase and move changes forward.

johnfwhitmore:
Actually if I do that will I destroy this PR? Scissors I'm not sure I should merge the latest into my branch and push it to this PR. Let me know.

chschlue:
This branch is messed up in several ways.

For starters, F3 fields of all Amplifier_Operational.lib have been filled. Are you using a KiCad nightly build?

And I just noticed that is also the reason there's a merge conflict affecting NCS2325DM. Alright, it's only messed up in one way ;)

cpresser:
One way to fix this is:
- close KiCad
- `git reset --hard HEAD^` (to undo the last fucked up commit in your local tree)
- `git fetch origin`
- `git rebase origin/master`
- add the description changes again
- `git add -p` (this adds changed lines individually)
- `git commit`
- `git push <yourremote>` -f 

A force push is not the best way, since it basically means I have to re-check every commit again. Maintainers don't like force-push at all.
I personally think rebasing is acceptable and prefer that over merging the master into the local branch.

chschlue:
Wait!
It's just https://github.com/KiCad/kicad-symbols/pull/2479/commits/258b3f0d0f6b3a2fa0b899ab9eff2507de8fefa9.

I suggest reverting that commit and pushing the changes to the dcm file again.

cpresser:
> Wait!
> It's just [42224be](https://github.com/KiCad/kicad-symbols/commit/42224be102454983d3c5f2ab696fab2b5f655ee8).
> 
> I suggest reverting that commit and pushing the changes to the dcm file again.

That is what `HEAD^` does. its the previous commit.

johnfwhitmore:
My origin/master is going to be well ahead of that branch for the LF352 which was probably created a while ago. Feb 5th according to this PR.
Best thing I can do is to kill this PR and restart from the current repo state.

johnfwhitmore:
Think I might have gotten away with it. But the lib file I should re-check that out.
I am using a very recent version of Kicad as I re-installed a distro on this last week. When I do that I normally pull from git and build kicad from source. The version in distro repos is usually a bit behind.

cpresser:
> Think I might have gotten away with it. But the lib file I should re-check that out.

Yep, it looks good.

> I am using a very recent version of Kicad as I re-installed a distro on this last week. When I do that I normally pull from git and build kicad from source. The version in distro repos is usually a bit behind.

KLC mandates to use the latest stable version. Depending on the exact branch you build from the git version might be to new.

Unfortunately I am still not happy with the description.
- [ ] The package should not be in parenthesis, and it should be comma separated.
- [ ] There should be commas as delimiters in the description (https://kicad-pcb.org/libraries/klc/S6.2/ , Point 6)

johnfwhitmore:
Got a message to say that the symbol isn't loading correctly. I probably started this in Feb on a different computer with a different version of KiCad. I'm going to close this PR and start again. I'll install a better version of KiCad.

PR #2482, opened 2020-02-07
Add LA1185 FM mixer/receiver/amplifier chip from Sanyo

dvdesolve:
Hello! I've created symbol for Sanyo's LA1185 FM mixer/receiver/amplifier chip. There is no direct link for official datasheet PDF so I'll leave [this page](https://www.alldatasheet.com/datasheet-pdf/pdf/39974/SANYO/LA1185.html) as datasheet reference. Symbol screenshot:

symbol


myfreescalewebpage:
Hi @dvdesolve , thanks for contributing,

- [ ] The description should ends with the package: `FM receiver front-end IC for radio-cassette recorder, music center applications, SIP-9`
- [ ] The footprint should be `Package_SIP:SIP-9_22.3x3mm_P2.54mm`
- [ ] The footprint filter should be `SIP*22.3x3mm*P2.54mm*`
- [ ] Pin 2 should be Passive
- [ ] The symbol is a bit large, you should reduce the width

Ok for this datasheet link because no other is available.

Cheers,
Joel

dvdesolve:
Fixed all your points. I think this width is optimal because shrinking it even further will cause labels overlap

dvdesolve:
Hmm, it seems like suggested footprint doesn't exist at all

myfreescalewebpage:
Hi @dvdesolve 
Thanks for the quick fix. All points OK, the width is nice  like that.
The footprint should be submitted in kicad-footprints. We always create the exact footprints, we do not accept approximation.
Joel

dvdesolve:
So, should I prepare this footprint first and make pull request in `kicad-footprints`?

myfreescalewebpage:
Yes please. The symbol footprint can't be merge now.

myfreescalewebpage:
I recommend you to use the previous SIP-9 package, duplicate and fix the length which is a little bit longer. That's all.

dvdesolve:
Okay, I've done that. You can check my pull request [here](https://github.com/KiCad/kicad-footprints/pull/2175)

myfreescalewebpage:
I will let footprint specialists check it.

chschlue:
FP merged.

myfreescalewebpage:

PR #2484, opened 2020-02-08
Add LM2931 low-drop voltage regulator in TO220 package

dvdesolve:
Hello! I've created symbol for LM2931 low-drop voltage regulator chip in TO220 package (3.3 and 5.0 versions). [Datasheet](https://www.ti.com/lit/ds/symlink/lm2931-n.pdf). Symbol screenshot:

symbol

Symbol is slight off-centered but it was based upon existing LM7805 symbol so I think it should be ok to ignore check warning.


aewallin:
tab (pin4 of footprint?) is GND - so it should probably be stacked with pin2 ?

dvdesolve:
I don't see pin4 in TO220-3-Vertical package... Could you please make a screenshot? Actually this one is based on LM7805 symbol - I've just made some small fixes

aewallin:
right. If we go with atomic parts then maybe we need another symbol for the 'lay-down' version where the tab is in contact with the PCB and the footprint has 4 pins.
Should the symbol-name for this variant then include "Vertical"??

dvdesolve:
I haven't seen specialized symbols for lay-downs. I always thought that for
these kind of parts we could just choose between footprint orientations and
not specialized symbols. of course if we talk about the same package (to220
in this case). or may be I'm just too inattentive...

вт, 25 февр. 2020 г., 20:55 Anders Wallin <notifications@github.com>:

evanshultz:
This is a good discussion. Probably it is something to discuss to consensus and address library-wide. However, right now none of the TO220 horizontal footprints in the lib have copper so let's leave off a pin for the tab at this point.

@dvdesolve 
The symbol you copied from is for a generic part and includes the package name as a suffix on the symbol name (see KLC for details). Usually we would not allow this on manufacturer-specific parts, but LM2931 seems to be readily available from several vendors. Somebody feel free to check that or correct me. Just a note so you know.

dvdesolve:
so, if I get it right, in general all is good and no fixes are needed?

вт, 25 февр. 2020 г., 22:50 evanshultz <notifications@github.com>:

myfreescalewebpage:
Hello @dvdesolve 
This part should simply be an alias of LM7805_TO220 .
Joel

dvdesolve:
If I got it right, I should only leave edits in `Regulator_Linear.dcm` and add alias fields to `LM7805_TO220` entry in `Regulator_Linear.lib`?

myfreescalewebpage:
Yes exactly.

dvdesolve:
I've pushed fixes. I think that's all about this chip :)

myfreescalewebpage:
This is OK for me too, merging. Thanks. Joel

PR #2489, opened 2020-02-11
Add SN75162BDW

bastien-roucaries:
Add Octal general-purpose interface bus transceiver, SOIC-24 used as gpib bidirectional bus transceiver

Improve datasheet by naming pin with TERM or GPIB prefix depending of the side

Only SOIC is actually made by TI so create only SOIC symbol.

Datasheet  http://www.ti.com/lit/ds/symlink/sn75162b.pdf

Screenshot_20200211_223321

---


bastien-roucaries:
Move the control pin to the terminal side. Update the screenshot

myfreescalewebpage:
Hello @bastien-roucaries thanks for contributing,

It's not possible to add this kind of device in the lib, the device should fit the datasheet. The pin name can't reflect a specific application (GPIB in this case). If the device is used for antoher purpose, then it makes no sense to have pin called "GPIB".

Please review the symbol and I will perform the review again asap. You an fllow example of SN75160 device in the lib.

Joel

bastien-roucaries:
@myfreescalewebpage How can we differentiate in this case between TERM side and BUS side ?

Do you pefer to call it BUS side ? 
SN75160  has different name on the datasheet for BUS side and TERM side, not this component that use the same name for BUS and TERM side (unfortunatly)

To my knowledge this component is only used for GPIB control (see datasheet).

Bastien

myfreescalewebpage:
Hi @bastien-roucaries , thanks for contributing,

Sorry I just realize the pin name was GPIB specific. So yes we can consider this will be used only for GPIB, we keep the pin naming you have done.

A singlecomments I have during my review:

- [ ] Pins 11, 14 and 23 are missing (NC, Invisible, Passive, on the border outline of the symbol)

Cheers,
Joel

myfreescalewebpage:
@bastien-roucaries ping

PR #2495, opened 2020-02-16
added usb-c-6pin

aster94:
see https://github.com/KiCad/kicad-footprints/pull/2075


poeschlr:
please add a screenshot of your creation

aster94:
image

evanshultz:
Is this 6-pin version an official part of the USB spec?

See https://github.com/KiCad/kicad-symbols/pull/937. There might be something useful in the original PR that added USB symbols too.

aster94:
I had a quick scan of one of these documents and i didn't saw it, maybe @diegoherranz checked it in depth and can give some suggestion

diegoherranz:
I didn't remember seeing this connector version in the standard, but just in case I've checked it (Universal Serial Bus Type-C Cable and Connector Specification, Release 2.0, August 2019).

There is a power-only plug, but not power-only receptacle. And the power-only receptacle needs more pins anyway:

> The USB Type-C Power-Only plug is a depopulated version of the USB Full-Featured Type-C plug or the USB 2.0 Type-C plug. [...] Contacts for CC, VBUS, and GND (i.e., A1, A4, A5, A9, A12, B1, B4, B9, and B12) shall be present. Physical presence of contacts in the other 15 contact locations is optional. The USB Type-C Power-Only plug shall only be used on a non-charger captive cable application. Implementation of Rd or CC communication on pin A5 is required in the application.

So, I don't think it is compliant.
What do we do in this case? Accept it as a USB standard symbol? Or create a symbol specific for this part? Even if not compliant other manufacturers may start making similar parts?

aster94:
even if not a standard i would support for the add of the symbol/footprint. I am seeing this part used more and more on aliexpress and similar supplier

chschlue:
`Shall` indicates a requirement.

From IEC directives:
shall

diegoherranz:
Within the same document (Universal Serial Bus Type-C Cable and Connector Specification, Release 2.0, August 2019):

> 1.4.2.7 Shall
> Shall is a keyword indicating a mandatory (normative) requirement. Designers are
> mandated to implement all such requirements to ensure interoperability with other
> compliant Devices.

Anyway, I think the question is not whether to add this to the library or not, but whether to define it as a USB generic symbol or manufacturer specific. Is it just this MPN with this 6-pin arrangement or are there other manufactures breaking the standard in the same way?

aster94:
Today i found another USB C with 6 pins. I don't know if the manufacter is the same

https://a.aliexpress.com/_BUzCnB

myfreescalewebpage:
Hello @aster94 

Not a full review but first comments:

- [ ] This should go in the "Connector library" with the other USB C symbols
- [ ] Is it a receptacle or plug ? You should follow the naming convention of the other USB C connectors
- [ ] Footprint to be completed
- [ ] Do you have a real datasheet for this symbol and for the footprint ? I mean with the pinout, dimensions, etc. If yes you should simply indicate it in the symbol
- [ ] Position of SHIELD and GND pins to be fixed in order to keep consistency with the other USB C symbols

Will continue the review after

Joel

aster94:
hello, unfortunately, I don't have time to change the library, so anyone reading this feel free to make the necessary changes

myfreescalewebpage:
@aster94 thanks for the update, I indicate it is Abandoned right now but anyone can continue later

aster94:
ok, please mark also this as abandoned https://github.com/KiCad/kicad-footprints/pull/2075

PR #2500, opened 2020-02-17
added NCP551

Knochi:
ON-Semi NCP551 series low dropout regulator
added common voltage versions 1.5, 1.8, 3.3 and 5.0V

---

Regulator_Linear_NCP551SN50T1G


chschlue (changes requested):
Please remove T1G packaging suffix and move NC pin onto the symbol's outline.

Knochi:
> 
> 
> Please remove T1G packaging suffix and move NC pin onto the symbol's outline.

What do you mean by "onto the symbols outline" it is on the outline.

chschlue:
No, it's not. It's 100 mils below.

If you were to draw a wire from GND to the right, you'd connect it. That's why NC pins have to be within the body.

Also, please remove automotive variants. Missed those before.

Knochi:
oh ok.. didn't get that the have to be within the body.. never saw that before
Am 14. März 2020, 19:32 +0100 schrieb Christian Schlüter <notifications@github.com>:

koenraad-voltza:
> Also, please remove automotive variants. Missed those before.

Why? I don't see anything in the KLC against automotive MPNs, and the base part number is different. Maybe I missed something?

chschlue:
That's to avoid bloat as per [S2.2](https://kicad-pcb.org/libraries/klc/S2.2/).
In this case we cannot reasonably use wildcards (who would search for `NCx551`?)
We don't have `SNx4*` symbols for `SN54*` and `SN74*` parts for that reason.

PR #2505, opened 2020-02-21
Add TS3L501E 8-channel mux

chschlue:
DS: http://www.ti.com/lit/ds/symlink/ts3l501e.pdf
FP: https://github.com/KiCad/kicad-footprints/pull/2101
Script: https://github.com/pointhi/kicad-footprint-generator/pull/501

Units A-D:
TS3L501ERUA
Channels in pairs; part is meant for differential signals and package is also laid out that way.

TS3L501ERUE
Additional unit for LEDs.

TS3L501ERUF
Power and control unit; VDD pins are not stacked because the DS specifically suggest using the most convenient one only (pages 17f).

---


cpresser:
- [x] Pin Names and Numbering are good
- [x] The split into units for differential signaling is okay. It means that the schematic for the LAN application will look ugly - but that's acceptable. Besides, if you had opted for a big symbol that has all 8 channels together it would be the other way around.
- [ ] I am not sure about the pin-types. Currently the `Analog_Switch` lib has both devices with mux-pins set to `passive` and `bidir`. This adds another flavor. 
- [ ] Having multiple power inputs where the user only connects one will generate multiple DRC errors. But  don't see how that can be improved with changes to the symbol. Any ideas?

chschlue:
> * [ ]  I am not sure about the pin-types. Currently the `Analog_Switch` lib has both devices with mux-pins set to `passive` and `bidir`. This adds another flavor.

Good catch. Don't know what I was thinking. Bidir makes more sense (tri-state throws a warning when connected to output). I would regard passive as kind of a last resort type only.

> * [ ]  Having multiple power inputs where the user only connects one will generate multiple DRC errors. But  don't see how that can be improved with changes to the symbol. Any ideas?

None other than that it's left to the user to add NC flags to the unused pins.

PR #2510, opened 2020-02-23
Add SC16IS752

ppielorz:
Added SC16IS572 SPI to dual UART interface.
Datasheet: https://www.nxp.com/docs/en/data-sheet/SC16IS752_SC16IS762.pdf
Screenshot 16IS752IPW: 
image

Screenshot 16IS752IBS: 
image

Pins are groupped by function, not by direction (similar to datasheet).
---


diegoherranz:
The screenshot in here doesn't match the latest data. Can you update the screenshot? It can be misleading otherwise. Also, can you add screenshots for both symbols being added?

Thanks!

diegoherranz:
Thanks for your contribution and sorry for the delay in reviewing this. It all look pretty good except for a couple of things.

Checks done which look OK:
- Descriptions look OK
- Footprints and footprint filters look OK except for comment below
- Pinout looks OK expect for comments below

Changes needed:
- [ ] Datasheet mandates connected the center pad to GND for the VQFN. Can you add it to the VSS pin stack?
- [ ] Footprint filter for the VQFN should include `1EP`. See [KLC S5.2](https://www.kicad-pcb.org/libraries/klc/S5.2/).
- [ ] I would set SDA/VSS as bidirectional instead of passive. Whether it is used as I2C or connected to GND, bidirectional should work and I find it more specific than passive.

Thanks!

ppielorz:
Everything should be ok now

diegoherranz:
Looks good. Thanks for the quick fix. Merging.

PR #2512, opened 2020-02-27
Add missing pins 3 and 5

evanshultz:
https://www.xppower.com/portals/0/pdfs/SF_ITX.pdf

When reviewing https://github.com/KiCad/kicad-footprints/pull/2079, I noticed that one of the symbols was missing pins 3 and 5. And the footprint has those pins so we need them. I think the others are OK but maybe my eyes are bleary.

---


chschlue:
ITX*SA don't have pins 3 and 5, or am I missing something?

evanshultz:
So `N.P.` means not populated? Didn't catch that before but I suppose it makes sense. Then we need a footprint without those pins instead... and it seems we already have one. Thanks for letting me know.

PR #2528, opened 2020-03-08
Add LMR33640 swtching regulator

nivekg:
Here's a symbol for the LMR33640ADDAR and LMR33640DDDAR. The LMR33640DDDAR is an alias LMR33640ADDAR since everything is the except for its switching frequency in the description.

LMR33640

Only violates rule S4.2 and thats because the vin and vcc are on the sides like other regulator symbols. 
Datasheet: http://www.ti.com/lit/ds/symlink/lmr33640.pdf


cpresser:

There are a few small things that need fixing before I can merge this:
- [ ] The `R` suffix is a packaging option and thus needs to be removed
- [ ] I think the VCC pin should go on the bottom left, as shown here: 
image
- [ ] That will allow to put the FB pin on the bottom right.Thus allowing to put a `R_Small` between FB and SW for nice wiring up in the schematic.
- [ ] AGND does not need the EP in its name.
- [ ] The pin-count should not be included in the footprint-filter
- [ ] The PG pin type is open-collector

nivekg:
Made the changes as requested. I did move the ground pins slightly to right, since after moving the VCC pin to the left, the writing was cramped. Here an updated screenshot:
lmr66340_updated

cpresser:
Thanks, that looks good!

Issue #953, opened 2018-09-18
Timer: S5.1 and S5.2 errors


antoniovazquezblanco:
KLC Rules:
http://kicad-pcb.org/libraries/klc/S5.1/
http://kicad-pcb.org/libraries/klc/S5.2/

Splitting issue #28 I have found the following violations:
```
Checking symbol '8284':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'PDIP*W7.62mm*' does not match the footprint 'Package_DIP:DIP-18_W7.62mm' set for this symbol.
     - could not match 'DIP-18_W7.62mm' against filter 'PDIP*W7.62mm*'
     - could not match 'Package_DIP:DIP-18_W7.62mm' against filter 'PDIP*W7.62mm*'
    Symbol has a footprint defined in the footprint field, but several (2) footprint filters set. If the symbol is for a single default footprint, remove the surplus filters. If the symbol is meant for multiple different footprints, empty the footprint field.
checking ruleS5.2
checking ruleS5.1
Checking symbol '82C54':
  Violating S5.1
    For components with a single default footprint, footprint field is filled with valid footprint
    filename
    Footprint filter 'PDIP*W15.24mm*' does not match the footprint 'Package_DIP:DIP-24_W15.24mm' set for this symbol.
     - could not match 'DIP-24_W15.24mm' against filter 'PDIP*W15.24mm*'
     - could not match 'Package_DIP:DIP-24_W15.24mm' against filter 'PDIP*W15.24mm*'
    Symbol has a footprint defined in the footprint field, but several (2) footprint filters set. If the symbol is for a single default footprint, remove the surplus filters. If the symbol is meant for multiple different footprints, empty the footprint field.
```

serkanxselcuk:
Hi @antoniovazquezblanco , 

>Footprint filter 'PDIP*W7.62mm*' does not match the footprint 'Package_DIP:DIP-18_W7.62mm' set for this symbol.
     - could not match 'DIP-18_W7.62mm' against filter 'PDIP*W7.62mm*'
     - could not match 'Package_DIP:DIP-18_W7.62mm' against filter 'PDIP*W7.62mm*'

You filtered `PDIP*W7.62mm*` and set footprint as `DIP-18_W7.62mm`, which cannot be found using filter. KLC shows this as error. I would try to set filter `*DIP*W7.62mm*`' or `DIP*W7.62mm*`.

[...]

PR #2533, opened 2020-03-11
Fix S5.1 and S5.2 errors in Timer.lib (#953)

serkanxselcuk:
Fixed Issue: #953
Datasheets:
https://www.cpu-galaxy.at/cpu/ram%20rom%20eprom/other_intel_chips/other_intel-Dateien/D8284A_Datasheet.pdf
image

http://www.scs.stanford.edu/10wi-cs140/pintos/specs/8254.pdf
image
image

---


serkanxselcuk:
Travis Warning:
```
Checking symbol '8284':
  Violating S4.4
     - Pin CLK (8) @ (400,600) is type OUTPUT : suggested INPUT
```
CLK pin is really output on datasheet.

evanshultz:
@serkanxselcuk 
Thanks for taking time to improve the libraries!

Pins are moving, which we can't allow during major releases. Could you revert any pin location changes and then I review what's left? In this case, you can still make the pins 100mil long as long as the body size changes to the pin locations don't move. Does that make sense?

serkanxselcuk:
Hi,
> Thanks for taking time to improve the libraries!
> 
> Pins are moving, which we can't allow during major releases. Could you revert any pin location changes and then I review what's left? In this case, you can still make the pins 100mil long as long as the body size changes to the pin locations don't move. Does that make sense?
As I understand, we can change pin positions if it is needed, for major versions (next: 6.0.0) but not minor versions (next: 5.1.6).

Anyway how is workflow, if we know we need to change pin positions in future for an issue or PR?
- Option 1: We make other changes without to change pin coordinates in a PR. We open a new issue or reopen PR to change pin positions and fix it as soon as we start to work for a major version.
- Option 2: We make all needed changes, inclusive pin positions, but not merge until we start to work for a new major version. Maybe contributors need to inform in comments, if pins are moved.

Option 1 needs more effort, produces new issues form fixed PRs and new PR as well. In option 2, a PR must wait longer to merge, even if it is fixed.

I think @evanshultz decided for option 1, @myfreescalewebpage decided for option 2. What is the final decision? Should I revert pin positions as in Option 1 or should I leave it as is like option 2? Sure, I can revert pin positions, but I would prefer option 2 for less effort and less count of issues.

evanshultz:
Correct. This is because the symbols are not part of the project, so existing schematic will be updated and will not 'just work' anymore if they're using this library and the pins change. During major releases, as you noted, we will accept breaking schematics. And with v6 the symbol library stuff is changing so this may be a moot point after v6.

And pin position(s) can change any time if there is an actual bug that needs to be fixed.

Either way. It is up to you. All users should appreciate less known errors in the library, but if you don't want to make acceptable changes now then this PR won't get merged and the library will stay as it is. Unfortunately it's not quick and simple in this case to get everything you've proposed done.

cpresser:
@evanshultz @myfreescalewebpage Ping.
We could merge this now.

evanshultz:
@cpresser 
Yes. Thank you.

@serkanxselcuk 
From http://www.scs.stanford.edu/10wi-cs140/pintos/specs/8254.pdf, it looks like the three CLK* pins should be Input type. Can you check on that? Also, active low pins do not need a final `~` so it can be removed.

myfreescalewebpage:
@serkanxselcuk ping, are you still alive on this subject ?

evanshultz:
@serkanxselcuk 
Since I know you're still working on your PRs, are you able to make the two changes I mentioned above? I assume the CLK* pins are inputs, but don't do that and let me know why if I'm wrong. Then I think this is ready to merge.

serkanxselcuk:
Hi @evanshultz 

CLK is output on 8284, input on 8254.
One is a clock generator, which gives clock output. The other one is interval timer, which gives output if clock input's count is reached.

image
8284

image
8254

second `~` is removed.

evanshultz:
The two 8254 symbols have the CLK* pins set as Open Collector, not Input type. Can you fix that? Then this is ready.

serkanxselcuk:
@evanshultz 
they are already shown as input by me.
image
image

image
image


Do you mean they must be open collector?

evanshultz:
Oh. No. My mistake. It's because they have the clock input style. I don't see that KLC says pins can only be the Line style, but that's what I recall. So then there isn't any issue and sorry for not checking the pins closely enough. Thank you!

PR #2542, opened 2020-03-14
added TPS54061

rs-development:
http://www.ti.com/lit/ds/symlink/tps54061.pdf

image

Reused Symbols from TPS54360, only shifted GND Pins to fit VSENSE instead of FB.

image


myfreescalewebpage:

- [ ] The name of the device should be "TPS54061DRB"
- [ ] The footprint filter should be `Texas*S*PVSON*`
- [ ] The pin offset should be 20mil
- [ ] Pin 9 should be stacked with GND
- [ ] BOOT should be Passive
- [ ] Pin 4 should be Passive
- [ ] PH should be Power Output
- [ ] I'm not sure the position of the pin is the best according to the schematic reference in the datasheet page 16, I propose the following:

Capture

Cheers,
Joel

rs-development:
@myfreescalewebpage done as you have said.

image

myfreescalewebpage:
Thanks for the fixes. One travis error remaining about the stacking rule: pin 9 name must be "GND".

myfreescalewebpage:
Pin 9 Passive to respect stacking rule :)

myfreescalewebpage:
Thanks. Merging.

PR #2546, opened 2020-03-15
Add ESP32-S2-WROVER SoC Module

morganrallen:
Provides symbol for new ESP32-S2-WROVER module

---
https://www.espressif.com/sites/default/files/documentation/esp32-s2-wroom_esp32-s2-wroom-i_datasheet_en.pdf
Associated footprint PR https://github.com/KiCad/kicad-footprints/pull/2141

image


morganrallen:
Footprint has been accepted and merged in.
https://github.com/KiCad/kicad-footprints/blob/master/RF_Module.pretty/ESP32-S2-WROVER.kicad_mod

myfreescalewebpage:

myfreescalewebpage:
Hi @morganrallen , thanks for contributing,

- [ ] "Txx" name on pin 33 to 36 not useful, we usually keep only the first function
- [ ] Why an extra space between IO26 and IO33 ?
- [ ] I suggest to have EN in the top left corner has done for the other ESP32 in the lib
- [ ] You will be able to reduce with and height of the symbol

Cheers,
Joel

myfreescalewebpage:
```
Can you clarify what you mean by "GND pins should be stacked?"
```

All GND pins at the same place to automatically connect them together.
First GND pin is Power Input visible, the other GND pins are Passive, invisible.

```
The ESP32-S2 has moved the EN pin from top-left to top-right in addition to being in correct pin order.
```

OK in this case, not a problem.

myfreescalewebpage:
Some comments above not fixed.
GND pin can be centered horizontally it will be clearly better.
Joel

myfreescalewebpage:
Thanks, still remaining the position of EN pin.
This is not to trouble you have this, but if you look at all the other ESP32 devices in the lib, they have all the EN pin at the same place : in the top left corner. The job of reviewer is also to ensure the consistency across the library and this is why I request you to move the pins so that EN will be in the top left corner for this symbol too.
Joel

morganrallen:
https://github.com/KiCad/kicad-symbols/pull/2546#issuecomment-609944591
> ```
> The ESP32-S2 has moved the EN pin from top-left to top-right in addition to being in correct pin order.
> ```
> 
> 
> OK in this case, not a problem.

myfreescalewebpage:
oupss sorry :D let's go merging !

myfreescalewebpage:
I read again the thread.
I realize only now you were speaking of your own symbol when I said OK for the position of EN pin. When I previously responded OK I was thinking you were speaking of another symbol !
So, in the library today, all ESP32 have EN pin on the left.

myfreescalewebpage:
Thanks, and again sorry for the misunderstanding. Merging.

PR #2551, opened 2020-03-16
Add FAN7388 as alias of FAN7888

SchrodingersGat:
https://www.onsemi.com/pub/Collateral/FAN7388-D.pdf

Simply adds a new alias of an existing part, the FAN7388 as rated at 600Vbs

image

PR #2556, opened 2020-03-17
6EDL04N02PRXUMA1

SchrodingersGat:
Adds symbol for the 6EDL04N02PRXUMA1 gate driver:

Three-phase half-bridge gate driver

[Link](https://www.digikey.com/product-detail/en/infineon-technologies/6EDL04N02PRXUMA1/6EDL04N02PRXUMA1CT-ND/4806646)

[Datasheet](https://www.infineon.com/dgdl/6ED2-DB-Rev2_3.pdf?folderId=db3a30431a5c32f2011a77fa27a86cb5&fileId=db3a304336797ff901367c18c18445af)

image


cpresser:
Hi,

do you have a screenshot of the symbol wired up with transistors for the 3 half bridges?
My feeling is that this symbol would be nicer if it would have 4 units. 3x half-bridge, 1x control+power.
What do you think?

SchrodingersGat:
The design of this symbol was based on the existing triple gate drivers in the library (see #2551)

I personally prefer it this style, rather than split, but library consistency would take preference I imagine

cpresser:
> The design of this symbol was based on the existing triple gate drivers in the library (see #2551)
> 
> I personally prefer it this style, rather than split, but library consistency would take preference I imagine

That is a good point. Lets leave it like it is. I will do a full review soon.

cpresser:
- [x] Footprint and Filter are good
- [x] Pin names and numbers (funny side-node, the datasheet does not have pin17 as NC in the pin-table) check out.
- [ ] I am not sure about the pin-types for `VB1 ` and `VS1`. The datasheet says 'power'. However 'passive' also matches `FAN7888`.
- [ ] Partname. The `XUMA1`suffix should be dropped as it is non functional: https://forum.digikey.com/t/infineon-opn-suffixes-example-ntma1/161
- [ ] Description and keywords: The current does not match what I could find in the datasheet: `the typical output current can be given with 165 mA for pull-up and 375 mA for pull down.`

myfreescalewebpage:
@SchrodingersGat ping, are you still alive on this subject ?

myfreescalewebpage:
Non news of original author, indicate this is Abandoned

PR #2376, opened 2019-12-10
Add coupled inductors with 1-3-4-2 pinout

evanshultz:
I added Lodestone VTM footprints, like this one:
image

The pinout is great for a single winding because it allows an inductor winding going from pins 1 to 2 to work if the inductor is inserted into the board in either possible orientation.

But this inductor base is also commonly used for common mode chokes, and in that case the coupled inductor symbols didn't match the required pinout. So I've created them:
image

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [ ] Provide a URL to a datasheet for the symbol(s) you are contributing
- [ ] An example screenshot image is very helpful
- [ ] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [ ] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #2558, opened 2020-03-17
Added coupled inductor symbols, with different pinout.

cnieves1:
Datasheet: https://www.bourns.com/docs/Product-Datasheets/SRF1260.pdf

This is needed for footprint https://github.com/KiCad/kicad-footprints/pull/2152
Captura de pantalla de 2020-03-17 15-05-15
Captura de pantalla de 2020-03-17 15-05-37

---


cnieves1:
Regarding 'L_Core_Ferrite_Coupled_Small_13_24' violating S4.1, it has the same pin lenght as 'L_Core_Ferrite_Coupled_Small_13_24', and I think they should be the same.

evanshultz:
See https://github.com/KiCad/kicad-symbols/pull/2376. I think the symbols are complementary, but likely whichever PR gets merged first will result in branch conflicts in the other PR.

cnieves1:
@evanshultz: didn't find that PR before. Shall I close this PR? #2376 has one additional symbol this PR has not...

evanshultz:
I believe the symbols are different so both PRs are useful. Just a note for the future that two different PRs were submitting similar symbols.

evanshultz:
@cnieves1 
This looks fine to me except I'd like to name it without an underscore between the pin numbers. I can see why you grouped them this way, but I'd rather capture something in the description if this is necessary. Since users can see the symbol in the browser when selecting a part they will know the pins on each winding and other generic symbols put all the pin numbers together.

cnieves1:
Name changed now..

evanshultz:
Thanks!

PR #2560, opened 2020-03-17
LF351 symbol added

johnfwhitmore:
Re-added LF351 with correct version of Kicad

Datasheey - https://www.st.com/resource/en/datasheet/lf351.pdf

Screenshot from 2020-03-17 20-26-52


Amplifier_Operational.dcm
```diff
@@ -180,6 +180,18 @@ K single jfet opamp
 F http://www.ti.com/lit/ds/symlink/lf357.pdf
 $ENDCMP
 #
+$CMP LF351D
+D Wide bandwidth single JFET operational amplifier, SOIC-8
+K Singel, OpAmp, JFET
```
cpresser: typo. 'singel' vs. 'single'
Also commas don't belong into keywords.
But description should have commas.

Amplifier_Operational.dcm
```diff
@@ -180,6 +180,18 @@ K single jfet opamp
...
+F https://www.st.com/resource/en/datasheet/lf351.pdf
+$ENDCMP
+#
+$CMP LF351N
+D Wide bandwidth single JFET operational amplifier, DIP8
```
cpresser: DIP-8 instead of DIP8 please.

Amplifier_Operational.dcm
```diff
@@ -180,6 +180,18 @@ K single jfet opamp
...
+K Single, OpAmp, JFET
```
cpresser: same as above, no commas here.

johnfwhitmore:
Sorry meant to do that yesterday, but got distracted.
And sorry about the mistake in the first place. I got confused with the last PR, which I deleted. The last comment said something about commas and I added them to keyword field not remembering the exact request that was made. Having deleted the PR and the branch I couldn't get back to the exact comment.

cpresser:
There is no need to rush. It is nice to finish a PR fast, but most take weeks or even months.
Can you please fix the last thing from my review: `DIP-8 instead of DIP8 please`
Then this is ready to merge.

Issue #2552, opened 2020-03-16
Analog_Switch ADG728 wrong Pin to signal assignment


kagehisa:
In the official Datasheet the pinout is the following:
![Bildschirmfoto 2020-03-16 um 11 04 58](https://user-images.githubusercontent.com/37015428/76745570-636b4e00-6776-11ea-80de-5ee7f4c13a30.png)

In the Library the pinout for Signals 1 to 4 are correct but 5 to 8 are wrong.
![Bildschirmfoto 2020-03-16 um 11 08 33](https://user-images.githubusercontent.com/37015428/76745747-b04f2480-6776-11ea-9d73-a3d982e1c3f5.png)
Signal S8 should be on pin 9, S7 on pin 10, S6 on pin 11 and S5 on pin 12.
The other assignments are correct.

cpresser:
Can confirm, that looks broken.
Would you provide a fix for that? It is a rather simple change and a good way to start contributing to the library.

kagehisa:
Alright I will give it a try.

cpresser:
There are some instructions over here: https://kicad-pcb.org/libraries/contribute/

PR #2565, opened 2020-03-19
Changed pin to signal assignment for pins 9 to 12 according to the da…

kagehisa:
Corrected the pin to signal assignemnt for pins 9 to 12 in the following way:
-pin 9 to signal 8
-pin 10 to signal 7
-pin 11 to signal 6
-pin 12 to signal 5

Issue: #2552 

Datasheet: [Analog Devices ADG728](https://www.analog.com/media/en/technical-documentation/data-sheets/ADG728_729.pdf)

Changes


chschlue:
Thanks.
Could you also change `A1` to input?

PR #2569, opened 2020-03-23
Fix L_Core_xxxx_Coupled_1324 introduced in #2376

cpresser:
My KiCad version also does re-order some parts
Perhaps the sorting is dependant on locale?

This fixes the errors introduced in #2376
@myfreescalewebpage 
@evanshultz


evanshultz:
Thanks. My first commit was to see if GH would allow the PR to be merged. But then I was suddenly interrupted and didn't get to write the comment in my PR or finish checking things.

myfreescalewebpage:
Thanks for the quick fix !! It's now OK on my side too. Joel

PR #2571, opened 2020-03-24
Add AD654

evanshultz:
https://www.analog.com/media/en/technical-documentation/data-sheets/AD654.pdf
image
image

This felt like the best library but let me know if we have any V-F converters I didn't find or if there's a better place.

---


myfreescalewebpage:
Hello @evanshultz 

I have trouble with the transistor on pins 1/2. Sometimes pin 2 is not Power Input has shown on figure 11 page 9 of the datasheet. Seems the transistor wiring has no specific constraint. I suggest the pins 1 and 2 to be Passive:

Capture

Joel

evanshultz:
@myfreescalewebpage 
Both figures on page 9 show pin 2 going to ground. Can you please check that? Since the pin is named LOGIC COMMON I expect that while it's shown as the output transistor's emitter it's also the ground for all other circuitry and the IC won't work right if it's not grounded. There could be some application where it's not, but I've not seen it and the datasheet doesn't allude to it. Have I overlooked anything?

I don't find anything regarding pin 1's type like I noted above, but it seems Open Collector is valid and isn't likely to result in any ERC or wiring issues. And it probably is truly an open collector pin. Did you find a case where this would cause issues?

myfreescalewebpage:
@evanshultz oupsss sorry, have check again you're right, it's connected to ground.

Anyway, I'm not sure it is mandatory to connect it to the ground. First page of the datasheet it is indicated "The Logic Common terminal can be connected to any level between ground (or –VS) and 4 volts below +VS". I think emitter follower configuration is possible.

evanshultz:
Ah. Thanks for pointing that out. The change you're suggesting seems OK but I'm curious now.

My guess is that emitter follower is not possible as the pin is also the 'bottom' of internal current sources which need 4V+ to operate. I asked ADI for clarification.

evanshultz:
@myfreescalewebpage 
Response from ADI:
```
The FOUT signal is referenced to LOGIC COMMON. In the datasheet diagrams, the pin is connected to GND (0V) externally. The NPN emitter is connected to the same potential. 
The LOGIC COMMON can be connected to a DC voltage from –VS to (+VS – 4V). The reference voltage for FOUT changes accordingly.
The LOGIC COMMON is not connected to other circuitry internally.
```

So my guess was wrong. Since pin 2 needs to be connected to a DC voltage I'm still in favor of keeping Power Input type. And would Open Collector still be fine for pin 1?

myfreescalewebpage:
Ok with the current status, you can post a fix later if you have ERC issues for example.

PR #2572, opened 2020-03-24
Addition of LM3150 Buck Converter Controller

jamesrlucas:
Addition of LM3150 Buck Converter Controller from Texas Instruments

[Datasheet](http://www.ti.com/lit/ds/symlink/lm3150.pdf)

I've laid the pins out in a manner that gives a clean schematic look - this is why ILIM & FB are on the right, depsite being inputs, and VCC is on the left. (See screenshot)
Are you happy with the footprint filter? Or should I make it more specific?

image

---


jamesrlucas:
Is Travis on holiday or something?

chschlue:
Probably quarantined.

cpresser:
I like your pin-layout. Well though and even with an example how it can be wired up :heart: .

Here is my review:
- [x] Pin names, numbers, types are good
- [x] package and fp-filter are good
- [ ] Pin length must be 100mil (I wonder why travis does not detect that - perhaps we need a better test)
- [ ] `VIN` can go onto the left side. Actually most voltage-regulators in the library are build that way and we like consistency. It does generate travis warning, but is okay. (We want to rework that anyway https://github.com/KiCad/kicad-website/pull/478). Then you can also put `EN`all the way to the top, then have `VIN` and with enough space to put a `R_Small` resistor the `RON`pin. Similar to point 4 in the data-sheet.
- [ ] Please drop the word 'package' from the description, also `42-V` should be `42V` (without the dash)

jamesrlucas:
All done!

Travis didn't run when I previously submitted, so it didn't pick up the pin name offsets / lengths. All sorted now though.

PR #2577, opened 2020-03-26
Added LTC4314 I2C Mutiplexer

polwel:
Added LTC4314 I2C Multiplexer.

* **LTC4314xUDC** (QFN-20)
* **LTC4314xGN** (SSOP-20)

[Datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/4314f.pdf)

image


polwel:
Hi, anything I can do to help the reviewers? I don't want to rush. Just bumping because I haven't heard from anyone 😉

cpresser:
Hi @polwel 
You can support the review-team by doing reviews as well. But on a individual PR there is not much you can do since one is not allowed to review their own contributions.
In this case I picked your contribution because I am also looking at those LTC parts for a private project. And then I was lazy for a few days.

Anyway, here is my review:
- [x] I like the layout. It conforms to KLC and is basically the same as the one in the datasheet
- [x] Pintypes are good. Usually we have SCL as input, but here that does not apply. `EP` as passive is also a good choice. It does not need to be connected and we don't want a DRC warning in that case.
- [x] Pin-Numbers are good
- [x] Name, Description and Keywords
- [ ] The part should go into the `Interface_Expansion` library. Similar parts are already there (`PCA9544AD`, `TCA9548APWR`)
- [x] Footprints and filters are good

polwel:
Thanks @cpresser . No need to apologize. I appreciate your effort!

The `Interface_Expansion` library slipped my attention. I put the part in `Interface` because there I found another part of `LTC43xx` line of products (the 4317 to be precise). I will correct this shortly.

cpresser:
IMHO the 4317 should be moved as well. But that's a breaking change and should only be done once we are nearing the v6.0 release.

Out of curiosity, have you used that device yet? Does it support clock-stretching? I could not find that in the datasheet. But would assume yes, when looking at Figure10.

polwel:
Thanks for merging!

I agree, looking at it, the 4317 is out of place. I can open a ticket for that.

No, I haven't given the 4314 a try yet :)

I certainly expect it to support clock stretching. AFAICT, the SDA and SCL paths are identical (except for the fault recovery), so SCL forced low by a slave should be propagated to the master just fine. As you say, it should be possible for the master to be located on either the up- or downstream side of the switch.

PR #2586, opened 2020-03-28
Add split version of G5V-2 relay symbol

evanshultz:
https://omronfs.omron.com/en_US/ecb/products/pdf/en-g5v_2.pdf
image
image
image

This is simply a broken-up version of the existing G5V-2 relay symbol. Allows better placement options in some cases.

Three notes:
1. This is like the existing `FRT5_separated` symbol.
1. We use the `_Split` suffix in some other places, such as the Transformer library. Since that was shorter and more common I used that suffix here.
1. There are many relays that are essentially the same as this one. [Finder 30.22](https://www.finder-relais.net/en/finder-relays-series-30.pdf) is already in the library. [Panasonic DS](https://www.panasonic-electric-works.com/pew/eu/downloads/ds_61005_en_ds.pdf) as another. I wonder if it would be worthwhile to make generic symbols and footprints for common relays? @poeschlr already provided a standard for relay pin naming/numbering. I think naming for symbols and footprints has been settled already (the generic footprint simply leaves off manufacturer info).

---


myfreescalewebpage:
Hello @evanshultz 
Not bad to have this kind of possibility!
No comment on the symbol.
Joel

PR #1796, opened 2019-04-29
Mpl3115a2

jneiva08:
Add NXP MPL3115A2

Datasheet:https://www.nxp.com/docs/en/data-sheet/MPL3115A2.pdf

Symbol:
mpl3115

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

PR #2591, opened 2020-03-30
Add MPL3115A2

ottojo:
* Symbol based on MPL115A1
* [Datasheet](https://www.nxp.com/docs/en/data-sheet/MPL3115A2.pdf)

Travis seems to not work right now, this is the output of checklib.py:
```
Checking symbol 'MPL3115A2':
  Violating S3.1
    Origin is centered on the middle of the symbol
    > Symbol slightly off-center
    >   Center calculated @ (50, 50)
```
I would prefer to deviate from the 100mil grid to make the symbol symmetrical and centered, but S3.1 states

> For non-symmetrical symbols, or symbols where this requirement would move pins off the 100mil grid, then the symbol should be placed as close to the origin as possible while observing the pin grid alignment requirements.

Screenshot from 2020-03-30 20-11-37

---


ottojo:
didnt see that #1796 was merged, closing this

PR #2592, opened 2020-03-30
Add more generic ground symbols

evanshultz:
Often there may multiple grounds separated just for noise reasons. There could be a few ohms or a bead between them, for example. They aren't specifically GNDA or GNDD or GNDS, just a partition of a common ground. So here are a three symbols to assist with that.
image

---


power.lib
```diff
@@ -920,7 +920,7 @@ ENDDEF
 #
 # Earth
 #
-DEF ~Earth #PWR 0 0 Y Y 1 F P
+DEF Earth #PWR 0 0 Y Y 1 F P
```
cpresser: Can you explain why the `~ ` got removec

evanshultz: That was an old legacy naming convention from, I believe, KiCad v4. Probably this library hasn't been touched in a while so things were updated during the save in the UI.

cpresser:
Those 3 `~' are the only ones left in power.lib
There are a few more in Graphics.lib. And I confirmed that saving with 2.4 format removes them.

PR #2022, opened 2019-07-26
Add WE-TVS-82400102 TVS Diode

cp-aquila:
Add WE-TVS-82400102 TVS Diode

Datasheet: https://katalog.we-online.de/pbs/datasheet/82400102.pdf
grafik
grafik

This is pretty similar to USBLC6-2SC6, but has an additional diode, so I made a new symbol.

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] An example screenshot image is very helpful
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [ ] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [ ] Give a reason behind any intentional library convention rule violation.

PR #2598, opened 2020-04-01
USBLC6 ESD protection diodes

vlunyov:
Added very low capacitance ESD protection USBLC6-4SC6 (SOT-23-6) for 4 data-line from ST:
https://www.st.com/resource/en/datasheet/usblc6-4.pdf
USBLC6-4SC6

Added very low capacitance ESD protection USBLC6-2P6 (SOT-666) for 2 data-line from ST:
https://www.st.com/resource/en/datasheet/usblc6-2.pdf
USBLC6-2P6

Fixed URL to documentation and description of the USBLC6-2SC6 (SOT-23-6).

Added Automotive aliases USBLC6-2SC6Y (SOT-23-6) and USBLC6-4SC6Y (SOT-23-6).

So, this is all ESD data line protection USBLC series components from ST.

---


vlunyov:
A am going to pull request onlu for two last commits. Why the pull request contains 30 commits? How I can fix this?

myfreescalewebpage:
You need to rebase your branch. https://github.com/KiCad/kicad-symbols/pull/2598/files should display only you modifications.

vlunyov:
There are three violations:
Violating S4.4
     - Pin GND (2) @ (0,-500) is of type PASSIVE
But, in this case more correctly to keep the Pin electrical type as PASSIVE, because USBLC6-xxx components are passive components contains only diodes.

cp-aquila:
The pin-lengths look like they are to long. They should be 100mils.
That can also be fixed for the old part without breaking compatibility. Just increase the body size to accomodate for shorter pins.

For reference, I also added a similar part in the past (PR still pending) that has a different style:
https://github.com/KiCad/kicad-symbols/pull/2022

vlunyov:
When pin lengths reduced to 100mil pin names intersects with component body:
image
So, I can remove pin names or uncheck the Show pin names checkbox.
Unchecking the checkbox is preferred in my view.

cp-aquila:
Ah, I guess that was the original reason for the long pins.
I agree that unchecking show-pin-names is a good solution.

vlunyov:
I agree that the solution looks better with pin length 100mil and without pin names:
image

cpresser:
All right, here is my review:
- [ ] Please remove the `Y` variant. We don't include them
- [x] Footprints and filters are good
- [x] Description and Keywords are good
- [x] Layout, pin-types are good

rs-development:
There has been a time where a screenshot of the symbol in use had to be provided for a pull request. Don't get me wrong, thanks for contributing! But this seems unusable for me. How is it supposed to be integrated into the schematic flow?

Maybe someone can provide an example screenshot, thanks!

cpresser:
I am not aware of the 'screenshot in use rule'. I agree it is use-full to see how a symbol is to be used. As part of my review I ask for them if I have questions regarding the usage of symbols.

In this PR, the pinout-style was chosen because of *legacy* and *non breaking*.
Example on how this symbol can be used. Its ugly, but that the way it is.
image

If you want, you can use the symbol of `WE-TVS-82400102` instead. It has the same pinout but a different, more convenient style. I am saying that, because I designed it :grin: https://github.com/KiCad/kicad-symbols/pull/2022

rs-development:
It is always legacy :1st_place_medal: 

Your `WE-TVS-82400102` looks better for me, thanks for the tip!

PR #2618, opened 2020-04-07
SGTL5000XNAA3 Symbol

johnfwhitmore:
Added symbol for NXP SGTL5000. 

https://www.nxp.com/docs/en/data-sheet/SGTL5000.pdf

I've initially added the SGTL5000XNAA3, the 32QFN package just to go through the CI tests here. Once they check out and everything is in order I'll add the SGTL5000XNLA3 (20 Pin Package).

Screenshot from 2020-04-07 13-14-37


johnfwhitmore:
Screenshot from 2020-04-07 13-56-08

myfreescalewebpage:
Hi @johnfwhitmore , thanks for contributing,

SGTL5000XNAA3 (QFN-32):
- [ ] Description should ends with package name: `Low Power Stereo Codec with Headphone Amp, QFN-32`
- [ ] Footprint filter should be `QFN*1EP*5x5mm*P0.5mm*`
- [ ] Use capital letters for pin names
- [ ] VDDA should be Power Input
- [ ] You should group I2S pins together
- [ ] You can align LINE IN with LINE OUT for example
- [ ] VDDIO should be on the top left (Digital IO on the left, Analog IO on the right is great)
- [ ] Name of pin "VAG" is "VAG", not "VGA" and this pin should be Passive

QFN-20 version : same comments apply.

Cheers,
Joel

johnfwhitmore:
Thanks a million for your comments, and sorry about those mistakes. I did have both NC and Gnd pins stacked but got an error for the NC pins and unstacked everything. I'll sort those issues out now.

Thanks again

johnfwhitmore:
There's an error in the changes I've made, but I can't currently get to the Travis output as there's some error with it. I'll check back later and fix what ever the problem is.

myfreescalewebpage:
Thanks for the fixes.

- [ ] For the error on SGTL5000XNAA3 device, this is about pin stacking. The rule is the following: First pin visible, power input. Following pins invisible, Passive.
You just have to set pins 3 and 33 Passive to fix the error.

- [ ] Last comment : for both symbol can you center GND pin and align GND with VDDA?

Joel

johnfwhitmore:
Sorry about the delay on that correction

myfreescalewebpage:

PR #2619, opened 2020-04-07
Fix pin number in PESD3V3L4UG and PESD3V3L4UW

scibi:
Changed pin 6 to 5  in PESD3V3L4UG and PESD3V3L4UW. There is no pin 6 in SOT-353 and SOT-665.

Datasheet: https://assets.nexperia.com/documents/data-sheet/PESDXL4UF_G_W.pdf

image


evanshultz:
I believe I added those symbols so this was probably my mistake. Thanks for noticing and submitting the fix!

PR #1504, opened 2019-02-07
MCU Module: Create VisionSOM-RT

myfreescalewebpage:
Creation of VisionSOM-RT module from SomLabs.

500px-visionsom-rt-v_1_1

The module is pin-to-pin compatible with the VisionSOM-6UL I have previously created (https://github.com/KiCad/kicad-symbols/pull/1195), except that this one has no CSI interface (pins are NC) and that 4 GND pins are not used (they are NC too). This is established comparing the pinout at https://wiki.somlabs.com/index.php/VisionSOM-6ULL_Datasheet_and_Pinout and https://wiki.somlabs.com/index.php/VisionSOM-RT_Datasheet_and_Pinout.

For this module, I first copied the VisionSOM-6UL symbol and performed these modifications + description/doc link.

Note: there is no PDF datasheet available, I have proposed the SomLabs Wiki page, on chich the pinout is available.

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing

https://wiki.somlabs.com/index.php/VisionSOM-RT_Datasheet_and_Pinout

- [x] An example screenshot image is very helpful

capture

- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.

The module is inserted in a SODIMM connector. Strategy discussed for the VisionSOM-6UL module kept.

- [x] If there are matching footprint PRs, provide link(s) as appropriate

No PR.

- [x] Check the output of the Travis automated check scripts - fix any errors as required

No errors at this time.

- [x] Give a reason behind any intentional library convention rule violation.

No rule violation at this time except the footprint filter, see strategy applies for the VisionSOM-6UL module.

PR #2628, opened 2020-04-12
MCU Module: Add VisionSOM-STM32MP1

myfreescalewebpage:
This pull request adds VisionSOM-STM32MP1 from SOMLABS.
It has been designed in the continuity of VisonSOM-6UL/6ULL and VisionSOM-RT (I was also the author of these references, merged by @evanshultz at #1195 and #1504).

The power pins are the same (some GND are added), but multiple other functional pins have been modified so I haven't keep the possibility to make the symbol interchangeable.

Capture

There is no PDF datasheet available, reference is https://wiki.somlabs.com/index.php/VisionSOM-STM32MP1_Datasheet_and_Pinout

---


aris-kimi:
Hi @myfreescalewebpage ,

TOP pins checked.
εικόνα

GND Pins checked.

BOOT0/1/2 checked.

DSI- pins checked.

ENET pins checked.

GPIO pins checked.

JTAG pins checked. Question, `-nRST` that `n` declares negative logic?

LCD pins checked.

SD2 pins checked.

USB pins checked.

UART pins checked.

USART pins checked.

For the first round i left NC pins while deleting pins that passed check.
εικόνα

Pin types also checked and i think they are ok.

NC pins checked.

From similar merged PRs i saw no footprint so i think this follows same.

Only one suggestion, i think  it would be better with `V2.1 ` added in the description.

And one question about symbol name, `VisionSOM-STM32MP1` i think it would be better at keywords field with existing text.
Wouldn't be better to add naming from ordering info table with aliases?

εικόνα

I would like some feedback for this. I think it can still be merged and without applied suggestions.

This was a great experience i have to admit.

PR #2632, opened 2020-04-13
Add dual diodes in two new symbol styles

evanshultz:
At https://github.com/KiCad/kicad-library/issues/986 I had proposed new schematic symbols for dual diodes. I see general acceptance there but this never got finished before the library migrated. So, let's try again.

I've started with just two symbols as an example, based off the latest master, so show what it would look like. I think you can image what other variations would look like. If I also see support for one of both of these dual diode styles I'll make the other symbol versions.

image

The naming of this version seems challenging. What term is most obvious? I used `_Parallel` because the diode symbols are parallel with each other (as opposed to the existing symbol which has the diodes shown in series). But is there something better?

image
image

My one real concern is how to show the 'duplicate' pin on the `_Split` version. Leaving it missing seems bad and takes away the ability to show the electrical connection between the two diodes which we assume are going to be separated in the schematic, but is the fake 'pin' I've shown here confusing?

---


evanshultz:
All other versions are done, except for dual TVS which seem to uncommon in a generic form and not worth the time. We have many specialized symbols for voltage protection devices which I think is the proper way to go. Let me know if I'm wrong and they, or any other variant, needs to be added. Or if you have other questions or comments.

myfreescalewebpage:
Hello @evanshultz 

Really a great job here ! Is it scripted ?

Just a single question: Some `D_x2_*`  diodes have "schottky" in the description, is it expected ?

Joel

myfreescalewebpage:
@evanshultz ping

evanshultz:
It isn't scripted. But I think our primitive symbols (diode, FET, BJT, etc.) should be. Maybe after v6. :)

Yep. Some descriptions were messed up. I think they're fixed now.

myfreescalewebpage:
Thanks, I think it's correct.
Merging

PR #2647, opened 2020-04-17
Added Microchip PIC18F2xK22 I/P microcontrollers.

RICCIARDI-Adrien:
Added Microchip PIC18F2xK22 I/P microcontrollers. Includes PIC18F23K22 I/P, PIC18F24K22 I/P, PIC18F25K22 I/P, PIC18F26K22 I/P.

---

http://ww1.microchip.com/downloads/en/DeviceDoc/40001412G.pdf
Capture d’écran_2020-04-17_18-18-58

- [NA] If there are matching footprint PRs, provide link(s) as appropriate
- [NA] Give a reason behind any intentional library convention rule violation.


diegoherranz:
Thanks for your contribution.


- [ ] Naming: I think that the package suffix should be `SP` for SPDIP and not `P` for PDIP (see "Add to cart" on https://www.microchip.com/wwwproducts/en/PIC18F23K22). There is a dash and not an underscore on the original part number. Also, the temperature rating is replaced with a wildcard (see [KLC S2.2](https://www.kicad-pcb.org/libraries/klc/S2.2/)). The parts should be called `PIC18F23K22-xSP`, `PIC18F24K22-xSP`, etc.
- [ ] VSS pins should be stacked (see [KLC S4.3](https://www.kicad-pcb.org/libraries/klc/S4.3/))
- [ ] MCLR is input only, even when working as GPIO, more like GPI ;)
- [ ] It was agreed for Microcontrollers in the library, that we wouldn't list all the alternate functions of the pins because the symbol gets very verbose. However, for PICs it was accepted to include the alternate functions of the most important ones, e.g. oscillator, reset, debugging. You can see `PIC18F25K83-xSP` as a reference. Could you do something similar?
- [ ] I would place the MCLR pin on the top left. You can move RCx to the right hand side to avoid making the symbol longer. Again, you can use `PIC18F25K83-xSP` as a reference.
- [ ] Can you add `B` (bytes) as applicable in the descriptions?

I will review the pinout carefully after these changes (given that the pins will be modified).

Thanks again for your contribution and I hope it is the first of many!

RICCIARDI-Adrien:
Thank you for your precise answer. I should have fixed all the issues !

diegoherranz:
Looks OK. Just a minor one, to be closer to the datasheet, I think would use the following pin names:

- ~MCLR/RE3 or ~MCLR/VPP/RE3
- RB7/PGD
- RB6/PGC

That is, the Rxy first, except for the MCLR pin.
Could you change that?
Thanks.

diegoherranz:
- Pinout looks correct
- Footprint and footprint filter look OK

Thanks for the quick fix. Merging.

PR #2651, opened 2020-04-18
Added TP4056 Li-Ion charger symbol

ppielorz:
Hi,
I've prepared TP4056 symbol but I didn't find manafacturer webpage... And I found two different datasheets: one in English and one in Chinese I think. Only the second one specifies footprint and it is probably HTSOP-8-1EP_3.9x4.9mm_P1.27mm_EP2.4x3.2mm.
I've placed TEMP input on the right side because it would be easier to create schematic.

Datasheet: https://download.kamami.pl/p562208-TP4056.pdf and https://datasheet.lcsc.com/szlcsc/1912111437_TPOWER-TP4056_C382139.pdf (second one regarding footprint)

Screenshot: 
image

---


myfreescalewebpage:
Hi @ppielorz , thanks for contributing,

- [ ] You have to set the footprint field. It is SOP-8, not HTSOP. If you don't find the exact footprint in the library, you should submit it.
- [ ] The footprint filter is not correct (see travis log)
- [ ] Use capital letters for pin name VCC
- [ ] PROG is probably better at the bottom of the symbol to simply wire the resistor.
- [ ] PROG should be Passive
- [ ] CHRG and STDBY probably better on the left as shown on the example schematic
- [ ] CHRG and STDBY should be Open Colelctor
- [ ] Center the GND pin
- [ ] Pin 9 should be called "EP" and should be on the bottom

Cheers,
Joel

myfreescalewebpage:
@ppielorz ping

PR #2655, opened 2020-04-19
Fix generic USB connector pin types

### LINKED ISSUE #2520: USB3 B connector has incorrect pin power type
The USB3 B type connector symbol annotates the VBUS & GND pins a "Power Input" whereas they should be "Power Output". The USB2 B type connector has this correct.
(Annotating these pins as "Power Output" has been described as the correct annotation in issue #989)

Much less interesting, but still seemingly incorrect, the connector drawing on the symbol reflects a type A connector instead of a B type connector. The USB2 B type symbol has this correct.

Lastly, a question: is there a reason the SS Tx/Rx lines are labeled Bidirectional instead of Input/Output ?

chschlue:
* Fix USB3_B drawing

Fixes #2520

USB3_A
USB3_B
USB_A
USB_B
USB_B_Micro
USB_OTG

---


evanshultz:
Thanks! If I read the issue correctly the SSRX/TX pins should stay type B for alternate mode use. Otherwise it seems OK to me.

chschlue:
I read it to mean bidirectional for USB **C** alternate modes.
AFAIK, USB 3 over A/B connectors has no alternate modes. Not completely sure about that, though.

diegoherranz:
> I read it to mean bidirectional for USB C alternate modes.

Yes, only on type C as far as I'm aware. However, please note that the direction (RX/TX) is always described from the host perspective, so I think that the USB3_A is wrong now.

Thanks!

chschlue:
Right.
Fixed.

diegoherranz:
Taking advantage of this PR, a couple more comments:
- On every USB C symbol: VBUS is power input or output since power can be reversed. Should they be passive? What about GND in that case?
- Similar for USB_OTG, VBUS can be power input or output. In fact, given we have separate USB_OTG and USB_B_Mini/Micro symbols, I think that the OTG one should allow power in both directions. Otherwise, I don't see any point in having different symbols.

Thanks!

chschlue:
So we basically have three options:

1. Power output: Works if used on a device, throws false errors if used on a host.
2. Power input: Works if used on a host, works but requires power flags and largely defeats ERC if used on a device.
3. Passive: Works if used on a host, works but requires power flags and completely defeats ERC if used on a device.

(I know the "host supplies power" / "device draws power" concept is obsolete, but I think you get the idea.)

So going with power input seems to be the least ugly solution. On the other hand, one could argue that every _connector_ pin is passive per se.

aris-kimi:
Hi there @chschlue @diegoherranz and @evanshultz , if i may share an idea, two symbols with sink/source in the description or in the symbol's name, in order to show the difference, from the "pcb perspective"..

diegoherranz:
Hi, @aris-kimi. The thing with USB type-C is that the power direction is dynamic and it is very common to swap on the fly, so I think we need a single symbol, but something that works in both power directions. Same for the USB_OTG.

I think I would go for passive because I find that more clear than "Power input" or "Power output" given that you may need to use the opposite power direction which may look weird in the schematic. But there's probably no ideal solution. We would need a "Bidirectional power" type :)

chschlue:
Changed to passive for now.

diegoherranz:
Thanks, @chschlue.
Anyway, that's only my point of view. I leave it to @evanshultz to decide.

evanshultz:
Sounds fine to me. Thanks all, and apologies for the delay.

chschlue:
@evanshultz are you going to cherry pick this to v5?

evanshultz:
Sounds like you want it in the v5 branch. If you want to do it, go ahead. If you want me to do it, I can figure it out. It's up to you.

chschlue:
Thinking about it again, it's probably easier to cherry pick all bug fixes at once right before tagging 5.1.7.

@KiCad/librarians please make sure bug fixes are added to the 5.1.7 milestone.

PR #2657, opened 2020-04-19
Add symbol for Quectel L80 GPS module

thatoddmailbox:
Datasheet: https://www.quectel.com/UploadImage/Downlad/Quectel_L80-R_Hardware_Design_V1.2.pdf

Screenshot: 
image

I've made a new footprint for this too: KiCad/kicad-footprints#2228.

---

- [X] Check the output of the Travis automated check scripts - fix any errors as required - somehow Travis won't let me check the log, but I ran the script on my computer, and there were no errors - I think the Travis error is because the footprint is not merged 


myfreescalewebpage:
Hi @thatoddmailbox , thanks for contributing,

- [ ] Use CAPITAL letters for pin name
- [ ] NC pin should be moved on the border outline of the symbol
- [ ] Reserved pin should be Visible, type Passive, so they are available if a user has a specific application. You should group them, for example of the left (and move RESET on the right). I propose the following:

Capture

Else just need to wait the footprint.

Cheers,
Joel

thatoddmailbox:
Hi @myfreescalewebpage, thanks for the feedback! I made the changes you suggested, screenshot below:
image

I did make the symbol larger than in your suggestion because I thought it looked a little cramped to have the VCC and RESERVED text right next to each other. However, this meant that I had to push both the top and bottom out by 100 mils (to keep the VCC and GND pins on the 100 mil grid), so now it's a little large...I can change it back if you think it's better.

myfreescalewebpage:
The size is great. no more comment, just need to wait the footprint.

myfreescalewebpage:

myfreescalewebpage:
Merging following integration of the fp

PR #2660, opened 2020-04-20
add l7980 switching regulator

matthuszagh:
Adds [L7980](https://www.st.com/resource/en/datasheet/l7980.pdf) switching regulator.

l7980

[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/2233)

---


cpresser:
* Pin-pos-offset can be 20mil
* Since you added the HSOP-Part the correct name is `L7980A`
* I am not sure if the pin positions arre good. Can you show how you wire that part up in the schematic?

matthuszagh:
Here's an example I'm using.

l7980a

Here's what the datasheet gives for comparison.

l7980a

I agree... the pin positions are a bit awkward. Maybe switching FSW and COMP would help?

matthuszagh:
@cpresser I went ahead and made these changes, as well as switching pins 4 and 6. Let me know what you think.

l7980a

cpresser:
Sorry for the huge delay. I was lacking motivation to do reviews for a few months.

Anyway, I gave this another look:
- [ ] FSW is an input
- [ ] Layout: EN could be the top pin on the left side to avoid crossing the sync-pin if used. You are also allowed to put VCC on the left side as an exception to KLC S4.2 if you like the left-to-right style.
- [ ] I am not sure about the exposed pad. Is there a use-case where it is not connected to GND? I could not find one in the datasheet. The rule that applies here is 'if the manufacturer says its GND, stack it with the other GND pins'. What do you think?
- [ ] The footprint is missing. Is there a PR for that one? 
- [ ] Please use the non-thermal-vias footprint as default footprint.

matthuszagh:
@cpresser I've implemented your changes. I've also gone ahead and tied the EP to GND, since I can't think of a use case where this shouldn't be done and the datasheet explicitly mentions the benefit of tying it to gnd.

[footprint pr](https://github.com/KiCad/kicad-footprints/pull/2233)

l7980a

cpresser:
If you like you can put VCC and EN on the left side. Travis-CI will complain, but we explicitly allow that style for voltage regulators. For example this one
image

This is not mandatory. We already have both styles in the library, so we can't aim for consistency.

chschlue:
FP merged.

matthuszagh:
@cpresser yeah I think I actually prefer that.

sch

cpresser:
Almost :)
The Pins on the right side are not on the 100mil grid.Travis-CI also complained about that. I suggest to move `OUT` up so it is on the same height as `VCC`. Also move the `COMP` pin (I suggest down).

matthuszagh:
@cpresser fixed now. Thanks for catching that.

PR #2661, opened 2020-04-21
hmc431: NC pins should be tied to ground

matthuszagh:
Gerber files [here](https://www.analog.com/en/design-center/evaluation-hardware-and-software/evaluation-boards-kits/eval-hmc431lp4.html) show no-connect pins tied to center exposed pad and surrounding ground plane.

hmc431


chschlue:
The (spartan) datasheet doesn't mention anything of the sort and the (Hittite) eval PCB drawing also appears to show nothing but a guard ring, yet the newer eval board gerber seems plausible.

I would suggesting creating an NC pin stack (seperate from pins 15/25) to leave it to the user.

matthuszagh:
That seems sensible to me. I've also added some vertical space so the pin names have more room with the new NC pin. Let me know your thoughts. Here's the new symbol

hmc431

evanshultz:
Changing the pin locations isn't allowable during the KiCad v5 release. If they do, the symbol will change in schematics already designed and we won't break schematics since there is a way to add these NC pins without doing so. Please revert or undo those changes so the pins don't move. Thank you.

matthuszagh:
Done.

hmc431

chschlue:
Thanks.
Please change NC pin types to passive to avoid ERC errors.

matthuszagh:
Done as well.

PR #2664, opened 2020-04-22
Revert "Fix for LM4040 TO-92, pin 3 (NC) was used as anode (instead of pin 1)"

martinjankoehler:
Reverts KiCad/kicad-symbols#2663

Sorry, I must have overlooked the word "Bottom" in the data sheet's image caption,
in combination with KiCad's TO-92 footprint pin numbering  differing from TI's.

KiCad's original pinout was actually ok.

I tested it now with a LM4040CIZ10 10V reference to verify.

Bildschirmfoto 2020-04-22 um 15 10 31

image


chschlue:
Oops.

PR #2667, opened 2020-04-23
Add split resistor network symbols

evanshultz:
R_Network*_Split units A and B showing the 'dummy' pin on units B and above:
image
image

R_Pack*_SIP_Split units A and B:
image
image

R_Pack*_Split:
image
image

The R_Network* symbols really should have 'SIP' in their name because they are only for pinouts in a SIP package. There are DIP and SOIC and convex/concave packages with bussed resistor networks, but they come in other pinouts. See https://www.bourns.com/docs/Product-Datasheets/4100R.pdf and http://www.vishay.com/docs/31511/31511.pdf and https://www.ctscorp.com/wp-content/uploads/766.pdf for DIP/SOIC, and then https://www.seielect.com/catalog/sei-ravf.pdf and https://industrial.panasonic.com/cdbs/www-data/pdf/AOC0000/AOC0000C12.pdf and https://www.ctscorp.com/wp-content/uploads/74x.pdf for concave/convex packages. This is just to show that while these symbol names don't include the word SIP, they have a SIP FP filter and we would need symbols with different pinouts to cover non-SIP packages.

Note that R_Pack04_Split already existed, which oddly was the only one, and it changes notably just in having a RN ref des prefix like all other resistor networks and now has pin numbers visible.

I didn't do the US resistor shapes but by the time I finished pasting these sections into the huge Device.lib file it probably would have been easier...

Script PR: https://github.com/KiCad/kicad-library-utils/pull/325

---


evanshultz:
You know, I will go ahead and do the R_Network*_US symbols just to get them done. But after sleeping. And then I'll upload the DCM file which I forgot.

evanshultz:
Added the US symbols. The zigzag lines are slightly different than before, but very similar and are the same size as the non-US resistor rectangles (old on left, new on right):
image

I noticed that the non-US ones allowed the pin numbers to crash into the symbol body so they are updated too (again new is on the right):
image

That was buggish but not enough to mark this as a bug.

Also, the number of lines in the non-US resistors is getting annoying. I could have used more complex polylines that overlap with the resistor rectangles to reduce the number of lines in each symbol, but the added number of points in the polyline seemed like overall not worth the effort.

In all cases the pin locations didn't change.

evanshultz:
It looks like I missed a period in my email address so CLA is barfing. While Gmail doesn't care about that apparently the CLA does. What is the way to proceed? Obviously it's safe to say I agree.

chschlue:
Yes, that's probably safe to say.
But can't you add both variants to your GH account? Just to avoid the hassle in the future?

evanshultz:
I could, but since I just made a typo when setting up the Git client it should be OK in the future. Until I set up a new PC and my brain switches off again.

BTW, the split resistors with pin numbers having double digits are a bit ugly, but since they're 50mil long pins I don't see other options. Here is R_Pack11Split unit K:
image

cpresser:
For reference, the one existing R_Pack_split was added here https://github.com/KiCad/kicad-symbols/pull/2109 At the time I was not thinking of adding more scripted variants.

Regarding the email-address, you can do a interactive rebase to do a `--reset-author`. And force push.

myfreescalewebpage:
@evanshultz no comment to do.
Check if you can solve the mail address for the next issues :)
Merging.

PR #2677, opened 2020-04-27
Stack output and GND pins of TB6612

poeschlr:
Output pins that are always connected must be stacked for best ERC
results.
Stacked ground pins in sensible groupings.

Ensured at one pin of every signal is still on the same position as
before.

https://toshiba.semicon-storage.com/info/docget.jsp?did=10660&prodName=TB6612FNG
https://forum.kicad.info/t/conflict-problem-between-pins-severity-error-yet-again/22521

I could imagine that all power ground pins could be stacked. The datasheet is quite bad for this part so i can not really tell. Similarly with VM2 and VM3 possibly also VM1

pin list

Screenshot from 2020-04-27 22-46-51


vishnumaiea:
Changing the pin types got rid of the original ERC error. But hiding pins has a problem. Each pair of pins are there to handle larger currents. If you hide them, the user can not use them by default.

For example, I changed all pin types here and hidden one of the AO1 pin (pin 2).

image

Then the PCB layout will show,

image

Now obviously the pin 2 is not connected to anything.

poeschlr:
Check the symbol i made! It is different to what you did. I did not hide the pins i stacked them. See https://www.kicad-pcb.org/libraries/klc/S4.3/ and of course https://forum.kicad.info/t/electrical-type-of-schematic-symbol-pins/9439

vishnumaiea:
Got it! Thanks.

myfreescalewebpage:
Hello @poeschlr thanks for this enhancements

The modifications are breaking the design of schematic so it will be for v6.
I propose if you have time to fix the other KLC issues on this symbol (style of pin, length of pins, and also travis errors).

Else no comment on the modifications done to stack the pins.

Joel

evanshultz:
@myfreescalewebpage 
This is a bug because when connected properly in a schematic errors are creating. The pins are an incorrect type and this is the fix. On top of that, Rene stacked pins but made sure at least one of pins is still in the original location. So there could be floating bits of wire on a schematic but the ERC issues will be fixed and the symbol gets pins stacked as well which helps with new users of the schematic. But fixing pin types is the major error and why this is to be done now.

@poeschlr 
Thank you.

myfreescalewebpage:
It's right!

PR #2684, opened 2020-04-29
trf37a73: stack no-connect pins to allow connecting to ground

matthuszagh:
[Datasheet](https://www.ti.com/lit/ds/symlink/trf37a73.pdf)

No-connects are now passive and stacked, which allows either leaving
them floating or connecting to ground. The datasheet shows a ground
connection.

trf37a73

----

#

---


myfreescalewebpage:
Hello @matthuszagh the datasheet recommends to connect NC pins to the ground so I think we can reasonably stack them with GND. What do you think ?

matthuszagh:
@myfreescalewebpage yeah that's fine by me. I can't think of any reason why someone wouldn't want them tied to ground. I was basing this on the feedback I'd received on [this PR](https://github.com/KiCad/kicad-symbols/pull/2661), but I guess in that case the decision was more ambiguous. What's the general policy here? If the datasheet recommends a ground connection do that, otherwise stack them separately and leave it to the user?

myfreescalewebpage:
Yes if the datasheet indicates it's should be connected together, it's good to stack them. This is particularly true for devices on which the expose pad is connected to GND for example, and in this case we stack the EP pin with GND pin on the symbol.

evanshultz:
The datasheet indicates these pins are unconnected (not internally connected to ground) and does not say they are required to be connected to ground. So I would leave them as separate pins.

I could also see an argument to not stacking the NC pins since they are spread out around the chip and it may be desirable to not connect them all to ground. Since the EP is so close this doesn't really seem likely, but this part is outside my expertise so I'm not really sure.

myfreescalewebpage:
@evanshultz and @matthuszagh it's sure there are only two possibilities:
- All NC stacked with GND if we want to stack
- All NC pin not stacked at all
I agree we cannot stack NC pins together and just let it like that.

@evanshultz the datasheet indicates page 3:

> No electrical connection. Connect pad to GND for board level reliability integrity.

That's why I suggest to stack NC pins with GND.

Joel

matthuszagh:
@evanshultz's comment is making me question my original decision. If there is a reason to selectively not connect some of the pins to ground, which seems plausible, then its best to require a bit more work from the user instead of preventing them from using their desired, valid use case right?

Unfortunately I don't have much experience with this. I imagine other people will have a much better idea of the proper course of action here. But, given the general tradeoff of interpreting the use case and making the workflow slightly easier vs not getting in the user's way, I'll always opt for not getting in the user's way.

evanshultz:
@myfreescalewebpage 
I saw that. It does not say they must always be grounded, but only suggests grounding for strength. This is the rationale we've used on other symbols to use a separate NC pin. Only if it's declared that NC pins must be stacked are they stacked.

myfreescalewebpage:
Yes I agree. Actually the symbol is the following:

Capture

Any advice to place the NC pins (I know you have a large experience with the triangle symbols) ?

Joel

myfreescalewebpage:
@evanshultz ping, please can you have a look to my last comment on this pr ? Thanks!

evanshultz:
I don't see enough locations to nicely place 4 NC pins on the triangle. A separate unit with just the 4 NC pins on the bottom is an option, although a bit ugly. We do have a larger triangle symbol used in some of the power amp parts. I slightly prefer the 2 unit solution, but since we don't have a guideline or precedent I think it's best to choose a solution that you are happy with, @matthuszagh , and get this merged.

Also, should the PWDN pin by Input type?

myfreescalewebpage:
I agree a unit with only NC pins is ugly, this is not my preferred solution.
@matthuszagh maybe NC pins vertical on the left of GND is great ?
PWDN can be left floating so I think Passive is better.

myfreescalewebpage:
@matthuszagh ping
Thanks
Joel

matthuszagh:
@myfreescalewebpage thanks for the reminder. I personally really dislike 2-unit solutions, but I guess that's just a personal preference. Your idea is to stack them to the left of the bottom gnd pin? Yeah that seems fine.

myfreescalewebpage:
Yes exactly.

matthuszagh:
Hold on, wasn't @evanshultz's message above that these shouldn't be stacked? Isn't that why the choice is either to make the triangle bigger or use a second unit?

myfreescalewebpage:
We accepted recently another device with stacked NC pin because the datasheet recommended to connect them together to a fix voltage. Very similar to the device of this PR.

matthuszagh:
So evanshultz's comments above that NC pins should only be stacked if the datasheet says they must be stacked is no longer valid? This seems reasonable to me because this gives the user the option to connect the pins or not. @evanshultz want to weigh in here?

myfreescalewebpage:
> This seems reasonable to me because this gives the user the option to connect the pins or not. 

This is the conclusion we have in the other symbol I was referring to. Sorry I don't remember the PR number.

matthuszagh:
Ok.. So then we shouldn't stack the NC pins? The "this" in "This seems reasonable..." referred to not stacking pins. I'm not sure we're on the same page here. I thought Evan's comment was pretty clear about not stacking.

I think this would be a good candidate for updating the KLC. The current recommendation is:
> Pins must not be of type No Connect (these pins should never be connected together)

which doesn't seem to be the prevailing wisdom.

myfreescalewebpage:
How damm, yes we are not on the same line.
I'm looking at the symbol we have accepted recently and I have found this one: https://github.com/KiCad/kicad-symbols/pull/2661

evanshultz:
Yes, I am saying that unless the pins must be all connected together, they should be separate pins. That is the only way to give the user control to implement the part in any way that is acceptable. If the vendor doesn't require a certain connection, then the user must be free to connect the pins and this requires separate pins. There way be another way but that is the conclusion we reached in the past.

matthuszagh:
@evanshultz ok I'll either implement the NCs as a separate unit or part of a larger triangle symbol then. Since there isn't a standard procedure for whether to use a separate unit, is there a way we can try to standardize this in the KLC to avoid future ambiguity and promote consistency in the library? I also think the KLC section I linked above should be updated to reflect this. I'm happy to create PRs for this, but I'd want further input from the community about the separate unit vs. larger symbol decision. I personally dislike separate units since I find them visually confusing, but I also appreciate large triangle symbols aren't so nice either. Maybe we should open an issue to field opinions?

myfreescalewebpage:
I vote for the large triangle. I also don't like the solution with one unit only with NC pins. They are visually invisible, and thus this will make visually a unit with no pin on the schematic !

matthuszagh:
Here's a screenshot of the new version. Let me know what you think.

trf37a73

myfreescalewebpage:
@evanshultz what do you think ?
My opinion is not clear, it's not bad but looks... strange.
Joel

myfreescalewebpage:
@evanshultz ping

evanshultz:
Ugh. I does looks strange. But I would like to get this done for you, @matthuszagh , since you've spent so much time on this PR.

@myfreescalewebpage 
Perhaps you merge it like this to get it done? The two NC pins above RFIN could perhaps move to the bottom so all NC pins are closer to the GND pin? That really doesn't seem better, just difference. And I don't have any other ideas.

matthuszagh:
@evanshultz I'm happy to wait if you think there's a better solution. But if it's not clear what a better solution is, maybe it does make sense to just merge.

myfreescalewebpage:
I don't see a better solution right now and I agree we can merge to progress on this and permit the integration before moving to GitLab.
We can still change it later if we thing a better solution.
Joel

PR #2688, opened 2020-05-01
Add TLV320AIC3100 Audio Codec

einball:
This adds the TLV320AIC3100 audio codec to the symbol library. The datasheet can be found [here](http://www.ti.com/lit/ds/symlink/tlv320aic3100.pdf?ts=1588329811434). The footprint is a VQFN-32 with 5x5mm. the datasheet gives for the exposed pad 3.45mm with a tolerance of 0.1mm, so I used the 3.5mm EP package already present in kicad-footprints.

__Example Screenshot:__
Bildschirmfoto vom 2020-05-01 14-41-37

---


diegoherranz:
Thanks for your contribution.

Checks done which look OK:
- Symbol name matches datasheet
- Datasheet link works
- Footprint exists and I think it is close enough to the recommendation in the datasheet
- Footprint filter looks OK
- Keyworks look reasonable
- Pinout and signal names and types OK except for the notes below
- Pin arrangement looks reasonable


Fixes pending:
- [ ] Can you add the simplified footprint name (VQFN-32) at the end of the description separated by a comma? See [KLC S6.2](https://kicad-pcb.org/libraries/klc/S6.2/).
- [ ] Can you add an EP pin 33? The datasheet says that it "must be connected to analog output driver ground" so it could even be a stacked pin as per [KLC S4.3](https://kicad-pcb.org/libraries/klc/S4.3/). However, in this case, I think it's better to have the EP as a separate pin given there are different grounds.
- [ ]  Can you use Output instead of Power Output for SPKP and SPKM? Power output is only used for supply rails, regulators, etc. In this case I think that output is more adequate.
- [ ]  Can you stack the SPKP, SPKM, SPKVSS pins as per [KLC S4.3](https://kicad-pcb.org/libraries/klc/S4.3/)? The second pin for each as Passive and hidden. SPKVDD requires separate decoupling for both its pins and in those cases, we keep them separate in the symbol, so that one's OK.
- [ ] MICBIAS could be Power output.
- [ ] For I2C "slave" interfaces, we always set SDA as bidirectional and SCL as input only. SCL could behave as an output if using clock stretching (if supported) but that's the only case and for consistency with the library, can you set SCL as input?

Thanks!

diegoherranz:
Hi, @einball 
This repo is getting locked down. See https://github.com/KiCad/kicad-symbols/blob/master/README.md
Would you be able to finish this PR before the repo moves to gitlab?
Thanks!

einball:
Hi @diegoherranz,

thanks a lot for the review. I applied the corrections in the latest commit so this can be merged before the lockdown. Sorry it took so long!

Bildschirmfoto vom 2020-09-09 14-37-17

diegoherranz:
Thanks for all the changes. Looks good to me now. Merging.

PR #2689, opened 2020-05-01
ADL5801: rearrange pins for more logical interface

matthuszagh:
[Datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/ADL5801.pdf).

Datasheet describes RF pins as inputs and IF pins as
outputs. Therefore, I think it makes more sense for the inputs to go
on the left side and the outputs to go on the right
side.

ENBL is active low since pulling it low enables the internal detector
and bias circuit.

VSET is used by tying it via a resistor to the power supply, so it
makes more sense to place it near the top. Placing it on the right
side maintains stylistic compatibility with ADL5802, the dual channel
version of this chip. DETO can be tied to VSET so it should be placed
next to it.

adl5801

---


evanshultz:
We won't move pins for stylistic reasons except during major releases because that will affect existing schematics. So we can consider this for v6 but not before then.

Thanks for observing that things could be better and submitting a PR. Once we're close to v6 this PR should be reviewed and given appropriate feedback.

matthuszagh:
@evanshultz cool yeah I figured this wouldn't go into v5. Happy to leave this till then.

myfreescalewebpage:
No comment to do, can be merged when v6 will be there.
Joel

PR #2699, opened 2020-05-05
Add TPS54202

gcormier:
I modeled this after the existing `LMR33640ADDA`. I changed the pin numbers, but left them otherwise in the same position. I'm not saying there's others that aren't completely different, but I figured

1. If LMR33640ADDA is acceptable this might too.
2. It might let a designer swap out components easily if they are simplifying a circuit or switching components.

Existing LMR33640ADDA
image

New TPS54202 ( https://www.ti.com/lit/ds/symlink/tps54202.pdf )
image
image

image

---


myfreescalewebpage:
@gcormier thanks for this symbol. However, it is too large, even if it is good to have compatibility with another device, it is not justified here.
The name of your device should be "TPS54202DDC"
Finally I think you new device can be an alias of TPS54302 I think ?
Joel

gcormier:
Argh, that is a perfect alias :) I did some searching before hand for "simple switcher" and other keywords to try and find an existing one but I guess I missed it! I will close this PR and just submit a simple one with an alias.

gcormier:
#2703 open

PR #2726, opened 2020-05-14
ADM3057E isolated CAN transceiver

rleh:
*Add ADM3057E symbol*

grafik

- [x] Datasheet: https://www.analog.com/media/en/technical-documentation/data-sheets/ADM3055E-3057E.pdf
- [x] Screenshot
- [x] Footprint already exists


rleh:
> Could you convert this to our "modern" isolated bus transceiver style [...]

Like this?

grafik

rleh:
Or rather like this (with V_ISO_OUT and GND_ISO on the right side):

grafik


---


ab921f0 is shown in the upper picture.

chschlue:
Yes, I was thinking of something along those lines. Choose what's easier to wire up in the schematic.

Also, consider using passive type for the bus side "power" pins, as these aren't actually supposed to drive other power inputs or be driven by external power outputs. Passive avoids having to add power flags if used correctly and triggers ERC warnings if used incorrectly.

rleh:
@chschlue Good point setting the isolated side power pin to passive.

I think I like the variant with V_ISO_OUT and GND_ISO on the right side (see last comit 56d6aa6) better:
grafik

---

Ready to merge from my side.

rleh:
grafik
The `RS` pin belongs to the "isolated" side of the transceiver and thus should be located on the right side.

I will move it closer to the ground pins.

rleh:
grafik

rleh:
Thank you for reviewing!!!

PR #2730, opened 2020-05-15
Add AD637

evanshultz:
https://www.analog.com/media/en/technical-documentation/data-sheets/AD637.pdf

image
image

The pins aren't arranged exactly with inputs on the left and outputs on the right to be easiest to wire up:
- A cap should be between RMS_OUT and CAV
- DEN_INPUT comes from RMS_OUT
- BUFF_IN, if used, will come from one of the output pins

I think the pin types are reasonable, but some are interesting and could warrant discussion.

---

---


myfreescalewebpage:
Hello @evanshultz 
No comment to do, merging.
Joel

PR #2736, opened 2020-05-16
Gd32e23x

xjtuecho:
add GigaDevice GD32E23x MCU symbols.

---

---


myfreescalewebpage:
Hello @xjtuecho 

Not a full review, but some few comments right now:

- [ ] I suggest to rename the new library "MCU_GigaDevice" except if you want to add dozens of symbols ?
- [ ] VSS pins to be stacked
- [ ] GPIO on the left or on the right, not at the bottom or the top

Joel

xjtuecho:
> Hello @xjtuecho
> 
> Not a full review, but some few comments right now:
> 
> * [ ]  I suggest to rename the new library "MCU_GigaDevice" except if you want to add dozens of symbols ?
> * [ ]  VSS pins to be stacked
> * [ ]  GPIO on the left or on the right, not at the bottom or the top
> 
> Joel

I have pushed new commits to fix (1) and explain (3),  but sorry I can't understand (2)```VSS pins to be stacked```. Could you give some detail description? @myfreescalewebpage

myfreescalewebpage:
Sure, this is a common way to have a nice representation of the pin internally connected on MCU and modules. All VSS pins at the same place. See for example pin GND on module "Omega2S".
Joel

xjtuecho:
> Sure, this is a common way to have a nice representation of the pin internally connected on MCU and modules. All VSS pins at the same place. See for example pin GND on module "Omega2S".
> Joel

OK, now i know what `VSS pins to be stacked ` means. But  in my opinion, this is not a good idea, because:
[] On a published PDF schematic, it is difficult to figure out which pins are VSS, if they are stacked. Separated VSS pins are more unambiguous.
[] Wiring adjacent VSS pins on a schematic is effortless.
In fact, there are hundrads of `STM32` device in this library. VSS pins on those  device are separated. So i think separated VSS pins a better choice.

myfreescalewebpage:
Stacking is a convention we use now for a long time. Symbols without the stacking of pins are old.
I'm not sure of the exact roadmap, but a discussion was made to display the pins numbers of stacked pins on the schematic correctly. Maybe for V6 which is coming soon (should be available by the end of the year).
Joel

xjtuecho:
> Stacking is a convention we use now for a long time. Symbols without the stacking of pins are old.
> I'm not sure of the exact roadmap, but a discussion was made to display the pins numbers of stacked pins on the schematic correctly. Maybe for V6 which is coming soon (should be available by the end of the year).
> Joel
Stacking is not a good idea. If VSS should be stacked, Should VDD be stacked? Which PINs should be stacked? What is the thresholds of Stacking? no sense.

aris-kimi:
Hi @xjtuecho , please read this https://kicad-pcb.org/libraries/klc/S4.3/
It explains a lot about stacking pins and might clear things up.

xjtuecho:
> Hi @xjtuecho , please read this https://kicad-pcb.org/libraries/klc/S4.3/
> It explains a lot about stacking pins and might clear things up.

Think you. I have read S4.3.
```
2.Power supply pins must be stacked unless the datasheet specifies the need for decoupling capacitors on every pin.
```
On most ICs, separated VDD-VSS are designed for decoupling, this is the common case.
On FPGA or CPLD, VDDIO-GND pairs may belong to different IO banks. So they can't be stacked.
Stacked PINs only make a simple thing complicated and buggy.

aris-kimi:
You are correct, but since all VSS pins are connected to GND, those can be stacked, and VDDs can left seperated since datasheet does mention decoupling capacitors. So klc rules are not irrational about this, result is less wiring in the schematic and in my opinion this is a much better approach..

xjtuecho:
> Without giving my personal opinion, this is a kicad convention rule we should follow to have a consistent library. We cannot accept all contributions without a special attention to the integration with the library, else it will be a large mess.
> Joel

I respect your rule, Now you can close this PR.
But what I want to say is rule is not constant, especially that's not so reasonable.

myfreescalewebpage:
@xjtuecho I hope you understood we try to follow an objective. Here we are all people working for KiCad for free. We are volunteers. Upgrade of old symbols are also very appreciated to go in the right direction but it's impossible to do it quickly.
I indicate the PR is abandoned so if somebody wish to continue working on these devices it's still possible.
Joel

PR #2737, opened 2020-05-16
Add Ai-Thinker ESP32-CAM

stefanct:
- Schematic: http://wiki.ai-thinker.com/_media/esp32_cam%E5%BC%80%E5%8F%91%E6%9D%BF%E5%8E%9F%E7%90%86%E5%9B%BE.pdf
- Footprint is WIP

ESP32-CAM

It's not completely centered to minimize size while keeping pins on 100mils.
The logic pins are separated into left and right as follows: The pins on the left side are probably best left alone in most applications as they are required for programming the module (UART), accessing the PSRAM, and for generating the clock signal of the camera. The ones on the right are less restricted. They would be used to access the µSD card.

The power options are somewhat peculiar too. There is an LDO between 5V and 3V3 that can bypassed to supply 3.3V directly via the respective pin. Additionally there is a third power pin named VCC that is connected via a solder bridge to either of those.


myfreescalewebpage:
Hi @stefanct , thanks for contributing,

- [ ] This link seems better for the datasheet: "http://wiki.ai-thinker.com/_media/esp32/docs/esp32-cam_%E4%BA%A7%E5%93%81%E8%A7%84%E6%A0%BC%E4%B9%A6.pdf"
- [ ] Please use the pin numbering of the link above (last page of the pdf)
- [ ] 3V3 should be Power Output
- [ ] You can arrange the symbol with positive power pins centered and all IO on the right as done for several ESP32 device in the lib

Cheers,
Joel

myfreescalewebpage:
@stefanct ping, are you still working on this ?

stefanct:
@myfreescalewebpage it took you 4 months for the review so I deem this ping after two weeks a bit offensive TBH. That being said, I am currently not working on it and can't predict when I'll have time and motivation again to do it. I would not mind if anybody else finishes it instead.

myfreescalewebpage:
@stefanct sorry it's not the goal to be offensive. Just want to remember reviewers are all volunteers and 4 months to get the beginning of the review is not exceptional regardless to the amont of pull requests. On the other side we are preparing kicad v6 and for this all the repositories will be migrated to GitLab. We try to integrate the maximum of pull request we can before end of September, reason of this ping.

I indicate the PR is Abandoned right now, but anybody can continue if required.

Issue #2705, opened 2020-05-06
The version 6 wishlist


poeschlr:
This issue is here for every user and maintainer to add their wishes of how the library should change for the v6 release. There is no guarantee everything will be implemented but lets collect ideas here.

---

I suggest as deadline end of may until we make a decision which of the suggestions to target and with what priority. The first post will in the end be edited with the final rating

poeschlr:
My first ideas:
symbols
- Get rid of invisible power input pins for all symbols except maybe power symbols (such symbols should not create global labels)
- If the new file format has a better alternative to hidden power input pins for power symbols then use that instead
- Rework the switch and relay libs to use standardized pin "numbers" (11,12,14; 21,22,24; A1,A2...)
- Use the new naming convention for the switch and relay libs
- If the new symbol file format allows for "aliases" to have their own footprint field then maybe use more "aliases" than now
- If there is an alternative to stacking pins then use that

footprints
- Split DFN/QFN (generic libs plus maybe manufacturer spezific if the latter is still required, otherwise get rid of all the manufacturer specific footprints)
- Split SO lib (i have no clear plan yet how we would split it)
- Split the capacitor lib
- If there is a better alternative than directly linking the 3d model file then use that feature
- Make use of the new pad shape (chamfered pad)

calebreister:
I have a few ideas, though some of them may be beyond the KiCad library scope.

Symbols:
- Fix alignment issues with analog switches, provide additional aliases.
- Add an alternate function mechanism to allow pin names to changed for symbols such as microcontrollers. For instance, a user could select an alternate function by right clicking on a pin and selecting it. This could also improve rule checks.
- Test and fix broken datasheet links
- Overhaul 74x lib(s)

Footprints:
- Better connector organization (Samtec could really use some re-organization)
- Improve PCB edge footprints
  - Edge.Cuts layer in footprint files?
  - Create edge variants of connectors that are commonly placed at/near the edge of a board (such as barrel jacks)
- Footprint aliasing to support multiple 3D models (such as connectors with different heights)
- Create an official repo for footprint script + YAML (or JSON) files
  - Add Python and YAML/JSON conventions to the KLC (or another set of contribution guidelines)
  - One footprint per YAML entry!
  - Provide a way to easily identify script-generated footprints and track them back to associated source files

3D:
- Better/more detailed contribution guidelines
  - There should be a strict policy on the use of manufacturer-provided models and to what extent contributors are allowed to reference such models.
[...]

PR #2738, opened 2020-05-17
74x gxx series power pin cleanup

yankee14:
I'm working on fixing the power pin issues, broken datasheet hyperlinks, etc. with the 74xGxx library. I need advice on which KLC rules I'd be allowed to break, because I can't follow it and have the symbol look reasonable in all cases.

See https://www.kicad-pcb.org/libraries/klc/S3.8/

"Power pins in this case must be drawn on the same symbol as the logic pins."

That is very difficult to do while also keeping the symbol reasonably small. I would have to make the pin name font very small, or offset the power/ground pins to the left or right. Even if I move the pins to the left and right, there are cases where it still won't fit unless I make the symbol quite large.

---

---


chschlue:
`74LVC1G00/02` look good to me like that.
For multi-unit symbols, it looks like the screenshot in S3.8 is obsolete. I suggest using floating pins (like multi-unit opamps and such).

Generally, working logic symbols are going to be a brand new addition to v6, a black void where no man has gone before.
Because of that, I think the right way is making the symbols work and adapting KLC accordingly, not the other way round.

yankee14:
This PR after https://github.com/KiCad/kicad-symbols/issues/2705#issuecomment-629807359 :

c88f710347aed645fd9b42f4e937b5df

evanshultz:
Please don't misunderstand my comment. It's my opinion, as just one of the librarians, that having a consistently 'bad' library is best for users as they're used to the symbols and have implemented them in schematics now. This is a great improvement, but it's incremental and breeds entropy which may not be the best choice for the interim time left until v6 is released.

I'll own the scripts and make the new logic libraries which will allow us to address all symbols in one fell swoop in a maintainable way.

yankee14:
@evanshultz The scripts sound really good! I'm personally in no hurry, and I get what you mean by introducing entropy. In the meantime, I just edit the symbols individually in my own branch as I need them for work and such https://github.com/yankee14/kicad-symbols/tree/ed_devel.

I'm going to close this. Thanks.

PR #2739, opened 2020-05-17
added nrf52811QFxx QFN48 version

rs-development:
Added the nRF52811QFxx which is the QFN48 variante, the nRF52811QCxx which is QFN32 is already there. I have taken the QFN32 Symbol for reference.

I also updated the Description on the QFN32 Package since there is no FPU with the nRF52811 Variante.

https://infocenter.nordicsemi.com/pdf/nRF52811_PS_v1.0.pdf

image
image
image


daviid5:
@rs-development Pls check if 52810 and 52811 are pin compatible
If yes match the symbol
Note: VDD stacked in 52810 footprint
grafik

rs-development:
I think the footprints should be exchangeable since the chips are. Is it possible to adjust already merged symbols?

myfreescalewebpage:
Hello @rs-development and thanks for this symbol

No comment to do except the recommandation of @daviid5 is very good. It seems the pinout is exactly the same right ? In this case I suggest 52811 should be an alias of 52810.

Tell me.

Joel

myfreescalewebpage:
@rs-development ping

PR #2752, opened 2020-05-24
Update MCU_Module.dcm

mi-and-core:
BCM 2837 / BCM 2837B0 : quad cores ( 4 x Cortex A 53) 
Its the very same four cores as in Pi3


chschlue:
Confirmed, thanks.

PR #2759, opened 2020-05-28
Fix coupled inductors

cpresser:
It seems I missed a few items when working on 
https://github.com/KiCad/kicad-symbols/pull/2569 and https://github.com/KiCad/kicad-symbols/pull/2376

This was reported by user **zrb** in the KiCad IRC channel.

@evanshultz Can you take a look? Thanks.


evanshultz:
Sure!

The diff is massive. Can you point out what you intended to change and what KiCad changed on it's own? And what version of KiCad did you use to do this? Having a separate commit to capture the intentional changes would make this easier, as you know. But we'll figure it all out. :)

cpresser:
I used 5.1.6 from debian backports. 5.1.6+dfsg1-1~bpo10+1

It is already in two commits!?
The first one "Automatic reordering of lines" should have no functional changes.
The second one has the actual fixes. The naming on some devices did not match the pin-order of the symbol.

evanshultz:
Oh wow. Not my day. I think I clicked on the 'Files changed' link at the top of the GH page and was thinking that was a single commit. My mistake. Thanks and looks fine.

PR #2719, opened 2020-05-12
Add alias for tps56339

drid:
Add alias for tps56339

https://www.ti.com/lit/ds/symlink/tps56339.pdf

---

All contributions to the kicad library must follow the [KiCad library convention](http://kicad-pcb.org/libraries/klc/)

Thanks for creating a pull request to contribute to the KiCad libraries! To speed up integration of your PR, please check the following items:

- [x] Provide a URL to a datasheet for the symbol(s) you are contributing
- [x] Provide a screenshot of the symbol(s) from the symbol editor with the pin types visible
- [x] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
  - A new fitting footprint must be submitted if the library does not yet contain one.
- [x] If there are matching footprint PRs, provide link(s) as appropriate
- [x] Check the output of the Travis automated check scripts - fix any errors as required
- [x] Give a reason behind any intentional library convention rule violation.

---

Be patient, we maintainers are volunteers with limited time and need to check your contribution against the datasheet. You can speed up the process by providing all the necessary information (see above). And you can speed up the process even more by providing additional info like the screenshot of the symbol editor pin table (or for high pin counts converted to csv) sorted in the same way as the pin table in the datasheet and a direct link to the datasheet page that contains the pin table.

PR #2760, opened 2020-05-29
Fix TPS562200 pins

drid:
Change pin type
Fix pin offset

image

---

---


evanshultz:
@drid 
Thanks.

My only concern is about the SW pin. I think it should stay Output pin type. Because it can't supply current directly it shouldn't be Power Output. That would prevent an ERC error if SW was connected directly to Power Input pins, but there must be at least an inductor in series. So it seems most safe to not change this pin's type.

drid:
This was done per suggestions in https://github.com/KiCad/kicad-symbols/pull/2719 so both symbols should follow the same logic.
I do agree with you so I am changing this

evanshultz:
Thank you!

PR #2777, opened 2020-06-01
Add Senseair S8 family CO2 sensors

yankee14:
Add Senseair S8 family of CO2 sensors.

Symbol PR: https://github.com/KiCad/kicad-symbols/pull/2777
Footprint PR: https://github.com/KiCad/kicad-footprints/pull/2308

* 1 kHz PWM output
* Modbus
* Open collector alarm on CO2 threshold

Datasheet showing mount directions: https://rmtplusstoragesenseair.blob.core.windows.net/docs/publicerat/ANO102.pdf

Datasheet S8 Residential: https://rmtplusstoragesenseair.blob.core.windows.net/docs/publicerat/PSP107.pdf
Datasheet S8 Commercial: https://rmtplusstoragesenseair.blob.core.windows.net/docs/publicerat/PSP103.pdf
Datasheet S8 2%: https://rmtplusstoragesenseair.blob.core.windows.net/docs/publicerat/PSP108.pdf
Datasheet S8 5%: https://rmtplusstoragesenseair.blob.core.windows.net/docs/publicerat/PSP0113.pdf
Datasheet S8 LP: https://rmtplusstoragesenseair.blob.core.windows.net/docs/publicerat/PSP126.pdf

---

---


myfreescalewebpage:
Hello @yankee14 thanks for this symbols

Few comments:
- [ ] The footprint should be indicated in the properties. if the footprint are not the sames, the symbols can't be alias. One symbol per footprint.
- [ ] Please use capital letters for pin name.
- [ ] Pin can be grouped by functions, and the symbol a little bit larger. I suggest the following design:

Capture

Joel

yankee14:
@myfreescalewebpage thanks. I kept the pin ordering as I had it because of the way that it is likely to be used:

image

Filter capacitor on VCC_OUT facing downward to ground, open collector alarm pullup facing upward.

As for the footprint matching, I am a little confused. I know that if a certain part comes in multiple packages, there must be a unique symbol for each of the packages, and each of those symbols can contain their own matching criteria. In this case, however, there is only 1 package, and it can match to multiple footprints in the sense of, e.g. SOT-23-6 + SOT-23-6_HandSoldering. There is only one package, but multiple possible (nearly identical) footprints depending on the manner in which the user wants to mount the part. You guys have not created a unique symbol for SOT-23-6 and another unique symbol for SOT-23-6_HandSoldering. You leave the footprint field blank, and set the filter to match the two (nearly identical) footprints.

myfreescalewebpage:
@yankee14 thanks. Still missing the footprint in the properties, else OK for me.
Joel

yankee14:
Hi Joel,

> @yankee14 thanks. Still missing the footprint in the properties, else OK for me.
> Joel

I did not specify a footprint because there are multiple footprints to match. [KLC S5.1](https://kicad-pcb.org/libraries/klc/S5.1/) says "For generic symbols (those which map to multiple possible footprints), the Footprint field must be left blank". I have left it blank, and set only the alias.

myfreescalewebpage:
This is why we create one symbol per footprint when we want to have the footprint set in the properties. The users will use your symbol more easily because they will not make any error if you specify the footprint yourself. This way is generally better and is the preferable in the library.

If you want to keep like that anyway, it's correct regarding to the KLC.

Let me known of you decision.

Joel

yankee14:
Ah okay, thanks for giving me the option. Giving the KLC another read, [it seems to politely suggest](https://kicad-pcb.org/libraries/klc/G2.1/) that fully specified symbols are preferred, so I think I will create one symbol per footprint as you have suggested. It's been a while since I read through the whole KLC, so I forget things like that.

I'll get around to fixing that hopefully this weekend.

yankee14:
If I make dedicated symbols for each of the 3 footprints I made:

- Upward facing mount
- Downward facing mount
- Downward facing mount with board cutout

Then I will have 3 groups of symbols, one group for each footprint. I am not sure what naming standard to use for the symbols, maybe 004-0-0013_UPWARD and 004-0-0013_DOWNWARD and 004-0-0013_CUTOUT? I also don't know if my footprints are going to be accepted as drawn, because no librarian has commented on them yet.

myfreescalewebpage:
I don't really understand upward and downward ? Do you mean some mounting with the top of the device on the PCB ?
Joel

yankee14:
Hi Joel,

Sorry for the late reply, I got busy with other things.

It is possible to mount this part with the diffusion membrane facing upwards or facing downwards, as suggested on page 2 of this datasheet: https://rmtplusstoragesenseair.blob.core.windows.net/docs/publicerat/ANO102.pdf.

1. Upward facing (referring to the diffusion membrane), board cutout not needed.
2. Downward facing, diffusion membrane does not get to breathe very well.
3. Downward facing with Edge.Cuts layer polygons to cut out a section of the board to allow for better airflow through the board, to the membrane.

Does that make sense?

myfreescalewebpage:
Hello @yankee14 
Thanks for the link to the doc, it's very clear.
Ok to keep footprint empty with the current footprint filter.
Just need to wait the footprint now.
Joel

cpresser:
I am currently working on the footprint-PR.
My suggestion is to make this a fully specified part. The filter should match both variants of the footprint. Similar to devices that match a 'normal' and a 'thermal-vias' footprint. Then the designer can choose the proper footprint for the application.

myfreescalewebpage:
@cpresser not bad. I will check after footprint will be merged and requets the modifications.
@yankee14 I think the proposition of @cpresser is not bad. I let you checking together the right name to have on footprints to permit this.

yankee14:
@myfreescalewebpage @cpresser 

I have updated footprint in the last day or so, let me know what I should do to the symbol. Thanks.

myfreescalewebpage:
Thanks @yankee14, I let @cpresser finish with the footprint, then will continue here.

myfreescalewebpage:
Following integration of the footprint, just checked again the symbol. Looks ok for me. Merging.

PR #2778, opened 2020-06-03
Add Microchip MCP6V67 as alias of LM2904

yankee14:
Add Microchip MCP6V67 opamp as alias of LM2904

* "Zero Drift"
* Low input offset voltage

Datasheet: http://ww1.microchip.com/downloads/en/DeviceDoc/MCP6V66-Family-Data-Sheet-DS20006266A.pdf

Screencaps:
LM2904_UnitA

LM2904_UnitB

LM2904_UnitC

---

---


yankee14:
There is also a TDFN version of this part (with larger quantities available at Digi-Key), but I'm not sure how to handle that.

TDFN: MCP6V67T-E/MNY
MSOP-8: MCP6V67-E/MS

myfreescalewebpage:
Hello @yankee14 , thanks for contributing,

The policy is now generally to create one symbol for each package version. For MCP6V67 you should do it.

Joel

myfreescalewebpage:
Hello @yankee14 

Why MCP6V67EMS is not an alias of the same device too ?

Joel

yankee14:
@myfreescalewebpage I think EMS (the MSOP) is the one I did make an alias? I did not do the other one (the TDFN) because the footprint does not seem to exist in the library yet.

myfreescalewebpage:
It is in the git diff https://github.com/KiCad/kicad-symbols/pull/2778/files

yankee14:
@myfreescalewebpage yeah, that's definitely a bug with KiCad. It's happened to me several times, but I couldn't figure out how to recreate it. I didn't do anything in KiCad that should cause it to create a unique symbol out of the alias I added to LM2904.

The unique symbol is removed and only the alias to LM2904 is left, MCP6V67EMS.

myfreescalewebpage:
Yes I think I have already seen this 'behavior' of KiCad. I personally always verify the git diff just to ensure that's the wanted modification.

yankee14:
@myfreescalewebpage I found an issue with this PR during layout. The MSOP package for MCP6V67EMS exposed pad (would be pin 9) does not have a pin representation in the symbol, thus cannot be connected to a net.

PR #2780, opened 2020-06-04
WDC65C816S 8/16-bit microprocessor DIP-40

na-g:
WDC65C816S 8/16-bit static microprocessor DIP-40 package
Data sheet: http://archive.6502.org/datasheets/wdc_w65c816s_oct_11_2018.pdf
Screen shot:WDC65C816S-00

This conforms somewhat to the layout of the COMMODORE 6502 in obsolete/CPU, though a few pins needed to be regrouped for address validation (`VPA`,`VPD`,`RW`)

---

---


na-g:
This is my first part submission, and I would rather like this to be a high quality symbol. I plan on submitting the WDC65C02S static 65C02, interface adapter, and serial adapter. So I would like to get things correct now.

I have some questions for the reviewer:

1. Other packages are available, though I am don't believe them to be commonly used in the retro community. I am not opposed to adding them, if necessary. If that is the case, do they all need to be done in the same commit, or can they be split into multiple commits?

2. ϕ2 in some of the 68C816 literature it is an uppercase Greek phi, in this particular data sheet it is PHI2. I used the character for brevity. I noticed the Commodore 6502 part uses a Dutch zero for its clock ports. I would like to do whatever is considered clear, concise, and conventional.

3. The data pins are tri-state buffers, like the 6502. The symbol for the 6502 has them marked as bi-directional not tri-state. I initially used bi-directional to match the 6502. Is there a convention here that I am not aware of, or is that an oversight in the 6502 symbol?

4. The top 8-bits of the address bus are multiplexed onto the data bus. If there is a better convention around the naming or arrangement let me know. They could be moved to below the address pins, but for now they roughly match the 6502 part.

cp-aquila:
Hi, and thanks for you first submission 🎉 

Some answers:
1. No, it is not mandatory to include all variants. They can be added in different Pull-requests.
2. There is no explicit written rule for which characters are allowed (but a similar one: https://kicad-pcb.org/libraries/klc/G1.1/) I would prefer to see `PHI2` here. The KLC also states that the pin-name needs to be exactly the same as in the datasheet.
3. There is rule S4.4 (https://kicad-pcb.org/libraries/klc/S4.4/). Those pins here are not really programmable, so bidir does not apply. Tri-state seems to be fine at first glance. The reason we have pin-types is mostly ERC. So look at it from an ERC perspective. Think of which other pin-types will be connected to them and if that would generate an error. For example, adding an output-pin to a net with a tri-state-pin seems wrong and should at least generate a ERC warning.
4. No clue. I don't understand the multiplexim scheme here. Can you elaborate, or point me to the correct page in the datasheet?

na-g:
@cp-aquila thanks for your feedback!

myfreescalewebpage:
@na-g ping, please check my previous comment

na-g:
@myfreescalewebpage , thank you for your review! I haven't had time to work on this; I'll get to it this weekend. Then, I will close it as a duplicate of #2862 which is a superset of this work. This can be reopened if #2862 doesn't land, for some unforeseeable reason. Thanks again for the guidance, it will be useful for future contributions as well.

myfreescalewebpage:
@na-g thanks for the answer, will review again after the changes

myfreescalewebpage:
@na-g #2862 has been closed by the contributor

PR #2782, opened 2020-06-04
74xx123: fix location of RCext and Cext

hvraven:
previously Cext and RCext were flipped between the two units. Now both units have the identical layout. The pin assignment is/was correct, but due to the subtle layout changes it was very simple to introduce errors. [datasheet](https://assets.nexperia.com/documents/data-sheet/74HC_HCT123.pdf)

Old Unit B:
Screenshot_20200604_171035
New Unit B (old and new unit A layout):
Screenshot_20200604_171052

---

---


evanshultz:
@aris-kimi 
That is correct.

@hvraven 
Since the symbol would get changed on existing schematics, the parts connected to the pins would suddenly get flipped! We don't want to break existing schematics. But it is a good change and @myfreescalewebpage has marked this for v6 so we'll merge it when the time is right. Thanks!

myfreescalewebpage:
Yes exactly. Just made the review, no comment except we need to wait the v6 release.

PR #2784, opened 2020-06-05
Fix a typo in 8-bit ISA bus connector signals

davefiddes:
Fix-8-bit-ISA-bus-typo
The IRQ2 connection on B4 has a typo (INQ2). Cross
checked against the Wikipedia "datasheet" and the IBM PC
XT Technical Reference (p477 D-11) from
http://bitsavers.informatik.uni-stuttgart.de/pdf/ibm/pc/xt/1502237_PC_XT_Technical_Reference_Apr83.pdf


evanshultz:
Good catch. Thank you!

PR #2785, opened 2020-06-05
Fix the control outputs of 8288 for multi-master operation

davefiddes:
8288-bus-interface-pin-fix
The 8288 control outputs are incorrectly configured as
outputs. When used in a multi-master configuration this
means it is not possible to connect command outputs to a bus transceiver such as 74LS245. This is because the resulting
design will fail ERC due to each output being connected to a
tri-state output.

The command outputs should be tri-state. According to the
Command Enable section of the datasheet (page 18-121)  a
logic low on the CEN pin will activate this behaviour.


evanshultz:
Thanks and welcome!

It's always helpful to attach the datasheet you reference. The symbol currently links to http://www.datasheets360.com/pdf/7208679415653955300.

That section of the datasheet specifically says "not 3-state" but that doesn't make sense to me. Maybe it's a typo. The change you've made seems good to me. However, INTA is also part of the COMMAND BUS pins as shown on page 18-118. Should it be Tri-State too?

davefiddes:
Sorry for the slow response. Had to do quite a bit of digging to understand the datasheet and device operation. Ended up breadboarding it to be sure...

You are correct the INTA pin is also a Command Output.

All of the Command Outputs are tri-state. This is explained in the ~AEN~ pin description which controls the state of the command output pins. The behaviour of the CEN pin is very poorly explained. It affects the operation of the logic not the state of the pins directly. Apologies for not realising this and pointing you in the right direction in my original comment. Also, apologies for not including the datasheet URL.

I have updated the symbol to fix two errors highlighted by the Travis CI checks that seem to have changed in the past few days.

evanshultz:
@davefiddes 
Sorry that I missed your update. All looks good now. Sorry again for the delay and thank you!

PR #2787, opened 2020-06-06
Added a set of common generic relays

osholt:
I added generic relays to device library of the following popular configurations:
**Relay_DPDT**
image
**Relay_DPST_NC**
image
**Relay_DPST_NO**
image
**Relay_SPDT**
image
**Relay_SPST_NC**
image
**Relay_SPST_NO**
image

All symbols are consistent with the fully specified relays used in the relays library. Generic relays encounter similar problems to generic op amps regarding pin numbering: every package and every manufacturer uses different conventions. I picked a simple convention to follow in line with generic op amps.

All symbols were tested with the checklib.py script and comply with all KLC guidelines applicable to generic parts.

I have not included any latching relays but I (or someone else) may do this in the future.

---

- [N/A] Provide a URL to a datasheet for the symbol(s) you are contributing
- [N/A] Ensure that the associated footprints match the [official footprint library](https://github.com/kicad/kicad-footprints)
- [N/A] If there are matching footprint PRs, provide link(s) as appropriate

---


myfreescalewebpage:
@osholt thanks for this contribution.
However, I'm not sure it's really wanted because you are showing a specific configuration of pins and all relays do not have the same pin assignment. @evanshultz your point of view ? Maybe something we should define with generic pin name ?
Joel

osholt:
> @osholt thanks for this contribution.
> However, I'm not sure it's really wanted because you are showing a specific configuration of pins and all relays do not have the same pin assignment. @evanshultz your point of view ? Maybe something we should define with generic pin name ?
> Joel

Thanks for taking the time to review! I am happy to call the pins whatever works.

The precedent for transistors of including every pin permutation I did not think was appropriate as relay pin allocations are far from standardised between packages or manufacturers but the functions are highly standardised.

I created them because I was frustrated by a lack of generic relays to include in the design stage before having to commit to a specific part. 

There are certainly other generic parts that struggle from the same "generic function but no generic pin number problem" and I am not sure how they have been handled: coupled inductors, multi-tap transformers and and generic op-amps come to mind.

Oliver

myfreescalewebpage:
@chschlue also maybe your point of vieww (my comment at https://github.com/KiCad/kicad-symbols/pull/2787#issuecomment-687803001)

chschlue:
I'm generally in favor of the numbering suggested in https://github.com/KiCad/kicad-symbols/issues/2705#issuecomment-624892692, but I'm not quite sure if that's understood internationally, especially across the pond.
@evanshultz?

myfreescalewebpage:
I agree

PR #2794, opened 2020-06-09
Add Nexperia 74HC165 8-bit as alias to 74LS165

yankee14:
Add Nexperia 74HC165 8-bit parallel load shift register as alias to 74LS165.

Datasheet: https://assets.nexperia.com/documents/data-sheet/74HC_HCT165.pdf

kicad-symbols: https://github.com/KiCad/kicad-symbols/pull/2794
kicad-footprints: https://github.com/KiCad/kicad-footprints/pull/2314, https://github.com/KiCad/kicad-footprints/pull/2315
kicad-footprint-generator: https://github.com/pointhi/kicad-footprint-generator/pull/570, https://github.com/pointhi/kicad-footprint-generator/pull/571

I am pretty sure I am doing this wrong, and I'm not sure what to do. My intent is to add Nexperia's 74HC165 to the library in several different packages it comes in. There are the following manufacturer part numbers:

- `74HC165BQ,115`, is a VFQFN package with an exposed pad, so obviously would have to be submitted as a separate symbol due to the extra pin. **I am NOT trying to add that one**.
- `74HC165D,653`, is an SO-16 package. **I am trying to add this one**.
- `74HC165DB,118`, is an SSOP-16 package. **I am trying to add this one**.
- `74HC165PW,118`, is a TSSOP-16 package. **I am trying to add this one**.

So far, I've taken 74LS165, which is pin identical to these parts, and created a single alias called 74HC165. I also added footprint filters to match to these additional packages. It doesn't feel like this is right, so please let me know what I should do.

Screencaps:

74LS165

image

---

---


aris-kimi:
Hi @yankee14 , it seems like 74VHC9164FT symbol from your older PR, from a different branch:
( yankee14:toshiba_74VHC9164FT_universal_shift_register ), is still present at your local lib, also the one you have uploaded at the current PR (from another branch).

So in this PR there are presented 2 symbols.

I think you should edit your file again, then delete 74VHC9164FT symbol, and then upload it again on this branch: ( yankee14:74HC165_alias ).

aris-kimi:
Hi @yankee14 , I just saw #2772 that is similar with your PR, i am not quite sure about symbols with different packages as an alias, and also i am not sure about how two similar PRs proceed.
I think it is better to keep this PR open and wait for a librarian for suggestion.

poeschlr:
Normally we require a separate fully specified symbol per footprint. However, the logic libs are one of the exceptions to this rule.

aris-kimi:
Thank you @poeschlr , i thought that there is something going on with those libs, but i wasn't sure..

From what i saw on those PRs, this symbol introduced here is acceptable and may require less changes in contrast with the older PR, which has already been labeled.
How does this workflow goes?? Does timeline give one priority over other?

myfreescalewebpage:
@aris-kimi the other PR is not valid, only the lib file is modified and the 74HC165 should an alias of the existing symbol, which is done here so we will keep this job because it's done. I will reject the other PR.
Joel

PR #2798, opened 2020-06-12
Add shielded SIM card

Schlumpf:
it's a copy of the original generic SIM card symbol, but with SH pin

grafik

---

- [x] ~Provide a URL to a datasheet for the symbol(s) you are contributing~ **generic symbol**
- [x] ~If there are matching footprint PRs, provide link(s) as appropriate~
  - **generic symbol**
  - **variant of existing symbol**

---


myfreescalewebpage:
Hi @Schlumpf , thanks for contributing,

- [ ] Can you put SH on the left as done on "JAE_SIM_Card_SF72S006" symbol ?

Cheers,
Joel

Schlumpf:
Done :)

myfreescalewebpage:
Thanks. I have forgotten some Travis error on the last review:

- [ ] Pin offset to be fixed
- [ ] Footprint is not correct, should be empty for this device, like "SIM_Card" symbol.

The other warning are normal.

Joel

Schlumpf:
I changed the pin offset of both variants (shielded and unshielded) to 20mil.
But I don't get your second point. The footprint field is empty. There is set a footprint filter, the same as unshielded variant.

myfreescalewebpage:
Thanks for the fix. Sorry the second point is not correct, probably I was checking another symbol at the same time...
No more comment, merging.
Joel

PR #2799, opened 2020-06-12
Add IXYS CPC3703 N-Channel JFET, SOT-89

yankee14:
Add IXYS CPC3703, N-Channel JFET, 250V DSmax, 360mA Idss, SOT-89

Datasheet: http://www.ixysic.com/home/pdfs.nsf/0/B6BA0194695DEB9D85256DB00069D26D/$file/CPC3703.pdf

CPC3703

image

---

---


yankee14:
> better if
the description was: "360mA Idss, 250V Vbr, 4Ohm Rds, 15V Vgs, N-Channel JFET, SOT-89"

I like your description format better, but it is not really consistent with the existing JFETs. I made your change anyhow.

> and keywords were: "N-Channel FET Depletion Transistor

I'm not sure about other people, but I have never referred to this kind of transistor as a "depletion mode" transistor, except for maybe the 2 weeks we studied them in uni. I just say "JFET", and I think that's what most people around me call them. So, when I come into my schematic and I want to add a JFET, I don't think to search "depletion", I think to search "jfet". Since the keywords are mainly responsible for the search results that pop up, it also does not make much sense to capitalize any of them (see the screenshot here in [KLC S6.2](https://kicad-pcb.org/libraries/klc/S6.2/). I think it is important to just make a case insensitive list of all the things that people might type to find this transistor, which is why I specified "n channel n-channel jfet depletion mode". That may not be consistent with the rest of the JFETs in the library.

yankee14:
@aris-kimi of course and thank you very much for your reviews! I enjoy doing this.

yankee14:
> If not much of a trouble, i think it would be great if you could also change all depletion jfets' keywords,
in order to keep the library consistent.

I don't mind doing that at all, but I think I will let someone else do that who has more experience in how the keywords should be formatted.

myfreescalewebpage:
Hello @yankee14  and thanks @aris-kimi for the first step review.

Two comments for me:
- [ ] Where is it indicated to be a JFET ? The symbol and the description in the datasheet do not match your pull request.... ?
- [ ] Footprint filter never contain the number of pins, should be `SOT?89*` here

Joel

yankee14:
@myfreescalewebpage You are both right, i got that totally mixed up. I have never used a depletion mode transistor before, and I was led to believe that they only exist as JFETs. I am closing this.

PR #2801, opened 2020-06-12
Add Schaffner RN102 series common mode chokes

yankee14:
Add Schaffner RN102 series common mode chokes.

Datasheet: https://www.schaffner.com/products/download/product/datasheet/rn-series-common-mode-chokes-new/

I was pleasantly surprised to find that the footprint already exists for this series. Maybe this PR should be denied, in favor of the generic common mode chokes under Device?

RN102-0

---

---


cpresser:
Making a full specified symbol is a perfectly fine addition.

I am not quite sure about the name. I am thinking about adding a prefix perhaps like `Choke_RN102-0.3-02-12M`. Currently nothing is prefixed in that library. 
Contents so far in `Filter.lib` are SAW and AC-line-filters, nothing like this (common-mode-choke). But on the other hand I could not find a library that is a better fit for this addition.
What do you think?

yankee14:
@cpresser that sounds good to me! I went ahead and made the naming change as you suggested.

myfreescalewebpage:
Hello @yankee14 

I propose to follow naming used on footprint, starting the part name with "Choke_Schaffner_RN102...".

Else no comment to do. Please tell me what you think.

Joel

cpresser:
Perhaps even `Choke_CommonMode_Schaffner_RN102...`. Chokes that are not common mode do exist.
On the other hand, that information is already in the description and the keywords.
But more prefix makes it simpler to split the library in the future assuming there are more common-mode-choke additions.

myfreescalewebpage:
There is already RN102 which seems to be the part name so I suppose not common mode will have a different one.

yankee14:
Hi @cpresser and @myfreescalewebpage , would you like `Choke_Schaffner_RN102` or `Choke_CommonMode_Schaffner_RN102`?

myfreescalewebpage:
Footprint has "Choke_Schaffner_RN102" so I prefer having the same logic.

myfreescalewebpage:

PR #2803, opened 2020-06-13
Add EFM8UB31F40G-A

LeSpirou:
Datasheet:
https://www.silabs.com/documents/public/data-sheets/efm8ub3-datasheet.pdf

Added three MCU parts which did not exist in that library before.
Three different symbols, completly symbol compatible within the shared commons which are the majority of pins.
Checked pin numbers twice for each symbol.
Added apropriate footprint filters.
Associated compatible footprints from the standard libraries.
_The centered GND pad for the EFM8UB31F40G-A-QFN24 is 0.5mm greater in width and height (2.6x2.6mm instead of 2.55x2.5mm)._

grafik
grafik
grafik

---

---


LeSpirou:
Violation S4.1: The symbol keeps compact and in favor of S4.2 it's clear.
Alternative is already in work but would prefer the comitted one since the alternative is not as good looking (and symmetric) as the comitted one.

LeSpirou:
ok, where can I get those checker scripts? There are none in my installation folder (Win64).

LeSpirou:
grafik
grafik
grafik

PR #2807, opened 2020-06-16
Add Onsemi TLV431 precision shunt regulators

yankee14:
Add Onsemi TLV431 precision shunt regulators

Datasheet: https://www.onsemi.com/pub/Collateral/TLV431A-D.PDF

Comes in:

- TO-92
- SOT-23
- TSOP-5

Screencap:

TLV431xSNT1G

TL432DBZ

TL431LP

image

image

---

---


yankee14:
What happened to Travis?

aris-kimi:
> 
> 
> What happened to Travis?

I think since it is a migration time for symbol libs we should wait...

yankee14:
@aris-kimi okay, that's fine! Do you happen to know if the library migration is something that's going to be scripted? Or is there a lot of manual work involved. I need to read about the differences at some point. I heard it is going to be more of a lisp style format.

aris-kimi:
Hi there @yankee14 , i don't have a clue... I am trying to figure this out.. It is this s-expressions like the footprints use. I am sure it is going to be some scripting involved, but i'll stick to manual way at present..

cpresser:
- [ ] The `SOT-23` package has another pin-out (Ref-K-A) than the existing `TL431DBV` (K-Ref-A). Unfortunately it also needs a new symbol and can not be made as alias.
- [ ] Since this is a 'small' series, please add aliases for the A,B & C variants with matching descriptions. I would prefer that over the wildcard at this position.
- [ ] Some of the wilcards positions in the name seem to reference non-functional parts (amount or parts per reel/tube). Please remove that. The 2nd column in the table actually specifies the device-code - use that. 
  - `TLV431ALP` (TO-92)
  -  `TLV431ASN` (TSOP-5)
  - `TLV431ASN1` (SOT-23)
- [ ] The `TSOP-5_1.65x3.05mm_P0.95mm` package does not exactly match the datasheet. I suspect its a good enough fit for hobby-projects, but not for professional use. The real body-dimensions are 1.5x3.0mm. Perhaps head over to the footprint-generator repository to create a matching footprint.

yankee14:
> The SOT-23 package has another pin-out (Ref-K-A) than the existing TL431DBV (K-Ref-A). Unfortunately it also needs a new symbol and can not be made as alias.

I don't think mine is an alias of `TL431DBV`, but is an alias of `TL432DBZ`. It looks like the pinout matches.

image

image

image

cpresser:
Sorry about the SOT-23 confusion. My mistake, I looked at the wrong part.

The `TSOP-5` package does not seem to be generated by the footprint-generator scripts. That's why you could not find it.

Looking at `TSOP-5_1.65x3.05mm_P0.95mm`
* The pad-size is 0.7x0.51. The TLV431 datasheet say 1.0x0.7mm
* Thus, the center-position of the pads are not correct either.
* The pitch is correct (0.95mm)
* The body-size is slightly off. Its 1.65.x3.05, should be 1.5x3.0mm (according to calculations done in my head)

yankee14:
I agree the footprint is not correct, but I'm not sure what I can do about that at this point.

evanshultz:
I'm almost done with an update to the gullwing generator which will allow scripting these kinds of footprints. Also discussed at https://github.com/pointhi/kicad-footprint-generator/pull/586. Our scripted footprints use IPC rules to determine the pad size which may or may not match what a datasheet says. But we use the industry-standard info because, well, it is standard in the industry. Just for clarity and to reinforce the point @cpresser made above. The SOT-23 package apparently is captured in JEDEC TO-236, from which we can get dimensions and tolerances for the script. I don't see any reference to a JEDEC or JEITA standard for the TSOP-5 package, but I can look more.

Also, I think we're at a point to invoke [KLC S2.1 clause 3](https://kicad-pcb.org/libraries/klc/S2.1/). See TI's TLV431 datasheet at https://www.ti.com/lit/ds/symlink/tlv431.pdf.  Also https://www.diodes.com/assets/Datasheets/TLV431.pdf. I believe they are all compatible parts but the suffix differs for the same package. I didn't look closely so please check that.

cpresser:
Regarding the Footprints: I suggest to wait for the generated footprints.

> 
> Also, I think we're at a point to invoke [KLC S2.1 clause 3](https://kicad-pcb.org/libraries/klc/S2.1/). See TI's TLV431 datasheet at https://www.ti.com/lit/ds/symlink/tlv431.pdf. Also https://www.diodes.com/assets/Datasheets/TLV431.pdf. I believe they are all compatible parts but the suffix differs for the same package. I didn't look closely so please check that.

Not sure about that.
This PR (onsemi) adds different variants `A`, `B` and `C` (voltage tolerances). Texas Instruments has identical specs for A and B, but no C variant. Diodes Inc has another suffix `T`for the 0.2% variant  (https://www.diodes.com/assets/Datasheets/TLV431.pdf).


How would the resulting symbols be named?* 
* `TLV431_SOT-23` (only Texas)
* `TLV431A_SOT-23`
* `TLV431B_SOT-23`
* `TLV431C_SOT-23`
* `TLV431A_SOT-23-5`
* ...

I think we could also solve that with aliases if we want to add variants of other manufacturers.
What do you think?

evanshultz:
Initial submission for scripting footprint support is at https://github.com/pointhi/kicad-footprint-generator/pull/604.

Something like that is what I was thinking. The list above looks fine to me. I didn't look at the tolerances to compare, but I have used TI for TLV431 and so I thought all these companies probably want to be an option for supply chain and so the suffixes would be for the same electrical specs. Maybe I'm wrong. But in this case if the letter and tolerance match, but the suffix for the package differs, I think this clause of KLC applies. Or maybe not. Take a look at the generic opamps in the library (TL072 and MC33078 are examples, I think) and see if this works.

evanshultz:
The footprint generator script can now handle this TSOP-5 footprint.

I compared Diodes, TI and ON datasheets. A is always 1%, B is always 0.5%, Diodes makes a T with 0.2%, and TI makes a no-suffix with 1.5%. And all the pinouts match. I don't find any other vendors at Digikey for TLV431. So I think either way works and choosing MPN suffixes with ALIASes or using the package suffix as I mentioned above works.

One other note: It appears the 5-pin package can have the NC pin floating or connected to anode. That is mentioned clearly in the non-ON datasheets but if it is in the ON datasheet I've missed it.

yankee14:
@evanshultz thanks for updating the scripts to support this. Can you let me know which .yaml file I should edit?

cpresser:
I suggest to add a new file
`kicad-footprint-generator/scripts/Packages/Package_Gullwing__QFP_SOIC_SO/size_definition/tsop.yaml`

There are quite a few files that already have SOP or TSOP footprints in them, but none of them are a good match for this part.

You can look at `test_hidden_deleted_pins.yaml` to get an idea how to use the new generator features.

yankee14:
I apologize, but I will not have time to work on this before the 1 October deadline.

PR #2829, opened 2020-06-28
added Panasonic ALFG relay symbols

aracosta1:


aris-kimi:
Hi @aracosta1 , i noticed that you have some regulator symbols also added in the files changed. I would advise you to avoid mixing different symbols from different libs in a single PR.

aracosta1:
I've seen this and I.m trying to redo the pull request with only the needed changes

aracosta1:
Need to select only correct changes

PR #2832, opened 2020-06-28
St l6491 mosfet driver

aracosta1:
Symbol for ST high and low side gate driver

image

footprint is standard SO-14


aracosta1:
A symbol with same name added with #2877. 
This can be closed to avoid conflict

myfreescalewebpage:
Hello @aracosta1 and so much sorry for this conflict. Your fixes were correct. I have reviewed your PR and also the other one but I have realized it was the same device!
I indicate this is duplicated and again I apologize.
Joel

PR #2834, opened 2020-06-28
added ST LD39150 regulator

aracosta1:
ST Micro LD39150 ultra low drop regulator.

datasheet at [https://www.st.com/resource/en/datasheet/ld39150.pdf](https://www.st.com/resource/en/datasheet/ld39150.pdf)

footprint is DPAK 

image


myfreescalewebpage:
Hello @aracosta1 thanks for this contribution. No comment. Joel

PR #2835, opened 2020-06-30
Bugfix: fix position of VZCR pin on ACS730 (and aliases)

gkeeth:
ACS730xLCTR-20AB (and its aliases) has a pin placed inside the body. The pin is intended to be connected to the external circuit so it shouldn't be inside the body. Also, the incorrect placement causes a cosmetic issue with overlapping pin name text.

Git archaeology suggests this symbol was pulled in from the old repos. I didn't track the history further back.

Datasheet link:
https://www.allegromicro.com/~/media/Files/Datasheets/ACS730-Datasheet.ashx

Before:
before

After:
after

---

---


gkeeth:
Travis suggests VOUT should be type power output instead of output -- in this case travis doesn't know what it's talking about :)

aris-kimi:
Hi @gkeeth , i would suggest to rename it as VIOUT, same as the datasheet...

gkeeth:
Done. Updated screenshot: 

renamed

Datasheet for comparison:
renamed

gkeeth:
I understand moving pins is a breaking change, but the symbol is essentially unusable in its current state. In cases like that don’t bug fixes get merged even if they’re breaking?

aris-kimi:
Based on older similar PRs, i think no.
If you draw the wire you can connect to that pin (old symbol), so i do not think it is unusable,
so the change proposed here, will breake current schematics.
Keep this PR open and wait for a librarian to label and comment. I hope this helped.

gkeeth:
Understood. Thanks for the review.

myfreescalewebpage:
Hello @gkeeth 
Thanks for this fix. I think @aris-kimi is right, merging for v6 is preferable.
I have no comment to add, just put tags and milestone.
Joel

gkeeth:
Great, thanks!

evanshultz:
I would consider this worth doing now. If the wire goes directly right, it will still be connected. And if it doesn't, ERC will catch an unconnected pin. IMO it's a clear bug fix so moving pins is fine. But it's assigned to you, @myfreescalewebpage , so your call.

myfreescalewebpage:
Considering v6 is very closed now we may think this will not a be an important disturbance, so merging according to your opinion @evanshultz

PR #2837, opened 2020-07-02
Add female variants of coaxial connector symbols

ferdymercury:
Generic connectors in the symbol library usually have male and female variants. Coaxial connectors were missing the female variant, which is now added in this PR.

image

- [x] A new fitting footprint must be submitted if the library does not yet contain one.


ferdymercury:
Side note, this is how other software tends to depict coaxial connectors:
image

ferdymercury:
We could discuss if we want to change it to something like:
image

ferdymercury:
This is the variant in https://www.ohwr.org/projects/pmtlib
image

myfreescalewebpage:
Hello @ferdymercury  and thanks for contributing,

Does the design refer to a specific norm or usage ? Can you provide examples ?

The current "Conn_Coaxial" symbol can be used with male or female parts. I personally used with UFL, SMA, RPSMA etc connectors without any trouble and I don't think I have seen your design on schematics in the past.

Thanks to clarify,

Joel

ferdymercury:
Thanks!

I understand that one can use the same connector symbol for male or female parts.

But why is there then a male and female variant for a normal connector, and not for a BNC one?

image

What I did was just to copy these symbol style, having the open circle as a mark for female.

ferdymercury:
Concerning the norm, I am not sure, I just followed the style that was in Conn_01x01_Female adapted on top the current BNC symbol.

Maybe this discussion helps: https://forum.kicad.info/t/correct-connector-symbols-to-use-iaw-iec-60617/3923

myfreescalewebpage:
In this case I suggest to rename the existing symbol with "_Male" and to have something similar to the DB connectors: males with plain circle and females with background color circle.

I think the idea is not bad I just look after consistency across the library and references to norm.

ferdymercury:
Thanks for the input!

Concerning renaming the existing symbols as _Female, wouldn't that break compatibility with older schematics?

myfreescalewebpage:
It will. But we are very closed to KiCad V6, and breaking changes are permitted for that.
The only consequence for you is that we will have to wait a bit before merging your PR.
The agenda is to start merging breaking change for v6 beginning of September, it's very closed !

ferdymercury:
I was checking, and the problem with using a filled circle for male, is that it would conflict with what we have now for Coaxial_Power. So I am not sure what to do...

myfreescalewebpage:
I'm not sure either.
@poeschlr and @evanshultz if you have time to think about this issue ?
Joel

ferdymercury:
(The semicircle idea I had proposed, I had copied from Conn_01_Female connector style)

PR #2838, opened 2020-07-02
Add high voltage variants of coaxial connectors

ferdymercury:
Add high voltage variants of coaxial connectors

image

- [x] A new fitting footprint must be submitted if the library does not yet contain one.


ferdymercury:
Builds on top of https://github.com/KiCad/kicad-symbols/pull/2837

myfreescalewebpage:
See comment at https://github.com/KiCad/kicad-symbols/pull/2837#issuecomment-679329282

Not sure the Coaxial HV symbol is something useful, do you have example of devices from some manufacturers to illustrate the symbol ?

You can use one of the "SYM_Flash" symbols together with coaxial connector. If we go in this direction we will add High Voltage version for all connectors and I'm not sure it's wanted...

Joel

ferdymercury:
I like the solution with SYM_Flash. Thanks for the idea.

PR #2840, opened 2020-07-02
new ldo regulator XC6210B332MR

TheSillyMouse:
New LDO regulator

Manufacturer: TOREX
Part Number: XC6210B332MR

Datasheet:
https://www.torexsemi.com/file/xc6210/XC6210.pdf

Screenshot:
image

---

---


aris-kimi:
Hi @TheSillyMouse , you should move not connected pin inside symbol's body, and also move part name and reference on top, outside of body.

TheSillyMouse:
> Hi @TheSillyMouse , you should move not connected pin inside symbol's body, and also move part name and reference on top, outside of body.

Hi @aris-kimi the symbol is exactly the same as the existing one "XC6220B331MR". Which is already in the list.
I'll fix the reference and part name.
Thanks!

aris-kimi:
You have moved part name and footprint value, you have left reference inside the body.

diegoherranz:
If I'm not mistaken, this could be an alias of XC6220B331MR not to have another symbol. Could you turn it into an alias instead?

The NC placement is wrong on XC6220B331MR as mentioned above. That should be fixed, but it can go on a separate pull request (it's a bug). I'm happy to review that one too, ping me if it gets submitted.

Thanks!

TheSillyMouse:
> If I'm not mistaken, this could be an alias of XC6220B331MR not to have another symbol. Could you turn it into an alias instead?
> 
> The NC placement is wrong on XC6220B331MR as mentioned above. That should be fixed, but it can go on a separate pull request (it's a bug). I'm happy to review that one too, ping me if it gets submitted.
> 
> Thanks!

Hi, @diegoherranz.
I turned it into an alias.

diegoherranz:
Thanks for turning this into an alias. It makes it much easier to maintain the library if we reduce the number of symbols.

Just one comment. The datasheet URL points to the XC6220 instead of the XC6210.

Thanks!

TheSillyMouse:
> Thanks for turning this into an alias. It makes it much easier to maintain the library if we reduce the number of symbols.
> 
> Just one comment. The datasheet URL points to the XC6220 instead of the XC6210.
> 
> Thanks!

Done.

diegoherranz:
Thanks for the fix, it looks OK. Merging.

If you wanted to fix the XC6220B331MR on a separate PR, let me know and I would review it.

Thanks!

PR #2843, opened 2020-07-06
SoC PolarFire: Add library for 5 PolarFire SoC parts

pthomas:
Add the following parts in SoC_PolarFire.lib:
MPFS250T-FCVG484
MPFS250T-FCSG536
MPFS250T-FCVG784
MPFS250T-FCG1152
MPFS460T-FCG1152

---
Folks,

Some may be aware of Microchip's new RISC-V based SoC FPGAs:
https://www.microsemi.com/product-directory/soc-fpgas/5498-polarfire-soc-fpga#overview

Anyway I wanted to get a KiCad library and didn't see anything readily available, so I ended up playing with the tools in kicad-library-utils, and I'm reasonably happy with the result. For each part there is a released Excel spreadsheet in the Packing section of the link below:
https://www.microsemi.com/product-directory/soc-fpgas/5498-polarfire-soc-fpga#resources

Basically using a python script each part is built using the spreadsheets for name references and for the pad location. The ordering and how to break up the units is based both on the symbol in the dev kit schematic[1] and KLC[2]. Here is a listing of the biggest differences from the symbol in the reference schematic:
1) U1J, U1K & U1H are all combined to one POWER unit using pin stacking for each net.
2) For the KiCad library Bank 0 and Bank 1 Order Top->Bottom Left->Right, which is opposite from the Icicle reference.

I put this as an RFC because I know there are a couple of issues and wanted some feedback on other items.
*Packages: Right now none of the parts have packages. This is because there is some python issue with my footprint wizard, and I can't seem to figure out how to use SymbolGenerator.py to set the package anyway.
*Library name: I put it in as a unique library SoC_PolarFire.lib which is very convenient for generating the file. Other name would be fine too.
*NC Pins: if the part had NC pins (not all do) I ended up creating a separate unit and put them all in as non-stacked but invisible. That seemed to satisfy checklib.py, but I'm not sure what is most desired here?
Power pin location: When running checklib.py it complains that "Positive power pins should be placed at top of symbol". This would create a very strange power unit where it's very short and wide with tons of pins on top and a single one on the bottom?
*Power unit: as mentioned above I wasn't completely sure if I did the pin stacking correctly.

I know this is a lot to take in, let me know if there is anything else I can do.

thanks,
Paul

[1] https://www.microsemi.com/document-portal/doc_download/1244587-polarfire-soc-icicle-kit-schematics
[2] https://kicad-pcb.org/libraries/klc/


myfreescalewebpage:
Hello @pthomas 

Not a full review but some first comments:

- [ ] sym-lib-table file to be updated
- [ ] Description and keywords can be enhanced, see for example MCU
- [ ] Description should be terminated by the package name
- [ ] Footprint and fp filter to be completed

Joel

myfreescalewebpage:
@pthomas ping

pthomas:
Hi Joel,

Thanks for the review, I incorporated the changes into a v2 branch,
but I see everything is on hold until next month.

-Paul

myfreescalewebpage:
You can still push to this PR right now, the repo is locked so we cannot created new pull request, but we merge the maximum of pull request we can by the end of the month.

pthomas:
OK, Do I just commit to pthomas:master instead of the v2 branch? Should it
be 1 commit or 2?

-Paul

myfreescalewebpage:
since the previous commit of this PR were on your master branch, do it on your master branch to continue. 1 or 2 commits does not matter

pthomas:
OK, the master branch is updated, I didn't submit the pull request for
the packages yet, but the two new packages are here:
https://github.com/pthomas/kicad-footprints/tree/polarfire-footprints-v1

Let me know if I need to do anything else.

-Paul

myfreescalewebpage:
Will perform full review soon, but this is a large PR so it might take a while.
You can create a PR for the footprints now, so someone can also review them in parallel.

PR #2850, opened 2020-07-09
Fix 8-pin mini DIN connector

OYTIS:
Signed-off-by: Anton Gerasimov <tossel@gmail.com>

Fixed 8 pin mini DIN symbol to better match the actual view of the female connector.

Here's the datasheet referenced by the symbol: http://service.powerdynamics.com/ec/Catalog17/Section%2011.pdf
Here are some other typical ones: [one](https://eucon-elektronik.com//katalog_uk/catalog_all_chaps/09_01-10_mini-din.pdf), [two](http://www.alicat.com/wp-content/uploads/2012/05/Eight-Pin-Mini-DIN-Pin-outs.pdf).

Pin numbering is a bit questionable, because different sources show different (reflected one against another) numbering schemes, and I couldn't find the actual standard, so kept it as it was.

PS:
also, doesn't DCO (aka `Signed-off-by`) work with CC?


myfreescalewebpage:
Hello @OYTIS 

Good catch, and thanks for this enhancement.
It does not break existing designs so it's possible to merge it right now.

I don't understand you PS, what do you mean ?

Joel

OYTIS:
Great, thank you!

> I don't understand you PS, what do you mean?

There is an option to git command `--signoff` that adds a line to your commit message, e.g. `Signed-off-by: Anton Gerasimov <tossel@gmail.com>` in this one. It has an effect (I'm not a lawyer by the way, but that's how it's used) of signing the [Developer Certificate of Origin](https://developercertificate.org/), which is more or less the same as the [CLA](https://cla-assistant.io/KiCad/kicad-symbols) for this project. Also, signing CLAs sometimes makes people nervous, while DCO goes without saying in open source community.

myfreescalewebpage:
I'm not aware of the details, we require people to sign the CLA today.

PR #2852, opened 2020-07-10
OPA818 op-amp in 3x3mm WSON-8

aewallin:
https://www.ti.com/lit/ds/symlink/opa818.pdf

OPA818IDRGx

some open questions:
- pin1 ~PD has an internal pull-down, so maybe it should be passive type? (datasheet suggests it can be left floating)
- not sure about the convention for exact part-number e.g. OPA818IDRGx (some other parts seem to replace "I" by "x" - but the datasheet only shows this part-number.)
- I reduced size of PD and EP names, so they fit the symbol.

footprint PR https://github.com/KiCad/kicad-footprints/pull/2223

---

---


myfreescalewebpage:

- [ ] Name of the symbol should be "OPA818xDRG"

Else just need to wait the footprint

Cheers,
Joel

aewallin:
footprint now merged!

chschlue:
Indeed.
@cpresser was that an oversight?

cpresser:
We are talking about https://github.com/KiCad/kicad-footprints/pull/2223 right?
And https://github.com/pointhi/kicad-footprint-generator/pull/633

TBH, I did not put to much thought into that. On the generator PR it was already prefixed Texas. Most likely since its a copy-paste variant of `Texas_S-PWSON-N8`, just with the EP added. That seems plausible, I did not question that since it was consistent with the existing variant.

What do you suggest we do about that? Since all those footprints are build according to ipc spec and not texas-spec, I think removing the manufacturer prefix for them is the way to go.
The only other use of the existing `Texas_S-PWSON-N8_EP1.2x2mm` is  `LP2951-3.0_WSON`

chschlue:
Yes, I think we best fix both.

cpresser:
That should do the trick:
https://github.com/pointhi/kicad-footprint-generator/pull/642
https://github.com/KiCad/kicad-footprints/pull/2494
https://github.com/KiCad/kicad-symbols/pull/3094

aewallin:
footprint now changed back to the generic name.

myfreescalewebpage:

PR #2853, opened 2020-07-10
Adding all BDxxFC0 series LDO regulator

JonathSpirit:
Adding all BDxxFC0 series LDO regulator

datasheet : 
https://fscdn.rohm.com/en/products/databook/datasheet/ic/power/linear_regulator/bdxxfc0wefj-e.pdf

Annotation 2020-07-10 160453

Variable version (BD00FC0WFP, BD00FC0WEFJ) :

varaible to252

varaible htsop

Fix without enable version (BDxxFC0FP) : 

fix sans w

Alias : 

fix sans w alias

Fix with enable version TO-263-5 (BDxxFC0WFP) : 

fix avec w to252

Alias : 

fix avec w to252 alias

Fix with enable version HTSOP-8 (BDxxFC0WEFJ) : 

fix avec w htsop

Alias : 

fix avec w htsop alias

---

---


JonathSpirit:
Violating S3.1
    Origin is centered on the middle of the symbol
I've taken as an example other symbols.

Violating S4.2
    Positive power pins should be placed at top of symbol
The power pins are often on the sides for regulators.

aris-kimi:
Hi @JonathSpirit , thankyou for your contribution.

Footprint filters should be:
line 570: `HTSOP*1EP*3.9x4.9mm*P1.27mm*`
line 594: `TO?263*`
line 641: `TO?252*`
line 661: `HTSOP*1EP*3.9x4.9mm*P1.27mm*`
line 686: `TO?263*`

And also a suggestion, i think it would be better for those symbols without enable, to totaly remove from decription, and leave it only to those symbols which applies.

Not a full review, only a few quick comments..

JonathSpirit:
Thanks for the review,

I'm just confused for the TO263/TO252 filters. For example, TO-263-5_TabPin3 is not compatible with "WFP" symbols because pin 3 is not connected instead of GND so the filter should contain "TabPin6" ?

And the filter should also logically contain the number of pins ? (TO-263-5)

myfreescalewebpage:
@JonathSpirit the fixes of @aris-kimi are correct. Fp filter are sometimes very permissive but it is normal + fp filter never contains the number of pins.

One comment I have to do looking at the screenshots: the size of the outline is not usual for regulators. Height should be increased by 100mil.

Joel

PS @aris-kimi would you liek to continue the review here or not ?

evanshultz:
It looks like the width can be decreased as well.

JonathSpirit:
image

image

image

I resize all of the symbols by taking the existing regulators as an example (The only concern is that they are all different) and modify the footprint filters

Regulator_Linear.lib
```diff
@@ -581,6 +581,51 @@ X VI 3 -300 0 100 R 50 50 1 1 W
...
+X GND 9 0 -300 100 U 50 50 1 1 P N
+ENDDRAW
+ENDDEF
+#
+# BD00FC0WFP
+#
+DEF BD00FC0WFP U 0 20 Y Y 1 F N
+F0 "U" 0 300 50 H V C CNN
+F1 "BD00FC0WFP" 0 200 50 H V C CNN
+F2 "Package_TO_SOT_SMD:TO-263-5_TabPin6" 0 100 50 H I C CNN
```
aris-kimi: package must be TO-252-5
εικόνα
There is no 263 at the ds. Please fix those parts that are wrong (if others) and adjust filters.

PR #2861, opened 2020-07-16
add STGP7NC60HD

BAHOO08:
I added new symbol to Transistor_IGBT.lib.

---
datasheet http://www.farnell.com/datasheets/2309889.pdf
изображение

after symbol checking  i got  this message
"Checking symbol 'STGP7NC60HD':
  Violating S3.1"

But i think this symbol set symmetrical.


cp-aquila:
TBH, I think this looks rather ugly with the big circle. The body-diode is also way bigger than the other diode-arrows in the symbol.

How about something like this?
grafik
I did start from `Q_NIGBT_GCE` to create that.

BAHOO08:
изображение
Made the circle smaller

chschlue:
It would be great if we could stick with all the other transistor symbols' pin positions for interchangeability, like @cpresser's mockup.

BAHOO08:
@chschlue, I dont understand, what do you mean?

chschlue:
Transistor pins in the lib are usually at -200/0, 100/200, 100/-200 while yours are at -300/0, 50/200, 50/-200.

myfreescalewebpage:
Hi @BAHOO08 , thanks for contributing,

- [ ] Description should ends with package name: `25A, 600V, N-Channel IGBT, TO-220`
- [ ] Can you follow @chschlue recommandation at https://github.com/KiCad/kicad-symbols/pull/2861#issuecomment-675596301 please ? The design of @cpresser at https://github.com/KiCad/kicad-symbols/pull/2861#issuecomment-669077271 correspond to these constraints, please carefully check the position of the pins

Cheers,
Joel

BAHOO08:
I changed position pins into -200/0, 100/200, 100/-200 and add TO-220 to discription.
P.S. Sorry for taking so long to edit. There was no way to go to the laptop on vacation.

myfreescalewebpage:
No worries :) Thanks for the update. Merging.

PR #2866, opened 2020-07-18
Fix wrong footprint of Finder 32.21.x000

Werni2A:
The symbol previously used the footprint of the Finder 32.21.x**3**00 that has one pin less than the 32.21.x**0**00.

[Datasheet](https://gfinder.findernet.com/assets/Series/355/S32EN.pdf)

Old:
image

New:
image


cpresser:
Good catch. And thanks for the fix!

PR #2868, opened 2020-07-19
RF_Filter - 1090MHz SAW filters 8-pin STA1090EC/MP08167, and 6-pin ST…

chmorgan:
Add SAW filter symbols

---

---


aewallin:
- if you search for "bandpass" in the current existing symbols you will find they all use more or less the same symbol, is there any reason not to use that one here also?
- I would include 'bandpass' in the description and/or keywords
- GND pins should be named GND and stacked?
- for the balanced input, is there a preferred direction for the pins? should they be named +/- or IN and ~IN or similar?

(I am not a reviewer, just commenting from the sidelines..)

chmorgan:
Hi @aewallin.

For not a reviewer this is good stuff!

- Good idea on bandpass, the datasheets do describe them as bandpass filters as that's what they do.
- None of the existing symbols have the correct pinout but their symbols do make sense for these SAW filters. I'll update to use the same shape.
- Correct on the ground pins and stacking, I'll fix that as well.
- There is nothing mentioned on the data sheets about how to connect the balanced inputs. Maybe IN_1 and IN_2 is ok?
- What would the footprint name be then?

Chris

chmorgan:
@aewallin, updates pushed and screenshots above updated.

myfreescalewebpage:
Hi @chmorgan , thanks for contributing, and thanks to @aewallin for the first review

- [ ] Footprint filter: `Filter*SAW*3.8x3.8mm*`
- [ ] I agree with the alias, but I don't see the where the name of the symbol come from... ? Where STA0232A and STA1090EC come from ? Have you got a datasheet ?

Cheers,
Joel

chmorgan:
Hi @myfreescalewebpage, for whatever reason Sawtron doesn't have datasheets for their parts online although they've sent me the datasheets upon request for info about those parts. Equivalent Golledge part datasheets (which are aliases), are the ones I've cited but I'm not sure how to cite data sheets for parts that I can't find on the web. The STA1090EC and STA0232A are quite popular and used all over avionics circuits too...

myfreescalewebpage:
Thanks for the detail. I suggest to remove the datasheet link for the STA devices in this case.

PR #2876, opened 2020-07-25
Add Traco TBA 1 Series

Werni2A:
This is the first part of a larger set that adds Traco DC/DC converters. As this is the first part it would be great if you could already take a look at it and point out some possible mistakes or aspects that need improvement.

[Datasheet](https://www.tracopower.com/sites/default/files/products/datasheets/tba1_datasheet.pdf)
[Footprint PR](https://github.com/KiCad/kicad-footprints/pull/2372)

The symbol was copied from TMR_1-0511 in the Regulator_Switching lib.

image


Werni2A:
It looks like there went something wrong with the TMR part where I copied the symbol from. It should not be contained in my changes. I used "copy and save as" to create the new symbols. Does this affect the old symbol somehow? I don't understand why all these parts were changed/added...

PR #2880, opened 2020-07-28
Re-Push of ADuM764x

Werni2A:
This is a re-push of #1105. I fixed the outstanding issues pointed out by @myfreescalewebpage , except readding ADuM7643 as I think its a good choice to replace it with more specific symbols. But if still desired I can change it back.

image
image
image
image


myfreescalewebpage:
Hello @Werni2A and thanks to continue on this subject.

Just one comment about this:
- [ ] Footprint filter should be `QSOP*3.9x8.7mm*P0.635mm*`

This will be merged with milestone v6 because this breaks the existing library, but I think it's good way to do it.

Joel

Werni2A:
@myfreescalewebpage thank you for the review, I updated the filter.

myfreescalewebpage:
Thanks for the quick fix!

myfreescalewebpage:
Can now be merged when v6 will be ready.

PR #2891, opened 2020-07-31
Dmg9926 udm

Electro707:
Adding the DMG9926UDM common-drain n-channel MOSFET.
Screenshot of the symbol in KiCad:
image
Screenshot of the pins table:
image
Link to the part's datasheet:
[https://www.diodes.com/assets/Datasheets/ds31770.pdf](https://www.diodes.com/assets/Datasheets/ds31770.pdf)

---

---


myfreescalewebpage:
@Electro707 thanks for this symbol,

The design is not consistent with the other dual mosfet we have in the library. I think it will be better with something looking like IRFI4019H.

Will do full review after.

Joel

Electro707:
@myfreescalewebpage My reasoning behind the filled background box is to distinguish a dual mosfet in one package versus 2 individual mosfets connected together (at least on the schematic side). But if it's more conventional not to have such an outline, I could remove it.

myfreescalewebpage:
Yes please, for library consistency

cp-aquila:
IMHO it would be best to copy `IRFI4019H` and change it. So that the two symbols can be exchanged against each other without rearranging the schematic.
The description should also mention `Dual Half Bridge` instead of just `Dual` so its obvious that the individual FETs are connected.

Electro707:
@cp-aquila My symbol was originally based on that I think. The reason I changed the symbol is due to this part having 2 pins for both mosfet drains. So unless there is a way to stack up pins, I think the 2 pins should be separate.
I changed the description to `Dual Half Bridge`.
For reference, this is what the footprint current looks like after my last commit:
image

cp-aquila:
You can and should stack the pins. See https://kicad-pcb.org/libraries/klc/S4.3/

Electro707:
Screenshot of part after modification to make one of the drain pins invisible and stack it with the other drain pin. 
image

myfreescalewebpage:
Hello @Electro707 and thanks @cp-aquila for the first steps of the review.

Last comment:

- [ ] Footprint filter should be `SOT?23*`

Joel

Electro707:
Done!

myfreescalewebpage:
Thanks, merging.

PR #2898, opened 2020-08-05
Adedd nRF52820

G-Pereira:
Added nRF52820-QDAA
[Datasheet](https://infocenter.nordicsemi.com/pdf/nRF52820_PS_v1.0.pdf)
nrf52820

---


aris-kimi:
Hi @G-Pereira , i have noticed pin26 and pin38 noted in the datasheet to be connected, so i think those need to be stacked. Also not sure if those DECx pins should be passive or power output.

Great job i think, thank you for this. But better wait for a full review from an experienced librarian.

G-Pereira:
Hi @aris-kimi , thanks for reviewing this! 

> i have noticed pin26 and pin38 noted in the datasheet to be connected, so i think those need to be stacked

I have noticed that too but it appears that the symbol for nRF52840 (similar IC) appears to have those pins separate due to optional independent decoupling. I ended up leaving them separate for this IC as well since the same applies here.

> Also not sure if those DECx pins should be passive or power output.

I also went by this one by comparing it with the nRF52840, nRF52810, nRF52811 and nRF52832. They all use "Passive" for this pins and it made sense to me because they are not meant to be used as power supply pins not even for "pin configuration" (set some pins to high level voltage)

myfreescalewebpage:
Hi @G-Pereira , thanks for contributing,

- [ ] Name of the symbol should be "nRF52820-QDxx"

Else, ready to go.

Cheers,
Joel

G-Pereira:
Fixed!

myfreescalewebpage:

PR #2357, opened 2019-12-04
Added TVSxx00 family as alias of TVS0500

cnieves1:
Added TVSxx00 devices family as alias of TVS0500.
All these share footprint and symbol:
http://www.ti.com/interface/circuit-protection/esd-protection-and-tvs-surge-diodes/products.html#p3197=26;43&p2192=Surge%20Protection&p810=Uni-Directional&sort=p3285;desc

imagen

PR #2900, opened 2020-08-06
Added BZV55xxxx Zener Diode full list

aris-kimi:
As a `ZMDxx` aliases.
It is the full list of zener diodes expanded.
If this is not acceptable inform me to change to one alias addition in the form of `BZV55xxxx`
Also not sure about `, MiniMELF(SOD-80C)` description ending.

Datasheet: https://assets.nexperia.com/documents/data-sheet/BZV55_SER.pdf

---

---


cpresser:
> It is the full list of zener diodes expanded.
> If this is not acceptable inform me to change to one alias addition in the form of `BZV55xxxx`

It is acceptable, there is already a series as individual aliases in the lib (`DZ2S030X0L`). Other rather recent commits also added whole series (https://github.com/KiCad/kicad-symbols/pull/2357).

On the other hand, we also merged wildcard style series: https://github.com/KiCad/kicad-symbols/pull/2137
This PR is actually pretty similar to the existing `BZM55xxx` series.

I would rather include the wildcard version. It adds less clutter. I don't see the big benefit for the user of having each voltage as own variant.
Any additional thoughts or input @evanshultz ?

> Also not sure about `, MiniMELF(SOD-80C)` description ending.

I suggest to take `, MiniMELF` to be consistent with other parts already in the lib. Also, the `SOD-80C` string is nowhere to be found in the official libraries.

aris-kimi:
@cpresser , thank you for your answer, i included `SOD-80C` found in the datasheet because it could have saved me some search before i found the correlation to `MiniMELF`. I do like the consistency in libraries, so i will remove it and commit again.

evanshultz:
https://kicad-pcb.org/libraries/klc/G2.1/ is clear that new contributions should use individual parts as ALIASes and not wildcards. So this then looks good. This is relatively new and I believe the rationale is that it's more helpful since users don't have to guess and try where wildcards might be. At the cost of bigger files. There's always a tradeoff but that's KLC as of today.

I would think having both package names in the description, or a keyword, is helpful too. If it's in the description please use parenthesis since we use a forward slash to denote available packages on symbols which aren't fully-specified.

aris-kimi:
Thank you @evanshultz , if i understood correct 
`10V, 500mW, 2%,  Zener diode, MiniMELF(SOD-80C)`
is the acceptable description?
If yes then @cpresser , can you choose to merge only two previous commits? In which i think (beside some mess i've fixed) description fits?

I do want to thank you once again for making me feel welcomed here!!

myfreescalewebpage:
@aris-kimi I have checked this PR too, I think the current description is good.
Just one comment: there is a double space before "Zener" in the description for all the aliases, can you remove one space please ?
Then ready to merge.
Joel

myfreescalewebpage:

PR #2905, opened 2020-08-08
AD8313 RF Power detector

chmorgan:
image

---

---


aris-kimi:
Hi @chmorgan , i think this addition should be made under RF lib and also seems that symbol's dimensions could be reduced.
Just a suggestion, better wait for a review...
Thank you for your time.

chmorgan:
Hello!

I think there are other similar parts in this library that’s why I added it
there. For example ADL5513 (I think is one). I’m using both on an eval
board so we can see the performance differences between the two.

Chris

myfreescalewebpage:
Hi @chmorgan , thanks for contributing,

- [ ] I agree with @aris-kimi for two comments.
- [ ] Name of the symbol should be "AD8313xRM"
- [ ] 8 GHz in the description is wrong, should be 2.5 GHz
- [ ] Fp filter should be `MSOP*3x3mm*P0.65mm*` (no "*" at the beginning)
- [ ] Name of pin 6 should be COMM
- [ ] Pin 7 should be Input, not Power Input

Cheers,
Joel

PR #2916, opened 2020-08-18
Use updated KiCad .gitignore from https://github.com/github/gitignore/blob/master/KiCad.gitignore

McNeight:
Use updated KiCad .gitignore from https://github.com/github/gitignore/blob/master/KiCad.gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,28 @@
-*.pyc
+# From https://github.com/github/gitignore/blob/master/KiCad.gitignore
+# Dated 20 Aug 2019
+# For PCBs designed using KiCad: http://www.kicad-pcb.org/
+# Format documentation: http://kicad-pcb.org/help/file-formats/
+
+# Temporary files
+*.000
 *.bak
 *.bck
+*.kicad_pcb-bak
+*.sch-bak
+*~
+_autosave-*
+*.tmp
+*-save.pro
+*-save.kicad_pcb
+fp-info-cache
+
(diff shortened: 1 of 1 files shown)


chschlue:
However, this .gitignore template is meant for versioned KiCad project directories, not libraries.
The file extensions you added wouldn't usually turn up in a symbol library, or am I missing something here?

McNeight:
Generally speaking, you are correct that most wont be seen in a normal library. I guess I look at it as 2 separate use cases:
1) Regular users who install a release version of KiCad which doesn't have any git information.
2) Power users and developers who use this as a versioned KiCad project.

The file is only seen in the second case. I don't recall the specific reason I needed to change the file, but I figured I'd update it to the latest from the public gitignore repo.

myfreescalewebpage:
I'm not really in favor of this. If we go in this direction, plenty of extension will need to be escaped. *.docx, mp3, and so more.
People committing to the repo should just look after their commits. It's a very good exercice.

diegoherranz:
@McNeight, for your second case with power users, if anyone creates a project including a versioned version of this library, the .gitignore should belong to that project and not to the libraries if I understand your use case correctly.

I agree with @chschlue and @myfreescalewebpage and I would vote to reject this PR. In fact, I don't see the need for a .gitignore at all, same as @myfreescalewebpage.

Thanks.

chschlue:
Fully agree. If you're mixing library and project directories, you're probably doing it wrong.
Closing.

PR #2922, opened 2020-08-20
Added ISL3295ExHZ-T and ISL3298ExRTZ-T to Interface_UART.lib

UsairemAlamgeer:
I have added ISL3295ExHZ-T and ISL3298ExRTZ-T to Interface_UART.lib as suggested before.

image

image


aris-kimi:
Hi @UsairemAlamgeer , some tips:

Datasheet needs to be from a manufacturer and not distributors, links from distributors are accepted only if manufacturer does not provide one.

Description should be something like `RS485, RS422, 20Mbps Transceiver, 3.0V to 7V, Low-Power, SOT-23`

Keywords should be something like `RS485 RS422 transceiver Class 3 ESD Level`

Package should be `Package_TO_SOT_SMD:SOT-23-6`

Footprint filter should be `SOT*23*`

Same logic should be applied to both symbols but with appropriate fields naming accordingly for TDFN package.

I am not so sure for the existing (SOT-23) package though, some dimensions do not seem to fit with values mentioned in the datasheet. My lib might needs an update, but to be sure you better wait for a full review from an experienced librarian.

UsairemAlamgeer:
> Hi @UsairemAlamgeer , some tips:
> 
> Datasheet needs to be from a manufacturer and not distributors, links from distributors are accepted only if manufacturer does not provide one.
> 
> Description should be something like `RS485, RS422, 20Mbps Transceiver, 3.0V to 7V, Low-Power, SOT-23`
> 
> Keywords should be something like `RS485 RS422 transceiver Class 3 ESD Level`
> 
> Package should be `Package_TO_SOT_SMD:SOT-23-6`
> 
> Footprint filter should be `SOT?23*`
> 
> Same logic should be applied to both symbols but with appropriate fields naming accordingly for TDFN package.
> 
> I am not so sure for the existing (SOT-23) package though, some dimensions do not seem to fit with values mentioned in the datasheet. My lib might needs an update, but to be sure you better wait for a full review from an experienced librarian.

Thanks for the feedback @aris-kimi.

I will update just like you have pointed.
Also for the packages, I have designed a new SOT23 as per the Dimensions of the datasheet.

aris-kimi:
There are still some things that need to change, for example, 3298 part needs to change in order to allow interchangeability between those two. DE, DI, Y and Z could be at the same place on both symbols. Also two GND pins have to be stacked.
 
I do not quite understand what those links mentioned here are about..

I could suggest to spend some time reading this: https://kicad-pcb.org/libraries/klc/ there is a lot of useful info.

UsairemAlamgeer:
image
image

PR #2935, opened 2020-08-26
Create single- and double-row symbols

evanshultz:
As promised at https://github.com/KiCad/kicad-symbols/issues/2705#issuecomment-629806378, here are the first batch of DIN 41612 symbols. These cover the single- and double-row footprints we currently have in the library (https://github.com/KiCad/kicad-footprints/pull/1076).

image
image
image

My only real concern is about the footprint filters. It appears CvPcb only looks at the number of pins and not the pin names, which means that a 2x16 connector using pin numbers a1 through a16 and b1 though b16 get matched just like one that has a2-a32 and b2-b32 with only even pins. I don't see a way around that so this may be the best we can do. If not, let me know.

Script PR: https://github.com/KiCad/kicad-library-utils/pull/332

As described in the PR, I wanted to try and get this in since it takes care of most of the footprints we have without requiring any framework changes. That is a bigger process and likely will take more time. So I'd really like to get this in now and do the triple-row symbols in a later PR.

Edit: Related issue and PR:
- https://github.com/KiCad/kicad-symbols/issues/1157
- https://github.com/KiCad/kicad-symbols/pull/1286

---

## :warning: Deprecation warning
In preparation for the KiCad v6 release this repository will be locked down on Sept 1, 2020. No new pull requests will be accepted. Existing pull requests can be worked on until Oct 1, 2020. Changes breaking v5.1 compatibility can be merged starting Sept 1, 2020.

On Oct 1, 2020 this repository, including issues and pull requests, will be archived and transferred to [gitlab.com](https://gitlab.com/kicad/libraries/kicad-symbols/). In order for your pull requests and issues to be imported into GitLab you must set your email address on GitHub to public and use the same address for your GitLab account. Or login to GitLab at least once using the GitHub icon. Otherwise the importer can't correlate the account information and the issues/comments on GitLab will be owned by `kicad-bot` ([importer documentation](https://docs.gitlab.com/ee/user/project/import/github.html#how-it-works)).

We plan to convert the library to the new v6 S-expr format after it is imported to GitLab. That also means that old pull requests will need to be redone with the v6 format if they are not merged before then.

---

---


myfreescalewebpage:
Hello @evanshultz 

One comment about the naming of the symbols: why "DIN41612_02x16_AB" is not called "DIN41612_02x16_AB_OddPins" ?
Else no comment on the symbosl themselves, do you plan to have more parts in this PR ?

Joel

myfreescalewebpage:
@evanshultz ping

evanshultz:
Thee "_EvenPins" versions only have even pin numbers (2, 4, 6...). No suffix means all pin numbers are included (1, 2, 3...). There aren't any symbols with only odd pins because we don't have any footprints with only odd pins. Please check the footprints merged at https://github.com/KiCad/kicad-footprints/pull/1076 or look at them in the library. I do believe that is correct and these symbols will support all single- and dual-row DIN41612 footprints we have in our library.

The script framework needs enhancement to support the triple-row symbols, so I was hoping to stop here and get these merged. Then update the script framework and come back later with another PR for the triple-row symbols. If that's OK with you, then no, I don't plan to add more now.

myfreescalewebpage:
@evanshultz thanks for the clarification and for the lesson about naming !
Merging now.

PR #2938, opened 2020-08-27
Added π120U30 and π120U31

G-Pereira:
Alias for this two digital isolators.
[Datasheet](http://www.rpsemi.com/public/doc/2Pai_%CF%8012xUxx_20200416_EN.pdf)


chschlue:
Guess this had to happen... there was a discussion about manufacturers' sanity just yesterday...

@stambaughw @sethhillbrand are there still reasons not to include multibyte chars in symbol names or do you believe it's safe?

@G-Pereira how about calling them τ/2?

sethhillbrand:
U+0C30 is in our standard font, so it should be no problem.

aris-kimi:
I want to suggest to add a translated version of MPN at the keyword field as `pi120u30/1`. I think this might ease the search process.

G-Pereira:
> I want to suggest to add a translated version of MPN at the keyword field as `pi120u30/1`. I think this might ease the search process.

I was about to suggest the same

aris-kimi:
@G-Pereira one last change, parent symbol's name offset i think it is good to change to 20mil / 0.508mm

G-Pereira:
Hi, I already fixed that on a separate PR that already has been merged, check https://github.com/KiCad/kicad-symbols/pull/2939

G-Pereira:
@aris-kimi and @chschlue is this ready to merge?

chschlue (approved):
Yes, thanks.

PR #2941, opened 2020-08-27
Added AP3402 buck converter

marecl:
Added AP3402 buck converter from Diodes Incorporated
Datasheet: https://www.diodes.com/assets/Datasheets/AP3402.pdf

Symbol:
image

Footprint: builtin TSOT-23-6 (TSOT-26) (yes, they are the same. naming in KiCAD is confusing sometimes)
Footprint filtering includes HandSoldering variant

Travis throws out warning that power pins are not at the top of the symbol. I thought that if it worked for AP63200WU, then I can violate this rule too >:)
[edit] Forgot to explain. Except from the reason above, this layout would match the one in the datasheet and it's quite readable anyway, separating input from output stage.

---

## :warning: Deprecation warning
In preparation for the KiCad v6 release this repository will be locked down on Sept 1, 2020. No new pull requests will be accepted. Existing pull requests can be worked on until Oct 1, 2020. Changes breaking v5.1 compatibility can be merged starting Sept 1, 2020.

On Oct 1, 2020 this repository, including issues and pull requests, will be archived and transferred to [gitlab.com](https://gitlab.com/kicad/libraries/kicad-symbols/). In order for your pull requests and issues to be imported into GitLab you must set your email address on GitHub to public and use the same address for your GitLab account. Or login to GitLab at least once using the GitHub icon. Otherwise the importer can't correlate the account information and the issues/comments on GitLab will be owned by `kicad-bot` ([importer documentation](https://docs.gitlab.com/ee/user/project/import/github.html#how-it-works)).

We plan to convert the library to the new v6 S-expr format after it is imported to GitLab. That also means that old pull requests will need to be redone with the v6 format if they are not merged before then.

---

---


aris-kimi:
Hi @marecl , thank you for your contribution.

Description should ends with `, TSOT-26`
Pin names must be capitals on all characters.

Also not sure about footprint selection. I think a new scripted one must be submitted. (Meaning also fp filter must change but better wait for more ipnut in this)

@cpresser can you please confirm this?

https://www.diodes.com/assets/Package-Files/TSOT26.pdf :
εικόνα

https://www.jedec.org/sites/default/files/docs/MO-193D.pdf (current TSOT-23-6)

marecl:
thank you @aris-kimi I forgot about committing that one!
Also yes, they're the same package but with different names. Because everyone needs their own standards, eh?
image
https://www.mouser.com/pdfdocs/r3152n-sot-23-6.pdf

aris-kimi:
There are some small differences though, at A and A2 dimmensions. Small difference but i think this requires a new fp. But we better wait for a confirmation.

cpresser:
You are right, this looks exactly like `MO-193AA`.
But isn't there already a PR to fix that package? This one? https://github.com/KiCad/kicad-footprints/pull/2412

Thus, merging the symbol with footprint set to `TSOT-23-6` is okay. The current footprint is okay-ish, the new one will be an exact fit.

aris-kimi:
Thanks. Description then should ends as decribed at the datasheet `TSOT-26`? or as the fp applied `TSOT-23`?

I think it woulh be great if `TSOT26` added at new fp's description..

cpresser:
I looked over this again and now I am sad. I would be happy if somebody has more insight into this.

There is also a `SC-74-6_1.5x2.9mm_P0.95mm` in the footprint-repo.
Here is a screenshot of it overlaying the current `TSOP-23-6`:
image

TBH, I am not sure which package-name to put into the description. This is the first time I see `TSOT26`. We do have `TSOP-6` in the library. Not sure why they are named like that.
Then there is `SC-74`. Which is also called `TSOP6` or `SOT457` (see https://www.nxp.com/docs/en/package-information/SOT457.pdf)

Why are there so many packages that are almost identical, and have overlapping names? :angry: 
Does anybody know why the `TSOT-23-6` package in our repo is named the way it is? 

My current assumption is that `MO-193AA` is not the same as `SOT457`. I based that on the fact that the `SOT457` datasheet has an empty JEDEC field.

---

Since my comment from above still seems to be correct ( `MO-193AA` has the same dimensions as diodes datasheet) we should stick with that.
Perhaps we should put the KiCad package name and the manufacturers name into the description. Like `, TSOT26 (TSOT23-6, MO-193AA)`

aris-kimi:
If we stick with:

`F2.3 - Manufacturer specific version of generic footprints`

would then be a good approach to copy scripted TSOT23-6 footprint and rename it to DIODES_TSOT-26 and mention its origin in its description?

chschlue:
@cpresser 
It's... complicated.

The SOT-###, TO-### naming is theoretically obsolete (not used by JEDEC anymore but replaced by https://www.jedec.org/sites/default/files/docs/JESD30E.pdf).
MO-### are basically just document numbers for mechanical outlines (hence MO) and their JESD30 designator (the newer ones, that is).

There a many packages with very well known (although obsolete) names like TO-220, for example, but also SOT-23, SOT-89 and many more that stuck and are used by many manufacturers and the libs.

Then, there's JEITA, which is basically the Japanese JEDEC, with registered packages like SC-70 and SC-74 which are, I believe, mechanically identical to their JEDEC counterparts but also has a SOIC registration which differs from JEDEC MS-012, for example (I don't have access to those, though).

On the flip side, we have manufacturers making up internal package names that conflict with those common/ex-official names, like NXP (which calls everything SOT###, but at least tends to reference JEDEC/JEITA registrations if available) or Diodes (which apparently opted to shorten TSOT-23-6 to TSOT26).

I suggested footprint aliases to be able to use MF designations in symbols without having to actually duplicate everything (including 3D models) over and over, but these at least won't be in v6, if coming at all.

If you can come up with a one-size-fits-all solution to this mess, be my guest.

That being said, I'm against adding a Diodes_TSOT26, at least as long as it's just a duplicate of TSOT-23-6.

marecl:
just for the record, I didn't want to put a stick into a beehive :'D

I have an idea that each generic footprint could have a tag with information to which standard it was created (or alternative name maybe?) or split them into libraries like Package_JEDEC.

I think duplicates would be fine at early stage of adding these. there's a royal ton of generic footprints and compatibility between these would probably have to be manually checked. dunno.

aris-kimi:
@marecl , i felt the same way... It is really a mess out there...

For this PR i want to ask for some last changes,

- [ ]  to move EN down left, PG one up and LX also one up
In order to allow easy change of symbols between similars like:
εικόνα

- [ ] Footprint filter should be `TSOT?23*`


Thanks for the typo correction.

aris-kimi:
Thanks @marecl , symbol looks good.

εικόνα

@cpresser , i want to ask about this lib table issues noted, i am merging this one after confirmation that we are good to go. Symbol has been checked everything is ok.

PR #3094, opened 2020-09-26
Make Texas_WSON-N8 variants non vendor specific

cpresser:
LP2951-3.0_WSON was using the vendor specific footprint.

https://github.com/KiCad/kicad-symbols/pull/2852


chschlue (approved):
Thanks.

myfreescalewebpage:
Hello @cpresser @chschlue have you check the part fit the wanted footprint exactly. I reported in the past an issue with Texas footprints having the same name but very different dimensions, at https://github.com/KiCad/kicad-footprints/issues/1536

cpresser:
No, I had not, I just changed the naming.

It seems like the previous vendor specific texas footprint was wrong.
* The EP in the LP2951 datasheet is 2.6x3.0mm. 
* Prior and after this merge, this symbol had a FP with 1.2x2mm
* So it was wrong to begin with, and I kept it that way (I should have checked)
* I will add the missing WSON footprint and make another PR to fix this symbol.

image

cpresser:
The correct footprint is already in the lib, but it is not generated. And the name is wrong. I will replace it.
It is also used by some other parts, I will update them accordingly:
image

cpresser:
Sorry, I have more confusion to add:
There is `LP2951` and `LP2951-N`. 
image
image

The datasheet link of  `LP2951-3.0_WSON` is wrong. It points to the `N` version.
A fix for that is here: https://github.com/KiCad/kicad-symbols/pull/3096

cpresser:
PRs related to improving the WSON8-4x4mm (list will be extended):
https://github.com/pointhi/kicad-footprint-generator/pull/643

PR #3096, opened 2020-09-28
Fix datasheets of LP2951-***

cpresser:
The datasheet pointed to the -n variant which is a different part.


myfreescalewebpage:
@cpresser no comment here, merging
Joel

PR #3097, opened 2020-09-28
Replace WSON-8_4x4mm with generated footprint

cpresser:
The existing footprint did not have the EP in its name.
Its filename also differed from the kicad-name (WSON8_4x4mm_P0.8mm), there
is a missing dash.


myfreescalewebpage:
@cpresser review of each modifications:

- [ ] LP2987-5.0_WSON8 and LP2988-5.0_WSON8 the EP size is not correct, another fp should be created (EP 2.2x3mm)
- [ ] LM5017SD the EP size is not correct, another fp should be created (EP 1.98x3mm)

Joel

cpresser:
Looks like there were quite a few bugs hidden with the existing WSON8 parts.
I will also create FPs for those variants later today.

myfreescalewebpage:
Yes it seems. Thanks.

cpresser:
More footprints are here. https://github.com/KiCad/kicad-footprints/pull/2503
@myfreescalewebpage  Can you please check the naming?

myfreescalewebpage:
Name looks great @cpresser thanks. Wait review of footprint by @chschlue you have assigned on

myfreescalewebpage:
Merging following integration of the fp

